PR #6, opened 2021-01-03
fixes typo

IronicBadger:
sorry it's such a small contribution but this fixes a little typo. looks like a very interesting project!!!


hay-kot:
Ha! You'd think I could spell my own app name right. Thanks for the commit, I feel like I've made it now having a podcast host contribute to my project.

PR #26, opened 2021-01-04
Update docker-compose

Dulanic:
Updated docker-compose to use one higher level mount instead of 2 lower level mounts. This will also store the debug json locally if imports fail vs having to go into the container.


Dulanic:
Closing for now, caused unforseen issues. Will test further before submitting,

hay-kot:
Odd, I'm actually using a higher level mount on my "production" instance. I suspect is has something to do with the folder not existing? There's a few problems that are caused by that, I thought I cleaned them all up but maybe not.

PR #33, opened 2021-01-05
Added Simple Field Validation to New Theme Modal

zackbcom:
With out the validation, a user can enter a blank theme name. When trying to delete the theme with no name, the call fails, leaving the blank theme there forever.

Blank Theme name: Unable to remove.
image

First open: Disabled Create
image

Error state: Added field validation for required.
image

Successful validation:
image


hay-kot:
Good catch, Thanks!

PR #39, opened 2021-01-06
Added System Dark Mode, Updated theme settings, Persistent Vuex

zackbcom:
Added option for System specific dark mode.
Updated Themes to have one color palette for dark and light to align with the database.
Added confirmation for deleting a theme.

image

image

### DIFF
--- a/dev/dev-notes.md
+++ b/dev/dev-notes.md
@@ -46,3 +46,5 @@ General
 - Improved documentation
 - Fixed hot-reloading development environment - [grssmnn](https://github.com/grssmnn)
 - Added Confirmation component to deleting recipes - [zackbcom](https://github.com/zackbcom)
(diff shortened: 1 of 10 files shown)


zackbcom:
@hay-kot Let me know what you think with the progress. There are some edge cases I still need to fix. Like deleting the current theme, It should fall back to default.

hay-kot:
@zackbcom  looks great so far, thanks for cleaning up my garbage!

zackbcom:
1. Added in persistence to vuex, everything in state will rehydrate on pageRefresh / browserReload.
2. Cookie code was removed from codebase.
3. need to lock down the changing of default theme.

zackbcom:
Updated store persistence only for the userSettings module. (Selected Theme, Darkmode Setting)

PR #52, opened 2021-01-09
Frontend Organization

hay-kot:
Pages now stored in independent folders
General UI components in components/UI
Dedicated components in components/{page}


zackbcom:
lgtm

PR #57, opened 2021-01-09
Changed uvicorn port to 80

### LINKED ISSUE #55: Dockerfile exposes port 80, but runs on port 9000
The Dockerfile exposes port 80 (Inherited from tiangolo/uvicorn-gunicorn-fastapi which in turn inherits it from tiangolo/uvicorn-gunicorn), but the cmd in the dockerfile runs uvicorn on port 9000. I don't think there's any problem in running on port 80 in docker, since nothing else will run on the containers port 80 - and users can map the port however they like (The docker-compose.yml already maps to a different port than 9000).

kentora:
Closes #55


docker-compose.yml
```diff
@@ -9,13 +9,7 @@ services:
...
     ports:
-      - 9090:9000
-    environment:
-      db_username: root
-      db_password: example
-      db_host: mongo
-      db_port: 27017
-    volumes:
+      - 9090:80
       - ./mealie/data/img:/app/data/img
```
hay-kot, line 13: Your docker-compose will not work.

hay-kot (commented):
While I understand the concern for an additional exposed port I feel a better approach would be to rebuild the container from the ground up on an alpine image. Currently, the image is over 1gb. Which is insane for what's going on in the background. I plan to revisit the container build after I get some more patches out and it's a little more stable but am open to PRs to improve the docker container overall. 

Removing port 9000 would break everyone's docker-compose and I don't think the benefits outweigh the downsides. I think it makes more sense to do what I've described above. 

Thanks!

kentora:
I do appreciate that an effort could be done to slim down the final image - an easy step would be to change the base image to use alpine (a version that already exists). This will have the same port 80 exposed, and currently there is no way of "un-expose" a port via dockerfile, so I still think the port change is valid.

When recreating docker containers, and pulling the latest image the CMD does not seem to change (just tested this, though not through compose), so changing the CMD should only affect new containers.

The motivation behind using port 80, instead of exposing another port is to better support reverse proxies such as traefik, which by default passes requests to the exposed port. This can be overridden, but is not as user-friendly. 

If you change your mind, and want this change based on the above, let me know, and I'll take care of the conflicts and the oopsie I made in docker-compose...

Best Regards

hay-kot:
Interesting, I wasn't aware of the problem with Traefik. I think that alone warrant the change. If you can please re-work the PR on the `dev` branch and I'll merge it with the `master` in the next release as a breaking change. 👍

PR #72, opened 2021-01-14
Translations

### LINKED ISSUE #50: Multilanguage support possible?
Hello,

i would like to know if multiple languages are (or will be) supported? I would like to use this for our cooking-club with some friends, but i would like to know if this application can be set to another language (dutch for example) because some of them have limited knowlegde of english. I would gladly like to help with the translations...

keep up the great work!

kentora:
(Somewhat) closes #50 

Initiation of translations, along with a danish translations, that might be improved upon.
Right now language can be added with a new .json file - but as the project grows and new strings should be translated to existing translations, it might be nice to have a translation management tool in place.


hay-kot:
Thanks for cleaning up those merge conflicts!

I'd like to see some sort of documentation added to the developer guide docs on how to contribute more languages and how to maintain this moving forward.

dpieski:
I am not too familiar with code tests that are run on commit (maybe the "Checks" tab?), but I wonder if a check can be made when any language file is modified, to make sure that all language files have the same keys and have a value for each key? e.g., for each file in "/locales/".

kentora:
> I am not too familiar with code tests that are run on commit (maybe the "Checks" tab?), but I wonder if a check can be made when any language file is modified, to make sure that all language files have the same keys and have a value for each key? e.g., for each file in "/locales/".

This could be one way to go, but as the project (hopefully) grow, i don't think it's viable to reqiure all files to have the exact same keys (and that they are filled). This would mean two things:

1. When a helpfull persons starts to translate, it has to be done in full
2. When someone adds a functionality that requires a new string in UI, they need to provide translations for this string, for every existing language.

I think a better option is to use some sort of translation management system to have users translate each string, view progress, maybe even suggest better translations for already translated strings, and (this is the kicker for me) some sort of notification when a new string should be translated to a language you are helping out with.

kentora:
Updated to match dev branch again, and added some initial documentation. Let me know what you think :)

PR #106, opened 2021-01-21
Add a health check

goose-ws:
Adds a health check to ensure the container's webserver is running


hay-kot:
Thanks for the PR. Currently the docker container on the master branch is not reflective of my overall goals. in the dev branch I've slimmed it down using the alpine image and plan to remove as many dependencies as possible. As such, I'd like to not rely on curl to perform the health checks. I'm open to implementing something in python using existing packages.

PR #123, opened 2021-01-24
First take - reuse categories

### LINKED ISSUE #83: Present users with existing categories and tags, when adding or editing a recipe
Currently users will have to type in the exact same tag/category to get the (not yet implemented) advantages of tags and categories.

To help the user, they should be presented with existing categories and tags in the combobox. The user should then be able to search and select among the existing (or, of course as it is possible today: add a new one).

kentora:
Needs:
* Migrations
* Tags
* Pulling in dev and resolving conflicts

Close #83 when merged


hay-kot:
Closing, as I was able to take your changes and merge them into #141, thanks for the hard work!

PR #128, opened 2021-01-29
translation: add swedish

dekvall:
- Also change da to Danish and not dutch


hay-kot:
> * Also change da to Danish and not dutch

Well that's embarrassing! Thanks for the contribution.

Issue #124, opened 2021-01-24
[META] Language Support


hay-kot:
Language Support is now done through a 3rd party site that will automagically create and submit a PR for you! Checkout [Mealie at Crowdin](https://crowdin.com/project/mealie) for how you can support your language or others.

[![Crowdin](https://badges.crowdin.net/mealie/localized.svg)](https://crowdin.com/project/mealie)

<s>
See the [documentation](https://hay-kot.github.io/mealie/contributors/translating/) for how to contribute additional languages for the web UI. PR Request should be made on the `dev` branch to keep up with rapid development changes. If you'd like to contribute/support additional languages comment on this issue. 

# Current Translation Status
## Currently Supported
Current Languages in `frontend/src/store/modules/language.js`:
* Danish (da) - @dekvall #72
* English (en) - general
* French (fr) - @Raghnarok05, @Batgame  #93 
* Swedish (sv) - @dekvall #128
* 简体中文 (zh-CN) - @wengtad #133
* 繁體中文 (zh-TW) - @wengtad #133

## Prepared Pull Requests
Pull Requests with `language.js` and locale file:

## Currently Supported Needing Attention
* Danish (da) 
* Swedish (sv) 
* French (fr) [Updated with #146]
</s>

dpieski:
I can do a Spanish translation if no one else is already working on that

beefsaus:
I can do a Dutch translation if you need help

ilco3:
ive added a dutch translation  i translated myself 
mealie/frontend/src/locales/ could sombody check if  if it works -if my usage of words is correct

> I can do a Dutch translation if you need help

ilco3:
> ive added a dutch translation i translated myself
> mealie/frontend/src/locales/ could sombody check if if it works -if my usage of words is correct
> 
> > I can do a Dutch translation if you need help

uploaded it trough the create a new file function on mealie/frontend/src/locales/

mindesbunister:
#160

dpieski:
# Current Translation Status
## Currently Supported
Current Languages in `frontend/src/store/modules/language.js`:
* Danish (da) - @dekvall #72
* English (en) - general
* French (fr) - @hay-kot, @Raghnarok05, @Batgame  #93 
* Swedish (sv) - @dekvall #128
* 简体中文 (zh-CN) - @wengtad #133
* 繁體中文 (zh-TW) - @wengtad #133

## Prepared Pull Requests
[...]

PR #154, opened 2021-02-10
Add polish translation

retmas-gh:
As the title says; reread it twice so it should be at least usable and everything should be spelled correctly.


dpieski:
Reference #124

PR #177, opened 2021-02-21
iOS shortcut documentation

rastacalavera:
the slug for the screenshot image will need to be modified


hay-kot:
Awesome, thanks for this! If you could add the link to the [mkdocs.yml](https://github.com/hay-kot/mealie/blob/master/docs/mkdocs.yml) so it'll show up on the page, I'll get this merged.

rastacalavera:
Sounds good! I added it in what made sense to me but feel free to change the order or tell me where to put it and I can change it too.

PR #188, opened 2021-02-25
fix docs ios image

wengtad:


rastacalavera:
Thanks for doing this!! I was going to work on polishing it up today! I too was going to change it to a block quote after I saw how messy it was as inline code. I was just reading about defining image size too on a reddit thread :P Although when I previewed your md in your master branch, it doesn't look like that code is working properly.

wengtad:
> Thanks for doing this!! I was going to work on polishing it up today! I too was going to change it to a block quote after I saw how messy it was as inline code. I was just reading about defining image size too on a reddit thread :P Although when I previewed your md in your master branch, I see your image html. Does that syntax go away when actually viewing it in production? I think it should cause it looks like ` - attr_list` is in the main yaml

This should thanks to @hay-kot for his great work. I'm just changing the image path to make the images show out. I'm previewed it on mkdocs and I believe the actual result on production will be like this after PR merge:
image

With `attr_list` enabled, the image and text will be align side by side.

PR #219, opened 2021-03-22
Feature/authentication

hay-kot:


hay-kot:
+ Dropped fastapi-login
+ Added own jwt creation/management
+ Proper dependency injection for testing
+ fix various login bugs.

PR #224, opened 2021-03-27
Feature/submit on enter key

### LINKED ISSUE #223: Modal dialogs such as login do not submit by pressing enter
**Describe the bug**
Modal dialogs such as the login modal and import by link modal do not perform the expected action when pressing enter.
- The login dialog does nothing on enter, would expect it to submit the form
- The link import modal refreshes the page but does not submit the form, this is especially confusing since one may assume it was imported.

**Steps To Reproduce**
1. Press `Log In` from menu (or import from recipe add)
2. Enter credentials / url
3. Hit the enter key

**Expected behavior**
Web form is submitted, login / import is performed.

**Actual Behavior**
Nothing happens for login / page refreshes without import. 

**Device Information (please complete the following information):**
 - OS: Ubuntu 20.04
 - Deployment: "dev"
 - Browser: Chrome
 - Version: dev branch @ 223c198e83, Chrome: 89.0.4389.90 (Official Build) (64-bit)

**Additional context**
I believe the issue is present in both master & dev branches.

hay-kot:
Add submit on enter and proper forms for...
  - Add recipe by URL
  - Login Modal + Page
  - Sign Up Page
  - Add User
  - Add Theme
  - Add Group

Fixes #223

### DIFF
--- a/frontend/src/components/UI/AddRecipeFab.vue
+++ b/frontend/src/components/UI/AddRecipeFab.vue
@@ -21,9 +21,8 @@
 
           <v-spacer></v-spacer>
         </v-app-bar>
-
-        <v-card-text>
-          <v-form ref="urlForm">
+        <v-form ref="urlForm" @submit="createRecipe">
+          <v-card-text>
             <v-text-field
               v-model="recipeURL"
               :label="$t('new-recipe.recipe-url')"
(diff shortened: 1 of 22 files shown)


frontend/src/components/UI/AddRecipeFab.vue
```diff
@@ -35,29 +34,29 @@
...
-        </v-card-actions>
+          <v-card-actions>
+            <v-spacer></v-spacer>
+            <v-btn color="grey" text @click="reset">
+              {{ $t("general.close") }}
+            </v-btn>
+            <v-btn
+              color="success"
+              text
+              @click.prevent="createRecipe"
```
djmattyg007, line 53: This click handler shouldn't be necessary. Instead of preventing the form's submit action from running only to run the same method it would normally run, the click handler should be removed entirely and replaced with `type="submit"`. This turns it into a native form submit button, utilising native browser functionality.

hay-kot: Ah, you are right. I misremembered the requirements of the `.prevent` and where it was needed to keep the SPA from reloading. I've addressed this in #225. Thanks for the input!

PR #237, opened 2021-03-31
Typo with Tuesday

rrikesh:


sephrat:
Is this string still used? From my quick search, it looks like it's leftovers from before we localized days, that I shamefully did not take the time to clean up.
Also, please create pull requests against dev branch.

hay-kot:
> Is this string still used? From my quick search, it looks like it's leftovers from before we localized dates, that I shamefully did not take the time to clean up.
> Also, please create pull requests against dev branch.

I just pushed a new release, the live release was probably still using it. I did notice that it was still there. I was a little afraid to remove it, but now that the backend is in a much better place I can focus on cleaning up the frontend. I did however fix the typo prior to release v0.4.0 just in case.

Issue #248, opened 2021-04-01
Recipes imported from Nextcloud have invalid times


dietrichmd:
**Describe the bug**
When importing a recipe from Nextcloud, the time of the recipe is shown as PTxHxxM where x=number.  See https://i.imgur.com/h000JCX.png for example

**Steps To Reproduce**
Please be specific!
1. migrations,
2. import recipes from nextcloud'
3. open recipe
4. see invalid time

**Expected behavior**
Time should be displayed in hours and minutes

**Actual Behavior**
Time is shown only in minutes, if that, preceeded by unneeded characters

**Screenshots**
https://i.imgur.com/h000JCX.png

**Device Information (please complete the following information):**
 - OS: Ubuntu 20.04
 - Deployment: docker-compose
 - Browser: chrome
 - Version: 89

sephrat:
I also faced this when importing my Nextcloud recipes. I just learned to live with it until I manually edited all my recipes.
This is due to Nextcloud Cookbook overwriting the times with this custom format (technically it should be an ISO 8601 duration format: https://schema.org/Recipe). 

Ideally we could amend the times when importing from Nextcloud Cookbook to overcome this. It's not a complicated logic but it will need some (if not a lot) rework in Mealie.
For now I would recommend manually editing the json files before importing them into Mealie.

hay-kot:
Can you provide a nexcloud file or backup that I can test? I think this might be an easy fix.

dietrichmd:
Sure thing:  http://cloud.dukefamily.net/s/Y9jHccw6RgiCzrF

hay-kot:
Fixed with #248. Will try to get a patch out in the next day or so 🎉

hay-kot:
This has been fixed in [v0.4.1](https://hay-kot.github.io/mealie/changelog/v0.4.1/) 🎉 along with some other NextCloud migration fixed

PR #250, opened 2021-04-01
Nextcloud Import Bugs - #248

hay-kot:
- Parses timedelta strings properly...ish? Minimal testing was done
- Fix failed imports when category was an empty string


sephrat:
Shouldn't we try to parse any time to ISO 8601 duration format in json file, and prettify the date on the frontend? For the sake of sticking to the schema standards (and avoid a Nextcloud-like situation where this software will have its own time format), and also for the sake of an easier localization.

hay-kot:
> Shouldn't we try to parse any time to ISO 8601 duration format in json file, and prettify the date on the frontend? For the sake of sticking to the schema standards (and avoid a Nextcloud-like situation where this software will have its own time format), and also for the sake of an easier localization.

That's probably the 'right' way to do this. Using strings for times offer more flexibility as to what you want to put like '3-4 hours'. If we were to go about formatting them as ISO 8601 duration we would need to validate the input on the backend, change the input types to sanitize on input with a time picker, and come up with some sort of migration path for existing Mealie recipes. 

Since storing as strings is working for me, this change isn't super high on my radar.

sephrat:
Great, just wanted to make sure this was on your roadmap as well. :-) 
Let's keep it simple for now.

PR #264, opened 2021-04-05
Fix docker dev db persistence

sephrat:
- Changed app_data to /app/dev/data
- app.py is now the single entrypoint: initialization is done there. This was needed because the RUN commands in Dockerfile were broken (initialization was done prior to volume mounts).

I still need to test the docker production build, not sure yet if the initialization in run.sh is still needed.

### DIFF
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -19,4 +19,4 @@ COPY ./mealie /app/mealie
 RUN poetry install 
 
 RUN chmod +x /app/mealie/run.sh
-CMD /app/mealie/run.sh
+CMD /app/mealie/run.sh reload
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -18,4 +18,5 @@ COPY ./mealie /app/mealie
 
 RUN poetry install 
 
-CMD ["poetry", "run", "python", "mealie/app.py"]
+RUN chmod +x /app/mealie/run.sh
+CMD /app/mealie/run.sh
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -14,12 +14,8 @@ RUN curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/get-
 # Copy poetry.lock* in case it doesn't exist in the repo
 COPY ./pyproject.toml /app/
 
-# RUN poetry install 
-
 COPY ./mealie /app/mealie
 
 RUN poetry install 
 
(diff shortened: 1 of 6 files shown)


hay-kot:
> * app.py is now the single entrypoint: initialization is done there. This was needed because the RUN commands in Dockerfile were broken (initialization was done prior to volume mounts).

I'd prefer to keep the pre-start functions outside off the app.py. 

I've had numerous issues in the past with the database not initializing properly due to import order or some other code locks out the database initialization. I think I'd prefer to use the `run.sh` as a source of truth for startup. We can read in the production environmental variable and determine how to run the server in the run.sh. I think that gives the most flexibility long term and helps me stay sane when it comes to weird startup bugs and import errors.

If we make those changes to run.sh then we can use a single entry point for the production and development container, and use run.sh to start the backend natively on the machine. 

What do you think?

sephrat:
I actually started by trying to use the run.sh script but ran into some issues and ended up with this solution. What you say makes sense, I'll give it a second try.

hay-kot:
I dug around a bit and found that, identifying a boolean value in bash is tenuous. 

The docker images for FastAPI uses 2 scripts that look for a pre-start script to run prior to actually running. Having a pre-start.sh, a start.sh, and a start-reload.sh may be the way to go if there isn't a clean way to do it in one script. 

Some other ideas...
- use a `MODE` env variable to set reload. `if [ ${MODE} == "reload"]; then ...` something like that.
- Pass an argument into the script to run either in reload or production mode. `run.sh --reload`

Source -> https://github.com/tiangolo/uvicorn-gunicorn-docker/tree/master/docker-images

sephrat:
Well I thought it was my code that didn't work but for some reason, even my production build from dev branch will not start:
`/bin/sh: /app/mealie/run.sh: not found`
I'm obviously too novice to figure this stuff out quickly and I'd rather not waste hours into this :-) I'll push my latest changes for someone else to pick up where I left and hopefully we can all figure this out.

sephrat:
Wait, just found out my issue: my shell script had CRLF instead of LF... Damned, I wish I realized that earlier.
I can get the images to run now, let me get back to the initial topic!

sephrat:
OK I'm done, thanks for bearing with me.
I did not include the dynamic DIR thing you brought up as it was not needed.

I was about to adapt the 'backend' part in makefile, but it turns out it's not that straightforward to execute a shell script on Windows, so I left it as-is.

PR #282, opened 2021-04-10
Fix syntax error from new portuguese translations

sephrat:
@pmatarodrigues FYI


sephrat:
Woops, forgot to merge from upstream.

PR #289, opened 2021-04-11
feature/editor-improvements

hay-kot:
- [x] Pin Editor Buttons
- [x] Cleanup Funcional Code
- [x] Split up router
- [x] Fix Locale 
- [x] Update Locale selector
- [x] Refactor component locations
- [x] Fix Mobile Cards
- [x] Recipe Time Card Layout


hay-kot:
> Actually the solution probably is to fix the way these items are generated rather than finding a way to set the locale early. It's more rework but these really are anomalies compared to the way all other strings are handled.
> Let me know if you need a hand.

I was able to get the language to set from the store prior to navigation in the index.js with a router guard. This is a fix for now. The core issues is reactivity of nested computed properties. I'm not sure what the 'best practice' is for having nested computed properties be reactive.

sephrat:
I'm afraid it's still not enough. If you open Mealie home page for the very first time, the sidebar still renders in English on first load. F5 fixes it. 
At least it's more acceptable than before. I assume the next step should be making the sidebar reactive like you said.

sephrat:
Odd indeed, it's pretty consistent on my end with Firefox and Chrome, both on Windows and Android. It's always the sidebar that sticks in English despite my settings being in French. 
image
The rest is quite invisible as it's either reactive or rendered after the language intialization.

PR #296, opened 2021-04-13
Enrich page title with context

sephrat:
- Static pages have their own titles
- The name of the recipe is displayed when viewing it

image

The logic behind the setting of the title may be a bit complicated but it's the only one I found able to render dynamic title (for recipes) and that would not break the titles in the browser history.


hay-kot:
Snazzy!

PR #319, opened 2021-04-20
Crowdin integration

sephrat:
- Locale codes moved to 4-letter code standards.
- Crowdin configuration
- Fixed portuguese files


sephrat:
About /frontend/src/locales/dateTimeFormats/ :
They're not to be translated, but integrating them into Crowdin will simplify the rollout of a new language.
We can hide those strings from the UI here:
image

Issue #249, opened 2021-04-01
RP4/docker - error: can't initialize time (workaround included)


cklann1:
**Describe the bug**

Container fails startup on RP4 2g running:
Linux rp4-2g 5.10.17-v7l+ #1403 SMP Mon Feb 22 11:33:35 GMT 2021 armv7l GNU/Linux

Container log shows errors:

```bash
    Fatal Python error: init_interp_main: can't initialize time,
    Python runtime state: core initialized,
    PermissionError: [Errno 1] Operation not permitted,
    ,
    Current thread 0xb6f8a390 (most recent call first):,
    <no Python frame>,
    {"level":"info","ts":1617282122.7082403,"msg":"using provided configuration","config_file":"./Caddyfile","config_adapter":""},
    {"level":"warn","ts":1617282122.7122238,"logger":"admin","msg":"admin endpoint disabled"},
    {"level":"info","ts":1617282122.7126932,"logger":"tls.cache.maintenance","msg":"started background certificate maintenance","cache":"0x3cba140"},
    {"level":"info","ts":1617282122.7139606,"logger":"tls","msg":"cleaned up storage units"},
    {"level":"info","ts":1617282122.715005,"msg":"autosaved config","file":"/root/.config/caddy/autosave.json"},
    {"level":"info","ts":1617282122.71505,"msg":"serving initial configuration"},
    Successfully started Caddy (pid=14) - Caddy is running in the background,
    Fatal Python error: init_interp_main: can't initialize time,
    Python runtime state: core initialized,
    PermissionError: [Errno 1] Operation not permitted,
    ,
    Current thread 0xb6f33390 (most recent call first):,
    <no Python frame>,
```
**Steps To Reproduce**
1) Built container using docker-compose from <https://hay-kot.github.io/mealie/getting-started/install/>:
  ```bash
    pi@rp4-2g:~/mealie$ docker pull hkotel/mealie
    pi@rp4-2g:~/mealie$ nano docker-compose.yml
    pi@rp4-2g:~/mealie$ mkdir data
    pi@rp4-2g:~/mealie$ docker-compose up -d
  ```
2) Started container (using Portainer)

3) No joy :(

**Additional context**
Successfully implemented this workaround (fix?) on the host rp4-2g found here: <https://github.com/linuxserver/docker-papermerge/issues/4#issuecomment-735236388>
```bash
[...]

PR #335, opened 2021-04-21
Add tip to fix rpi host running docker container

cklann1:
Per issue #249, tip gives guidance for running Mealie in a docker linux/arm/v7 container.


cklann1:
Thanks for the noob advice @sephrat  - hope this works (more) better.

sephrat:
Yep that's better!

Issue #337, opened 2021-04-21
Custom DEFAULT_USERNAME doesn't take effect


waywardone:
[Documentation](https://hay-kot.github.io/mealie/getting-started/install/) indicates that you can customize the default username and password using `ENV DEFAULT_USERNAME` and `DEFAULT_PASSWORD` but it looks like `ENV DEFAULT_EMAIL` is what produces the desired effect and not `DEFAULT_USERNAME`. 

**Steps To Reproduce**
1. Set `DEFAULT_USERNAME` to `foo@bar.com` and `DEFAULT_PASSWORD` to `topsecret` in `docker-compose.yml`
2. Start mealie and attempt to login with `foo@bar.com` and `topsecret`. This fails.
3. Attempt to login with `changeme@email.com` and `topsecret`. This works.

**docker-compose.yml**
```
    mealie:
    image: hkotel/mealie:v0.4.3
    container_name: mealie
    hostname: mealie
    restart: always
    ports:
      - 4015:80
    environment:
      DB_TYPE: "sqlite"
      TZ: "${TZ}"
      DEFAULT_PASSWORD: "topsecret"
      DEFAULT_USERNAME: "foo@bar.com"
    volumes:
      - ${USERDIR}/docker/mealie:/app/data
```

**mealie.log**
```
INFO: 	------APP SETTINGS------
INFO: 	{
    "PRODUCTION": true,
    "IS_DEMO": false,
    "API_PORT": 9000,
    "API_DOCS": true,
    "DATABASE_TYPE": "sqlite",
    "SQLITE_FILE": "/app/data/db/mealie_v0.4.0.sqlite",
    "DEFAULT_GROUP": "Home",
    "DEFAULT_EMAIL": "changeme@email.com",
    "TOKEN_TIME": 2
}
```

waywardone:
Docs updated to reflect the right ENV variable.

PR #338, opened 2021-04-22
Issue #337 - Update docs for user env

waywardone:
Update docs to reflect the env variable that is actually used.


hay-kot:
Hey, Thanks for filing the issue and follow up with a PR!

PR #341, opened 2021-04-22
Localize more dates and texts

sephrat:


sephrat:
BTW, is /frontend/src/pages/Admin/Backup/BackupCard.vue obsolete? It looks a lot like AvailableBackupCard and I couldn't find its uses.

frontend/src/utils/index.js
```diff
@@ -56,22 +17,6 @@ export default {
     const uniqueKey = `${item}-${index}`;
     return uniqueKey;
   },
-  getDateAsText(dateObject) {
```
hay-kot, line 59: Are you sure these don't do anything? Last time I tried to remove them, I remember them being used to do something in the Meal Plan components.

sephrat: They used to, but I made sure I replaced their uses everywhere in favor of $d function. It's a standard function in vue-i18n, more powerful and natively translated.
https://kazupon.github.io/vue-i18n/guide/datetime.html

hay-kot: Fancy! Thanks for all work on localization, it makes a huge difference.

PR #346, opened 2021-04-23
Fix infinite loop in meal planner randomizer

sephrat:
Infinite loop would happen if there were more days than recipes.

It would probably be better to enhance the algorithm or at least warn the user, but since it's an edge case it's low prio.


sephrat:
I forgot to mention that there is no regression on regular cases, the while loop actually was useless due to filteredRecipes being automatically updated everytime usedRecipes is populated, so the condition was always false (until the list got empty)

PR #347, opened 2021-04-23
App Bar Rewrite

hay-kot:
### Mobile 
- [x] App Bar Rewrite
    - [x] Consolidate App/Fab
    - [x] Top/Bottom Bar
- [x] Sidebar Rewrite
    - [x] Consolidate category/admin sidebar
    - [x] Make reactive / fix reactive language values 
- [x] Fab Rewrite

- [x] Misc Cleanup
    - [x] Fix Open/Close on Navigation to Recipes Pages
    - [x] New Icon for Menu
    - [x] Fix no images on some meals
    - [x] Fix Image display error for users
    - [x] Fix double press on add recipe fab in mobile
- [x] Mobile Search Bar


hay-kot:
@sephrat, Do you mind testing out the translations loading for the new sidebar on this branch? it *should* be reactive now. 🤞

sephrat:
It's still failing to initialize on first load. However if you convert the arrays to computed ones, it works! 

```
  computed: {
    baseMainLinks() { return [
      {
        icon: "mdi-home",
        to: "/",
        title: this.$t("page.home-page"),
      },
      {
        icon: "mdi-view-module",
        to: "/recipes/all",
        title: this.$t("page.all-recipes"),
      },
      {
        icon: "mdi-magnify",
        to: "/search",
        title: this.$t("search.search"),
      },
    ]},
```
As for the App bar, I like the new bay better. It makes it more obvious to know where to click to show options.
Sticking it to the bottom can be a good idea (Firefox way!), however the search needs to expand to the top in that case. At the moment it's expanding out of view.
Screenshot_20210425-085012_Firefox


Also, unrelated to these changes but this made me realise one thing: the top-right icon image is quite confusing, you're expecting something about your user (logout/profile), but you can actually find the meal planner and the site options there. This and the fact that the sidebar items differ if you're on the home page or if you're in settings, makes me realize there are two places where you can browse through the pages, and you need to remember where each of these are hidden. 
One possible option could be to move some options to the sidebar besides the user avatar, and top-right navigation could be only about the meal planner? There's probably a better idea but it can be a start.

hay-kot:
> It's still failing to initialize on first load. However if you convert the arrays to computed ones, it works!

Now that you point that out, it's become very obvious why this hasn't been working. I've updated it and it should now work as expected.

> however the search needs to expand to the top in that case. At the moment it's expanding out of view.

I took a different approach. There's not a "search dialog" that overtakes the screen on mobile that given a better view and resolves the issue at hand.

Screen Shot 2021-04-25 at 12 51 20 PM

> One possible option could be to move some options to the sidebar besides the user avatar, and top-right navigation could be only about the meal planner?

I'm trying not to overcrowd the sidebar too much and keep it focused on recipe content. My thought with the user icon is that all the menu items that appear are derived from a user. If you aren't logged in there's nothing in that menu, if you are logged in you get links that are tied to your account. I think this is inline with a typical setup, for example the Icon in the upper right on Github has links for "Your Repos" and other content specific to your account.

PR #357, opened 2021-04-27
Other databases support + alembic

Diaoul:
Leverage on the ORM to provide support for multiple databases + Alembic

Quickly tested that the app starts with:
```bash
docker run --name postgresql --rm -e POSTGRESQL_USERNAME=mealie -e POSTGRESQL_PASSWORD=mealie -e POSTGRESQL_DATABASE=mealie -p 5432:5432 bitnami/postgresql:latest
export DB_URL=postgresql+psycopg2://mealie:mealie@localhost:5432/mealie
alembic upgrade head
uvicorn mealie.app:app --reload
```
:warning: Important notes:
- Some schemas needed to be modified due to FK constraints on mismatching types.
- I removed the sqlite database versioning as this is incompatible with all the other databases, remove the old one and relaunch the app will recreate the new database. Alembic does this.
- Default path for sqlite database has been modified to go to data directory.

More testing is needed, feedback is welcome.

Suggestions out of scope for this PR:
- Use a data directory for persistence only (logs, database, user uploads, exports, imports) currently there is also resized images I think and a few other things that can be regenerated? I am not sure. This would make the data volume something that we might want to backup without taking up too much disk space with ephemeral things or pictures that can be regenerated.
~~- Alembic for smoother database migrations (although major changes to the schemas need to be handled manually through some alembic migration scripting)~~


hay-kot:
> I removed the sqlite database versioning as this is incompatible with all the other databases, remove the old one and relaunch the app will recreate the new database.

The upgrade process for users would be too SSH into their server and manually remove the database before restarting that container? That does too much to reduce usability by the end user. 

> Default path for sqlite database has been modified to mealie.db relative path.

If you use a relative path from a python file the database will be not be stored in persistent storage.

Diaoul:
> The upgrade process for users would be too SSH into their server and manually remove the database before restarting that container? That does too much to reduce usability by the end user.

How do you currently handle database migration? Is it not just creating a new empty database? If so, we can detect the schema has changed, probably create_all will fail, catch that and do the removing in the startup script.
Basically the idea is to make the code database agnostic and handle that at the user level.
This is a temporary issue anyways if alembic comes to the rescue.

> If you use a relative path from a python file the database will be not be stored in persistent storage.

I can fix that for docker.

hay-kot:
> How do you currently handle database migration? Is it not just creating a new empty database? If so, we can detect the schema has changed, probably create_all will fail, catch that and do the removing in the startup script.
> Basically the idea is to make the code database agnostic and handle that at the user level.
> This is a temporary issue anyways if alembic comes to the rescue.

There is no migration, users us the built in import/export functionality during upgrades, but the other database isn't deleted so if there's a problem the user can always roll back to the previous version. I'm weary about arbitrarily removing user data during startup. Alembic would solve this, but until then we need a workable solution. 

> I can fix that for docker.

Just to clarify, this isn't a docker issue. You should really be setting an absolute path to the data volume as it was before. With the additional database settings, I think a good approach would be to fall back to SQLite if not PG configuration is not detected. Then you could use the existing programming to set the absolute path of the database. You could also accomplish this with a class @property or a validator.

hay-kot:
I don't think the Alembic script is setting up relationships appropriately. When I run from a fresh database I can't login, and get the following error

```log
  File "/Users/hayden/Projects/mealie/.venv/lib/python3.9/site-packages/uvicorn/protocols/http/httptools_impl.py", line 398, in run_asgi
    result = await app(self.scope, self.receive, self.send)
  File "/Users/hayden/Projects/mealie/.venv/lib/python3.9/site-packages/uvicorn/middleware/proxy_headers.py", line 45, in __call__
    return await self.app(scope, receive, send)
  File "/Users/hayden/Projects/mealie/.venv/lib/python3.9/site-packages/uvicorn/middleware/debug.py", line 81, in __call__
    raise exc from None
  File "/Users/hayden/Projects/mealie/.venv/lib/python3.9/site-packages/uvicorn/middleware/debug.py", line 78, in __call__
    await self.app(scope, receive, inner_send)
  File "/Users/hayden/Projects/mealie/.venv/lib/python3.9/site-packages/fastapi/applications.py", line 199, in __call__
    await super().__call__(scope, receive, send)
  File "/Users/hayden/Projects/mealie/.venv/lib/python3.9/site-packages/starlette/applications.py", line 111, in __call__
    await self.middleware_stack(scope, receive, send)
  File "/Users/hayden/Projects/mealie/.venv/lib/python3.9/site-packages/starlette/middleware/errors.py", line 181, in __call__
    raise exc from None
  File "/Users/hayden/Projects/mealie/.venv/lib/python3.9/site-packages/starlette/middleware/errors.py", line 159, in __call__
    await self.app(scope, receive, _send)
  File "/Users/hayden/Projects/mealie/.venv/lib/python3.9/site-packages/starlette/exceptions.py", line 82, in __call__
    raise exc from None
  File "/Users/hayden/Projects/mealie/.venv/lib/python3.9/site-packages/starlette/exceptions.py", line 71, in __call__
    await self.app(scope, receive, sender)
  File "/Users/hayden/Projects/mealie/.venv/lib/python3.9/site-packages/starlette/routing.py", line 566, in __call__
    await route.handle(scope, receive, send)
  File "/Users/hayden/Projects/mealie/.venv/lib/python3.9/site-packages/starlette/routing.py", line 227, in handle
    await self.app(scope, receive, send)
  File "/Users/hayden/Projects/mealie/.venv/lib/python3.9/site-packages/starlette/routing.py", line 41, in app
    response = await func(request)
  File "/Users/hayden/Projects/mealie/.venv/lib/python3.9/site-packages/fastapi/routing.py", line 201, in app
    raw_response = await run_endpoint_function(
  File "/Users/hayden/Projects/mealie/.venv/lib/python3.9/site-packages/fastapi/routing.py", line 150, in run_endpoint_function
    return await run_in_threadpool(dependant.call, **values)
  File "/Users/hayden/Projects/mealie/.venv/lib/python3.9/site-packages/starlette/concurrency.py", line 34, in run_in_threadpool
    return await loop.run_in_executor(None, func, *args)
  File "/opt/homebrew/Cellar/python@3.9/3.9.4/Frameworks/Python.framework/Versions/3.9/lib/python3.9/concurrent/futures/thread.py", line 52, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/Users/hayden/Projects/mealie/mealie/routes/users/auth.py", line 26, in get_token
    user = authenticate_user(session, email, password)
  File "/Users/hayden/Projects/mealie/mealie/core/security.py", line 30, in authenticate_user
    user: UserInDB = db.users.get(session, email, "email", any_case=True)
  File "/Users/hayden/Projects/mealie/mealie/db/db_base.py", line 96, in get
    return self.schema.from_orm(result[0])
  File "/Users/hayden/Projects/mealie/.venv/lib/python3.9/site-packages/pydantic/main.py", line 621, in from_orm
    obj = {ROOT_KEY: obj} if cls.__custom_root_type__ else cls._decompose_class(obj)
  File "/Users/hayden/Projects/mealie/.venv/lib/python3.9/site-packages/pydantic/main.py", line 741, in _decompose_class
    return cls.__config__.getter_dict(obj)
  File "/Users/hayden/Projects/mealie/mealie/schema/user.py", line 57, in getter_dict
    "group": ormModel.group.name,
AttributeError: 'NoneType' object has no attribute 'name'
INFO: 	127.0.0.1:0 - "GET /api/users/self HTTP/1.1" 500
ERROR: 	Exception in ASGI application
  File "/Users/hayden/Projects/mealie/.venv/lib/python3.9/site-packages/uvicorn/protocols/http/httptools_impl.py", line 398, in run_asgi
    result = await app(self.scope, self.receive, self.send)
  File "/Users/hayden/Projects/mealie/.venv/lib/python3.9/site-packages/uvicorn/middleware/proxy_headers.py", line 45, in __call__
    return await self.app(scope, receive, send)
  File "/Users/hayden/Projects/mealie/.venv/lib/python3.9/site-packages/uvicorn/middleware/debug.py", line 81, in __call__
    raise exc from None
  File "/Users/hayden/Projects/mealie/.venv/lib/python3.9/site-packages/uvicorn/middleware/debug.py", line 78, in __call__
    await self.app(scope, receive, inner_send)
  File "/Users/hayden/Projects/mealie/.venv/lib/python3.9/site-packages/fastapi/applications.py", line 199, in __call__
    await super().__call__(scope, receive, send)
  File "/Users/hayden/Projects/mealie/.venv/lib/python3.9/site-packages/starlette/applications.py", line 111, in __call__
    await self.middleware_stack(scope, receive, send)
  File "/Users/hayden/Projects/mealie/.venv/lib/python3.9/site-packages/starlette/middleware/errors.py", line 181, in __call__
    raise exc from None
  File "/Users/hayden/Projects/mealie/.venv/lib/python3.9/site-packages/starlette/middleware/errors.py", line 159, in __call__
    await self.app(scope, receive, _send)
  File "/Users/hayden/Projects/mealie/.venv/lib/python3.9/site-packages/starlette/exceptions.py", line 82, in __call__
    raise exc from None
  File "/Users/hayden/Projects/mealie/.venv/lib/python3.9/site-packages/starlette/exceptions.py", line 71, in __call__
    await self.app(scope, receive, sender)
  File "/Users/hayden/Projects/mealie/.venv/lib/python3.9/site-packages/starlette/routing.py", line 566, in __call__
    await route.handle(scope, receive, send)
  File "/Users/hayden/Projects/mealie/.venv/lib/python3.9/site-packages/starlette/routing.py", line 227, in handle
    await self.app(scope, receive, send)
  File "/Users/hayden/Projects/mealie/.venv/lib/python3.9/site-packages/starlette/routing.py", line 41, in app
    response = await func(request)
  File "/Users/hayden/Projects/mealie/.venv/lib/python3.9/site-packages/fastapi/routing.py", line 191, in app
    solved_result = await solve_dependencies(
  File "/Users/hayden/Projects/mealie/.venv/lib/python3.9/site-packages/fastapi/dependencies/utils.py", line 548, in solve_dependencies
    solved = await call(**sub_values)
  File "/Users/hayden/Projects/mealie/mealie/routes/deps.py", line 32, in get_current_user
    user = db.users.get(session, token_data.username, "email", any_case=True)
  File "/Users/hayden/Projects/mealie/mealie/db/db_base.py", line 96, in get
    return self.schema.from_orm(result[0])
  File "/Users/hayden/Projects/mealie/.venv/lib/python3.9/site-packages/pydantic/main.py", line 621, in from_orm
    obj = {ROOT_KEY: obj} if cls.__custom_root_type__ else cls._decompose_class(obj)
  File "/Users/hayden/Projects/mealie/.venv/lib/python3.9/site-packages/pydantic/main.py", line 741, in _decompose_class
    return cls.__config__.getter_dict(obj)
  File "/Users/hayden/Projects/mealie/mealie/schema/user.py", line 57, in getter_dict
    "group": ormModel.group.name,
AttributeError: 'NoneType' object has no attribute 'name'

```

mealie/db/models/db_session.py
```diff
@@ -1,16 +1,16 @@
...
 
 def sql_global_init(db_url: str):
+    thread_safe = True
+    if "sqlite" in db_url:
+        thread_safe = False
+
     engine = sa.create_engine(
         db_url,
         echo=False,
+        connect_args={"check_same_thread": thread_safe},
```
Diaoul, line 13: This doesn't work for other databases as check_same_thread is not an option.

wengtad:
Ermm.. this will eventually merge to dev branch first right? I saw it is now pointing to master branch.
Btw great job!

Diaoul:
This PR is bigger than I wanted it to be but I think it is in a working state atm.
I had to disable the reload option as it was constantly reloading due to logs and db activity. I don't know how it is supposed to work with data being under /app.
I suggest this can be moved to /data on the docker image, this would avoid this kind of conflicts

mealie/schema/settings.py
```diff
@@ -15,18 +15,6 @@ class SiteSettings(CamelModel):
     class Config:
         orm_mode = True
 
-        schema_extra = {
```
hay-kot, line 18: Why did you remove these?

mealie/run.sh
```diff
@@ -7,20 +7,18 @@ ARG1=${1:-production}
...
+poetry run alembic upgrade head
 poetry run python /app/mealie/services/image/minify.py
 
-# Migrations
 # TODO
-    # Migrations
-    # Set Port from ENV Variable
+# Set Port from ENV Variable
 
-if [[ "$ARG1" = "reload" ]]
```
hay-kot, line 18: If the application is constantly reloading you need to properly setup you development environment by setting the ENV variables, not remove the code for development. 

```env
# Determines Production Mode, This will set the directory path to use for data storage
PRODUCTION=False
```

template.env
```diff
@@ -3,7 +3,7 @@ DEFAULT_GROUP=Home
 ENV=False
 API_PORT=9000
 API_DOCS=True
-DB_TYPE=sqlite
+DB_URL=sqlite:///data/mealie.db
```
hay-kot, line 6: This should default to None and allow Pydantic to properly set the URL accordingly.

Dockerfile
```diff
@@ -1,54 +1,51 @@
+# build
```
hay-kot, line 1: Can you provide the difference in build times and/or the image size when compared to the Alpine image?

hay-kot (changes requested):
A couple additional things 

1. Tests don't run or pass with a SQL error in CI/CD.
```
E   sqlalchemy.exc.OperationalError: (sqlite3.OperationalError) no such table: groups
E   [SQL: SELECT groups.id AS groups_id, groups.name AS groups_name, groups.webhook_enable AS groups_webhook_enable, groups.webhook_time AS groups_webhook_time
E   FROM groups]
E   (Background on this error at: http://sqlalche.me/e/14/e3q8)
```

2. As a maintainer, what do I need to know/do to setup migrations in the future? I'm assuming I need to generate the migration files that are then shipped to the end user? 
3. The [Contributor Guide](https://hay-kot.github.io/mealie/contributors/developers-guide/code-contributions/) outlines that all PRs should be made against the `dev` branch.
4. Given the amount of code that have been added or changed I'd like to see some additional testing added to verify functionality. [pytest-alembic](https://pypi.org/project/pytest-alembic/) As I said, I will still look into running the tests against a Postgresql Database as well. 

Thanks again for all the hard work on this

mealie/schema/settings.py
Diaoul: Those are now part of the seed data in alembic migrations

mealie/run.sh
Diaoul: Not sure how this is supposed to work, I'll put it back.

Dockerfile
Diaoul: 1.09GB for mealie-api:dev
886MB for python:3.9-buster
not sure what previous sizes were
maybe slim could be used but I've had some troubles with it: reinstalling build tools and dev libraries increase pretty much to the size of regular debian

mealie/schema/settings.py
hay-kot: > Those are now part of the seed data in alembic migrations

These are for providing documentation for the API. 

https://fastapi.tiangolo.com/tutorial/schema-extra-example/

Diaoul:
1. Should be better but CI is blocked for me due to the limitation you put:
>  First-time contributors need a maintainer to approve running workflows. Learn more. 
2. This is regular alembic, for more details I would suggest the docs but I used `alembic revision --autogenerate -m "Initial revision"`. You should make changes to the schemas and run this command again with a different message it will generate the migration for you. Check everything is OK and that's it. Run `alembic migrate head` to apply to your db.
3. I will rebase
4. I fear this goes beyond what I can offer with this PR which goal is to provide the groundwork for multi database support and minimal changed for end users on sqlite (well, initially at least.. :stuck_out_tongue_winking_eye:)

hay-kot:
I haven't been able to get the app working from this branch. I think there's just too many changes to track overall to narrow down where each error is occurring. I'm going to grab code from this PR and incorporate them in smaller chunks into others. I'll start with the docker image changes, then support for additional databases, and then add alembic. I've started with the docker image over at  PR #369.  

I'll try keep this updated on what I've been able to accomplish...
- [ ] Docker Changes
    - [x] Switched to Debian base image
    - [x] Fixed package installation on production image
    - [x] switched to buster-slim
    - [x] Additional cleanup after installation
    - [ ] Add UID/GUID support 

Thanks again!

Diaoul:
What is the error you're getting? I can run `make docker-dev` and get it to work. For testing without docker you can run:
```bash
alembic upgrade head
make test
```
Running the alembic migrations first is necessary. This is due to a preexisting bug related to the way import is done in the app, actual code is run before test settings are applied, hence the regular database is needed. Previous code created the db on the fly (without alembic) so this was invisible:
In dev branch try yourself:
```bash
$ rm dev/data/db/*
$ poetry run pytest
$ ls dev/data/db/
mealie_v0.5.0.sqlite  # this should not be there
```

Yes this PR grew too much due to various dependencies issue: fixes in the schemas, then alembic, then docker image, then the test bug... Walking back that path seems reasonable.

hay-kot:
> What is the error you're getting? I can run make docker-dev and get it to work. For testing without docker you can run:

There are several. The production image no longer runs with the changes with the new docker due to poetry not installing the app package. When running on the base OS I couldn't get it to run either with errors similar to those SQL errors above, and when it did run the database wouldn't be setup correctly and I couldn't login. 

> Running the alembic migrations first is necessary. This is due to a preexisting bug related to the way import is done in the app,

I've *fixed* this for the time being by setting the database path prior to importing the rest of the files in conftest and not starting the background scheduler during tests. Not ideal, but it works for now.

Diaoul:
Issues are fixed. As for the local run, did you run `alembic upgrade head`? What is the output? You should have 
```
INFO  [alembic.runtime.migration] Context impl SQLiteImpl.
INFO  [alembic.runtime.migration] Will assume non-transactional DDL.
INFO  [alembic.runtime.migration] Running upgrade  -> 89e8733c36f2, Initial revision
```

hay-kot:
UPDATE: 

Postgres support has been added as apart of PR #371 including integration with CI/CD for automated testing.

PR #358, opened 2021-04-27
More localization

sephrat:
- [X] A few strings missing here and there
- [X] Localize import UIs
- [x] Localize sign-ups UIs
- [x] Localize toolbox
- [X] Handle API error texts in frontend instead of backend ⚠️ May cause side-effects for people directly consuming the API
- [X] Better status codes from API
- [X] Warning message when trying to display 'dinner this week' or 'dinner today' when no meal is planned


frontend/src/components/Login/SignUpForm.vue
```diff
@@ -132,18 +133,20 @@ export default {
         admin: false,
       };
 
-      let successUser = false;
       if (this.$refs.signUpForm.validate()) {
         let response = await api.signUps.createUser(this.token, userData);
-        successUser = response.snackbar.text.includes("Created");
+        if (response.status != 200) {
```
hay-kot, line 138: What a much better way of handling responses! 

Could we extract this functionality into a function in utils? Then we can call this from the API calls directly so every time we use `api.recipes.delete(var)` to get the response processing for free. We also don't have to write if/else checks in every component that makes an API call. 

```js
function processResponse(response, successCode, successText, failText) {
  if (response.status != successCode) {
    utils.notify.error(successText);
  } else {
    utils.notify.success(failText);
  }
}
```

sephrat: I see what you mean. I wanted to do this but couldn't wrap my head around how to do this while allowing for specific messages based on error codes. 
I eventually got an idea: https://github.com/sephrat/mealie/commit/e92b55feb88bf3c3cf747ff91558af49d43b0658
Let me know what you think before I refactor the rest of the API calls.

hay-kot: Passing a function to get the text looks like a great option. Thanks for all your hard work!

PR #381, opened 2021-05-01
Remove former profile image upon new upload

sephrat:
When uploading a new profile image with a different extension, the old image is not removed.
Looks like it was supposed to be handled but it silently fails.


sephrat:
This made me realise python tests use the dev persisted folder. Would be great if we could make it use a temporary folder for tests.

hay-kot:
> This made me realise python tests use the /dev/data persisted folder. Would be great if we could make it use a temporary folder for tests.

Yeah, this isn't ideal. Unfortunately, I'm kind of stuck on how to properly do this. The code that sets the data directory runs on import [See config.py](https://github.com/hay-kot/mealie/blob/212ca64e276bb4904fdcf6333afb60877182bd64/mealie/core/config.py#L44). The same is done for the app_settings variable [See config.py](https://github.com/hay-kot/mealie/blob/212ca64e276bb4904fdcf6333afb60877182bd64/mealie/core/config.py#L134). 

Along the same lines I'm trying to resolve an issue with being able to run tests against a Postgres database but am running into similar issues with the database being initialized on import, so when you run the tests it populates a SQLite database. I've sort of got this working by overriding the database URL with a [pre_start.py](https://github.com/hay-kot/mealie/blob/feature/additional-db-uri-support/tests/pre_test.py) that I import prior to testing.

Unfortunately, I'm having a hard time finding any alternatives to running the code at import. If you have any ideas on how to approach this, I'd love to figure this out.

PR #422, opened 2021-05-23
Add web manifest

pR0Ps:
This provides a more app-like experience for mobile users that use the "add to homescreen" functionality of their web browsers.

### DIFF
--- a/frontend/public/index.html
+++ b/frontend/public/index.html
@@ -5,6 +5,8 @@
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <meta name="viewport" content="width=device-width,initial-scale=1.0">
     <link rel="icon" href="<%= BASE_URL %>favicon.ico">
+    <link rel="apple-touch-icon" sizes="256x256" href="<%= BASE_URL %>favicon.ico">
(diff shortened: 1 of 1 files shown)


hay-kot:
Cool! Is that all it takes for basic PWA functionality?

Can you change the target branch to dev? That's where we stage all the changes for the next version.

Thanks!

pR0Ps:
Yep! With this change modern mobile browsers that support PWAs will open the site in a separate view and hide the normal browser chrome, making it look and feel much more native.

There are more properties that can be added to the manifest (see https://developer.mozilla.org/en-US/docs/Web/Manifest ), but the minimal set I've added gives you most of the functionality (title, icon, opening external pages in a normal web browser).

I've changed the PR to target the dev branch

zierbeek:
This isnt the same as a PWA i think. There is a difference between 'add to homepage' and 'install the app'

pR0Ps:
I'm not that knowledgeable when it comes to what constitutes a PWA so you might be right.

Functionally what this change does is make it so when a mobile user uses their browsers "add to home screen" function, the icon on the home screen has the Mealie icon and name. Tapping that icon will open the website, but in a chrome-less browser without any other tabs that makes it look and feel more like a normal installed application.

zierbeek:
Screenshot_20210523-220649_One UI Home

Here you can see some apps that are real progressive web apps and ones that are just an 'add to home screen' 
I'm not totally sure how they work behind the scenes. The ones without chrome logo are the real pwa's which are installed

zierbeek:
https://developers.google.com/web/updates/2015/12/getting-started-pwa

PR #427, opened 2021-05-25
PWA attempt 

zierbeek:
All of the editing is done in the frontend folder

Create manifest.json
create main.js to call service-worker
create sw.js, the service worker itself
Folder for different image sizes (still needs a high quality upload of images)
Edit index.html with ref to images, manifest and sw.

Problem: PWA works when running it locally, localhost. When creating it with docker-prod, it can't find the manifest and such, can't find the sw.


hay-kot:
Hey! Thanks for all the work on this. Part of my github workflow is to use the `dev` branch for all pre-release code. So anything that is a feature and not a critical bug fixes goes into the `dev` branch. That means when you start a PR or start working on a potential PR you'll want to select the `dev` branch as your base. That way you don't have to sort out of a bunch of weird conflicts later down the line. 

This isn't a huge deal, but similar to your docs updates if you could reset you base to the `dev` branch so you can capture the latest changes that would be extremely helpful. Hopefully I can dig into this this weekend 👍

PR #434, opened 2021-05-26
First try making mealie a PWA

zierbeek:
All of the editing is done in the frontend folder

Create manifest.json
create main.js to call service-worker
create sw.js, the service worker itself
Folder for different image sizes (still needs a high quality upload of images)
Edit index.html with ref to images, manifest and sw.

Problem: PWA works when running it locally, localhost. When creating it with docker-prod, it can't find the manifest and such, can't find the sw.


zierbeek:
The other one is closed :)

wengtad:
Hi, I had created my version of PWA in Vue way. It's support install app to desktop and "Add to Home screen" as well. You may have a look on my fork repo [here](https://github.com/wengtad/mealie/tree/pwa).

From what I know about PWA only works on HTTPS but no issue if using localhost. Correct me if I'm wrong. Happy coding.

zierbeek:
Hi thanks! It works mainly. Just not when hosted, it can't seem to find thé manifest. Maybe you can take a look?

And die you do it with thé vua PWA plugin? I tried it with that but no succes :)

wengtad:
I just try to hosted my version, it is working fine.
Is your result like mine below?
splash
app

zierbeek:
If your way works, I think we should take your way :) does it cache offline? If not, you can grab a part of mine.

PR #437, opened 2021-05-26
Support PWA

wengtad:
Add support for PWA.

- Make Mealie installable on mobile or desktop via PWA way.
- "Add to Home screen", create a shortcut of your Mealie on your mobile.
- PWA offline cache, thanks to @zierbeek

Tested on devices/platforms:
- PC, Windows 10 using Brave browser
- Mobile, Android 9 using Brave browser

I don't have any Apple's product on hand so will need someone to help to test it.


wengtad:
BTW, @hay-kot , I notice the latest changes is breaking the DB.
Hit `no such column: users.username` on sqlite or `column users.username does not exist` on postgres. Not sure you aware on it or not.
Since it is in development so I just delete the DB and let Mealie create again. Is that any DB migration without manual work?

hay-kot:
> BTW, @hay-kot , I notice the latest changes is breaking the DB.
> Hit `no such column: users.username` on sqlite or `column users.username does not exist` on postgres. Not sure you aware on it or not.
> Since it is in development so I just delete the DB and let Mealie create again. Is that any DB migration without manual work?

I still haven't implemented any sort of database migration so the approach for development and production is to backup, delete your database and import. I'd like to get alembic working, but if I keep adding things to this release I'll never get it out the door.

PR #450, opened 2021-05-30
Bug/general fixes

hay-kot:
- [x] Fix No Mealplan Returned error
- [x] Fix Asset links
- [x] Dates off in UI For Meal Planner
- [x] Profile Image Fix


hay-kot:
@sephrat I'm running into an issue with the localization for the meal-planner. On the demo-site it's displaying the incorrect dates when creating a meal-plan. I was able to fix the quick-week generation but it looks like the localization of the date is messing with the display if it's off from the UTC date. 

I'm kind of stumped on this one, any ideas?

sephrat:
Ah so this timezone issue is back. I had to do some dirty hack here for the quick week to work on the right side of the globe: https://github.com/hay-kot/mealie/blob/2b97af572850b2fe100031f870837640d7bc37f6/frontend/src/components/MealPlan/MealPlanNew.vue#L221-L221
I'm not sure what's the right approach to fix this. [This SO question](https://stackoverflow.com/questions/29174810/javascript-date-timezone-issue) describes the issue we're facing: actualStartDate and actualEndDate are instantiated in UTC but all subsequent usages consider the user's timezone, so it's displaying the day before for people on the left side of the globe. 
The dirty fix of setting the time to noon solves it. A better approach could be to use a dedicated library but that's probably overkill for our use.

hay-kot:
Awesome. That really cleared things up for me. I *think* I worked out a hack around the UTC problem by replacing the "-" with "/". Which is the first work around in SO question you listed. My understanding of javascript dates is pretty low, so you'll have to confirm for me that that change doesn't break you guys on the other side!

PR #505, opened 2021-06-13
fix docs

wengtad:


makefile
```diff
@@ -30,37 +30,37 @@ purge: clean ## ⚠️  Removes All Developer Data for a fresh server start
...
 	poetry run pytest
 
-lint: ## 🧺 check style with flake8
+lint: ## 🧺 Check style with flake8
 	poetry run black .
 	poetry run black . --check
 	poetry run flake8 mealie tests
 
-coverage: ## ☂️  check code coverage quickly with the default Python
+coverage: ## ☂️ Check code coverage quickly with the default Python
```
hay-kot, line 57: Some of these had extra spaces because when rendered in the terminal they are not consistent, at least for me. Do these render equally spaced for you?

wengtad: oh, I will add it back

PR #509, opened 2021-06-13
handle recipeInstructions that are given as json objects / name-value…

tklingenberg8472:
I had trouble migrating recipes from nextcloud cookbook. I split down the zip that I tried to import so that each zip file contains a single recipe. I found that some recipes could be migrated like that and some could not. Those that could not be migrated use a different format for the recipe instructions like so:
`"recipeInstructions":{"0":"Einen Topf ...","2":"In der Zwischenzeit ...", "6":"Auf dem Grill ..."}`

I made this change directly on my docker container. It works for me. I hope it breaks no other migrations.

### DIFF
--- a/mealie/services/scraper/cleaner.py
+++ b/mealie/services/scraper/cleaner.py
@@ -70,6 +70,10 @@ class Cleaner:
         if not instructions:
             return []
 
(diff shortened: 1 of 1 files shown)


hay-kot:
Looks like a nice fix, thanks!

PR #526, opened 2021-06-16
Fix Users API Token API path 

sephrat:
Not sure if I did it right since it says the files are auto-generated, but I couldn't find the process that generates them and my changes were not overwritten by make frontend or make backend.


wengtad:
how about `make code-gen`, does it work?

sephrat:
Yes that's the one! Thanks for pointing it out. This does break my change*, so my fix is not the proper one, although it does fix the issue on the frontend. 
In that case there must be an error in the script. It doesn't like that the general prefix /api is reused in the rest of the route (/users/api-tokens) and drops it. If I rename the route from api-tokens to whatever-tokens, then it works. I'll look into it.

\* The JS file I changed here does not get overwritten. The script writes files to scripts/output/javascriptAPI. Not sure what's the workflow from there. @hay-kot do you manually copy files from that directory to frontend/src/api?

hay-kot:
The code-gen is very much an experiment of me not wanting to write a bunch of api route boilerplate. It's definitely experimental and I don't expect anyone to go through it and fix it. If there's something wrong with the generation you can just flag it, and manually make the changes if you don't want to get into editing the templates and such or go through that mess. Looks like you got it figured out though, so that's awesome! 

That said, my work flow is to copy verify the routes after generating and then copy them over. I don't trust it enough yet to overwrite the existing files automatically. As for the other js files, I'm only using those to grab new routes that are created from fast API and bootstrap creating the API call functions.

PR #529, opened 2021-06-17
dev-feature/analyze-bundle

hay-kot:
Bundle size from 1.7mb -> 1.45mb.
Removed 2mb extra css from material design icons


wengtad:
after this onwards, new icon will have to define at `globals.js` first before use?

PR #561, opened 2021-06-18
Lazy load translations

### LINKED ISSUE #527: [Task] Lazy Load Localization
Description: Lazy load localizations to reduce initial bundle size -> [See here for details on how](https://kazupon.github.io/vue-i18n/guide/lazy-loading.html)

sephrat:
- [x] Lazy load messages
- [x] Lazy load dateTimeFormats

I'm a bit puzzled with the results I've got so far. I followed the documentation but the webpack plugin still shows every language file, only regrouped differently. 
image

Am I reading this wrong, or does this make no difference on the bundle size? I'm open to feedback.

Fixes #527

### DIFF
--- a/frontend/src/i18n.js
+++ b/frontend/src/i18n.js
@@ -1,5 +1,7 @@
 import Vue from "vue";
 import VueI18n from "vue-i18n";
+import Vuetify from "@/plugins/vuetify";
+import axios from 'axios';
 
 Vue.use(VueI18n);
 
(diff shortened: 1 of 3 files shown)


hay-kot:
>  the webpack plugin still shows every language file, only regrouped differently

I'm pretty sure this is correct. All the files are still apart of the entire bundle, but lazy loading them splits them into separate payloads so each payload is only downloaded as required by the browser. When I was doing them before, they were shown as apart of the app.js bundle so they'd be apart of the initial js payload.

When you run the build command it should print to console what the initial load bundle size is, I was getting 1.48mb before. Should be small now I'd assume.

sephrat:
Makes sense, thanks for the clarification. I can indeed see that the app bundle size has been reduced from 1.48MiB to 1MiB, and down to 1020KiB with dateTimeFormats (this was mostly for fun 😄 ).

hay-kot:
Whooo! Checkout that bundle. Almost a reasonable size, 😁. Thanks for all the work on this!

PR #571, opened 2021-06-18
API security hardening

sephrat:
- [x] Prevent a regular user from promoting themself to admin
-  [x] Prevent an admin from demoting themself
- [x] Refactor token fixture to admin + regular user tokens
- [x] More security-related tests based on user rights
- [x] Secure admin routes
  - [x] Events
  - [x] Groups
  - [x] Category and tag update and deletion 
  - [x] Debug
  - [x] Migration
  - [x] Backup
  - [x] Site settings
  - [x] Sign up
- [x] Secure regular user routes
  - [x] Shopping list
  - [x] Themes
  - [x] User favorites
- [x] Refactor API routers into dedicated public/user/admin routers
- [x] Frontend non-regression tests


sephrat:
I'm not very happy with the poor factorization of [function 'login' ](https://github.com/hay-kot/mealie/pull/571/files#diff-e52e4ddd58b7ef887ab03c04116e676f6280b824ab7469d5d3080e5cba4f2128R52) which is due to my poor Python skills. It's not that bad but I wanted this function to have a lighter signature. I couldn't get my head around how to pass parameters from/to fixtures.
I believe it's acceptable as it is, but I'd be happy to learn a better way to do this.

sephrat:
I think I've tested everything on the frontend to make sure there is no regression.
One thing that's 'broken' is that a regular user cannot update the mealplan settings nor test webhooks anymore. I believe it's a good thing but the UI still lets you click on the buttons and you get a 403. This pull request is big enough as it is, I'll leave it as is for now.

hay-kot:
Awesome! Love all of this, it all looks really solid thanks for all the work on this. I agree with the comments on the group section. I think we may need another category of users that are group-admins. That way they can make changes to Group settings but not operate on the entire instance. We could even allow them to generate invites or something of the like. Something to look at later for sure. Thanks!

PR #572, opened 2021-06-18
Printview columns

zierbeek:
Add multiple columns depending on list length.


hay-kot:
Had to revert. I didn't catch it at first, but when printing (Firefox) I wouldn't get all the ingredients. they would overflow outside of the browser window. can you take another look at this?

zierbeek:
Aha! Will take a look and maybe use the grid approach in stead of max height

Issue #550, opened 2021-06-17
[Task] Add Gunicorn and Webworks to Dockerfile


hay-kot:
Description: Add Gunicorn and webworkers to dockerfile for improved performance

https://gunicorn.org/

Issue #429, opened 2021-05-25
Not able to run correctly in docker if user permissions are specified


RichardMore:
**Describe the bug**
When I try to run the container as a specific user due to folder permissions there are errors and login issues caused by the user the container runs as cannot access folders/files
Tried all run as user options, the preferred one would be like the linuxserver.io docker images PUID and PGID

**Steps To Reproduce**
Please be specific!
1. Run docker image as user specified
2. See logs and try to log in as well - it will fail

**Docker compose section**
  mealie:
    image: hkotel/mealie:latest
    user: 1000:1000
    ports:
      - 80:80
    environment:
      db_type: sqlite
      TZ: America/New_York
      UID: 1000
      PUID: 1000
      GUID: 1000
      PGID: 1000

**Expected behavior**
Create files and processes as the specified user

**Actual Behavior**
Errors

**Docker-compose logs**
Attaching to mealietest
mealietest   | /app/mealie/run.sh: line 10: poetry: Permission denied
mealietest   | /app/mealie/run.sh: line 11: poetry: Permission denied
mealietest   | Production
mealietest   | {"level":"info","ts":1621958901.1885612,"msg":"using provided configuration","config_file":"./Caddyfile","config_adapter":""}
mealietest   | {"level":"warn","ts":1621958901.1905532,"logger":"admin","msg":"admin endpoint disabled"}
mealietest   | {"level":"info","ts":1621958901.1908958,"logger":"tls.cache.maintenance","msg":"started background certificate maintenance","cache":"0xc000380a80"}
mealietest   | {"level":"error","ts":1621958901.1919827,"msg":"unable to create folder for config autosave","dir":"/.config/caddy","error":"mkdir /.config: permission denied"}
mealietest   | {"level":"info","ts":1621958901.1920056,"msg":"serving initial configuration"}
mealietest   | {"level":"info","ts":1621958901.1921408,"logger":"tls","msg":"cleaned up storage units"}
mealietest   | Successfully started Caddy (pid=15) - Caddy is running in the background
mealietest   | INFO:    Adding job tentatively -- it will be properly scheduled when the scheduler starts
mealietest   | INFO:    ----INIT SCHEDULE OBJECT-----
mealietest   | INFO:    Adding job tentatively -- it will be properly scheduled when the scheduler starts
mealietest   | Pending jobs:
mealietest   |     update_webhook_schedule (trigger: interval[0:30:00], pending)
[...]

PR #575, opened 2021-06-19
Optimize Docker

### LINKED ISSUE #350: Initialization scripts fail to run when not executing as root user inside container
Thanks so much for the work you do on mealie!

**Describe the bug**
If a user id and group id other than root are set for the mealie process it will fail to run some startup scripts. I noticed this when starting v0.4.3 in a clean environment when it failed to run the `init_db.py`. I had been running the 0.3.0 version of mealie with a non-root user.

I know the docs don't mention this as supported so this may actually be a feature request. Anyways the solution is probably rather simple. I think the `init_db.py` and `minify.py` files need to be marked executable for "other" and the `/.config` folder also needs to be accessible to the alternate user. It may also be good to check the exit code of things necessary for startup and fail fast with a message rather than start the process half-baked.

**Steps To Reproduce**
Run mealie in docker with a user other than root.

**Sample Code**
My docker compose file:

```
version: "3.1"
services:
  mealie:
    container_name: mealie
    image: hkotel/mealie:v0.4.3
    user: "989:988"
    environment:
      db_type: sqlite
    volumes:
      - <dir owned by mealie user>:/app/data
    ports:
      - 9000:80
    restart: unless-stopped
```

**Expected behavior**
The mealie image could be started and run as a user other than root.

**Actual Behavior**
Mealie technically starts but it's clear things are not healthy. Here's a sampling of my docker logs:

```
$ docker logs -f mealie
/app/mealie/run.sh: line 10: poetry: Permission denied
/app/mealie/run.sh: line 11: poetry: Permission denied
Production
{"level":"info","ts":1619286430.7026923,"msg":"using provided configuration","config_file":"./Caddyfile","config_adapter":""}
{"level":"warn","ts":1619286430.7050028,"logger":"admin","msg":"admin endpoint disabled"}
{"level":"error","ts":1619286430.7062602,"msg":"unable to create folder for config autosave","dir":"/.config/caddy","error":"mkdir /.config: permission denied"}
{"level":"info","ts":1619286430.706292,"msg":"serving initial configuration"}
{"level":"info","ts":1619286430.7065897,"logger":"tls","msg":"cleaned up storage units"}
{"level":"info","ts":1619286430.7066443,"logger":"tls.cache.maintenance","msg":"started background certificate maintenance","cache":"0xc000360bd0"}
Successfully started Caddy (pid=14) - Caddy is running in the background
...
INFO:   ------APP SETTINGS------
INFO:   {
    "PRODUCTION": true,
    "IS_DEMO": false,
    "API_PORT": 9000,
    "API_DOCS": true,
    "DATABASE_TYPE": "sqlite",
    "SQLITE_FILE": "/app/data/db/mealie_v0.4.0.sqlite",
    "DEFAULT_GROUP": "Home",
    "DEFAULT_EMAIL": "foo@example.com",
    "TOKEN_TIME": 8760
}
INFO:     Application startup complete.
INFO:   Application startup complete.
INFO:     Uvicorn running on http://0.0.0.0:9000 (Press CTRL+C to quit)
INFO:   Uvicorn running on http://0.0.0.0:9000 (Press CTRL+C to quit)
```

On my system I have a user specifically for mealie:

```
# Outside the container
$ id mealie
uid=989(mealie) gid=988(mealie) groups=988(mealie)

# Inside the container
$ id
uid=989 gid=988
$ ls -al
total 156
drwxr-xr-x    1 root     root          4096 Apr 20 15:24 .
drwxr-xr-x    1 root     root          4096 Apr 24 12:47 ..
-rw-r--r--    1 root     root           275 Apr 20 14:59 Caddyfile
drwxrwxr-x   10 989      988           4096 Apr 24 12:47 data
drwxr-xr-x    5 root     root          4096 Apr 20 15:24 dist
drwxr-xr-x    8 root     root          4096 Apr 20 15:24 mealie
-rw-r--r--    1 root     root        127707 Apr 20 15:04 poetry.lock
-rw-r--r--    1 root     root          1328 Apr 20 14:59 pyproject.toml
```

You can see here the data folder is accessible because it's mounted from the host where the ownership is correct.

**Device Information (please complete the following information):**
 - OS: Ubuntu 20.04.2
 - Deployment: docker-compose 1.28.2 and docker 20.10.6
 - Version: 0.4.3

wengtad:
- [x] Optimize Docker Dev Size (Frontend: from ~1.52GB to ~429MB | API: from ~657MB to ~380MB)
- [x] Optimize Docker Prod Size (from ~542MB to ~373MB)
- [x] Add Gunicorn #550
- [x] Add Docs for Gunicorn
- [x] Add PUID/PGID to Docker. Fixes #350, #429
- [x] Merge Dockerfile.dev into Dockerfile (dev shared same base together with prod)
- [x] Add Docs for PUID/PGID
- [x] Add Docker healthcheck (for this is not necessary, I could remove if you want)


wengtad:
I have done with this PR, hope you like it :)

docker-compose.yml
```diff
@@ -17,6 +18,13 @@ services:
       POSTGRES_SERVER: postgres
       POSTGRES_PORT: 5432
       POSTGRES_DB: mealie
+
+      # WORKERS_PER_CORE: 0.5
+      # MAX_WORKERS: 8
+
+      WEB_CONCURRENCY: 2
+    tmpfs:
```
hay-kot, line 26: Is the tmpfs required? Looks like it's only supported if Docker is running on Linux. Did you look into any alternatives? I'm happy to look into it if you can let me know what it's used for.

docs/docs/changelog/v0.5.1.md
```diff
@@ -1,4 +1,4 @@
-# vx.x.x COOL TITLE GOES HERE
```
hay-kot, line 1: Ha! Thanks!

hay-kot (commented):
Wow! This looks awesome and addresses so many things that I was dreading. I think I just have the one question about the temporary file system otherwise it looks great 👍

docker-compose.yml
wengtad: By default is using `/dev/shm` but the codefactor throwing issue for it. It is use by Gunicorn, you could look at the `gunicorn_conf.py` and search for `worker_tmp_dir`. Maybe could just remove that line then I guess no need to add tmpfs already.

hay-kot: Great, I'll look into this after it's merged. Thanks again!

wengtad: Thanks!

PR #577, opened 2021-06-20
Revert "Printview columns"

hay-kot:
Reverts hay-kot/mealie#572


zierbeek:
i could just increase the max height of the block but I think that isn't the right solution? 
I will try to use a grid class

PR #579, opened 2021-06-20
Printview columns

zierbeek:
Tried to solve it in a different way. Could you test with your long recipe? In my instance, it works fine.


zierbeek:
Updated now @hay-kot

hay-kot:
Looks better, you were able to fix the text wrapping problem. I still think the layout isn't ideal. It seems difficult to read to me. Maybe there's a better layout we haven't considered? 

Screen Shot 2021-07-01 at 4 04 59 PM

PR #621, opened 2021-07-09
Update ios.md

rastacalavera:
putting this PR in Dev rather than master


hay-kot:
Awesome! Thanks for updating this

PR #622, opened 2021-07-09
Added iOS pictures for documentation

rastacalavera:
forgot to add them in the original


sephrat:
I think you used the wrong source branch. If you wanted to add stuff to your other PR, you can simply commit to the same branch you used there and it will show up in the PR.

Issue #657, opened 2021-08-27
Container running as root user


usma0118:
I have tried to deploy with:
securityContext:
  runAsNonRoot: true

this has resulted into groupmode error

**Expected behavior**
Container and application should run fine.

**Actual Behavior**
container fails with following error:
Production
groupmod: Permission denied.
groupmod: cannot lock /etc/group; try again later.

**Device Information (please complete the following information):**
 - OS: Ubuntu 20.04
 - Deployment: Microk8s

hay-kot:
Caveats 
1. I'm don't know anything about Microk8s
2. I'm not a container expert, so I'm not sure how much help I can be here. 

---
Currently, the user is created at runtime with the set PUID/GUID variables so without running as an admin I'm not sure this will work. Obviously running as root is not ideal, but I'm unsure how to allow both PUID/GUID settings and running as non-root. 

I will gladly accept any PRs or support on making this work as it's not my primary area of expertise.

usma0118:
in cady file you have, can't it run on any other port instead of 80?
`:80 {
	@proxied path /api/* /docs /openapi.json

	@static {
		file
		path *.ico *.css *.js *.gif *.jpg *.jpeg *.png *.svg *.woff *.woff2 *.webp
	}

	encode gzip zstd

	# Handles Recipe Images / Assets
	handle_path /api/media/recipes/* {
		header @static Cache-Control max-age=31536000
		root * /app/data/recipes/
		file_server
	}`

hay-kot:
Is that related to this issue?

You should be able to set port 80 to anything you want.

usma0118:
I had been working on fixing this. but right now blocked due to scheduler.db error.
is it intentional to have scheduler.db as 644 permissions while mealie.db as 755?

644 mealie.log
755 mealie_v0.5.0.db
644 scheduler.db
644 scheduler.db-journal

I have non-root PR ready but this issuer is a blocker for testing.

hay-kot:
> is it intentional to have scheduler.db as 644 permissions while mealie.db as 755?

No, it isn't. scheduler.db is deleted and recreated at runtime. This could potentially be moved to a non-exposed volume. I'm not sure why the logger wouldn't have the correct permissions.

It may be necessary to call 

```python
os.chmod({dir}, 0775)
```

on whatever files have improper permissions.

usma0118:
I think moving it to a non-exposed volume is good idea as it holds no critical data.

usma0118:
[...]

PR #686, opened 2021-09-19
Run Docker/runasnonroot

usma0118:
from discussion: #657 

- Modified to run docker image as user mealie (PUID: 911, PGID: 911)
- Changed caddy port from 80 to 9080 for non-root binding (breaking change for existing deployments)
- removed groupmod from run script and handled in docker file.
- Bumped image to python:3.9.6
- moved scheduler.db to /app/temp as no exposed volume, Thanks to @articuno1au
- bumped poetry version.

I have tested this on docker and kubernetes (via helm).


Caddyfile
```diff
@@ -3,7 +3,7 @@
 	admin off
 }
 
-:80 {
+:9080 {
```
hay-kot, line 6: Can you set this as an ENV variable that defaults to 80? That way we can avoid the breaking change but still allow others to configure as necessary. Once we move to v1 I can make the port change by default.

https://caddyserver.com/docs/caddyfile-tutorial#environment-variables

mealie/run.sh
```diff
@@ -5,20 +5,6 @@ set -e
...
-add_user() {
-    groupmod -o -g "$PGID" abc
-    usermod -o -u "$PUID" abc
-
-    echo "
-    User uid:    $(id -u abc)
-    User gid:    $(id -g abc)
-    "
-    chown -R abc:abc /app
-}
```
hay-kot, line 21: Why isn't this needed anymore? Don't we still need to set permissions on the mounted directories?

hay-kot (changes requested):
Overall I think it looks good. You do need to set the PR to work from the dev branch as is described in the[ contributions guide ](https://hay-kot.github.io/mealie/contributors/developers-guide/code-contributions/)

Once those are done, I'll pull it down and do a thorough testing. Thanks for the work on this!

Caddyfile
usma0118: good idea, will do.

PR #688, opened 2021-09-20
Missing minus in Shopping List UI

BryceStevenWilley:
Simple change to an icon variable so it would appear.

Before:

Screenshot from 2021-09-19 23-55-15

After: 

Screenshot from 2021-09-19 23-54-54

I split up this change from the poetry update that I needed to get things compiling, I can drop the poetry change if desired.

### DIFF
--- a/.github/workflows/test-all.yml
+++ b/.github/workflows/test-all.yml
@@ -39,7 +39,7 @@ jobs:
       #  -----  install & configure poetry  -----
       #----------------------------------------------
       - name: Install Poetry
-        uses: snok/install-poetry@v1.1.1
+        uses: snok/install-poetry@v1
         with:
           virtualenvs-create: true
           virtualenvs-in-project: true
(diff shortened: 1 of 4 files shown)


BryceStevenWilley:
The failing test is:
```
  AttributeError

  'EmptyConstraint' object has no attribute 'allows'

  at ~/.poetry/lib/poetry/_vendor/py3.9/poetry/core/version/markers.py:279 in validate
      275│ 
      276│         if self._name not in environment:
      277│             return True
      278│ 
    → 279│         return self._constraint.allows(self._parser(environment[self._name]))
      280│ 
      281│     def without_extras(self):
      282│         return self.exclude("extra")
      283│ 
Error: Process completed with exit code 1.
```

Is apparently related to this issue: https://github.com/python-poetry/poetry/issues/4402, which is fixed in poetry 1.1.8. Should be able to fix by adding: `  version: 1.1.8` to https://github.com/hay-kot/mealie/blob/0addb3fe6b1e8ad7cdd9125ed097bc2154abe376/.github/workflows/test-all.yml#L42. I can try to make a PR, but won't be til later this week.

hay-kot:
Sorry about the CI/CD been having a lot of problems with it on multiple branches. Thanks for cleaning this up!

PR #714, opened 2021-10-01
feat: Add basic support for Web Share Target API

korapp:
This is a simple change that allows an installed PWA to receive recipe links for import via the Web Share Target API

https://user-images.githubusercontent.com/3818002/135656533-23a5abfa-53d0-4239-b12e-9f350d67d5ce.mp4


hay-kot:
Very cool, thank you!

PR #728, opened 2021-10-07
Fix docker-compose files

tribut:
Frontend-Dockerfile is now simply called `Dockerfile`, the [build/target parameter](https://docs.docker.com/compose/compose-file/compose-file-v3/#target) has been added in compose file version 3.4.

### DIFF
--- a/docker-compose.dev.yml
+++ b/docker-compose.dev.yml
@@ -1,5 +1,5 @@
 # Use root/example as user/password credentials
-version: "3.1"
+version: "3.4"
 services:
   # Vue Frontend
   mealie-frontend:
@@ -7,7 +7,7 @@ services:
     image: mealie-frontend:dev
     build:
       context: ./frontend
-      dockerfile: Dockerfile.frontend
+      dockerfile: Dockerfile
     restart: always
     ports:
       - 9920:8080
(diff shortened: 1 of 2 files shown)


hay-kot:
Hey! Thanks for the PR. The docker-compose file you updated is for development, but the dockerfile is the one used for production, as such it won't hot reload when making changes, so it doesn't really apply to the compose file.

Did you want to take a stab at implementing the hot reload for the docker-compose.dev.yml?

tribut:
Hey, thanks for taking the time. I really just wanted to get *any* dev environment working so I could try out some things. I understand that `frontend/Dockerfile` is meant for production, but I don't think there is another Dockerfile available right now?
With the changes from this PR, `make docker-dev` doesn't throw any errors at least. Is there some other way to get a mealie-next dev environment (using Docker) I'm missing?

hay-kot:
> but I don't think there is another Dockerfile available right now?

Correct. I haven't been using docker for development for some time, so I haven't been keeping it up to date. 

> With the changes from this PR, make docker-dev doesn't throw any errors at least. 

The image will build just fine and "work" but for development in the Frontend, you'll need to create a new Dockerfile to enable hot reloading. If you're only doing backend stuff, you probably don't need to worry about this for now.

hay-kot:
Closing since I think this PR is addressing the issue. Still a WIP though. 

- #756

PR #732, opened 2021-10-11
Fix issue with parsing scraped nutrition

cadamswaite:
The scraper often contains the nutrition information, this should be populated by default (then could be hidden if not needed)

Found a small bug in mealie/db/models/recipe/recipe.py

### DIFF
--- a/mealie/db/models/recipe/recipe.py
+++ b/mealie/db/models/recipe/recipe.py
@@ -112,7 +112,7 @@ class RecipeModel(SqlAlchemyBase, BaseMixins):
         self.image = image
         self.recipeCuisine = recipeCuisine
 
-        self.nutrition = Nutrition(**nutrition) if self.nutrition else Nutrition()
+        self.nutrition = Nutrition(**nutrition) if nutrition else Nutrition()
 
         self.tools = [Tool(tool=x) for x in tools] if tools else []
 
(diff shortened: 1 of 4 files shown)


cadamswaite:
Ah, I noticed the nutrition actually returns a string which causes some issues when editing the recipe, so this is not ready to merge. .e.g `"fatContent": "15g"` = `Fat 15g grams` etc.

Not sure the best way to parse this, since could be in any unit

hay-kot:
> Ah, I noticed the nutrition actually returns a string which causes some issues when editing the recipe, so this is not ready to merge. .e.g `"fatContent": "15g"` = `Fat 15g grams` etc.
> 
> Not sure the best way to parse this, since could be in any unit

Looks like you resolved this, at least for enough cases. Is this ready to review/merge?

cadamswaite:
Yeah it works for most cases, I just keep finding rogue ones like:
`"carbohydrateContent": "34g carbohydrate (21g sugars",` when everything get stripped out, this ends up being parsed as `3421g` which isn't ideal..

Not sure how best to deal with it - potentially a regex that matches the first `\d+(\.\d+)?`

cadamswaite:
@hay-kot I think this is now ready for review/merge. 
It works for me™ on the sites I've tried - the missing info is generally a scraper issue

PR #733, opened 2021-10-11
Fix issues parsing times not in "PT.*H.*M" format

cadamswaite:
Added more tests for the time cleaner. 
Allow parsing of a wider range of times by reusing regex in `parse_duration` (I was seeing issues for <1h)

I wasn't sure what to do about invalid times, so I logged them as an error. 
Happy to pass instead if you think that would be better

### DIFF
--- a/mealie/services/scraper/cleaner.py
+++ b/mealie/services/scraper/cleaner.py
@@ -179,9 +179,6 @@ def clean_time(time_entry):
         return str(time_entry)
 
 
-# ! TODO: Cleanup Code Below
-
-
 def parse_duration(iso_duration):
     """Parses an ISO 8601 duration string into a datetime.timedelta instance.
     Args:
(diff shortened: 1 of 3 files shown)


mealie/services/scraper/cleaner.py
```diff
@@ -167,65 +170,51 @@ def clean_time(time_entry):
...
-    if m[4]:
-        hours = int(m[4])
-    if m[5]:
-        minutes = int(m[5])
-    if m[6]:
-        seconds = float(m[6])
+    times = {"days": 0, "hours": 0, "minutes": 0, "seconds": 0}
+    for unit, value in times.items():
+        if m.group(unit):
+            times[unit] = int(float(m.group(unit)))
```
cadamswaite, line 206: If we want to maintain milliseconds in a recipe(?), can change this to be a cast of some sort e.g.
```
times = {"days": 0, "hours": 0, "minutes": 0, "seconds": 0.0}
...
times[unit] = type(value)(m.group(unit))
```

PR #743, opened 2021-10-16
[WIP] Adding a route for CSV imports

chollinger93:
# Feature Proposal
Importing recipes from a list of URLs (e.g., browser-agnostic bookmarks) is currently not possible. I've got this "brilliant" system of keeping bookmarks and physical printouts and wanted to consolidate everything into one application.

I've worked around this by writing a script in `go` that uses the API, but figured a quick new Migration class would be useful for the project as a whole.

# Functionality
Imports lists of URLs from a CSV file, with optional tags and categories.

The CSV file should be pipe (`|`) separated and contain one url per row. Tags and categories should be comma separated and are optional.

```
links|tags|categories
https://www.smoking-meat.com/september-24-2015-smoked-pork-tenderloin|bbq,smoking,american|main
https://example.com
```

The ZIP file should contain the CSV files in a flat directory. The files can end in any `.*sv` pattern.

```
recipes.zp
  ├── recipes_american.csv
  └── recipes_bbq.tsv
```

# Implementation 
Because I wanted to avoid adding a dependency, I parsed the CSV by hand using the `csv` standard module. This has been done in `services.migrations._migration_base.py`. I've aded a custom Migration implementation, called `csv.py`. The reader is a generator.

The frontend uses the same migration card, and I've added en-US, en-GB, and de-DE (the latter might be clunky - I'm not sure what one would call the "pipe symbol").

# Caveats
Ideally, I think it would be ideal for people to import the `bookmarks.html` that browsers produce, but after spending 30 minutes on figuring out a way to deterministically parse those, I've decided that copying all URLs from said file and tagging them in a text editor was easy enough. 

I think there are also points to be made with regards to the format of the CSV, but pipes seem like an unlikely candiate to be present in a URL.

It should also really be a multi-threaded process.

The migration API also only allows for zip files, which isn't great. Other file types get an incorrect success message. Matter of fact, the method for the `/upload` endpoint is called `upload_nextcloud_zipfile` - I'm not familiar enough with the codebase to touch that, though...

I also think the scraper should return a conflict code (409?) instead of a 500 for a duplicate.

And lastly, I think customizable `vue` card that allows for different CSV separators and the like would be useful.

# Next Steps
**I haven't added any tests yet**. Importing a handful of recipes works well:

Screenshot from 2021-10-16 11-50-42

I've used my actual bookmarks, about ~120 recipes, *and that returns a 500, despite the import running in the background*. My guess is the connection just times out, due to the time it takes to sequentially scrape all the recipes.

An idea was to batch the scraping and imports:
```python
def migrate(session: Session, csv_path: Path) -> Generator[MigrationImport,None,None]:
    csv_migration = CsvMigration(migration_file=csv_path, session=session)

    batch_size = 10
    all_recipes = []
    c = 0
    with csv_migration.temp_dir as dir:
        csv_dir = Path(dir)
        for f in csv_dir.glob("*.*sv"):
            for row in csv_migration.csv_reader(csv_file=f):
                # Scrape the recipe
                try:
                    recipe = create_from_url(row["url"])
                except Exception:
                    continue
                if not recipe or recipe.name is None:
                    continue
                # Keep count
                c += 1
                # Set the tags and categories
                recipe.tags = row["tags"]
                recipe.recipe_category = row["categories"]
                # Save it
                all_recipes.append(recipe)
                # Flush the Buffer
                if c >= batch_size:
                    csv_migration.import_recipes_to_database(all_recipes)
                    # Clear the buffer
                    all_recipes = []
                    c = 0
                # Yield
                yield csv_migration.migration_report
    # Import the rest
    csv_migration.import_recipes_to_database(all_recipes)
    yield csv_migration.migration_report
```
The problem here is that `migrate` returns a List, not a generator, so the frontend would be able to render imports results as they come in. Changing that, or even adding another dialog that essentially says "This will take a while", would be a bigger change.

Screenshot from 2021-10-16 12-23-47

I might need some help with that, assuming this feature is deemed useful.

Wanted to get this PR out here for discussion - there's clearly stuff to improve. :)


hay-kot:
Thanks for the work on this! Lots of cool stuff in this PR. A couple of things I want to bring up. Would you consider adding this as a feature to mealie-next? It's the major rewrite branch for Mealie, and I'd like to capture this for v1. 

> Imports lists of URLs from a CSV file, with optional tags and categories.
>
> The CSV file should be pipe (|) separated and contain one url per row. Tags and categories should be comma-separated 
> and are optional.

Is this a format from something or just one you've added? I'm hesitant to add support for a specific format like this for importing data, I think that the usage of a feature goes down once you add the non-standard requirements. What would you think of providing a text-field input on the recipe creation page [See Here](https://beta.mealie.io/recipe/create?tab=url) that would allow for a comma-separated list of URLs? Then submitting those to the backend.

You'd miss out on the tags/categories, but perhaps there could be a UI for this? Maybe after the user inputs the URL, we could provide them an editable data table to add Categories/Tags? I'm not sure

> I also think the scraper should return a conflict code (409?) instead of a 500 for a duplicate.

That would be better

> The problem here is that migrate returns a List, not a generator, so the frontend would be able to render imports results > as they come in. Changing that, or even adding another dialog that essentially says "This will take a while", would be a 
> bigger change.

This is a shortcoming of the existing infrastructure, there is no Job Que mechanism for this kind of thing. On my list of things to solve for v1. 

Will take a deeper dive on this tomorrow

chollinger93:
> Thanks for the work on this! Lots of cool stuff in this PR. A couple of things I want to bring up. Would you consider adding this as a feature to mealie-next? It's the major rewrite branch for Mealie, and I'd like to capture this for v1.
Didn't know that existed!

I just did a quick test merge locally, not too many conflicts, but:
```
mealie git:(feature/csv_import) ✗  make docker-dev
docker-compose -f docker-compose.dev.yml -p dev-mealie down && \
docker-compose -f docker-compose.dev.yml -p dev-mealie up --build
Removing network dev-mealie_default
Creating network "dev-mealie_default" with the default driver
Building mealie-frontend
ERROR: Cannot locate specified Dockerfile: Dockerfile.frontend
make: *** [makefile:99: docker-dev] Error 1
```

That file appears to be missing?

> > Imports lists of URLs from a CSV file, with optional tags and categories.
> > The CSV file should be pipe (|) separated and contain one url per row. Tags and categories should be comma-separated
> > and are optional.
> 
> Is this a format from something or just one you've added? I'm hesitant to add support for a specific format like this for importing data, I think that the usage of a feature goes down once you add the non-standard requirements. What would you think of providing a text-field input on the recipe creation page [See Here](https://beta.mealie.io/recipe/create?tab=url) that would allow for a comma-separated list of URLs? Then submitting those to the backend.
>
> You'd miss out on the tags/categories, but perhaps there could be a UI for this? Maybe after the user inputs the URL, we could provide them an editable data table to add Categories/Tags? I'm not sure

It's a fairly generic format, but admittedly one I hacked together just to get my bookmarks imported. I just have a feeling I can't be the only one who keeps their recipes in bookmarks - the barrier of entry into a new tool (at least for me) is raised dramatically if you can bring your own data. I've several datasets in CSV/XLS that I've never migrated to any organization tool myself (not just recipes).

I've see these types of non-standard imports in a lot of tools, 1Password comes to mind, because CSVs are easy to edit, portable, lightweight, and can be treated as both a flat list (i.e., just a list of URLs), as well as a semi-relational structure (like the one I used here that adds tags and categories).

That being said, I think both would be ideal - a migration/import tool for large (organized!) collections, as well as a quick add type feature that just adds a bunch of recipes you tag later.

hay-kot:
> That file appears to be missing?

Yeah, I'm not using docker for development anymore, and I haven't gotten around to adding a new file for the new frontend set-up some comments on that in 👇 

- #728

> It's a fairly generic format, but admittedly one I hacked together just to get my bookmarks imported.

There's a [SO Question](https://codereview.stackexchange.com/questions/246002/parse-html-bookmarks-file-to-json-using-python) on importing bookmarks, I gave the code a quick test and it seemed to work, might be something to look at (up to you)

> That being said, I think both would be ideal - a migration/import tool for large (organized!) collections, as well as a quick,> add type feature that just adds a bunch of recipes you tag later.

I think I'm convinced! I like the idea of being able to use a CSV to bulk import form URLs. I do think it would lead to less confusion if we used "," as separators as the default and used the "|" character for separating tags/categories. 

Moving forward, I think to keep things simple...
1) You should probably limit this PR to getting a working example for 5 rows or so. Just to prove that it works, and then I can work on setting up a Job Que system in the background to ensure that it runs in an async loop/background thread. (Unless you want to tackle that in another PR)
2) Adding a section in the UI for uploading a CSV
3) Bonus points if you can include a UI for doing the same sort of input within the UI, but that should be fairly simple for me to put together once all the backend APIs are strung together...

chollinger93:
> > That file appears to be missing?
> 
> Yeah, I'm not using docker for development anymore, and I haven't gotten around to adding a new file for the new frontend set-up some comments on that in point_down
> 
>     * [Fix docker-compose files #728](https://github.com/hay-kot/mealie/pull/728)
Gotcha. Are you just running on bare metal? If so, can you share the commands or add something to the makefile?

I've enjoyed the docker/makefile combination for the hour I spent on it! I've actually re-purposed that idea at work just today - it hadn't occurred to me that that might be easier than a long `sbin/` folder that works because of my `zsh` history - for a large `go`/K8s/Docker application. Works like a charm.

> > It's a fairly generic format, but admittedly one I hacked together just to get my bookmarks imported.
> 
> There's a [SO Question](https://codereview.stackexchange.com/questions/246002/parse-html-bookmarks-file-to-json-using-python) on importing bookmarks, I gave the code a quick test and it seemed to work, might be something to look at (up to you)

The issue with the `bookmarks.html` is a functional one - I keep mine in a folder called "Recipes" in the bookmarks toolbar in FF, with generic sub-categories. Other people might have a different logic here - I suggest having the bookmarks and csv import separately.

> > That being said, I think both would be ideal - a migration/import tool for large (organized!) collections, as well as a quick,> add type feature that just adds a bunch of recipes you tag later.
> 
> I think I'm convinced! I like the idea of being able to use a CSV to bulk import form URLs. I do think it would lead to less confusion if we used "," as separators as the default and used the "|" character for separating tags/categories.

That's fair. I've changed that in this PR.

> Moving forward, I think to keep things simple...
> 
>     1. You should probably limit this PR to getting a working example for 5 rows or so. Just to prove that it works, and then I can work on setting up a Job Que system in the background to ensure that it runs in an async loop/background thread. (Unless you want to tackle that in another PR)
> 
>     2. Adding a section in the UI for uploading a CSV
> 
>     3. Bonus points if you can include a UI for doing the same sort of input within the UI, but that should be fairly simple for me to put together once all the backend APIs are strung together...
1. If you don't mind, let me just add a size check in the backend and have the frontend complain about it when there's too many rows, which can be removed once the async queue is implemented. `read_csvs` can still be a generator - reading it is cheap. I've set the current limit to 10 rows, which works on my machine, a 2019 System76 Gazelle.
2. Do you mean outside the normal migration-frontend route? Currently, it exists in the migration UI.
3. I do Data Engineering as my day job, so I'm not sure how well I'd put together a new UI to be honest. I write my own website via vim over ssh and render my blog via markdown...

I've updated this PR (going to dev) with the following changes:

- Max 10 records per import; an error message after importing those 10 is displayed in the report (quickest hack for now)
- Updated docs
- Changed standard separator to `,` with `|` for the tags/categories
- Codestyle fix

...but my fork maintains a [version](https://github.com/chollinger93/mealie/tree/feature/csv_import_rewrite) that matches the rewrite, but I'd need to test it using the new deployment/dev setup.

I've still yet to add unit tests (I know, I know...).

Here's the test data (some real, some 404, some quoted, some not, with one record too many):
```csv
links,tags,categories
https://www.smoking-meat.com/september-24-2015-smoked-pork-tenderloin,bbq|smoking|american,mains
https://example.com/smoked-whiskey-cheeseburgers/,bbq|smoking|american,mains
https://www.virtualweberbullet.com/brisket-high-heat/,bbq|smoking|american,mains
https://honestandtruly.com/30-minute-homemade-cherry-syrup-recipe/,bbq|condiments,"condiments"
https://www.savoringthegood.com/cherry-syrup/,bbq|condiments,condiments
https://mykoreankitchen.com/korean-tofu-spicy-ketchup/#more-16619,"asian|korean|vegan|healthy",mains
"https://mykoreankitchen.com/bean-sprout-rice-kongnamul-bap/",asian|korean|vegan|healthy,mains
https://www.koreanbapsang.com/haemul-pajeon/#wprm-recipe-container-6822,asian|korean|seafood,mains
https://www.aspicyperspective.com/korean-bibimbap/,asian|korean,mains
https://mykoreankitchen.com/kimchi-fried-rice/,asian|korean|vegan|healthy,sides
https://www.chefkoch.de/rezepte/386231125240691/Feine-Linsensuppe.html,german|soup,mains
```
Result
Screenshot from 2021-10-19 21-07-36
Screenshot from 2021-10-19 21-16-32
:

hay-kot:
> Gotcha. Are you just running on bare metal? If so, can you share the commands or add something to the makefile?

`make setup` `make frontend` and `make backend` should get you up and running provided you have poetry and yarn and the correct Python version. I'll try to take a stab at a docker file tomorrow. 


> 1. If you don't mind, let me just add a size check in the backend and have the frontend complain about it when there are too many rows, which can be removed once the async queue is implemented. read_csvs can still be a generator - reading it is cheap. I've set the current limit to 10 rows, which works on my machine, a 2019 System76 Gazelle.

That works! 

> 2. Do you mean outside the normal migration-frontend route? Currently, it exists in the migration UI.

That is what I meant, but honestly, I'm rewriting the whole migration UI, so leave it wherever you had it and I'll probably end up moving it anyways

> 3. I do Data Engineering as my day job, so I'm not sure how well I'd put together a new UI to be honest. I write my own website via vim over ssh and render my blog via markdown...

Fair enough! 

Overall I really like everything that's going on here, thank you!

chollinger93:
Opening this back up, as I've just added unit tests for the dev branch. For the re-write, I'd still suggest a separate PR from here: https://github.com/chollinger93/mealie/tree/feature/csv_import_rewrite

PR #745, opened 2021-10-17
Feature: Pick best image from list

cadamswaite:
[Google's recipe properties](https://developers.google.com/search/docs/advanced/structured-data/recipe) allow for multiple images to be defined such that the best one can be chosen for a given resolution screen.

Currently, the first image is chosen. This can lead to very low quality images being captured (e.g. `https://www.sbs.com.au/food/recipes/filipino-chicken-adobo`)
This PR checks all the images within the list and returns the largest one.

Also, some sites require a `UserAgent` header when fetching images (e.g. `https://www.gonnawantseconds.com/sloppy-joe-casserole/`), otherwise giving a 403.
From the websites I've tried, it doesn't matter if this is empty - but some may care.
It would be good if some people were willing to beta test this on a wider range of websites.

### DIFF
--- a/mealie/services/scraper/scraper.py
+++ b/mealie/services/scraper/scraper.py
@@ -135,7 +135,7 @@ def clean_scraper(scraped_data: SchemaScraperFactory.SchemaScraper, url: str) ->
     return Recipe(
         name=try_get_default(scraped_data.title, "name", "No Name Found", cleaner.clean_string),
         slug="",
-        image=try_get_default(scraped_data.image, "image", None),
+        image=try_get_default(None, "image", None),
         description=try_get_default(None, "description", "", cleaner.clean_string),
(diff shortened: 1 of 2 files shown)


mealie/services/scraper/scraper.py
```diff
@@ -135,7 +135,7 @@ def get_instructions() -> list[dict]:
     return Recipe(
         name=try_get_default(scraped_data.title, "name", "No Name Found", cleaner.clean_string),
         slug="",
-        image=try_get_default(scraped_data.image, "image", None),
```
jayaddison, line 138: @cadamswaite hey - this change was quite a long time ago, so I'd understand if you can't recall, but: do you remember why this method call was removed?

My guess at the moment is that it's because `scraped_data.image()` only returns a single result, whereas the logic in this PR adds the ability to pick from multiple items to display.

Ref: trying to figure out #2547

PR #768, opened 2021-11-01
Implement ShoppingLists and Advanced Ingredients

Grygon:
Will fill out this draft PR more later with details from https://github.com/hay-kot/mealie/discussions/761


hay-kot:
Wanted to check in, I think this is one of the few big ticket Items I have before a release candidate for v1. Are you still planning on tackling this?

Issue #776, opened 2021-11-04
Up/Down arrow keys do not work in recipe editor


asymworks:
### First Check

- [X] This is not a feature request
- [X] I added a very descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Mealie documentation, with the integrated search.
- [X] I already read the docs and didn't find an answer.

### What is the issue you are experiencing?

**Bug Description**
When editing recipes, the up and down arrow keys do not work within the textarea to move the cursor up and down.

**Steps to Reproduce**
Edit an existing recipe which has more than one (possibly wrapped) line in the Recipe Description.  Position cursor on a line and press up or down arrow key.

**Expected Behavior**
The cursor moves up or down as requested, and no error is observed.

**Actual Behavior**
Cursor remains on the same line and an error is logged to the JavaScript console.

**Error Information**
```
Uncaught TypeError: Cannot read properties of undefined (reading 'focus')
    at a.selectRecipe (SearchDialog.vue:164)
    at a.onUpDown (SearchDialog.vue:148)
```

### Deployment

Docker (Linux)

### Mealie Version

0.5.x

PR #777, opened 2021-11-04
Allow arrow keys to function when SearchDialog is not open

asymworks:
Proposed fix for #776.  I have tested this with a custom Docker build and it seems to work for me, both in recipe editor and within the search dialog to select recipes from search results.


hay-kot:
Nice catch, Thanks!

PR #681, opened 2021-09-09
Add LDAP authentication support

dvdkon:
This is a minimal PR adding support for authentication via LDAP by trying to bind to an LDAP server if internal Mealie authentication fails. If the user is on the LDAP server but not in the Mealie database, a new user object is inserted into the database.

I've made no UI changes as they aren't strictly necessary, but they might be useful, like disabling password changes for LDAP users altogether or checking the current password also via LDAP.

PR #803, opened 2021-11-15
Add LDAP authentication support (v2, onto dev)

dvdkon:
A continuation of #681, retargeted to the `dev` branch.


hay-kot:
I'm not an LDAP user, is this pretty self explanatory to use? Is there a need for any kind of documentation other than the ENV variables? 

Also, I've captured these changes and have a PR ready for the `mealie-next` branch as well. 

- #826

dvdkon:
I think it's fine as-is, since any further documentation would have to go into explaining LDAP. These options are fairly typical for authentication, I understand this isn't ideal for users without previous LDAP experience, but I feel that LDAP auth is better explained elsewhere.

hay-kot:
Awesome, thanks for your work on this!

Northguy:
> I think it's fine as-is, since any further documentation would have to go into explaining LDAP. These options are fairly typical for authentication, I understand this isn't ideal for users without previous LDAP experience, but I feel that LDAP auth is better explained elsewhere.

Hi, could you add some addtional information? Trying to get this to work, but authentication fails and I fail to find any debug information in the log file

rafaelmathieu:
> > I think it's fine as-is, since any further documentation would have to go into explaining LDAP. These options are fairly typical for authentication, I understand this isn't ideal for users without previous LDAP experience, but I feel that LDAP auth is better explained elsewhere.
> 
> Hi, could you add some addtional information? Trying to get this to work, but authentication fails and I fail to find any debug information in the log file

I am in the same boat. not sure what I did wrong.

@hay-kot  Is there no bind user and password ?

Northguy:
@rafaelmathieu I had the same question. Assuming that an anonymous connection is required, I tried by allowing anonymous connections. But also this does not work.

dvdkon:
> Is there no bind user and password ?

I'll try to explain: Some (most?) software that uses LDAP authentication does what I'd call "two stage" auth: They first bind to the server with a service account, find the actual account with a filter, then open another connection and bind to that account (/object) with the user-provided password.

What some other software does and what Mealie does now is slightly different: Instead of using a service account to find the appropriate account, it directly gets its DN by templating a string (e.g. `cn={},dc=example,dc=com` turns into `cn=dvdkon,dc=example,dc=com` if I'm logging in).

There is no need for a service account (or anonymous bind) and authentication is slightly faster and simpler, but all users have to have predictable DNs. I think this tradeoff is acceptable for software mostly ran by smaller subjects, since most small deployments don't have people's accounts in different `ou`s. If your LDAP directory is like that, we'll have to add support for multi-stage auth to Mealie.

dvdkon:
By the way, I think it would be great if a user of the LDAP auth wrote or at least drafted some documentation, it's clear that I'm not a great judge of what's obvious when I develop :)

rafaelmathieu:
@dvdkon thanks for the update. that makes more sense. I had never encountered this setup before. 

Update: I got it to work on my setup. Thanks @dvdkon

SurlyB:
@dvdkon wondering if you can help me with the LDAP_ADMIN_FILTER. I've managed to get my user accounts bound to LDAP, but cannot figure out how to identify administrator users. As best I can tell, the issue seems to be that my LDAP implementation (QNAP NAS; some variant of OpenLDAP I think) does not expose a memberOf attribute on user objects. I think this means the logic would need to be inverted, looking instead at the group for the memberUid attribute(s). Any thoughts or suggestions?

PR #847, opened 2021-11-28
Add support for raspberry pi

rolfberkenbosch:
Add support for raspberry pi


hay-kot:
This will not build. [See here](https://hay-kot.github.io/mealie/changelog/v0.5.0/#armv7-support).

Also, you can use it on a Raspberry Pi if you have a 64 bit operating system.

PR #914, opened 2022-01-08
Use composition API for more components, enable more type checking

PFischbeck:
This turned out to be much larger than I originally anticipated, sorry for that! Hopefully we can merge it anyways.

What I did:
- activated typescript linting with some default rules
- improved several type definitions
- switched a lot of components to composition API with typescript
- regenerated the type definitions with `dev/scripts/gen_global_components.py`. I think a lot of type definitions are also in `frontend/api/class-interfaces`, and maybe we should remove the double definitions at some point, as this leads to inconsistent types.

Turning on the type checking also revealed some bugs, which I fixed. For example, did you know that until now one could not see the "favorite" symbol in the recipe card views? Or that one could not drag any item in the mealplan from the first day?


hay-kot:
Wow, this is amazing. Thank you so much for all the work you put into this. Clearly I need to spend a little more time with the typescript docs! 

You seriously put my ahead 2-3 weeks easy with all this stuff. Still looking through it but will get it merged soon!

PR #925, opened 2022-01-11
Enable localization based on browser settings, add language selector

### LINKED ISSUE #1067: [v1.0.0b] [Task] - Localization Support

Localization in v1 isn't setup to work properly. 

## Frontend 
- [x] Use the [Nuxt Vuei18](https://i18n.nuxtjs.org/) module to manage localization (setup)
- [x] [Setup Browser Language Detection ](https://i18n.nuxtjs.org/browser-language-detection)
- [x] Setup Language Selector using locale storage
- [x] Ensure languages are lazy-loaded
- [x] Link language selection to 'Accept-Language" header on Axios instance for backend messaging
- [x] Force reload on language change to avoid complex reactive text management
- [x] Auto Generate list from Crowdin API

## Backend
- [x] Setup Translation Provider
- [x] Automatically read language preference from "Accept-Language" header. Fallback to English (or maybe ENV?) 
- [x] Setup Crowdin translations to feed in backend messages

This solution would allow Mealie to offer a "stateless" multilingual support where each client could specify their own language and Mealie would translate on the fly. I'd prefer to keep it flexible, but we may need to set a language for the group so server-side scheduled notifications can be translatable as well.

PFischbeck:
This is to enable localization on mealie-next. I think previously the language was chosen globally by an admin, but I thought it makes more sense to have it per browser session. Thus, a cookie is used to save it, and on first visit, it is calculated from the browser settings. The user can change the language any time.

Right now, it looks like this:
language-selector

But I'm not happy with it:
- We should somehow use the language names instead of codes. Where do we maintain them?
- Where should this dropdown be? Maybe it makes more sense to have it in the top bar, like on the [Vuetify Docs](https://vuetifyjs.com/en/).


hay-kot:
> We should somehow use the language names instead of codes. Where do we maintain them?

Probably easiest to snag them from the `dev` branch 
- https://github.com/hay-kot/mealie/blob/dev/frontend/src/store/modules/language.js

> Where should this dropdown be? Maybe it makes more sense to have it in the top bar, like on the Vuetify Docs.

I could see this in a few different spots
1. As you suggested in the top row like it is in the Vuetify docs. If we go that route, we should probable change the logout button to be just an icon to match both.
2. Leave it where it's at but instead have it be a matching button style instead of the v-select and have it open a dialog with a list to select? 
3. Move it to the users page (probably bad for accessibility)

I think either option one or two are both solid choices. I tend to learn away from dialogs, but in this case it might make sense, especially if we want to put a note in there about how to provide additional translations.

PFischbeck:
I've implemented option 2, so we now have a dialog for choosing the language.
grafik
I've also expanded the language list from the master branch to include all language that are on Crowdin, although some of them are at 0%. Maybe we should comment them out for now?

Apart from that, this is ready for review! This solves all frontend tasks from #1067 except for the "Accept-Language" header.

hay-kot:
> I've implemented option 2, so we now have a dialog for choosing the language.

That looks slick!

> I've also expanded the language list from the master branch to include all language that are on Crowdin, although some of them are at 0%. Maybe we should comment them out for now?

We'll just leave them for now. I might take a stab at automating the generation of the Locales based off certain % of translation today or tomorrow using the Crowdin API.

Huge step forward, thank you!

PR #941, opened 2022-01-17
Update the mealie.core.config to have the current version

BryceStevenWilley:
The info box in the corner / the about page of the server is a version behind, claiming that there's a newer release, even though it's the most up to date version.

Not sure how much it'll help given that 0.5.4 is already released, but might as well bump it anyway. `¯\_(ツ)_/¯`


hay-kot:
I'm going to just bump the version and push some additional updates and bump the minor version. Thanks for this though!

PR #964, opened 2022-02-04
Adding Ukrainian language support

Toucani:
Fixes #963 

Hi! I am not a Vue/js developer so I might've missed something. However, I verified the solution manually and was able to switch to Ukrainian and browse the site with Ukrainian translation.


Toucani:
The [CI test results](https://github.com/hay-kot/mealie/runs/5103916147?check_suite_focus=true) are identical to the [13ebbe0](https://github.com/hay-kot/mealie/commit/13ebbe0c6f1605eafea466ac7a2bfd8830f2a96c) ones(latest commit at the time of commenting), so it looks to me like the MR doesn't break any tests.

Tell me if there is anything I should fix or add, otherwise, this MR looks ready to me:)

PR #989, opened 2022-02-22
Feature/new-login-page

hay-kot:


zierbeek:
HI, based on the new login page and the changes in dark mode you've made recently, I get this on a first page visit when cleared cache. It's a mix between the dark and light mode

Schermafbeelding 2022-03-01 om 14 43 11

Issue #832, opened 2021-11-24
[v1.0.0b] [Task] - Async Input Validators


hay-kot:
### What is the problem this task addresses?

There are several inputs that are restricted by unique values 

- Recipe Names
- User Names
- Group Names
- ?? Probably others

Often times no user feedback is provided until the form is submitted and sometimes not at all. 

### Proposed/Possible Solution(s)?

Several endpoints should be created to do Async validation on input fields throughout the applications. 

Endpoints should follow a pattern as follows `/api/validators/{validator-type}` 

Endpoint should take in a post with a value attributed and return a { valid: bool } payload.

Implementation of async validators on the frontend is still a little unclear. It appears that you would have to manually use a validation on blur to verify that input and manually mark it invalid. This can likely be done in a reusable way by using the composition API and a closure function.

m42e:
@hay-kot With recipe do you mean the `name` or are you referencing to the `slug`?

hay-kot:
> @hay-kot With recipe do you mean the `name` or are you referencing to the `slug`?

Would be slug. Uniqueness is determined by slug and group_id in the database, so the check would need to validate that the recipe could be created.

m42e:
I assumed that.

Miroito:
@hay-kot What's left on this issue other than group names? I can probably take the rest over

hay-kot:
Nothing has been done on the frontend so we need some way to add an async validation for the Vuetify inputs. I haven't looked into how to do that, but I suspect it will require some manual work to get it done and displayed correctly as their validators don't support async functions. I would like to see some additions made to the validation folder using the composition API to make these validations reusable across the application.

hay-kot:
@Miroito, not sure if you're working on this, but I'm working on this as apart of the registration rewrite and will include it in the PR

Miroito:
I hadn't started anything, please go ahead.

PR #1052, opened 2022-03-14
Added validators for users and recipes

m42e:
provide a simple get api, allowing to test for existence of
- user by username
- recipe by slug
- group by name (not tested yet)

first implementation for #832

### DIFF
--- a/mealie/repos/repository_recipes.py
+++ b/mealie/repos/repository_recipes.py
@@ -1,5 +1,5 @@
 from random import randint
-from typing import Any
+from typing import Any, Optional
 from uuid import UUID
 
 from sqlalchemy import and_, func
@@ -143,3 +143,10 @@ class RepositoryRecipes(RepositoryGeneric[Recipe, RecipeModel]):
             .order_by(func.random())  # Postgres and SQLite specific
             .limit(limit)
         ]
+
+
+    def get_by_slug(self, slug: str, limit=1) -> Optional[Recipe]:
+        dbrecipe = self.session.query(RecipeModel).filter(RecipeModel.slug == slug).one_or_none()
+        if dbrecipe is None:
+            return None
+        return self.schema.from_orm(dbrecipe)
--- a/mealie/repos/repository_recipes.py
+++ b/mealie/repos/repository_recipes.py
@@ -144,8 +144,12 @@ class RepositoryRecipes(RepositoryGeneric[Recipe, RecipeModel]):
             .limit(limit)
         ]
 
-    def get_by_slug(self, slug: str, limit=1) -> Optional[Recipe]:
-        dbrecipe = self.session.query(RecipeModel).filter(RecipeModel.slug == slug).one_or_none()
+    def get_by_slug(self, group_id: UUID, slug: str, limit=1) -> Optional[Recipe]:
+        dbrecipe = (
+            self.session.query(RecipeModel)
+            .filter(RecipeModel.group_id == group_id, RecipeModel.slug == slug)
+            .one_or_none()
+        )
         if dbrecipe is None:
             return None
         return self.schema.from_orm(dbrecipe)
--- a/mealie/repos/repository_recipes.py
+++ b/mealie/repos/repository_recipes.py
@@ -144,7 +144,6 @@ class RepositoryRecipes(RepositoryGeneric[Recipe, RecipeModel]):
             .limit(limit)
         ]
 
-
     def get_by_slug(self, slug: str, limit=1) -> Optional[Recipe]:
         dbrecipe = self.session.query(RecipeModel).filter(RecipeModel.slug == slug).one_or_none()
         if dbrecipe is None:
--- a/mealie/routes/validators/validators.py
+++ b/mealie/routes/validators/validators.py
@@ -1,16 +1,11 @@
-from typing import Union
 from uuid import UUID
 
 from fastapi import APIRouter, Depends
 from sqlalchemy.orm.session import Session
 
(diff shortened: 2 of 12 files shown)


mealie/repos/repository_recipes.py
```diff
@@ -143,3 +143,10 @@ def get_random(self, limit=1) -> list[Recipe]:
             .order_by(func.random())  # Postgres and SQLite specific
             .limit(limit)
         ]
+
+
+    def get_by_slug(self, slug: str, limit=1) -> Optional[Recipe]:
+        dbrecipe = self.session.query(RecipeModel).filter(RecipeModel.slug == slug).one_or_none()
+        if dbrecipe is None:
+            return None
+        return self.schema.from_orm(dbrecipe)
```
hay-kot: You'll need to consider the group_id here since the recipe table is unique based off of group_id and the slug. This would return a results for any recipe even if it's in another group.

I don't think function is necessary because the recipes repo uses the slug as a primary key. Just be sure to call `by_group` to filter by the group.

mealie/routes/validators/validators.py
```diff
@@ -0,0 +1,45 @@
...
+from mealie.db.models.recipe.recipe import RecipeModel
+
+
+router = APIRouter()
+
+
+def create_response(value, type: str):
+    if value is None:
+        return SuccessResponse.respond(message=f"No {type} exists")
+    return ErrorResponse.respond(message=f"{type.capitalize()} already exists")
```
hay-kot: SuccessResponse and ErrorResponse are used for flash messages on the frontend. There's some middleware on the frontend that will check for the message property and display them to the user. 

Since the intention of these is to provide some real-time validation for text fields it would be better to use a dedicated model that returns a valid: true/false payload so we can handle the error messages on frontend on a case by case basis.

```python
from pydantic import BaseModel

class ValidationResponse(BaseModel):
    valid: bool = False


```

mealie/routes/validators/validators.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+@router.get("/user/{name}", response_model=Union[ErrorResponse, SuccessResponse])
+async def validate_user(name: str, session: Session = Depends(generate_session)):
```
hay-kot: Since these methods aren't async they shouldn't be declared as async. FastAPI will execute async methods on the main thread and these will all block. If you remove the async they'll be thrown into a threadpool which will yield better performance.

hay-kot (changes requested):
Looks good overall, just a few small things that need to get changed, thanks for jumping on this! Let me know if you need any help.

mealie/repos/repository_recipes.py
m42e: Ok, do you mean to pass it as parameter or use `self.group_id`?

frontend/composables/use-validators.ts
```diff
@@ -3,11 +3,18 @@ const EMAIL_REGEX =
 
 const URL_REGEX = /[-a-zA-Z0-9@:%._+~#=]{1,256}.[a-zA-Z0-9()]{1,6}\b([-a-zA-Z0-9()@:%_+.~#?&//=]*)/;
 
+validate(what: string, value: string): boolean {
+	const response =  fetch(`/api/validators/${what}/${value}`);
+  const data = response.json();
+	return ! data.error;
+}
+
```
hay-kot: need to use the function keyword 

```ts
function validate(what: string, value: string): boolean {
	const response =  fetch(`/api/validators/${what}/${value}`);
  const data = response.json();
	return ! data.error;
}
```

frontend/composables/use-validators.ts
```diff
@@ -3,11 +3,18 @@ const EMAIL_REGEX =
...
 export const validators: {[key: string]: (v: string) => boolean | string} = {
   required: (v: string) => !!v || "This Field is Required",
   email: (v: string) => !v || EMAIL_REGEX.test(v) || "Email Must Be Valid",
   whitespace: (v: string) => !v || v.split(" ").length <= 1 || "No Whitespace Allowed",
   url: (v: string) => !v || URL_REGEX.test(v) || "Must Be A Valid URL",
+	username: (v: string) => await
```
hay-kot: I'm not sure that this will work with the current implementation of errors within Vuetify (the UI framework). I suggest we address this in another PR since it's probably going to require some custom error handling with advanced usage of the composition API. I know you're not super into TS/JS so I totally get it if you don't want to jump into this.

tests/fixtures/fixture_users.py
```diff
@@ -28,6 +28,7 @@ def build_unique_user(group: str, api_client: requests) -> utils.TestUser:
         _group_id=user_data.get("groupId"),
         user_id=user_data.get("id"),
         email=user_data.get("email"),
+        username=user_data.get("username"),
         token=token,
     )
```
hay-kot, line 34: The TestUser is created in multiple places you'll need to add this in a few more places.

frontend/composables/use-validators.ts
m42e: Well, won't work anyway. sorry I accidentally added the fronting files.

frontend/composables/use-validators.ts
m42e: sorry, that was a mistake. I did not want to commit that

tests/fixtures/fixture_users.py
m42e: copy that.

hay-kot:
Looks good,  thanks for sticking with it!

PR #1060, opened 2022-03-16
Improve devcontainer and configure VSCode debugging

### LINKED ISSUE #1058: [v1.0.0b] [Task] - Improve Developer Experience

Starting development on Mealie requires a whole set of tools to properly execute the Frontend and Backend. It can also be difficult to reproduce CI requirements locally so developer can break something and not know until they push and their workflow is approved. 

## General Improvements
- [x] Implement a VS Dev Container
    - [x] Speed up Development Container  
- [x] Update makefile help and naming for consistency
- [x] Setup pre-commit

## Python Development
- [x] Add MyPy type checker
    - [x] Fix Existing Errors
    - [x] Remove `strict_optional = false` flag #1759 
    - [x] CLI 
    - [x] CI
- [x] Add Linters to Pre-commit
    - [x] Black
    - [x] Flake8 / Flake8 Print
    - [x] <s> MyPy </s> - couldn't get it to respect config

## Vue / TS / Frontend Development
- [x] Add Vitest as Testing Framework
    - [x] Testing in CI
- [ ] Setup components structure to support testing

danielpalstra:
This PR improves the devcontainer by:

* Using python from the container instead of creating another virtual environment (using Poetry)
* Configures VSCode debugging
* Adds tasks to init the database
* Adds volumes to the docker container which should improve the performance when building the frontend

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -10,21 +10,24 @@ RUN if [ "${NODE_VERSION}" != "none" ]; then su vscode -c "umask 0002 && . /usr/
 
 # install poetry - respects $POETRY_VERSION & $POETRY_HOME
 
+RUN echo "export PROMPT_COMMAND='history -a'" >> /home/vscode/.bashrc \
+    && echo "export HISTFILE=~/commandhistory/.bash_history" >> /home/vscode/.bashrc \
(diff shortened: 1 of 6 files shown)


hay-kot:
Looks great, thanks for digging into this one!

PR #1072, opened 2022-03-20
Fix docker-compose.dev.yml reference

namelivia:
The command `make docker-dev` for setting up the development environment is failing on the next branch because `docker-compose.dev.yml` is pointing to an unexisting Dockerfile on the frontend folder, probably because it was renamed at some point but this dockerfile was not updated.
This PR updates the reference.

### DIFF
--- a/docker-compose.dev.yml
+++ b/docker-compose.dev.yml
@@ -7,7 +7,7 @@ services:
     image: mealie-frontend:dev
     build:
       context: ./frontend
-      dockerfile: Dockerfile.frontend
+      dockerfile: Dockerfile
     restart: always
     ports:
       - 9920:8080


hay-kot:
Hey thanks for this! Unfortunately, this doesn't do much for the docker-dev compose file. The Dockerfile you reference is for the production build and doesn't contain any of the bindings or setup for a development stack so any changes to the frontend file doesn't yield any changes in the container. I did start work on implmeneting a fix for this, but ran into an odd CORS issue and ended up abandoning the PR since I don't use docker for development and we've recently added dev-containers to the repo. 

Here's my PR for reference

- https://github.com/hay-kot/mealie/pull/756

If you want to take a stab at getting the docker-compose.dev.yml file all set up for development that would be great!

namelivia:
Sure, I noticed this yesterday afternoon and did a bunch of fixes that I have in my own branch to restore the docker dev environment. I'll read the changes in your branch and try to come back with another PR.

PR #1134, opened 2022-04-06
Optional Import Keywords as Tags

Miroito:
# Related
fixes https://github.com/hay-kot/mealie/issues/1105 - Optional Import Keywords as Tags

# Feature description

- A new checkbox has been added to the recipe creation page to allow the user to optionally tell the api to import the original recipe's keywords as tags in mealie.

### DIFF
--- a/frontend/api/class-interfaces/recipes/recipe.ts
+++ b/frontend/api/class-interfaces/recipes/recipe.ts
@@ -77,8 +77,8 @@ export class RecipeAPI extends BaseCRUDAPI<Recipe, CreateRecipe> {
     return await this.requests.post<Recipe | null>(routes.recipesTestScrapeUrl, { url });
   }
 
-  async createOneByUrl(url: string) {
-    return await this.requests.post<string>(routes.recipesCreateUrl, { url });
+  async createOneByUrl(url: string, include_tags: boolean) {
+    return await this.requests.post<string>(routes.recipesCreateUrl, { url, include_tags });
   }
 
   async createManyByUrl(payload: BulkCreatePayload) {
--- a/frontend/api/class-interfaces/recipes/recipe.ts
+++ b/frontend/api/class-interfaces/recipes/recipe.ts
@@ -77,8 +77,8 @@ export class RecipeAPI extends BaseCRUDAPI<Recipe, CreateRecipe> {
     return await this.requests.post<Recipe | null>(routes.recipesTestScrapeUrl, { url });
   }
 
-  async createOneByUrl(url: string, include_tags: boolean) {
-    return await this.requests.post<string>(routes.recipesCreateUrl, { url, include_tags });
+  async createOneByUrl(url: string, includeTags: boolean) {
+    return await this.requests.post<string>(routes.recipesCreateUrl, { url, includeTags });
   }
 
   async createManyByUrl(payload: BulkCreatePayload) {
--- a/mealie/schema/recipe/recipe.py
+++ b/mealie/schema/recipe/recipe.py
@@ -27,6 +27,7 @@ class RecipeTag(MealieModel):
     id: UUID4 = None
     name: str
     slug: str
+    group_id: UUID4 = None
 
     class Config:
         orm_mode = True
@@ -43,9 +44,10 @@ class RecipeTool(RecipeTag):
 
 class CreateRecipeByUrl(BaseModel):
     url: str
+    include_tags: int
 
     class Config:
-        schema_extra = {"example": {"url": "https://myfavoriterecipes.com/recipes"}}
+        schema_extra = {"example": {"url": "https://myfavoriterecipes.com/recipes", "include_tags": True}}
 
 
 class CreateRecipeBulk(BaseModel):
@@ -102,7 +104,7 @@ class RecipeSummary(MealieModel):
     @validator("recipe_category", always=True, pre=True, allow_reuse=True)
     def validate_categories(cats: list[Any]):  # type: ignore
         if isinstance(cats, list) and cats and isinstance(cats[0], str):
-            return [RecipeCategory(id=uuid4(), name=c, slug=slugify(c)) for c in cats]
+            return [RecipeCategory(id=uuid4(), name=c, slug=slugify(c), group_id=uuid4()) for c in cats]
         return cats
 
     @validator("group_id", always=True, pre=True, allow_reuse=True)
--- a/mealie/schema/recipe/recipe.py
+++ b/mealie/schema/recipe/recipe.py
@@ -42,12 +42,12 @@ class RecipeTool(RecipeTag):
     on_hand: bool = False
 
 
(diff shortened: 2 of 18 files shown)


mealie/schema/recipe/recipe.py
```diff
@@ -43,9 +44,10 @@ class RecipeTool(RecipeTag):
 
 class CreateRecipeByUrl(BaseModel):
     url: str
+    include_tags: int
```
hay-kot: Why is this an integer? 

Also if it truly does need to be a boolean, it needs to default too false. The debug endpoint also uses this model for the debug payload. We may want to split them in the future, but for now it's probably fine.

mealie/services/scraper/cleaner.py
```diff
@@ -297,3 +297,21 @@ def pretty_print_timedelta(t: timedelta, max_components=None, max_decimal_places
...
+
+    if isinstance(text, list):
+        return text
+
+    tag_list = text.split(",")
+
+    for i in range(len(tag_list)):
+        tag_list[i] = tag_list[i].strip().capitalize()
+
+    return tag_list
```
hay-kot: Given your assumption I think you need to do a little more to verify the shape here. It's possible you get a dictionary or something and throw an exception. I would add two things.

1) The last if isinstance should check to make sure it is a string, if it isn't it should just return an empty list. 
2) Do some additional validation on the list type check to make sure it is a list of strings, otherwise it'll throw an exception later down the line. 

There's probably some other things to consider, but hard to say for sure. Unfortunately you can't rely on websites to adhere to the common schema so you can get all kinds of things.

mealie/schema/recipe/recipe.py
```diff
@@ -102,7 +104,7 @@ def validate_tags(cats: list[Any]):  # type: ignore
     @validator("recipe_category", always=True, pre=True, allow_reuse=True)
     def validate_categories(cats: list[Any]):  # type: ignore
         if isinstance(cats, list) and cats and isinstance(cats[0], str):
-            return [RecipeCategory(id=uuid4(), name=c, slug=slugify(c)) for c in cats]
```
hay-kot: I thought you were going to remove this and the other changes to the group_id requirement and fix that later down the line?  I'm concerned that generating a random group_id is going to cause errors that we haven't tested for or something else down the line.

frontend/api/class-interfaces/recipes/recipe.ts
```diff
@@ -77,8 +77,8 @@ export class RecipeAPI extends BaseCRUDAPI<Recipe, CreateRecipe> {
     return await this.requests.post<Recipe | null>(routes.recipesTestScrapeUrl, { url });
   }
 
-  async createOneByUrl(url: string) {
-    return await this.requests.post<string>(routes.recipesCreateUrl, { url });
+  async createOneByUrl(url: string, include_tags: boolean) {
+    return await this.requests.post<string>(routes.recipesCreateUrl, { url, include_tags });
```
hay-kot: Should be camelCased

hay-kot (changes requested):
Some other things I noticed

1. If you scrape a recipe and then delete it an re-scrape it it won't associate keywords. Presumably this is because it's not creating them so something else is going on there. 
2. It looks like there's some changes you made to get the group_id but fixed that we talked about but you didn't revert. Can you provide some context around that? Are those changes necessary for this PR or should they be removed?

frontend/api/class-interfaces/recipes/recipe.ts
Miroito: If I change this, the backend is unable to translate the camel case into pascal case. It works fine for other routes but I'm missing it on this one.

Miroito:
> If you scrape a recipe and then delete it an re-scrape it it won't associate keywords. Presumably this is because it's not creating them so something else is going on there.

I am not experiencing this issue. Can you confirm these steps to reproduce it:
1. Create a recipe by url and include the tags
2. Open this recipe and delete it
3. Create it like in step 1

Miroito:
> It looks like there's some changes you made to get the group_id but fixed that we talked about but you didn't revert. Can you provide some context around that? Are those changes necessary for this PR or should they be removed?

I have put in this PR only what I found to be strictly necessary to finish the group_id implementation on category/tag/tools. Since all the automated tests passed and my own tests were successful, I left it like this.

mealie/schema/recipe/recipe.py
Miroito: Changed it to a bool and default value to False

frontend/api/class-interfaces/recipes/recipe.ts
hay-kot: In the schema model you need to derive from the base `MealieModel` instead of `BaseModel` that will take care of the casing.

mealie/services/scraper/cleaner.py
Miroito: I have changed it to support lists of strings and strings.

hay-kot:
> I am not experiencing this issue. Can you confirm these steps to reproduce it:
>
> Create a recipe by url and include the tags
> Open this recipe and delete it
> Create it like in step 1

That's exactly what I did. Here's the recipe. https://www.bonappetit.com/recipe/crispy-mushroom-quesadillas

> I have put in this PR only what I found to be strictly necessary to finish the group_id implementation on category/tag/tools. Since all the automated tests passed and my own tests were successful, I left it like this.

You'll need to remove the the injection of the random UUID's when creating the category model. There shouldn't be a need for that if you create the tags in the database before you assign them to the recipe. If you can't figure it out, let me know and I'll take a look.

mealie/schema/recipe/recipe.py
Miroito: I have removed it from the validator. Backend tests all pass and some testing on my side seems to confirm that it doesn't have much of an impact.
It's been a while now and I'm not sure why I added the group_id to this validation function. Could have been necessary or maybe I just stumbled upon it and decided it had its place there

Miroito:
> That's exactly what I did. Here's the recipe. https://www.bonappetit.com/recipe/crispy-mushroom-quesadillas

The scraping is incredibly slow for me, it takes more than 1 minute (maybe because of the amount of pictures?). But I just did that exact procedure and I had the tags on the second time too.
I've updated the cleanup function in the meantime, maybe that had an impact.

hay-kot:
> The scraping is incredibly slow for me, it takes more than 1 minute (maybe because of the amount of pictures?). 

Yeah, has to do with the images, the current implementation grabs all the images and compares the sizes _(i think)_ which takes a while with them. There's probably a better way to do that, maybe parsing the integers in the URL and sum them? That's a future problem

> But I just did that exact procedure and I had the tags on the second time too.
> I've updated the cleanup function in the meantime, maybe that had an impact.

Can confirm it's fixed 👍

hay-kot:
Closing - see PR below for changes

- #1170 

Thanks for all the work on this!

PR #1142, opened 2022-04-09
Allow tags/categories/tools deletion

Miroito:
# Related
fixes https://github.com/hay-kot/mealie/issues/1110 - Allow Category/Tag/Tool Deletion

# Feature description

- A new three dots menu is now visible on the pages for categories, tags and tools. It contains an item to delete the element.
- The menu is its own component so that it can be expanded if needed in the future (like the context menu for recipes).


hay-kot (changes requested):
LGTM. Can you make one small change and update the styling to match the existing context menu? Basically just remove the colors from the trash icon

CleanShot 2022-04-19 at 17 15 17

CleanShot 2022-04-19 at 17 16 28

PR #1157, opened 2022-04-16
Fix failed tests when env default email/password were changed

Miroito:
A lot of tests would fail when running in a dev environment with non default DEFAULT_EMAIL and DEFAULT_PASSWORD environment variables.

This PR **does not** fix https://github.com/hay-kot/mealie/issues/1156, this issue will need extra attention.

### DIFF
--- a/mealie/schema/user/user.py
+++ b/mealie/schema/user/user.py
@@ -76,7 +76,7 @@ class UserBase(MealieModel):
         schema_extra = {
             "username": "ChangeMe",
             "fullName": "Change Me",
-            "email": settings.DEFAULT_EMAIL,
+            "email": "changeme@email.com",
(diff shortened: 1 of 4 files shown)


mealie/schema/user/user.py
```diff
@@ -76,7 +76,7 @@ def getter_dict(cls, name_orm: User):
         schema_extra = {
             "username": "ChangeMe",
             "fullName": "Change Me",
-            "email": "changeme@email.com",
+            "email": settings.DEFAULT_EMAIL,
             "group": settings.DEFAULT_GROUP,
             "admin": "false",
```
hay-kot: This would expose the default email to anyone who has access to the API docs. We should leave it hardcoded.

hay-kot (changes requested):

PR #1167, opened 2022-04-22
mealie/run: only modify the uid and gid if necessary

mathstuf:
This allows the Docker image to run without write access to its image in
the default case.

Also manually create the group in the Docker image to avoid auto
assignment of a different id.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,7 +30,8 @@ ENV PYTHONUNBUFFERED=1 \
 ENV PATH="$POETRY_HOME/bin:$VENV_PATH/bin:$PATH"
 
 # create user account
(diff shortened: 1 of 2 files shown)


hay-kot:
Have you done anything to verify that this doesn't effect anything at runtime? I know in the past we had someone draft a PR that worked with the UID and GID, but there was too many errors because the python process wasn't creating files using the same permissions so it ran into all kinds of conflict.

mathstuf:
I got it to at least start up with `podman run --read-only=true` and a read/write volume for `/app/data`. What other situations should I test?

mathstuf:
Unstaling. I still have an open question here as to what exactly needs testing.

hay-kot:
Addressed in 

- #1395

Issue #814, opened 2021-11-23
[v1.0.0b] - Error 500 when information for new user is not unique 


gabt92:
### First Check

- [X] This is not a feature request
- [X] I added a very descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Mealie documentation, with the integrated search.
- [X] I already read the docs and didn't find an answer.

### What is the issue you are experiencing?

As of today, those users exists: 
```
0: {username: "Change Me", fullName: "Change Me", email: "changeme@email.com", admin: true, group: "Home",…}
1: {username: "Im The User1", fullName: "Im The User1", email: "email@email.com", admin: false,…}
2: {username: "Im the User2 FullName", fullName: "Im the User2 FullName", email: "user2@email.com",…}
```

Creating a new user with one of those email already assigned to a user generate a error 500.
![image](https://user-images.githubusercontent.com/38195308/142960143-02f3e521-9a4c-4a41-8288-b4f684de5004.png)

Same issue when the fullname already exist (possibly because the username already exist as explained in https://github.com/hay-kot/mealie/issues/813 ) 

There is no frontend indication that there was an error.

### Deployment

Other

### Deployment Details

Beta on https://beta.mealie.io/

PR #1168, opened 2022-04-22
fix: error on duplicate username/email registration

Miroito:
# Related
#814 - Error 500 when information for new user is not unique

# Fixes
Trying to register with an already used username or email was not handled and resulted in the frontend being unresponsive.
Now, the registration is done only if the username and email are unique records in the database and a feedback is sent to the user trying to register. 

# Note(s)
- I understand the error message is coming from the backend and is therefore not localized.
- **More important in my opinion**, duplicate records should return the http error code 409. I could unfortunately not find where I could define it.

### DIFF
--- a/mealie/services/user_services/registration_service.py
+++ b/mealie/services/user_services/registration_service.py
@@ -59,7 +59,7 @@ class RegistrationService:
         if self.repos.users.get_by_username(registration.username):
             raise HTTPException(status.HTTP_409_CONFLICT, {"message": self.t.t("exceptions.username-conflict-error")})
         elif self.repos.users.get(registration.email, "email"):
-            raise HTTPException(status.HTTP_409_CONFLICT, {"message": self.t.t(key="exceptions.email-conflict-error")})
+            raise HTTPException(status.HTTP_409_CONFLICT, {"message": self.t.t("exceptions.email-conflict-error")})
 
         self.logger.info(f"Registering user {registration.username}")
         token_entry = None
--- a/mealie/services/user_services/registration_service.py
+++ b/mealie/services/user_services/registration_service.py
@@ -55,9 +55,9 @@ class RegistrationService:
         self.registration = registration
 
         if self.repos.users.get_by_username(registration.username):
-            raise HTTPException(status.HTTP_400_BAD_REQUEST, {"message": "A user with this username already exists"})
+            raise HTTPException(status.HTTP_409_CONFLICT, {"message": "A user with this username already exists"})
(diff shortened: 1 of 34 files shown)


mealie/services/user_services/registration_service.py
```diff
@@ -54,6 +54,11 @@ def _register_new_group(self) -> GroupInDB:
     def register_user(self, registration: CreateUserRegistration) -> PrivateUser:
         self.registration = registration
 
+        if self.repos.users.get_by_username(registration.username):
+            raise HTTPException(status.HTTP_400_BAD_REQUEST, {"message": "A user with this username already exists"})
+        elif self.repos.users.get(registration.email, "email"):
+            raise HTTPException(status.HTTP_400_BAD_REQUEST, {"message": "A user with this email already exists"})
+
```
hay-kot: > duplicate records should return the http error code 409. I could unfortunately not find where I could define it.

You should be able to use status.HTTP_409 instead here right?

mealie/services/user_services/registration_service.py
```diff
@@ -70,6 +75,8 @@ def register_user(self, registration: CreateUserRegistration) -> PrivateUser:
         else:
             raise HTTPException(status.HTTP_400_BAD_REQUEST, {"message": "Missing group"})
 
+        self.logger.info(self.repos.users.get_by_username(registration.username))
+
```
hay-kot: Did you mean to leave this in?

hay-kot (commented):
There is now a backend translation framework thats available inside the `mealie/lang` folder with a provider and locale files. You should use those to define new error messages for users. I just added this a week or so ago so not all messages have been moved over.

mealie/services/user_services/registration_service.py
Miroito: No sorry it will be gone in the final PR

mealie/services/user_services/registration_service.py
Miroito: status.HTTP_409 is undefined. I understood from the fastapi docs that is was defined in the `response_model`

hay-kot: It's status.HTTP_409_CONFLICT, you can define the return success message in the decorator, if you're raising an error you'll pass it into the HTTPException from the fastapi module.

Miroito:
Took me the whole week end to find how to call the translation function.. Hope I got it :)

hay-kot:
LGTM. FYI, for next time, you don't need to update all the language files, just the en-US one and the Translation services will automatically populate all the other files and submit a PR.

PR #1180, opened 2022-04-28
Edit_uri + Navigation assistant and IOS image

zierbeek:
Change edit_uri to be sure that nightly docs can be edited 
Activate navigation.instant
Fix link in Ios shortcut


zierbeek:
if you could, rerun the test, I was wrong with the merging of the branch at first

PR #1235, opened 2022-05-16
SSL/No Auth Email Support

### LINKED ISSUE #1187: Feature Request: V1 - Add SSL Email Support
**Before submitting this feature request I have**
_COMPLETE AND FILL THESE OUT_
- [x] Reviewed the documentation and verified that my feature request isn't already a feature.
- [x] Reviewed the existing feature requests and my feature has not been requested.
- [x] Checked the [tasks taged](https://github.com/hay-kot/mealie/issues?q=is%3Aissue+is%3Aopen+label%3Atask+) issues and verified my feature is not covered

**Please Describe The Problem To Be Solved**
Essentially my email server/setup uses the standard email protocol SSL/port 465. It looks like mealie supports SmartTLS and unauthenticated email, just not authenticated SSL SMTP. 

**(Optional): Suggest A Solution**
I'm not super proficient in python, but I think this would be relatively easy to implement:

1. According to [python email's docs](https://python-emails.readthedocs.io/en/latest/), you just need to add a `dict` value of `'ssl': True`. It looks like that could go somewhere around [here](https://github.com/hay-kot/mealie/blob/67178f9b7455972f7730a699018a72cf39190d18/mealie/services/email/email_senders.py#L27-L28).
2. That probably means that there should be a [new environment variable/config option](https://github.com/hay-kot/mealie/blob/67178f9b7455972f7730a699018a72cf39190d18/mealie/core/settings/settings.py#L68-L89) for `SMTP_SSL` and would need to be added to the constructor and also the docs & docker compose files. There should probably be logic that does not allow both `SMTP_TLS` and `SMTP_TLS` to both be true. 

I think there is also an opportunity for creating a better developer experience by changing the email authentication strategy to a single environment variable: `SMTP_AUTH_STRATEGY` (for example), that takes either `None` (default), `TLS`, or `SSL`. That may be a bigger change since that would break any current users. You could probably do something where if someone set TLS = true you use that to default SMTP_AUTH to TLS, but again that depends on how you want to transition support for this. 

**Additional Information**
- [x] If the feature is accepted I'm willing to submit a PR to provide this feature. 
-> Sure, I would want to know which of the above two strategies you want this to go. :)

nkringle:
- [X]  Add new setting type for `SMTP_AUTH_STRATEGY`, which will default to `TLS`. Also accepts `SSL` or `NONE`. 
- [X] Modify the `SMTP_ENABLED` checks to not require user/pass if auth strategy is none. 
- [x] Update docs
- [x] Modify Email sender to support no auth/ssl.  

fixes #1187 & fixes #1218


mealie/core/settings/settings.py
```diff
@@ -68,23 +68,23 @@ def DB_URL_PUBLIC(self) -> str | None:
...
     SMTP_FROM_NAME: Optional[str] = "Mealie"
-    SMTP_TLS: Optional[bool] = True
+    SMTP_TLS: Optional[bool] = False
     SMTP_FROM_EMAIL: Optional[str]
     SMTP_USER: Optional[str]
     SMTP_PASSWORD: Optional[str]
+    SMTP_AUTH_STRATEGY: Optional[str] = "TLS"  # Options: 'TLS', 'SSL', 'NONE'
+
+    if SMTP_TLS:
+        SMTP_AUTH_STRATEGY = "TLS"
```
nkringle: I'm wondering if we even need to do this? 

Today, we default to use TLS, so I am defaulting the new Auth to TLS. No change in this case. 
Today, if someone set `SMTP_TLS` to False, then should I be setting Auth Strategy to none?

nkringle: I'm also wondering if we need to do a warning console statement if SMTP_TLS is used? What is the preferred deprecation strategy?

mealie/services/email/email_senders.py
```diff
@@ -24,8 +24,10 @@ def send(self, email_to: str, subject: str, html: str) -> bool:
         )
 
         smtp_options: dict[str, str | bool] = {"host": self.settings.SMTP_HOST, "port": self.settings.SMTP_PORT}
-        if self.settings.SMTP_TLS:
```
nkringle, line 27: Is there a good way to test the DefaultEmailSender? I see the test makes its own implementation and I don't want to rewrite the production code in the test...

nkringle:
@hay-kot , I know you're probably really busy. When you get a chance could you approve running the workflows on my PR?

mealie/services/email/email_senders.py
hay-kot: I usually just use the UI on the settings page when I'm testing the mail server that it actually sends mail.

mealie/core/settings/settings.py
hay-kot: I think at this point we could probably just remove the SMTP_TLS and just put a deprecation notice in the change-log. From your description it sounds like it shouldn't be too much of an issue, even if folks were using it.

mealie/services/email/email_senders.py
nkringle: Yeah I did that on my own machine, was just wondering if we could get an automated test to run in the pipeline.

mealie/core/settings/settings.py
nkringle: That's what I'll do then to just avoid creating tech debt.

hay-kot:
Wrapped up the last few things on this one, Thanks for doing all the legwork! 🚀

nkringle:
> Wrapped up the last few things on this one, Thanks for doing all the legwork! 🚀 

Thanks! Sorry, life got busy, thanks for finishing it up!

PR #1245, opened 2022-05-19
Consolidate frontend types

### LINKED ISSUE #915: [v1.0.0b] [Task] - Consolidate Frontend Types

As referenced in the PR below, there duplicate types being used on the frontend which will lead to confusion.

- https://github.com/hay-kot/mealie/pull/914

These types should be merged and the auto-generated types should be used. Additionally, it may be beneficial to update the make-file to include code-generation commands for easy usage.

PFischbeck:
This fixes #915.

I changed the `BaseCRUDAPI` to have `CreateType, ReadType, UpdateType`. In addition, I removed (almost) all interfaces that were defined twice (explicitly, and generated from the backend). I also added some schemas that were not generated for the frontend yet.

In total, we now have 300 fewer lines of code 🎉


hay-kot:
CleanShot 2022-05-21 at 11 21 42


😍  It's beautiful. 

Thank you!

PR #1254, opened 2022-05-23
Docs/update release notes

hay-kot:


Miroito:
Now that the vetur config file is gone, my vscode is all red because Vetur is lost. Is there a reason to have removed it? Do you have a replacement for vetur?
It's a real nightmare to write frontend code right now with basically everything wrongly highlighted in red...

PR #1283, opened 2022-05-25
refactor: split up recipe create page

### LINKED ISSUE #1083: [v1.0.0b] [Task] - Create Page Should Use Nested Routes Instead of Tabs

https://beta.mealie.io/recipe/create?tab=url is managed by a tab view instead of nested routes which reduces performance and increases load times. The code is also fairly complicated because of the multi-purpose usage of the route.

Use Nuxt nested routes to reduce code complexity and split code.

See This Route for Example Implementation:

- https://beta.mealie.io/group/data/foods

PFischbeck:
This fixes #1083.


hay-kot:
Looks great, thanks for your work on this one!

PR #1312, opened 2022-05-29
fix: update issue links in v1.0.0beta-2 changelog

PFischbeck:
Seems like `git-cliff` was misconfigured and pointed to non-existent issues.


hay-kot:
I guess I should have read the documentation... thanks for fixing that!

PR #1331, opened 2022-06-01
Bugfix: handle authentication issues when dealing with active directory

lupolio:
This pull request adds code to allow users of an active directory to authenticate against it and assign admin rights based on the given filter.


hay-kot (changes requested):
Looks pretty good. Please add a test-case to cover the changes you've made.

Elegant996:
Made some improvements on this implementation over in #1487. Take a look @lupolio, they may be more in line with your intent.

Thanks!

PR #1338, opened 2022-06-02
security: delay server response whenever username is non existing

### LINKED ISSUE #1336: [v1.0.0b] - Time based user enumeration in the /api/auth/token endpoint
### First Check

# Time-based user enumeration 
The login functionality of Mealie allows an unauthenticated user to enumerate existing accounts by timing the server's response time.

The average time it takes the API server to respond is significantly lower for non-existing users than existing users. 

Pasted image 20220602164724

## Impact 

The impact on Mealie is low (3,7); there will only be a handful of users on any deployment of Mealie. I am assuming that the probability of finding valid usernames on a randomly exposed instance is low. 

The attack complexity is high, and the impact on confidentiality is low.

CVSS:3.1/AV:N/AC:H/PR:N/UI:N/S:U/C:L/I:N/A:N - Low 3,7
https://www.first.org/cvss/calculator/3.1#CVSS:3.1/AV:N/AC:H/PR:N/UI:N/S:U/C:L/I:N/A:N

The vulnerability is related to CWE-203.

### Perform cryptographic operation 
Is also possible to separate the logic of checking passwords and usernames inside the core/security/security.py into two different checks.

Then adding the verify_password function with some random invalid data to make sure that both checks will take a similar amount of time to complete  
``` python 
    if not user:
        verify_password("abc123cba321", "$2b$12$JdHtJOlkPFwyxdjdygEzPOtYmdQF5/R5tHxw5Tq8pxjubyLqdIX5i")
        return False
    
    if not verify_password(password, user.password):
        return False
```

### Set minimum  response time
A possible fix would be setting a minimal randomized time that is required before the auth function returns a response code. The time should be between the fastest and slowest time it takes to calculate and compare the password hash. 

Pasted image 20220602164711

# References 
https://www.hacksplaining.com/prevention/user-enumeration
https://cwe.mitre.org/data/definitions/203.html 
https://www.cvedetails.com/vulnerability-list/cweid-203/vulnerabilities.html

### Deployment

Docker (Linux), Other

### Deployment Details

_No response_

underfl0w:
My attempt at fixing the enumeration issue reported at #1336

Importing two additional libraries, time and randrange, from random. 
Random is used for creating a random time that get_token will use before returning a response whenever the user authentication fails. 
Time is used for getting a timestamp used for crudely timestamping the execution time of the get_token function.

### DIFF
--- a/mealie/core/security/security.py
+++ b/mealie/core/security/security.py
@@ -79,15 +79,21 @@ def authenticate_user(session, email: str, password: str) -> PrivateUser | bool:
     settings = get_app_settings()
 
     db = get_repositories(session)
-    user: PrivateUser = db.users.get(email, "email", any_case=True)
+    user = db.users.get_one(email, "email", any_case=True)
 
     if not user:
-        user = db.users.get(email, "username", any_case=True)
+        user = db.users.get_one(email, "username", any_case=True)
 
     if settings.LDAP_AUTH_ENABLED and (not user or user.password == "LDAP"):
         return user_from_ldap(db, session, email, password)
 
-    if not user or not verify_password(password, user.password):
+    if not user:
+        # To prevent user enumeration we perform the verify_password computation to ensure
+        # server side time is relatively constant and not vulnerable to timing attacks.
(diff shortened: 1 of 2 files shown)


hay-kot:
Thanks for both reporting and implementing a fix for this. I made a small change and used the other recommendation in your Issue.

PR #1355, opened 2022-06-08
Extend Apprise JSON notification functionality with programmatic data

michael-genson:
Overview
---
While the existing Apprise notification implementation works great for general use, it's not as useful for programmatic use. This PR adds programmatic data to the [Apprise custom JSON notification](https://github.com/caronc/apprise/wiki/Notify_Custom_JSON) implementation. This was done to solve for [feature request #1350](https://github.com/hay-kot/mealie/discussions/1350).

Functional changes
---
When a user creates a notification using either the `json://` or `jsons://` endpoint, the message contains a custom `mealie-event-source` header. As an example, here is the `mealie-event-source` header for when a new recipe is created:
```json
{
   "event_type":"create",
   "item_type":"recipe",
   "item_id":"b13a63a1-66a1-46f6-941d-544bcd87f950",
   "slug":"pasta-fagioli"
}
```

This is implemented for all existing CRUD events (recipies, cookbooks, categories, tags, and shopping lists).

---

The event listener for shopping lists has been extended to include CRUD operations for shopping list items. The event source header is a bit more customized to accommodate this:
```json
{
   "event_type":"update",
   "item_type":"shopping-list-item",
   "shopping_list_id":"bca4bb04-0d6a-490f-84ad-09627d3a0a1f",
   "shopping_list_item_ids":[
      "9c81e510-9935-4742-899e-823ce9f7c5c2"
   ]
}
```

In order to avoid reworking the API for adding and removing recipe ingredients to a shopping list, the shopping list id and all shopping list item ids are added to the header, _**not just the ones that were modified**_. A different event type is used to convey this:
```json
{
   "event_type":"create-and-update",
   "item_type":"shopping-list-item",
   "shopping_list_id":"bca4bb04-0d6a-490f-84ad-09627d3a0a1f",
   "shopping_list_item_ids":[
      "9c81e510-9935-4742-899e-823ce9f7c5c2",
      "b9f3c293-0ac8-4c72-8c94-27593ad2ce77",
      "9055eabb-0141-4e95-b112-01fe0ccf3739"
   ]
}
```

Implementation
---
To achieve this, a new `event_source` arg was added to the `event_bus.dispatch` function. The dispatch function is updated to add the event source JSON data to the headers as a URL param (per the Apprise docs):
```python
def dispatch(self, group_id: UUID4, event_type: EventTypes, msg: str = "", event_source: str = "") -> None:
    self.group_id = group_id
    self.event_source = event_source

    def _dispatch():
        if urls := self.get_urls(event_type):

            # this is where we add the mealie-event-source header
            if self.event_source:
                urls = [
                    EventBusService.merge_query_parameters(url, {"+mealie-event-source": self.event_source})
                    for url in urls
                    if EventBusService.is_json_url(url)
                ]

            self.publisher.publish(EventBusMessage.from_type(event_type, body=msg), urls)

    self.bg.add_task(_dispatch)
```

What's Next?
---
The shopping list items are a bit of a hack; ideally the user is able to subscribe to just shopping list events, just shopping list item events, or both. However, this would require a database change and some frontend changes.

When a full recipe is added or removed to/from a shopping list, the API doesn't return the details of which items were impacted, it just returns the shopping list in its new state (including all items). A more proper solution would be to either send a mixed header (items a/b/c were created and items d/e/f were updated) or send two events: one for each CRUD operation. This requires a re-write of the relevant bits of the shopping list API and its dependencies.

### DIFF
--- a/mealie/services/event_bus_service/event_bus_service.py
+++ b/mealie/services/event_bus_service/event_bus_service.py
@@ -60,15 +60,7 @@ class EventBusService:
         def _dispatch(event_source: EventSource = None):
             if urls := self.get_urls(event_type):
                 if event_source:
-                    urls = [
-                        # We use query params to add custom key: value pairs to the Apprise payload by prepending the key with ":".
-                        EventBusService.merge_query_parameters(
-                            url, {f":{k}": v for k, v in event_source.dict().items()}
-                        )
-                        # only JSON, XML, and HTTP Form endpoints support the custom key: value pairs, so we only apply them to those endpoints
-                        if EventBusService.is_custom_url(url) else url
-                        for url in urls
-                    ]
+                    urls = EventBusService.update_urls_with_event_source(urls, event_source)
 
                 self.publisher.publish(EventBusMessage.from_type(event_type, body=msg), urls)
 
@@ -81,6 +73,16 @@ class EventBusService:
             event=EventBusMessage.from_type(EventTypes.test_message, body="This is a test event."),
             notification_urls=[url],
         )
+
+    @staticmethod
+    def update_urls_with_event_source(urls: list[str], event_source: EventSource):
+        return [
+            # We use query params to add custom key: value pairs to the Apprise payload by prepending the key with ":".
+            EventBusService.merge_query_parameters(url, {f":{k}": v for k, v in event_source.dict().items()})
+            # only certain endpoints support the custom key: value pairs, so we only apply them to those endpoints
+            if EventBusService.is_custom_url(url) else url
+            for url in urls
+        ]
 
     @staticmethod
     def merge_query_parameters(url: str, params: dict):
--- a/mealie/services/event_bus_service/event_bus_service.py
+++ b/mealie/services/event_bus_service/event_bus_service.py
@@ -72,7 +72,8 @@ class EventBusService:
 
                 self.publisher.publish(EventBusMessage.from_type(event_type, body=msg), urls)
 
-        self.bg.add_task(_dispatch(event_source=event_source))
+        if dispatch_task := _dispatch(event_source=event_source):
+            self.bg.add_task(dispatch_task)
 
     def test_publisher(self, url: str) -> None:
         self.bg.add_task(
--- a/mealie/services/event_bus_service/event_bus_service.py
+++ b/mealie/services/event_bus_service/event_bus_service.py
@@ -1,3 +1,5 @@
+from urllib.parse import parse_qs, urlencode, urlsplit, urlunsplit
+
 from fastapi import BackgroundTasks, Depends
 from pydantic import UUID4
 
@@ -15,6 +17,7 @@ class EventBusService:
         self._publisher = ApprisePublisher
         self.session = session
         self.group_id = None
+        self.event_source = None
 
     @property
     def publisher(self) -> PublisherLike:
@@ -29,11 +32,19 @@ class EventBusService:
 
         return [notifier.apprise_url for notifier in notifiers if getattr(notifier.options, event_type.name)]
 
-    def dispatch(self, group_id: UUID4, event_type: EventTypes, msg: str = "") -> None:
+    def dispatch(self, group_id: UUID4, event_type: EventTypes, msg: str = "", event_source: str = "") -> None:
         self.group_id = group_id
+        self.event_source = event_source
 
         def _dispatch():
             if urls := self.get_urls(event_type):
+                if self.event_source:
+                    urls = [
+                        EventBusService.merge_query_parameters(url, {"+mealie-event-source": self.event_source})
+                        for url in urls
+                        if EventBusService.is_json_url(url)
+                    ]
+
                 self.publisher.publish(EventBusMessage.from_type(event_type, body=msg), urls)
 
         self.bg.add_task(_dispatch)
@@ -44,3 +55,18 @@ class EventBusService:
             event=EventBusMessage.from_type(EventTypes.test_message, body="This is a test event."),
             notification_urls=[url],
         )
+
+    @staticmethod
+    def merge_query_parameters(url: str, params: dict):
+        scheme, netloc, path, query_string, fragment = urlsplit(url)
+
+        # merge query params
+        query_params = parse_qs(query_string)
+        query_params.update(params)
+        new_query_string = urlencode(query_params, doseq=True)
+
+        return urlunsplit((scheme, netloc, path, new_query_string, fragment))
+
+    @staticmethod
+    def is_json_url(url: str):
+        return url.split(":", 1)[0].lower() in ["json", "jsons"]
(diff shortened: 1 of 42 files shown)


michael-genson:
I guess this code was written before type checking was implemented. I'm going to leave it alone and add `type: ignore` to avoid breaking anything

hay-kot:
Hey! Just wanted to give you heads up, I'm probably not going to be able to dig into this PR (or your other ones) until this weekend. Cursory look pretty good though, thanks for all your work lately!

michael-genson:
No problem! I had some free time this week, and you guys made spinning up a dev environment insanely easy with v1

hay-kot:
Hmm. I didn't realize the limitation of the payload for the Apprise integration. Maybe that's not the best approach for this? My concern is that the head isn't a super expected place to find data, and it seems add some complexity to the overall usage. 

I wonder if it would make more sense to just use the requests package to make a post request with the `event_source` data instead of trying to leverage apprise. We'd have to add a new database column to the EventNotifier table to toggle the json payload, but maybe that would be a more extendible approach? - Alternatively if the `is_json_url` method is True we could extract the URL and override the Apprise Publisher and just post message directly. 

What do you think?

michael-genson:
I was just about to post a comment with a couple other alternatives, but then I stumbled upon this thread:
https://github.com/caronc/apprise/issues/541

Which led me here:
https://github.com/caronc/apprise/pull/547

Looks like you can add custom keys to the JSON payload, it's just not documented yet. I'm going to toy around with this

michael-genson:
This commit replaces the `json.dumps(...)` to the header and adds the `key: value` pairs to the body directly. Here's a sample notification:
```json
{
  "version": "1.0",
  "title": "Recipe Updated",
  "message": "Pasta Fagioli has been updated, http://localhost:8080/recipe/pasta-fagioli",
  "attachments": [],
  "type": "info",
  "event_type": "update",
  "item_type": "recipe",
  "item_id": "b13a63a1-66a1-46f6-941d-544bcd87f950",
  "slug": "pasta-fagioli"
}
```

mealie/services/event_bus_service/event_bus_service.py
```diff
@@ -23,17 +26,32 @@ def publisher(self) -> PublisherLike:
...
+                if self.event_source:
+                    urls = [
+                        # We use query params to add custom key: value pairs to the Apprise payload by prepending the key with ":".
+                        # As of 2022-11-06 this is undocumented, but discussed in this pull request:
+                        # https://github.com/caronc/apprise/pull/547
+                        EventBusService.merge_query_parameters(url, {f":{k}": v for k, v in self.event_source.items()})
+                        for url in urls
+                        if EventBusService.is_json_url(url)
+                    ]
+
```
hay-kot: Won't this filter out any non json_urls? Was that intended behavior here?

mealie/services/event_bus_service/event_bus_service.py
```diff
@@ -44,3 +62,18 @@ def test_publisher(self, url: str) -> None:
...
+        # merge query params
+        query_params = parse_qs(query_string)
+        query_params.update(params)
+        new_query_string = urlencode(query_params, doseq=True)
+
+        return urlunsplit((scheme, netloc, path, new_query_string, fragment))
+
+    @staticmethod
+    def is_json_url(url: str):
+        return url.split(":", 1)[0].lower() in ["json", "jsons"]
```
hay-kot: Can you add some basic unit-tests for these two methods?

mealie/services/event_bus_service/event_bus_service.py
```diff
@@ -23,17 +26,32 @@ def publisher(self) -> PublisherLike:
...
 
-    def dispatch(self, group_id: UUID4, event_type: EventTypes, msg: str = "") -> None:
-        self.group_id = group_id
+    def dispatch(self, group_id: UUID4, event_type: EventTypes, msg: str = "", event_source: dict = None) -> None:
+        self.group_id = group_id  # type: ignore
+
+        if event_source is None:
+            event_source = {}
+
+        self.event_source = event_source
```
hay-kot: I don't quite understand the need for this. Could you just use `event_source` as a function level variable and remove it from the class?

hay-kot (changes requested):
Overall, looks pretty good. Theres a few minor things that I think need to get fixed. 

I do think we should to utilize a class of some-sort to define what the event_source type looks like. I think having a simple class that allows for extension would be sufficient (see example below). That would give us a little type safety when calling and ensure that typo errors are minimal. 

```py

class EventSource:
    event_type: str
    item_type: str
    item_id: UUID4 | int
    kwargs: dict

    def __init__(self, event_type: str, item_type: str, item_id: UUID4 | int, **kwargs) -> None:
        self.event_type = event_type
        self.item_type = item_type
        self.item_id = item_id
        self.kwargs = kwargs

    def dict(self) -> dict:
        return {
            "event_type": self.event_type,
            "item_type": self.item_type,
            "item_id": str(self.item_id),
            **self.kwargs,
        }
```

Let me know if you need any support on this one. Happy to take on some of the rework as needed.

mealie/services/event_bus_service/event_bus_service.py
michael-genson: Yeah, custom key value pairs are only supported by the custom form, custom XML, and custom JSON endpoints. The original implementation used custom headers which was only supported by JSON, but I can add XML and form endpoints and leave a note explaining why.

mealie/services/event_bus_service/event_bus_service.py
michael-genson: I think I did that because of something to do with it getting passed to the `dispatch` function and then added to the async `_dispatch` function, but looking at it now I have no idea why I chose to make it a class attr as opposed to a function var ¯\\_(ツ)_/¯ I'll change it

mealie/services/event_bus_service/event_bus_service.py
hay-kot: I noticed that limitation too, kind of a bummer. I think the problem here is that if the is_json_url, function returns false we discard the url all together. Wouldn't that break regular Apprise integrations like discord where we're implementing them as just a message service? If we specify an Apprise URL for discord AND the event_source is defined, we'll never make the call. 

Or am I reading this wrong?

michael-genson: Oh wow I totally missed that, you're right. I was only testing the JSON ones. I can an else statement so we keep all URLs

michael-genson:
I tweaked the shopping list item event trigger slightly to make it more consistent with the new `EventSource` class. Once we're done I'm happy to document the custom endpoint changes somewhere.

hay-kot:
LGTM, I think we can do a separate PR for the documentation so we don't run into anymore merge conflicts with some other stuff that's going on. I'm thinking a "Webhooks" Page in the documentation that outlines the limitations of the Apprise integration and some of the extensions.

PR #1356, opened 2022-06-08
Implemented "order by" API parameters for recipe, food, and unit queries

michael-genson:
Added two new parameters for `get_all` queries:
`orderBy` - allows for sorting by any valid property name (e.g. `name` or `description`)
`orderDescending` - boolean toggle for the order direction

I figure someone can utilize these to implement sorting & paging on the frontend. This also enables incremental syncs by sorting by "dateUpdated".


hay-kot:
Sweet! Thanks for this, looks great!

I was working on a `pagination` method on the generic repository, but I haven't had a chance to test it out. We have pretty limited ordering and sorting in the API, something that's on my never ending list to improve.

PR #1368, opened 2022-06-10
Fix: Ingredient sections lost after parsing

### LINKED ISSUE #1232: [v1.0.0b] - Ingredient sections lost after parsing
### First Check

Sections created prior to ingredient parsing (mostly on recipes created via url) are lost after the ingredient parse operation.
To reproduce:
1. Create a recipe from a url
2. Edit it and create some random sections
3. Save
4. Edit again and click on parse to go to the ingredient parsing screen
5. here we can already see that the sections created before are gone
6. Modify the ingredients using the info from the parser and click Save All
7. Back in the recipe screen the sections are still gone

### Deployment

Other

### Deployment Details

Does not depend on deployment. I have tested this bug on my deployed instance and my development instance.

michael-genson:
As referenced in issue #1232, when parsing ingredients, the frontend extracts the notes from each ingredient, then sends a string array to the API to be parsed.

Since the ingredients are converted to strings, any extra attributes are lost. Generally this doesn't matter, but notably the title attribute is lost, which is where ingredient sections are stored.

This fix copies the title attribute from each unparsed ingredient over to its corresponding parsed ingredient.


frontend/pages/recipe/_slug/ingredient-parser.vue
```diff
@@ -140,6 +140,12 @@ export default defineComponent({
       const { data } = await api.recipes.parseIngredients(parser.value, raw);
 
       if (data) {
+        // When we send the recipe ingredient text to be parsed, we lose the reference to the original unparsed ingredient.
+        // Generally this is fine, but if the unparsed ingredient had a title, we lose it; we add back the title for each ingredient here.
+        for (let i = 0; i < recipe.value.recipeIngredient.length; i++) {
+          data[i].ingredient.title = recipe.value.recipeIngredient[i].title;
+        }
+
```
hay-kot: For the sake of my sanity, could you wrap this in a try/catch for a a TypeError incase the returned list is somehow not the same length? I think we can just carry on, but I'd like to have it just in case...

```js

try {
    for (let i = 0; i < recipe.value.recipeIngredient.length; i++) {
      data[i].ingredient.title = recipe.value.recipeIngredient[i].title;
    }
} catch (TypeError) {
   console.error("Index Mismatch Error") // or some better error message
}

```

hay-kot (changes requested):

michael-genson:
Ah yeah that's smart, if nothing else it prevents an error if something gets messed up in the future

hay-kot:
Thank you!

PR #1369, opened 2022-06-10
Fix: increased float rounding precision for CRF parser

### LINKED ISSUE #1358: [v1.0.0b] - Parser doesn't handle 1/8th properly
### First Check

When parsing a recipe that has 1/8 teaspoon in it.  Parser converts that to 0.1 instead of 0.125.   Then after recipe saves, it displays as 1/10th instead of original 1/8th.   If I override the 0.1 with 0.125, it will show properly as 1/8th.

image

### Deployment

Docker (Linux)

### Deployment Details

_No response_

michael-genson:
Issue raised in #1358: the CRF parser rounds the float value to the nearest 1/10; this causes issues for smaller fractions. I increased the precision to 3, which is accurate down to increments of 1/32.

I also limited the denominator to a max of 32 to prevent float precision artifacts like 5980780305148019/72057594037927936 (which is really just 1/12).

Full disclosure: I'm having trouble getting the CRF parser running in dev, so I instead emulated the process locally. Since it's all built-in decimal handling it shouldn't matter, but it's worth checking a few test cases in dev before merging (such as the one mentioned in #1358).

### DIFF
--- a/mealie/schema/recipe/recipe_ingredient.py
+++ b/mealie/schema/recipe/recipe_ingredient.py
@@ -74,7 +74,7 @@ class RecipeIngredient(MealieModel):
         empty string.
         """
         if isinstance(value, float):
-            return value
+            return round(value, 3)
         if value is None or value == "":
             return None
         return value
--- a/tests/unit_tests/test_ingredient_parser.py
+++ b/tests/unit_tests/test_ingredient_parser.py
@@ -26,8 +26,12 @@ def crf_exists() -> bool:
 test_ingredients = [
     TestIngredient("½ cup all-purpose flour", 0.5, "cup", "all-purpose flour", ""),
     TestIngredient("1 ½ teaspoons ground black pepper", 1.5, "teaspoon", "black pepper", "ground"),
-    TestIngredient("⅔ cup unsweetened flaked coconut", 0.7, "cup", "coconut", "unsweetened flaked"),
-    TestIngredient("⅓ cup panko bread crumbs", 0.3, "cup", "panko bread crumbs", ""),
+    TestIngredient("⅔ cup unsweetened flaked coconut", 0.667, "cup", "coconut", "unsweetened flaked"),
+    TestIngredient("⅓ cup panko bread crumbs", 0.333, "cup", "panko bread crumbs", ""),
+    # Small Fraction Tests - PR #1369
+    # Reported error is was for 1/8 - new lowest expected threshold is 1/32
+    TestIngredient("1/8 cup all-purpose flour", 0.125, "cup", "all-purpose flour", ""),
+    TestIngredient("1/32 cup all-purpose flour", 0.03125, "cup", "all-purpose flour", ""),
 ]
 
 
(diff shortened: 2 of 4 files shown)


mealie/schema/recipe/recipe_ingredient.py
```diff
@@ -74,7 +74,7 @@ def validate_quantity(cls, value, values) -> NoneFloat:
         empty string.
         """
         if isinstance(value, float):
-            return value
+            return round(value, 3)
```
hay-kot, line 77: I added this because in the case of 1/3 I was getting a value of `0.333333` on the frontend/backend and it seemed like a waste of space.

Do you see anything wrong with rounding here? Seems to operator the same to me as far as I've tested

tests/unit_tests/test_ingredient_parser.py
```diff
@@ -26,8 +26,12 @@ def crf_exists() -> bool:
...
     TestIngredient("½ cup all-purpose flour", 0.5, "cup", "all-purpose flour", ""),
     TestIngredient("1 ½ teaspoons ground black pepper", 1.5, "teaspoon", "black pepper", "ground"),
-    TestIngredient("⅔ cup unsweetened flaked coconut", 0.7, "cup", "coconut", "unsweetened flaked"),
-    TestIngredient("⅓ cup panko bread crumbs", 0.3, "cup", "panko bread crumbs", ""),
+    TestIngredient("⅔ cup unsweetened flaked coconut", 0.667, "cup", "coconut", "unsweetened flaked"),
+    TestIngredient("⅓ cup panko bread crumbs", 0.333, "cup", "panko bread crumbs", ""),
+    # Small Fraction Tests - PR #1369
+    # Reported error is was for 1/8 - new lowest expected threshold is 1/32
+    TestIngredient("1/8 cup all-purpose flour", 0.125, "cup", "all-purpose flour", ""),
+    TestIngredient("1/32 cup all-purpose flour", 0.03125, "cup", "all-purpose flour", ""),
```
hay-kot, line 34: I've added two test cases that pass on my machine and update the other two to work with the new significant digits we're using now.

hay-kot (approved):
LGTM. I added two test cases and fixed the failing tests since I have CRF++ installed on my computer. 

Could you just double check the validator I added on the Pydantic model and make sure I'm not missing anything obvious? If I didn't do anything stupid, I'll go ahead and merge after that.

mealie/schema/recipe/recipe_ingredient.py
michael-genson: Nope that looks good to me! That's pretty much the same thing I did in the parser anyway, so it makes sense here too

tests/unit_tests/test_ingredient_parser.py
michael-genson: Yeah that looks good to me, and the tests make sense so if they're passing I don't see anything wrong with it!

hay-kot:
Thanks for the sanity check! 🚀

PR #1351, opened 2022-06-06
Fixed text color on RecipeCard in RecipePrintView and implemented ingredient sections

michael-genson:
This implements ingredient sections in the recipe print view and fixes the grey text in the recipe card.

The recipe in the editor:
image

Old implementation ignores ingredient sections:
image

New implementation groups ingredients by section:
image

PR #1372, opened 2022-06-12
Avoid page breaks in sections when printing recipes and other CSS tweaks

michael-genson:
A user posted a delicious looking vegan ramen recipe to the discord, so I imported it into Mealie. It's a monster of a recipe, with 35 ingredients and 16 steps.

After parsing it and categorizing everything, I was unhappy with the print view at this scale. It was readable, but sections were broken up between pages (and some steps were cut in half) and overall it was pretty unwieldly if I were to try to cook it.

With this recipe as a base I made the following broad changes:
- brought back the sections computed property from PR #1351 that we cut as it was no longer needed
- wrapped each ingredient section, instruction section in a new class that avoids being split between two pages
- visually grouped each section with stronger headers
- tweaked the CSS to fit a bit more content on a page

And I'm pretty happy with the results. There's probably still room to improve some things, but the `page-break-inside: avoid;` CSS attribute makes printing much more consistent.

Here's a before and after on the recipe in question:
[Current View - Ultimate Vegan Ramen Recipe With Miso Broth.pdf](https://github.com/hay-kot/mealie/files/8886239/Current.View.-.Ultimate.Vegan.Ramen.Recipe.With.Miso.Broth.pdf)
[New View - Ultimate Vegan Ramen Recipe With Miso Broth.pdf](https://github.com/hay-kot/mealie/files/8886240/New.View.-.Ultimate.Vegan.Ramen.Recipe.With.Miso.Broth.pdf)

And I checked a couple of more basic recipes to make sure it still looks good. If there is no section title for the instructions, it treats each step as its own section (otherwise longer recipes with no instruction sections would always start on page 2).
[New View - Simple Honey Glazed Baby Carrots Recipe.pdf](https://github.com/hay-kot/mealie/files/8886249/New.View.-.Simple.Honey.Glazed.Baby.Carrots.Recipe.pdf)

### DIFF
--- a/frontend/components/Domain/Recipe/RecipePrintView.vue
+++ b/frontend/components/Domain/Recipe/RecipePrintView.vue
@@ -20,10 +20,7 @@
       <div v-for="(ingredientSection, sectionIndex) in ingredientSections" :key="`ingredient-section-${sectionIndex}`" class="print-section">
         <div class="ingredient-grid">
           <template v-for="(ingredient, ingredientIndex) in ingredientSection.ingredients">
-            <div v-if="ingredient.title" :key="`ingredient-title-${ingredientIndex}`" class="ingredient-title mt-2">
-              <h4>{{ ingredient.title }}</h4>
-              <hr />
-            </div>
+            <h4 v-if="ingredient.title" :key="`ingredient-title-${ingredientIndex}`" class="ingredient-title mt-2">{{ ingredient.title }}</h4>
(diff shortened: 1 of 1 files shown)


hay-kot:
Looks good! 

I made one small change (and some auto-formatting stuff) and used a reducer to implement the splitting instead of the multiple for loops. I find it a little more readable, but mostly personal preference I think. 

Thanks!

michael-genson:
Thanks! And TIL with reducers, that's much easier to read. I wasn't a fan of the index looping and slicing, I like your method much better. I'm still pretty iffy with JS.

PR #1379, opened 2022-06-13
feature: added "last-modified" header to supported record types

michael-genson:
As discussed in #1020, I've implemented support for the "last-modified" header on CRUD operations (when the schema supports it).

I achieved this by creating a custom `APIRoute` class that checks the response for the `updateAt` attribute. If it's present, it's added as a "last-modified" header.

While any model can hook into this, I only added it for shopping lists and shopping list items. In order to utilize this, the "updateAt" attribute from the `BaseModel` has to be exposed on the schema.

I also added support for the `HEAD` route for shopping lists/list items. FastAPI handles this automatically, all I had to do was duplicate the `@get(...)` decorator and replace it with `@head(...)`.

---

One issue that users will likely run into is that updates to shopping list items does not update its parent shopping list. We'll want to address this at some point (similar to limitations discussed in PR #1355) but probably after 1.0.0.

### DIFF
--- a/mealie/routes/_base/routers.py
+++ b/mealie/routes/_base/routers.py
@@ -1,6 +1,11 @@
-from typing import Optional
+import json
+from collections.abc import Callable
+from enum import Enum
+from json.decoder import JSONDecodeError
+from typing import Optional, Union
 
-from fastapi import APIRouter, Depends
+from fastapi import APIRouter, Depends, Request, Response
+from fastapi.routing import APIRoute
 
 from mealie.core.dependencies import get_admin_user, get_current_user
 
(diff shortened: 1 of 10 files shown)


michael-genson:
The test that's failing is asserting that the payload is serializable; it's getting caught up on the new datetime properties. Not sure why - if I test the route myself I get a serialized response just fine, but for some reason the test fails. Any ideas?

hay-kot:
> The test that's failing is asserting that the payload is serializable; it's getting caught up on the new datetime properties. Not sure why - if I test the route myself I get a serialized response just fine, but for some reason the test fails. Any ideas?

It's cause the endpoint uses the FastAPI serializer/deserializer which has much better support for complex objects. You can use the [Test utility here](https://github.com/hay-kot/mealie/blob/mealie-next/tests/utils/jsonify.py) to pre-jsonify the dict before passing it to the `put` call in the test.

michael-genson:
I'm lost on this one, tests worked fine locally, but there seems to be something wrong with Poetry. I found this GitHub thread with a similar symptom:
https://github.com/python-poetry/poetry/issues/2304

Doesn't really make any sense to me since it was working fine the other day, and nothing changed as far as I know

hay-kot:
Looks good to me.

@xxxserxxx - does this give you what you were looking for?

xxxserxxx:
@hay-kot Yes, from the description it does. I haven't tested it because I'm running Mealie out of a container and don't have an easy way to patch it; as soon as I find time I'll do a test.

Issue #1387, opened 2022-06-15
[v1.0.0b3] - Bulk Import bails if a URL fails to import


epheterson:
### First Check

- [X] This is not a feature request
- [X] I added a very descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Mealie documentation, with the integrated search.
- [X] I already read the docs and didn't find an answer.

### What is the issue you are experiencing?

I'm moving from a Google Doc to Mealie and copy/pasted over 100 URLs. The import went through a few then bailed out because it failed to import one. Since several will fail to import, this makes the bulk import process more tedious.

It'd be great if the import could proceed even if one fails, and present a list of URLs that failed to import so I can handle them manually.

### Deployment

Docker (Synology)

hay-kot:
Can you please provide some more information on your issue? 

- Application Details from the site settings page under bug report https://demo.mealie.io/admin/site-settings
- A Reproducible example, what steps do I need to take to reproduce the error described?

epheterson:
```
**Details**
Version: v1.0.0beta-3
Build: 56eb0bca7142715984c8028c4c881f086e979564
Application Mode: Production
Demo Status: Not Demo
API Port: 9000
API Docs: Enabled
Database Type: sqlite
Recipe Scraper Version: 14.3.0

**Checks**
Secure Site: No
Server Side Base URL: Yes
LDAP Ready: No
Email Configured: No
Docker Volumes: success
```

Reproduction Steps:

1. Copy a list of recipe URLs with one that will fail to import, including example below
2. Navigate to Mealie Home > Create > Import > Bulk URL Import
3. Click "Bulk Add" to paste a list of URLs, paste the list from below
4. Save, Submit and let it import
5. Notice it hits a failure and bails on the rest.

URL List that had a failure in it:
https://www.cookincanuck.com/teriyaki-turkey-rice-and-vegetable-casserole-recipe/
https://thepounddropper.com/skinny-sloppy-joes/
https://lifemadesimplebakes.com/crispy-lemon-pepper-chicken/
https://www.themediterraneandish.com/lemon-garlic-salmon-recipe/
https://www.delish.com/cooking/recipe-ideas/a28688815/spicy-shrimp-stuffed-avocados-recipe/
https://www.ffactor.com/recipes/sesame-snappeas/
[...]

PR #1396, opened 2022-06-15
fix: bulk URL imports give up if one URL fails

michael-genson:
As raised in #1387, if you supply the bulk URL importer a list of URLs, and one fails, it doesn't try to import the rest of them. This changes the behavior to log the failed URL and continue parsing the remaining URLs.


michael-genson:
You beat me by a couple of minutes :) #1394

Issue #1410, opened 2022-06-17
[v1.0.0b] [Task] - Recipes Should Use `created_at` instead of `date_added`


hay-kot:
### What is the problem this task addresses?

Recipes can appear in the "Recent" view in an odd order because they are not _truly_ sorted by created time, but by created date.

### Proposed/Possible Solution(s)?

We should expose the `created_at` field to the Schema and API and update the sort fields to reflect the `created_at` field. There is likely some edge cases on this one, so it will require some due diligence.

PR #1417, opened 2022-06-17
fix: sort recent recipes by created_at instead of date_added

michael-genson:
This is a fix for #1410. I changed the default sort to use created_at, and I also wrote it in the frontend to explicitly sort by created_at in case we ever change the default behavior.

Optionally we can switch this to the updated timestamp, which might be more desirable in recent, but I'm not sure. Regardless it's in the schema now so it's an easy toggle.


michael-genson:
These backend tests are out to get me

michael-genson:
@hay-kot do you know why the backend tests are failing? Am I doing something wrong? AFAIK they shouldn't be failing, the tests are running fine on my machine, and the errors seem package related

hay-kot:
The cache for the runner is probably broken, you can try rebasing and pick up the dependency changes and bust the cache.

michael-genson:
Ah I didn't think about the dependency changes, thanks!

PR #1447, opened 2022-06-24
feature: added route for deleting a recipe reference from a shopping list without modifying the list items

michael-genson:
Added a simple route to delete the entire recipe reference from a shopping list. This allows you to remove the reference without changing the items (which is useful when trying to clear a shopping list without deleting it)

### DIFF
--- a/mealie/routes/groups/controller_shopping_lists.py
+++ b/mealie/routes/groups/controller_shopping_lists.py
@@ -231,7 +231,7 @@ class ShoppingListController(BaseUserController):
         return data
 
     # =======================================================================
-    # Other Operations
+    # Recipe References
 
     @router.post("/{item_id}/recipe/{recipe_id}", response_model=ShoppingListOut)
     def add_recipe_ingredients_to_list(self, item_id: UUID4, recipe_id: UUID4):
@@ -256,6 +256,24 @@ class ShoppingListController(BaseUserController):
     @router.delete("/{item_id}/recipe/{recipe_id}", response_model=ShoppingListOut)
     def remove_recipe_ingredients_from_list(self, item_id: UUID4, recipe_id: UUID4):
         shopping_list = self.service.remove_recipe_ingredients_from_list(item_id, recipe_id)
+        if shopping_list:
+            self.event_bus.dispatch(
+                self.deps.acting_user.group_id,
+                EventTypes.shopping_list_updated,
+                msg=self.t(
+                    "notifications.generic-updated",
+                    name=shopping_list.name,
+                ),
+                event_source=EventSource(
+                    event_type="bulk-updated-items",
+                    item_type="shopping-list",
+                    item_id=shopping_list.id,
+                ),
+            )
+
+    @router.delete("/{item_id}/recipe-references/{recipe_reference_id}", response_model=ShoppingListOut)
+    def remove_recipe_reference_from_list(self, item_id: UUID4, recipe_reference_id: UUID4):
+        shopping_list = self.service.remove_recipe_reference_from_list(item_id, recipe_reference_id)
         if shopping_list:
             self.event_bus.dispatch(
                 self.deps.acting_user.group_id,
--- a/mealie/services/group_services/shopping_lists.py
+++ b/mealie/services/group_services/shopping_lists.py
@@ -47,9 +47,9 @@ class ShoppingListService:
 
     def consolidate_list_items(self, item_list: list[ShoppingListItemOut]) -> list[ShoppingListItemOut]:
         """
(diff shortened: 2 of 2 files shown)


hay-kot:
This

> This allows you to remove the reference without changing the items

and 

> (which is useful when trying to clear a shopping list without deleting it)

Seem opposite to me, don't you want to remove the reference and all linked items from the list then, like what you can do in the image below? 

image

Could you elaborate on the use case a little more here so I can understand it better? 

Also, if we are to add this it will need test coverage as well.

michael-genson:
Sure, it's more of a convenience than anything really. Say you want to add a recipe from Mealie to an external shopping list. One easy way to do this is:
1. Add the recipe to your Mealie shopping list
2. Receive the event that a recipe was added to a shopping list
3. Fetch all items from the shopping list
4. Clear the shopping list (by deleting all items and all recipe references)

The extra utility this route brings is that I can indiscriminately delete everything, then delete all of the references, as opposed to removing each recipe first, then delete any leftover items. The current implementation also only allows you to delete one recipe at a time, so if you were to add two of the same recipe you'd have to delete it twice.

The key complexity lies in the missing detail in step 2: I don't know which items were added recently, and which were already there. Ideally the event gave me more detail, but the route for adding a recipe only returns the entire list, not the new items. If you don't like the idea of this route I might tackle that instead, but there's some other database related stuff I think needs some work (i.e. treating shopping list item events different than shopping list events).

I _could_ track changes in some external database, then compare the previous state with the new state, but that's a bit more complex than what I'm using it for. It's easier to just be indiscriminate (the route is also faster because it doesn't do all the item comparisons).

---

What I really want to do is just keep the two lists in sync (the external list and Mealie) but Todoist, the list I'm syncing to, had some bad inconsistencies leading to duplicate items, so this was a fair compromise. When I have some more time I want to try to debug it, but I found other people having similar issues so I might just trash the idea.

What I really, _really_ want to do is integrate Alexa's shopping list with Mealie's shopping list, but that API is a monster (you can only access the shopping list through a skill, you can't access it directly, and the skill uses some fancy layer on top of OAuth that I haven't fully wrapped my head around yet) so I haven't had time to figure it out. Todoist syncs to it automatically, so it's my proxy.

---

Hopefully something in there made sense... :) I wouldn't mind ditching this and working on improving the existing endpoints instead, let me know what you think!

hay-kot:
I'm okay with merging this as a temporary stop-gap for your use case, but I think we can do a little better to allow proper synching across applications

With the new WebHook implementation, [I added a new field to leave the door open for extension](https://github.com/hay-kot/mealie/blob/2f7ff6d1780ef206f3f119041f2063e2b918555f/mealie/db/models/group/webhooks.py#L20-L23). Basically we have a new string field that we can specify to use Web Hooks on the event bus. I think we could refactor the event bus and add some functionality around Web Hooks to get you the information you need 

1. We reimplement the Event Bus to _actually_ work like an even bus, I.E. Controllers post events and we implement handlers that listen for the events and publish data based off that. 
2. I think off the bat we could just have two, Apprise and Web Hook listeners. This would allow us to use the same event system, but allow for generic web hooks to be used to embed more granular information. 
3. I think if we make those changes we'd just need to define what sort of data we'd need to embed in the web hooks. I'd have to defer to you on this one because you're actually trying to use the API so you'd know best. 

I don't think we'd need any database changes, but I can help facilitate any that do need to be made... What do you think, would that help you get closer to what you want with the shopping lists?

michael-genson:
>We reimplement the Event Bus to actually work like an even bus, I.E. Controllers post events and we implement handlers that listen for the events and publish data based off that.

Yeah I think this is great - so we'd unify the apprise notifications with webhooks. So we'd eliminate the `EventSource` class and add basically build that into the webhook?

>I think if we make those changes we'd just need to define what sort of data we'd need to embed in the web hooks. I'd have to defer to you on this one because you're actually trying to use the API so you'd know best.

That makes sense. I have a few questions on the overall structure of it all, but I can definitely take a crack at requirements to give us something to work off of. I can make a discussion post for that and we can move from there. Your comments here give me a vague idea of how it all fits together, but I think once we lay it out it'll make more sense.

---
Given all of that I'm happy to table this PR - at the very least we can reconsider when we implement some of the broader changes you suggested here, but my gut says we probably won't need it.

hay-kot:
I've tried to layout my thoughts in this issue

- https://github.com/hay-kot/mealie/issues/1463

Have a look and let me know what you think, we can continue the broader conversation of these ideas there. 👍

michael-genson:
Closing this since there's broader work to be done per #1463

PR #1450, opened 2022-06-25
Feat/event enhancements

michael-genson:
This adds several enhancements to custom event triggers (Forms/JSON/XML):
- Automatically generates a unique id to each event. This allows for uniquely identifying each event which is useful for deduplication (particularly great for integrations with SQS)
- Automatically attaches the actor triggering the event (i.e. the user's id)
- Adds API support to set the event source. This allows API consumers to specify whether they're acting as an integration, allowing for more complex routing logic (example usage below)
- Adds API support for an event source id. This allows API consumers to add their own identifier to an event (example usage below)
---
One issue I ran into when developing a shopping list integration was a feedback loop: if an external system updated the Mealie shopping list, then I received an event that the Mealie shopping list was updated, so I had to go check the external system to see if anything changed... etc. Now I can ignore events triggered by my integration.

Here's an example of how the `event_source` and `event_source_id` can be used to fix this, as well as enable advanced routing logic:
```python

def event_handler(event):
    if event.event_source == "integration" and event.event_source_id == "my_unique_integration":
        # do nothing with this event
        return

    # do something interesting with this event
    ...
```

This also enables one major enhancement to events: to make event subscriptions more granular. In Mealie I should be able to specify that I only want to subscribe to events from `generic` sources, not `integration` sources. But this requires some database stuff that I don't want to mess with yet.

This probably deserves some additional documentation, but I did put some hints in swagger.


hay-kot:
> Adds API support to set the event source. This allows API consumers to specify whether they're acting as an integration, allowing for more complex routing logic (example usage below)

Would it make sense to embed this into a JWT token on login/generation, so if you generate an API token you can specify a token as "Integration" type and it will set a property in the JWT. That way we'd be able to use it across all requests and wouldn't need to add Query to every route? 

> Adds API support for an event source id. This allows API consumers to add their own identifier to an event (example usage below)

Would it make more sense to use a custom header for this? Something like `X-Source-ID`. We can set that up on the base controller so it's accessible everywhere. 

---

Given all the work you're doing in this space we ought to have a documentation page on Integrations and what not - but that's an aside.

michael-genson:
> Would it make sense to embed this into a JWT token on login/generation, so if you generate an API token you can specify a token as "Integration" type and it will set a property in the JWT. That way we'd be able to use it across all requests and wouldn't need to add Query to every route?

Yeah, I like that waaay better than my implementation. And that probably removes the need to even have the source id, you can just leverage the JWT token and call that your source id. Should I extend the `LongLiveTokenIn` class and put the attributes there? Also, my experience with database migrations is with tiny Flask applications with 2-3 users, so I might need some help there (I'm assuming we'd need one to extend the object).

> Given all the work you're doing in this space we ought to have a documentation page on Integrations and what not - but that's an aside.

Yeah definitely, I think between this and your comments on #1447 we'll be in a stable spot and I can write those docs up!

hay-kot:
> Should I extend the LongLiveTokenIn class and put the attributes there? Also, my experience with database migrations is with tiny Flask applications with 2-3 users, so I might need some help there (I'm assuming we'd need one to extend the object).

Yes you can extend the LongLiveTokenIn to add a event_source property (or whatever), and then embed it into the JWT data [here]( https://github.com/hay-kot/mealie/blob/2f7ff6d1780ef206f3f119041f2063e2b918555f/mealie/routes/users/api_tokens.py#L22) and then it should be available to grab on the server side and access on any request. You shouldn't need to make any changes to the database though since all the data is stored in the JWT.

michael-genson:
Closing this since there's broader work to be done per #1463

PR #1468, opened 2022-07-04
Enhancement: Advanced Filtering API

michael-genson:
This extends the `get_all` routes to include a robust query filter. It allows the user to combine multiple relational and logical operators and group them arbitrarily.

Supported Relational Operators : `=, <>, >, <, >=, <=`
Supported Logical Operators: `AND, OR`

This enables simple query strings such as:
`name = "Pasta Fagioli" OR name = "My Favorite Recipe"`
`createdAt > "2021-02-22"`

But parenthesis can be used for more robust filters such as:
`name="cup" OR (useAbbreviation=f AND (name="ounce" OR createdAt > "2022-07-04T19:49:27Z"))`

The implementation uses a combination of enums (for operators), attribute validation (to prevent calling the database with an attribute that doesn't exist) and parameterized queries (to prevent SQL injection) for validation. It also throws a `400` error when predictable mistakes are made (invalid attributes, unbalanced parenthesis, or invalid date/datetime formats).

The meat of this PR is in `query_filter.py` which does all the parsing and validation; the remaining codebase remains largely unchanged.

---

Assuming this PR is accepted, I will do a proper writeup in the docs

### DIFF
--- a/tests/unit_tests/repository_tests/test_pagination.py
+++ b/tests/unit_tests/repository_tests/test_pagination.py
@@ -2,11 +2,13 @@ import time
 from random import randint
 from urllib.parse import parse_qsl, urlsplit
 
+import pytest
 from fastapi.testclient import TestClient
 from humps import camelize
 
 from mealie.repos.repository_factory import AllRepositories
-from mealie.schema.recipe.recipe_ingredient import SaveIngredientUnit
+from mealie.repos.repository_units import RepositoryUnit
+from mealie.schema.recipe.recipe_ingredient import IngredientUnit, SaveIngredientUnit
 from mealie.schema.response.pagination import PaginationQuery
 from mealie.services.seeder.seeder_service import SeederService
 from tests.utils.fixture_schemas import TestUser
@@ -117,7 +119,8 @@ def test_pagination_guides(database: AllRepositories, unique_user: TestUser):
         assert source_param in prev_params
 
 
-def test_pagination_filter(api_client: TestClient, database: AllRepositories, unique_user: TestUser):
+@pytest.fixture(scope="function")
+def query_units(database: AllRepositories, unique_user: TestUser):
     unit_1 = database.ingredient_units.create(
         SaveIngredientUnit(name="test unit 1", group_id=unique_user.group_id, use_abbreviation=True)
     )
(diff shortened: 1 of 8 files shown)


hay-kot:
This is nuts. Looks really powerful and something I want to get into Mealie, It looks like a pretty solid implementation I have a couple of questions. 

1. Did you base this off of any other Web API? 
2. Have you thought about how we would implement this on the frontend? At first glance it seems pretty complicated to wrap a UI around this other than exposing a "SQL Query Mode"
3. I do want to break up your test into more isolated cases, I can do that though if you don't want to take a stab at it. I know utilizing Pytest stuff can be a bit magical....

michael-genson:
Thank you! I was really happy with how it came out. I wrote the implementation myself, but I took inspiration from Freshdesk's relatively-new filter API which is functionally very similar:
https://developers.freshdesk.com/api/#filter_tickets

---

For a frontend wrapper, if we wanted to _fully_ support it I've seen things that are essentially query builders:
https://docs.oracle.com/en/cloud/saas/netsuite/ns-online-help/section_N647582.html

Or something like this (this is for calculations, not queries, but we could adapt a similar UI):
https://knowledge.hubspot.com/crm-setup/create-calculation-properties

The alternative would be something more lightweight that doesn't expose the full functionality. We could pull the schema, list all the fields, and allow users to select the operator/value. Something like:
```
[ ] Name       [=, <>, >, <] [Input]
[ ] Created At [=, <>, >, <] [Input]
[x] Updated At [=, <>, >, <] [Input]
...etc.
```

You can't get as granular with this (mainly because of parenthesis) but it would be simpler to implement.

---

I'm happy to add some more tests! Pytest is a bit of an enigma to me but using what you've already done as reference has made it pretty easy. What sort of isolated cases did you have in mind? Just more specific stuff?

tests/unit_tests/repository_tests/test_pagination.py
```diff
@@ -121,3 +115,79 @@ def test_pagination_guides(database: AllRepositories, unique_user: TestUser):
...
+
+    # invalid datetime format
+    qf = 'createdAt="this is not a valid datetime format"'
+    response = api_client.get(route, params={"queryFilter": qf}, headers=unique_user.token)
+    assert response.status_code == 400
+
+    # invalid attribute
+    qf = 'badAttribute="test value"'
+    response = api_client.get(route, params={"queryFilter": qf}, headers=unique_user.token)
+    assert response.status_code == 400
```
hay-kot: > I'm happy to add some more tests! Pytest is a bit of an enigma to me but using what you've already done as reference has made it pretty easy. What sort of isolated cases did you have in mind? Just more specific stuff?

I think you cover most of the cases you've introduced. Basically I'm looking for each one of these comments to be their own test so it's easier to isolate specific cases that I end up breaking. You could probably accomplish this by running the setup in a fixture and injecting what you need into the function from there and just set the fixture up to only run once.

hay-kot (commented):
Very Cool, I added some notes on how I'd like to see the test broken out, if you don't get to it by Sunday Morning, I'll take a stab at it some I can get this merged before I take some time off. 

Thanks for the additional context on your though process. A frontend query builder could be pretty interesting, I'll have to take a stab at that at some point.

tests/unit_tests/repository_tests/test_pagination.py
michael-genson: Ah I see, so separate them out into their own functions so it handles them as separate tests. That makes sense! I'll see if I can get it done tonight or tomorrow morning; I'll be out of town for a few days (leaving tomorrow) so if I don't get to it by then it's all yours :)

PR #1487, opened 2022-07-16
LDAP improvements

Elegant996:
This PR introduces two commits which improve the LDAP authentication system that mealie currently uses.

The first commit of the PR adds support for a `LDAP_BASE_DN` field which allows clients to properly specify the starting point of the directory for which to lookup clients. This **also** resolves an issue where Active Directory users would have issues authenticating as AD users use the form `{}@example.com` instead of the User DN. Both are now supported.

In addition to the above, the first commit also populates the `user` in the mealie DB with the approriate values from the LDAP server for `full_name` and `email` instead of just copying `username` for all fields. A check was also added to lookup the `user` email in AD and then authenticate against their `username` should there be a match in the mealie DB. This allows LDAP clients to login using their `email` on _subsequent_ logins.

The second commit adds support for secure LDAP connections. At present, mealie relies on non-secure LDAP but not all servers accept non-secure connections. This is achieved through two additional parameters: `LDAP_TLS_INSECURE` and `LDAP_TLS_CACERTFILE`, both of which are optional. Clients may either disable verification or specify the require certificate to establish the secure connection.

Thanks!


hay-kot:
From the discussion

> @hay-kot Reviewed the PR again, should be good to go. Not sure how to prevent the error regarding security.py from generating. Thanks!

What have you tried to fix them? 

Looks like you need to run black to fix your formatting.

mealie/core/security/security.py
```diff
@@ -47,23 +47,38 @@ def user_from_ldap(db: AllRepositories, session, username: str, password: str) -
...
+        user_bind = settings.LDAP_BIND_TEMPLATE.format(user.username)
+
     try:
-        conn.simple_bind_s(user_dn, password)
+        conn.simple_bind_s(user_bind, password)
     except (ldap.INVALID_CREDENTIALS, ldap.NO_SUCH_OBJECT):
         return False
 
-    user = db.users.get_one(username, "username", any_case=True)
+    user_dn, user_attr = conn.search_s(settings.LDAP_BASE_DN, ldap.SCOPE_SUBTREE, '(&(objectClass=user)(|(cn=%s)(sAMAccountName=%s)(mail=%s)))' % (username, username, username), ['name', 'mail'])[0]
```
hay-kot: Can you provide a code comment to explain what/why this string is needed? Also, you should probably use an f string instead of an interpolated string, unless there's a specific reasons that I'm missing?

hay-kot (changes requested):

mealie/core/security/security.py
Elegant996: Done and done.

hay-kot:
Hey, I rebased this PR with the intention of merging, but the LDAP tests are failing. 

Can you fix these and verify they're running on your machine and convert the back from a draft?

Elegant996:
Looks like we need to include the attributes when testing, should be good now...

Elegant996:
Resolved this but had to change the check for the result to false. The old method created the user and then searched against it in LDAP to determine if it had admin capability. The new method searches against the username to determine if it exists in LDAP before creating it. Since the user does not exist (dummy LDAP), an empty list of users is returned which ends in false.

PR #1497, opened 2022-07-19
Restore frontend sorting for all recipes

michael-genson:
This refactors a bunch of the frontend to support sorting for all recipes. Specifically, this leverages the new backend pagination API to sort using the backend (as opposed to sorting on the frontend).

I was unable to figure out how to get categories and tags to work properly, we might need to make some backend changes to get all recipes filtered by category/tag. Maybe there's a way to do it already, in which case we can merge the two solutions.

Since I only modified the behavior for all recipes, I introduced a few variables to accommodate the split logic:
1. In `RecipeCardSection.vue` I created a second filter which uses different callbacks. I renamed the old one to `sortRecipesFrontend` to indicate that we'll depreciate this once we merge categories and tags into the new logic.
2. I created a `usePagination` prop which helps preload the first page of data (if pagination is being used).

For the new logic: sorting by a new attribute requires us to clear the existing data and re-populate it with the new first page of data. While scrolling, we should extend this data. I created `replaceRecipes` and `appendRecipes` events/functions to accommodate this. I also added support for reverse sorting if you click the same sort twice (e.g. if you click on alphabetical sort it will sort A-Z, but if you click on it again it will sort Z-A).

I did my best to remove any unnecessary code following the all recipes logic refactoring, but I could've missed some of it. I also tried to place everything in the right file/section, but there might be some re-arranging to do since I'm still fairly new to Vue.

---

Next steps are to get categories and tags to use the same backend pagination/ordering methods. It might be an easy change, it might require some backend changes. Once that is done we can fully deprecate the old frontend sorting and remove the dual sorting functions/duplicate filter section.

### DIFF
--- a/frontend/components/Domain/Recipe/RecipeCardSection.vue
+++ b/frontend/components/Domain/Recipe/RecipeCardSection.vue
@@ -35,6 +35,48 @@
           </v-btn>
         </template>
         <v-list>
+          <v-list-item @click="sortRecipesFrontend(EVENTS.az)">
+            <v-icon left>
+              {{ $globals.icons.orderAlphabeticalAscending }}
+            </v-icon>
+            <v-list-item-title>{{ $t("general.sort-alphabetically") }}</v-list-item-title>
+          </v-list-item>
+          <v-list-item @click="sortRecipesFrontend(EVENTS.rating)">
+            <v-icon left>
+              {{ $globals.icons.star }}
+            </v-icon>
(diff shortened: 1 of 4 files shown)


hay-kot:
LGTM! Thanks for this an all your other PRs. I made two small changes

1. fix some of the linter warnings
2. Change the order of the content menu so it was last after the new sort.

I REALLY like that there is ascending and descending setup. I was a little confused on the frontend as to which was one was active so I also added a TODO to somehow indicate which direction is active.

PR #1506, opened 2022-07-27
Implemented local storage for sorting and dynamic sort icons on the new recipe sort card

michael-genson:
As referenced in 233eb1ef91a72306bb13b9a1c62df6d70606c425 :)


michael-genson:
I expanded this PR to include something else I was tinkering with - using local storage to store your sort and view choices. Originally I was planning on making it a different PR, but, because this makes use of a new icon variable, a separate PR would break once this was merged.

hay-kot:
Hey, thanks for this. I went in to make a couple changes and realized I ended up rewriting most of it :( - I didn't mean to take over your PR, sorry about that!

Here's what I changed

- Setup vueuse to use reactive local storage for preferences -  https://vueuse.org/core/useStorage/
- Refactored your switch statement to reuse logic across cases.

Really love that it's not tracking and saving the sort options. I think people are really going to like this 🚀

michael-genson:
Hah! No problem, I didn't know about useStorage, that's much cleaner. I'm still wading my way through Vue so I'm just happy to see there's a better solution

PR #1508, opened 2022-07-27
Added category and tag filters to Recipe Pagination Route

michael-genson:
This adds category and tag filters to the Recipes pagination route (`get /recipes`). Without a request body I don't think there's a way to input arrays, so I limited it to one tag and one category (both optional).

I also made it accept either the slug or the UUID, that way we can easily implement this on the frontend (see #1497).


hay-kot:
Pretty sure you can use Pydantic and/or FastAPI type annotations to pick out a list of strings from the Query parameters so we could support lists. 

Also, is there already existing test coverage for this route that exercises this code, or is that still needed?

michael-genson:
Ahh you're right, if you just include a list as a param FastAPI will move the param to the body, but if you explicitly mark it as a query param it works fine
https://github.com/tiangolo/fastapi/issues/321

Let me modify that and add some tests

michael-genson:
@hay-kot I think the error is a cache thing, I merged into mealie-next and ran the backend tests locally and it's fine. I did have to re-build the dev container though, probably for a similar reason. Is there something else I can do?

---

Unrelated to the error, I added the tests to the recipe repo tests and had to fix some types in older tests using casts. It's a bit messy but I think that's just a consequence of some of the overlapping pydantic models (e.g. `Recipe` vs `RecipeSummary`). We should probably try to unify those a bit at some point, or maybe just update the repo arg types to accept either. I've noticed this for a few models (tags, categories, etc.).

hay-kot:
> I think the error is a cache thing, I merged into mealie-next and ran the backend tests locally and it's fine. I did have to re-build the dev container though, probably for a similar reason. Is there something else I can do?

It is, I'm not sure why the merge didn't bust the cache because the poetry.lock file should have been updated. I'll pull this one down and test it manually. I think once we start branching. I haven't had any issues branching off the most recent mealie-next so hopefully it goes away after these open PR's.

> Unrelated to the error, I added the tests to the recipe repo tests and had to fix some types in older tests using casts. It's a bit messy but I think that's just a consequence of some of the overlapping pydantic models (e.g. Recipe vs RecipeSummary). We should probably try to unify those a bit at some point, or maybe just update the repo arg types to accept either. I've noticed this for a few models (tags, categories, etc.).

The reason behind some of the fragmentation has to do with the `from_orm` transformation that we use. If we use the Recipe model when we fetch large amounts of recipes it absolutely hammers the database with N+1 queries. Reducing the fields significantly speeds up query without having to write manual queries to isolate the values we need. 

The consequence of that is that typing becomes much more difficult. I've tried to get the Generic typing for the Repositories to work but there are many cases where typing could be improved to capture the _actual_ requirements instead of what is used most often. 

There is also the issue of circular imports in the schema folder but I think that is mostly relating to Categories and Tags.

michael-genson:
Sounds good! Let me know if you run into any test issues but I think it should be okay (and of course any feedback).

Regarding the ORM that makes sense, I wondered why it was like that. In that case yeah I think revisiting the typing requirements in the repositories would be a good idea (though probably a big undertaking) just to clean it up a bit, though in the meantime a few casts aren't horrible.

hay-kot:
Test all passed. Yay 🎉

PR #1511, opened 2022-07-28
Create new foods and units from their Data Management pages

michael-genson:
This adds a "Create" dialog to the food and unit data management pages.

One thing I noticed is that if you create a new food/unit, it doesn't appear in the list until you refresh. I'm unsure how to fix that, however this is also true if you delete one (which is an existing functionality) so it's not a new issue. I noticed there's a refresh on the `foodStore` and `unitStore`, and monitoring the network I can see that the new data is fetched, it's just not being re-rendered. Something to look into.

### DIFF
--- a/frontend/components/global/CrudTable.vue
+++ b/frontend/components/global/CrudTable.vue
@@ -3,7 +3,7 @@
     <v-card-actions>
       <v-menu v-if="tableConfig.hideColumns" offset-y bottom nudge-bottom="6" :close-on-content-click="false">
         <template #activator="{ on, attrs }">
-          <v-btn color="accent" class="mr-1" dark v-bind="attrs" v-on="on">
+          <v-btn color="accent" class="mr-2" dark v-bind="attrs" v-on="on">
             <v-icon>
               {{ $globals.icons.cog }}
             </v-icon>
--- a/frontend/pages/group/data/foods.vue
+++ b/frontend/pages/group/data/foods.vue
@@ -54,6 +54,30 @@
       </v-card-text>
     </BaseDialog>
 
+    <!-- Create Dialog -->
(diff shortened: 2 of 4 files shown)


michael-genson:
I'm not sure what the best way to fix the frontend lint errors are. There's probably a better way to handle `this.$refs`, but I'm not sure what it is.

hay-kot:
I'm actually surprised that accessing `this` worked in the setup, it must be because it was in a function? Either way it's not recommended. The recommended way to handle refs with the composition API is documented here -> https://vuejs.org/guide/essentials/template-refs.html

I made a small adjustment and added the `VForm` helper type as well. Good work!

hay-kot:
> I noticed there's a refresh on the foodStore and unitStore, and monitoring the network I can see that the new data is fetched, it's just not being re-rendered. Something to look into.

I noticed this too. It looks like it has to do with the store factory I wrote. When we pass in a `null` reference to the store as a starting `allRefs` variable it's never updated once it's set. We'd somehow need to link the store we pass into the factory to the `null` starter... or do some other refactoring to avoid this problem together. I took a stab at it but didn't get anywhere. Will leave it for later.

michael-genson:
> I'm actually surprised that accessing `this` worked in the setup, it must be because it was in a function? Either way it's not recommended. The recommended way to handle refs with the composition API is documented here -> https://vuejs.org/guide/essentials/template-refs.html

That's exactly what I was missing, thanks! I didn't fully understand `$refs`, I just found a stack overflow thread and it worked. Maybe it was some hoisted var that made it work here? Who knows.

Regarding the refresh, I'll see if I can tinker with it later as well, but you'll probably beat me

PR #1524, opened 2022-08-03
Add option to stay in edit mode after loading from URL.

jelicicm:
The functionality seems good except that I'd expect the checkbox on URL page to be checked by default which doesn't happen.

### DIFF
--- a/frontend/pages/recipe/create/url.vue
+++ b/frontend/pages/recipe/create/url.vue
@@ -151,7 +151,7 @@ export default defineComponent({
       }
       state.loading = true;
       const { response } = await api.recipes.createOneByUrl(url, importKeywordsAsTags);
-      handleResponse(response, stayInEditMode.toString());
+      handleResponse(response, stayInEditMode.value);
     }
 
     return {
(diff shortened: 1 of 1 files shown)


frontend/pages/recipe/create/url.vue
```diff
@@ -114,19 +115,32 @@ export default defineComponent({
...
+      },
+      set(keepEditMode: boolean) {
+        let stay_in_edit_mode = "0";
+        if (keepEditMode) {
+          stay_in_edit_mode = "1";
+        }
+        router.replace({ query: { ...route.value.query, stay_in_edit_mode } });
+      },
+    });
+
```
hay-kot: If you want a default value to be true, you have to initialize it when the component mounts otherwise it will be falsey by default.

jelicicm:
Behaviour now as expected.

hay-kot:
FYI whatever you were doing to resolve conflicts was making a mess of the Git History. I rebased against the `mealie-next` branch all it cleaned up everything.

PR #1530, opened 2022-08-07
Fix pytest app route generation

PFischbeck:
I noticed that some of the routes in `app_routes.py` were out of date, and the `dev/scripts/app_routes_gen.py` script was broken. I fixed it and noticed that some tests were using non-existent routes. I fixed them, although I am not sure how they were working before (in particular those using the non-existent `/api/groups` route to create new groups for test fixtures).

I also noticed that there is another script (`dev/code-generation/gen_pytest_routes.py`, which maybe tries to do the same thing. Is this obsolete, should we clean it up?


hay-kot:
So yeah. The Route generation is sort of a disaster right now. As the project got bigger I found it cumbersome to use it in tests. I had a grand idea of using the Open API Tags to generate modules and then I could import only the modules I needed and didn't have this giant class with a bunch of methods. 

I didn't make much of any head-way with that and only really achieved to fragment how we're doing both generation of test routes, and how we're using test routes. Some tests use a hand-written module and others use the generated test routes.

I'm open to ideas or opinions on which way to go with this. Clearly I can't be trusted anymore with this one alone anymore!

Side note - we should run the generators in CI and check to make sure there aren't diffs so we always know we're up to date...

hay-kot:
Fixed in #1736

PR #1532, opened 2022-08-08
chore: make vue domain components localizable

PFischbeck:
I went through all Domain components of the frontend to make them localizable. Still lots to do 😃


hay-kot:
Wow, awesome work!

PR #1548, opened 2022-08-11
Use `flex-basis`, not width to fix min size bug

BryceStevenWilley:
On mobile screens when following a registration invite link, the page
would be too wide to interact with, extending well over the sides of the
phone.

An image of a mobile size emulated browser, showing overflow

This is because the minimum size of content is set to `auto` (by
default), and accord to the spec (https://www.w3.org/TR/css-flexbox-1/#specified-size-suggestion)
the minimum size of the element is the mimimum size of it's content. The
password strength element in the panel had a width of 500px, making the
entire component overflow the screen.

Changing the width to `flex-basis` instead, allows for the password
strength element to shrink if it overflows the screen.

Screenshot from 2022-08-10 21-32-23

### DIFF
--- a/frontend/pages/register/register.vue
+++ b/frontend/pages/register/register.vue
@@ -173,7 +173,7 @@
                   @click:append="pwFields.togglePasswordShow"
                 />
                 <div class="d-flex justify-center pb-6 mt-n1">
-                  <div style="width: 500px">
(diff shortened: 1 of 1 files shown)


hay-kot:
Nice! Thanks for the fix.

PR #1564, opened 2022-08-15
Create shoppinglist from weekly mealplan

danielpalstra:
This PR adds a create button that adds all recipes from the mealplan to a new shopping list. 

image

### DIFF
--- a/frontend/pages/group/mealplan/planner.vue
+++ b/frontend/pages/group/mealplan/planner.vue
@@ -271,11 +271,11 @@
 </template>
 
 <script lang="ts">
-import { alert } from "~/composables/use-toast";
 import { computed, defineComponent, reactive, toRefs, watch, useContext } from "@nuxtjs/composition-api";
 import { isSameDay, addDays, subDays, parseISO, format } from "date-fns";
 import { SortableEvent } from "sortablejs"; // eslint-disable-line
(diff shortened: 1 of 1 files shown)


michael-genson:
Does the create button have other functions, or is it only for adding to the shopping list? If it's the latter, we should probably name it something like "Add to Shopping List". You can use the "formatListCheck" or "cartCheck" icon to match the shopping list icon

danielpalstra:
> Does the create button have other functions, or is it only for adding to the shopping list? If it's the latter, we should probably name it something like "Add to Shopping List". You can use the "formatListCheck" or "cartCheck" icon to match the shopping list icon

I would like to change the title of the button, but I couldn't find a way to set the text yet.

danielpalstra:
I've found an example so I renamed the Create button to "Add to Shopping List".

hay-kot:
Hey! Thanks for the PR. My main concern with this implementation is that it's possible to create a shopping list with duplicates since the scroller has some overlap when paging through the meal planner. I also think that that we should be able to add to an existing Shopping list since I've heard most folks have two or three shoppings lists they always keep around. 

Maybe we could implement a dialog that would let them select recipes from the current page and then add those to a selected shopping list or create one?

PR #1565, opened 2022-08-15
chore: make vue global components and layouts localizable

PFischbeck:
Step by step, we're getting closer to making the full frontend localizable 😄


hay-kot:
We don't deserve you! Thank you!

Issue #1463, opened 2022-06-27
[v1.0.0b] [Task] - Refactor Event Bus to Support Web Hooks


hay-kot:
### What is the problem this task addresses?

Currently Mealie has limited support for external integrations and does not provide any ability to adequately "sync" data between systems. The current even-bus implementation for Mealie is very limited and can't provide the necessary data that's needed to properly integrate deeply with different systems. 

Related PRs

- #1447

### Proposed/Possible Solution(s)?

**From PR 1447**

- We reimplement the Event Bus to actually work like an even bus, I.E. Controllers post events and we implement handlers that listen for the events and publish data based off that.
- I think off the bat we could just have two, Apprise and Web Hook listeners. This would allow us to use the same event system, but allow for generic web hooks to be used to embed more granular information.
- I think if we make those changes we'd just need to define what sort of data we'd need to embed in the web hooks. I'd have to defer to you on this one because you're actually trying to use the API so you'd know best.

**Here's a Diagram and what I'm thinking**

```mermaid
graph TD;
    Request --> B[Controller]
    B -- Publish Event --> C[Event Bus]
    C -- Event --> D[WebhookSubscriber]
    C -- Event --> E[AppriseSubscriber]
    D --> DB[Check Database For Entries]
    E --> DB
    DB -- Has Entries --> Publish
    DB -- No Entries --> NA[No Action]

```

Effectively the Controllers would publish events to the Event Bus which would kick off a background task that would call the subscribers allowing the Controllers to finish the request while the events are processed in the background. 

Next the EventBus class would call the subscribers and the subscribers would go take it from there. In the case of the Apprise we'd query the database for Notifies and perform the Apprise notifications for those events. In the case of Web Hooks, we query the Web Hooks table for Enabled listeners for specific events. 

---

Now that I write this down, It may make more sense to leave the Web Hook Table that way it is and instead add a field to the Notifiers table that just toggles a `is_webhook` field so if it's a Web Hook we send a POST request with the full data context, otherwise we use the apprise integration.

michael-genson:
Yeah, I like this. And I agree it would be simpler to allow notifies to be marked as a webhook - right now "webhooks" are only schedule-based, while "notifiers" are what I'm using for event-based integrations.

So if you marked a notifier as a webhook, you'd get some different configuration options (namely you're providing the `POST` URL instead of the apprise URL).

---

[...]

PR #1574, opened 2022-08-18
Event Bus Refactor

michael-genson:
The goal of this PR is to address refactoring the event bus, as outlined #1463. This PR does not fully address #1463, as it has left the existing webhook functionality untouched.

Changes to the Event
---

The Event (formerly a combination of `EventMessage`, `EventSource`) is a new pydantic model which unifies and extends the existing event payload. It's made up of several sub models:
- `EventBusMessage` (unchanged from its previous implementation)
- `EventTypes` (unchanged from its previous implementation, links to user-selected subscriptions)
- `integration_id` - metadata which can be set during api token generation (as discussed in #1450)
- `EventDocumentDataBase` - a base model that can be extended for each document type. More on this below

The new event document data param contains additional metadata for the notification subscriber. This aims to accomplish what the old `EventSource` did, but with cleaner structure. The base model `EventDocumentDataBase` is comprised of:
- `EventDocumentType` - an enum which is more granular than the `EventTypes` enum (more on this below)
- `EventOperation` - the specific CRUD operation (this can be different than the `EventTypes` implied operation, more below)

In addition to the base model, each record type has its own specific model. For instance, looking at `EventRecipeData`, there is an extra param `recipe_slug`. For `EventCookbookData` there is a `cookbook_id` param. The reason there are different models for each document type is because a "standard" model only covers a small number of documents (recipes, shopping list items, and several variations of bulk changes all have different models).

### The Document Type and Operation
The `EventTypes` enum gives information regarding the document type and the operation, so why introduce two explicit `EventDocumentType` and `EventOperation` params? This solves a few problems that `EventTypes` doesn't solve:

1. Child records should fire events if you are subscribed to parent records. Right now the only instance of this is shopping lists. If a shopping list item is updated, I should receive a notification that the shopping list was updated (besides the parent list not actually being changed). For instance, if I create a new shopping list item, this allows us to set the following params:
```python
event_type = EventTypes.shopping_list_updated  # the shopping list was updated with a new item
document_type = EventDocumentType.shopping_list_item  # the actual document type that was created is the list item
operation = EventOperation.create  # the item was created, even though the parent was updated
```
2. Sometimes certain operations have different return types. For instance, if recipes are added in bulk, it's done as a background task, so we don't know what the new slugs are (or if they were even successful). Because of this, the API returns a `job_id` instead of a list of slugs. This allows us to set the following params:
```python
event_type = EventTypes.recipe_created  # we are creating recipes
document_type = EventDocumentType.recipe_bulk_report  # we're returning a report, not a list of slugs
operation = EventOperation.create
```

Changes to the Event Bus
---
The Event Bus has been generalized to support multiple listeners/publishers. These changes aren't that drastic, I've just abstracted the workflow a bit.

The Event Bus now contains a list of listeners, each of which implement their own subscription and publishing functions. For now I've only created one listener: the `AppriseEventListener`, but it's a sub model of `EventListenerBase` which should be used to create new listeners (e.g. a webhook listener, eventually). The `AppriseEventListener` is mostly identical to the old Apprise implementation, with some tweaks to accommodate the new event format.

Known Issues
---
~~Because of the new event bus implementation, the old `test_notification` route broke. I think this route was being used on the frontend, so I want to fix it, but for now I've commented it out.~~ this works again, although we'll have to re-do it when we add more notifiers.

The recipe bulk actions endpoints don't fire events. This isn't a new issue, I actually only noticed that today. There's a `TODO` already there about having those endpoints return some data, which would be required for the event to fire. I've added a second `TODO` to address this eventually.

Next Steps
---
The obvious next step is to integrate the webhook functionality into this event bus. That will unify our two disparate notification workflows (which is the main goal of #1463).

Additionally, we should probably write tests for some of these new flows, but I'm not sure how to test the notifiers without actually sending notifications.


mealie/routes/groups/controller_cookbooks.py
```diff
@@ -51,27 +56,39 @@ def get_all(self, q: PaginationQuery = Depends(PaginationQuery)):
...
-                self.user.group_id,
-                EventTypes.cookbook_created,
-                msg=self.t("notifications.generic-created", name=val.name),
-                event_source=EventSource(event_type="create", item_type="cookbook", item_id=val.id, slug=val.slug),
+                request=request,
+                group_id=self.user.group_id,
+                event_type=EventTypes.cookbook_created,
+                document_data=EventCookbookData(operation=EventOperation.create, cookbook_id=cookbook.id),
+                message=self.t("notifications.generic-created", name=cookbook.name),
             )
```
hay-kot: This seems like a really verbose way to call this in each instance. I think there's going to be a ton of these in the code-base. Is there any way to abstract some of the common parameters? I have two ideas

1. Utilize the controllers to instantiate the EventBuss and have a `publish` event method that pre-captures the Request and group_id so we don't have to pass those in.
2. Have a class method on the EventBusService that requires a the group_id and request method be provided and then have a special method on the event bus that wraps `dispatch` but pre-sets the request and group_id.

mealie/routes/users/api_tokens.py
```diff
@@ -15,17 +15,22 @@ class UserApiTokensController(BaseUserController):
...
     ):
         """Create api_token in the Database"""
 
-        token_data = {"long_token": True, "id": str(self.user.id)}
+        token_data = {
+            "long_token": True,
+            "id": str(self.user.id),
+            "name": token_params.name,
+            "integration_id": token_params.integration_id,
+        }
```
hay-kot, line 27: I know adding this to the JWT was my idea, but I'm wondering if we _need_ that validation that using a JWT provides? Is there any security reason we couldn't just expose a `Mealie-Integration-Id` header and allow any user to provide it and just use the header instead of trying to grab the integration_id out of the JWT? 

I think that would simplify things a bit, but I'm not sure if there's a case I'm not considering.

hay-kot (commented):
Overall looks _really_ _really_ good. Super impressed with everything you've done so far. There are only really two things that stick out to me. I feel like there's probably a better way to encapsulate some more of the complexity of publishing an event to make it a littler more ergonomic to use.

If you're tapped out on this one, let me know and I can take a stab at improving the implementation you've got here. 

Great work!

mealie/routes/users/api_tokens.py
michael-genson: My thought with the JWT was that you can add it to your token and just forget about it, that way you don't have to add it to each call. Otherwise, I don't see an issue with using a header instead. Is there an easy way to add it to every route as a header param, like at the router level?

Another option would be to ditch the integration id entirely and just rely on the token name, though I feel dirty using the name for anything other than a human-friendly label.

mealie/routes/groups/controller_cookbooks.py
michael-genson: That makes sense, would that go in the router? i.e. the router captures the request/user and sends it to the event bus. That said I think the only thing I needed the request for was the token which has the integration id, so we may not event need the request if we're moving that to a custom header.

Actually, now that I think about it, isn't the controller instantiated with the group_id? Then we could just pass it to the event bus.

hay-kot: Yeah, the controller already has most everything you need. You could probably keep the JWT implementation but simplify it a bit by having a `integration_id` property on the controller that would return the `integration_id` from the JWT, and then have a method on the controller that contains those values to simplify the call in each route.

I don't think there is too many valid cases to reach for the router with the Controller based system we have. That sort of fills the void there.

mealie/routes/users/api_tokens.py
hay-kot: Since you've already done the work to put it in the JWT, I'm okay with going down that road. The overall difference between the JWT and header is pretty minimal so realistically it probably doesn't matter. It is probably more ergonomic for clients to not have to think about that.

mealie/routes/groups/controller_cookbooks.py
michael-genson: This is probably a stupid question, but how do I get the token/integration id to the controller? I see `BaseUserController` which has a `PrivateUser` that depends on `get_current_user` which gets the user from the token, but I don't actually see how the token gets passed to `get_current_user`. Looks like some FastAPI dependency magic.

michael-genson:
I'm not sure if this was the best way to get the integration id into the controller, let me know if there's a better way.

I created a CRUD controller to house the event bus and the new `publish` method that passes the integration_id and the group_id to the dispatch function on the event bus. Theoretically this could live on the event bus instead, but I feel like the event bus shouldn't serve as both the publisher and the dispatcher, if that makes sense. Curious to hear your thoughts on it.

The routes are much cleaner now: they don't require the request anymore, and the publish method is 3 params instead of 5:

```python
self.publish_event(
    event_type=EventTypes.cookbook_created,
    document_data=EventCookbookData(operation=EventOperation.create, cookbook_id=cookbook.id),
    message=self.t("notifications.generic-created", name=cookbook.name),
)
```

michael-genson:
Oh and I restored the test notification route, although we'll have to re-do it when we add more listeners (but at least it still serves its original purpose)

hay-kot:
This looks excellent, thank you!

PR #1587, opened 2022-08-21
refactor: recipe-page

hay-kot:


michael-genson:
Oh man this is looking much better. A couple small things that may or may not be worth considering:
- We pass a lot of recipe attributes between different components as props (slug, settings, recipe name, etc.), is there a reason against passing the whole recipe as a prop instead? Looking at `RecipeInstructions` for instance: it takes the assets, ingredients, the toggle for showing amounts, the id, and the slug. It might be cleaner to just pass the recipe and let the component access the relevant parts. I have a suspicion this is intentional, just wanted to mention it.
- I noticed most components take the slug instead of the id. I know the API uses the slug for recipes so this makes sense, but if we ever want to move away from recipe slugs to recipe ids now might be a good time.
- This is probably out of scope for this PR, but when messing with the print view I ran into a problem when trying to "preview" it, because of how we use `@media print` in the CSS. I was thinking about making a toggleable class that gets toggled for printing and for the print preview. Just mentioning it here in case you find a better way while refactoring this.

Love this way better so far than the current implementation!

hay-kot:
@michael-genson 

> We pass a lot of recipe attributes between different components as props (slug, settings, recipe name, etc.), is there a reason against passing the whole recipe as a prop instead? Looking at RecipeInstructions for instance: it takes the assets, ingredients, the toggle for showing amounts, the id, and the slug. It might be cleaner to just pass the recipe and let the component access the relevant parts. I have a suspicion this is intentional, just wanted to mention it.

That's an excellent point, I've done some updates to slim those down. Props are supposed to be "Read Only" by nature, but because it's an object we're kind of fudging some of the event handling and updating the object inside the children which is generally a bad pattern. I think in Vue 3 they have some better APIs for what we're doing so I was putting off that until we eventually migrate.

- For anything Read Only we absolutely can use the entire recipe
- For anything that we modify we _should_ use the v-model or the <name>.sync modifier to handle updates

Those are goals and not the reality though! 

> I noticed most components take the slug instead of the id. I know the API uses the slug for recipes so this makes sense, but if we ever want to move away from recipe slugs to recipe ids now might be a good time.

I'm more inclined to make a non-functional changes in this PR and use another one to submit a change in API design like that. We'll need to move that way eventually though, I'm sure. 

> This is probably out of scope for this PR, but when messing with the print view I ran into a problem when trying to "preview" it, because of how we use @media print in the CSS. I was thinking about making a toggleable class that gets toggled for printing and for the print preview. Just mentioning it here in case you find a better way while refactoring this.

You can actually use your dev tools to preview print mode for development, it was pretty life changing when I found out. 

https://stackoverflow.com/questions/9540990/using-chromes-element-inspector-in-print-preview-mode

---

Thanks for the feedback!

michael-genson:
> You can actually use your dev tools to preview print mode for development, it was pretty life changing when I found out.
> 
> https://stackoverflow.com/questions/9540990/using-chromes-element-inspector-in-print-preview-mode

Oh wow that's awesome, I had no idea! That makes working on it much easier.

Functionally I meant I need a "print preview" for something like #1576 where we have a custom modal that shows different print options.

PR #1593, opened 2022-08-24
Customizable Print Options

michael-genson:
This is a work-in-progress, need to phone a friend on this one.

---

This PR adds customizable print options, as discussed in #1576. It adds a new print dialog to the context menu (the button was always there, it just executed `window.print()`. The dialog has a print preview and a few customizations options (with room for more). It's also reactive:

_I don't speak German, I just had this recipe imported from a bug report a while ago :)_
2022-08-23_21h50_39

`RecipePrintView` dynamically renders the print preview, and `ContextMenu` handles the dialog and the config options (since that's where the print menu button lives, also to try to avoid conflict with the upcoming changes to the `_slug` page).

While the dialog looks nice, there's two primary issues preventing this from being complete:

1. I can't get the template to actually print. I tried using the `@media print` CSS from the previous print view, but I just get a blank page. I also tried wrapping it in a div set to `display: none` that toggles to `display: block` on print. Hoping for some help on this one. I think it has to do with the fact that I put this dialog on the context menu instead of the index, in which case this should probably wait for the `_slug` refactor.
2. We need to add a print button to the dialog menu (which calls `window.print()`, but when I tried I got some strange behavior. I might just be doing it wrong. It wouldn't close the dialog before printing, so the settings would print (even if I explicitly set the state to false first).

A couple other notes:
- Some of the new code might belong in composables. I'm not sure of the best place though.
- I had to duplicate some code from `RecipeTimeCard` to get the times to render properly (doing it via import made manipulating the CSS extremely messy, because of the use of `v-chip`).
- I want the print config options to save in storage, similar to what we did in #1506, but I haven't gotten to implementing that yet. Probably the last thing we want to do (or even punt that to a separate PR).


michael-genson:
merging this with the new recipe changes is going to be more of a hassle than redoing it, so I'm just going to start fresh

PR #1619, opened 2022-08-29
feat: add support for API extras on shopping lists, shopping list items, and food data

michael-genson:
This extends the db to support API extras on shopping lists, shopping list items, and food data. It functions the same way as the recipe extras.

I genericized the recipe `ApiExtras` table declaration to reference a simple base class to extend the life of my `c` and `v` keys. The actual table is unaffected. Similarly I moved the `recipe_extras` decorator for reuse, but otherwise left it functionally unchanged.

I added three new tables to support extras on additional documents:
## ShoppingListExtras
In order to keep the Mealie shopping list in sync with an external list, it's often helpful to store the external system's id.

While custom integrations for personal use can get around this by hardcoding list ids, a proper integration needs to dynamically store a reference that connects the two systems' ids. Allowing the id to be stored in Mealie avoids needing the other system to support storing ids, or building a db as middleware.

## ShoppingListItemExtras
Similar to above, but more critically, trying to keep items on lists between multiple systems can often be impossible without storing item ids. Some simple integrations can get around this by syncing by the item name, but our list items are more feature-rich (foods, labels, quantities, combining items, etc.) so that doesn't really work for us.

For instance, right now my integration is just one-way: it deletes the item in Mealie and adds it to my external list (which means I lose most of Mealie's cool list item features, like combining ingredients). 

## IngredientFoodExtras
This is mainly selfish, though I imagine others may use it too - when I go shopping there are certain items I don't buy because I always keep them in stock (e.g. butter) so I don't want them on my shopping list.

My original plan was to label these as "On Hand" and have my custom integration just ignore those items, but then I started using Mealie's food data to help categorize foods from other systems (in my case, my Alexa shopping list). So, if, for instance, I told Alexa to add butter to my shopping list, I still wanted that added to my list, and I want it categorized as "Dairy" not "On Hand". So I saw three options:
1. Modify the food schema to allow multiple labels (which is kind of messy, both for the user and because it's a db column change)
2. Add a new property for this specific use case (which has no flexibility and almost no one would use it)
3. Add support for custom data

---

For the alchemy exporter test, I changed the `ALEMBIC_VERSIONS` constant to the new db version; was this the right way to do that, or is there somewhere else that needs to be updated? I did a "search all" and didn't find it anywhere else.


michael-genson:
FYI @hay-kot this is ready :)

hay-kot (changes requested):
I think I'm okay with this one moving forward. It's quite a bit of complexity, but it makes sense for a rich shopping list integration. Could you add test coverage for the new tables you've created, and add a blurb in the documentation about how to use this? I think leaving it as API only for now is fine. If someone is really interested they can tack on a UI for it.

michael-genson:
Can do! And yeah I felt the same on the UI; no idea what it would be used for but if someone wants it they can open a FR or PR

michael-genson:
At some point I might create a new docs page detailing integration-related stuff (with the events API and these API extras) but I think I want to tackle webhooks first, and maybe build in some shopping list integrations. For now I've added it to the same section as the existing Recipe Extras docs

michael-genson:
@hay-kot this should be ready :)

hay-kot (changes requested):
LGTM. Need to rebase and redo your migration file with the latest changes. Let me know if you need any help with that. Easiest way would probably be to just rewrite the commit during the rebase to gen a new rev number.

michael-genson:
I think I did this right. I merged with `mealie-next` and manually adjusted the `down_revision` on these new changes to avoid a branch. I tested manually upgrading and downgrading the database to make sure it traverses correctly.

If I didn't let me know :)

PR #1635, opened 2022-09-09
fix: set meta description and image for shared recipes

### LINKED ISSUE #1620: [v1.0.0b] - Public share links should include image
### First Check

When sharing public links (added in recently nightly release) the name of the recipe shows up on platforms that show link previews, but not the recipe image. 

It’d be great if the recipe image appeared as well. I’m using iMessage mostly, others to try are Facebook messenger, Discord, Slack, etc…

### Deployment

Docker (Synology)

### Deployment Details

_No response_

PFischbeck:
This fixes #1620. Shared link previews should now show the proper recipe description and image - please test this, the HTML generated server-side looks good to me.

I had to change the `useRecipeMeta` composable a bit to make `useContext` work properly.

### DIFF
--- a/frontend/components/Domain/Recipe/RecipePage/RecipePage.vue
+++ b/frontend/components/Domain/Recipe/RecipePage/RecipePage.vue
@@ -267,8 +267,8 @@ export default defineComponent({
     /** =============================================================
      * Meta Tags
      */
-    const metaData = useRecipeMeta(ref(props.recipe));
-    useMeta(metaData);
+    const { recipeMeta } = useRecipeMeta();
+    useMeta(recipeMeta(ref(props.recipe)));
 
     const { user } = usePageUser();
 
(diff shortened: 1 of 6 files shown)


hay-kot:
I can't tell you how happy I am that I didn't have to figure this out. Cheers! 🍻

epheterson:
Hey I was checking out this change and it looks like the site icon for link previews loads from GitHub instead of the local instance (unchanged). Probably safe to always do that but should we change it to come from the server instead so it’s self-contained?

https://raw.githubusercontent.com/hay-kot/mealie/dev/frontend/public/img/icons/android-chrome-512x512.png

PFischbeck:
@epheterson This is a good idea, but I'm not sure how to achieve it. The problem is that the URL (called `og:image` in the `meta`) has to be a full, absolute URL, but the frontend currently doesn't know its domain. So this would require some more work, I think.

epheterson:
I guess it could be conditional if set, but it is defined on the api container ENV as BASE_URL. The API uses it to vend the invite links and share links with a full URL. 

Edit: Oh I see what you mean the api side knows but not the front end. Yeah not sure what’s required then. Maybe a new API call to vend the URL to the front end is necessary.

PR #1657, opened 2022-09-19
Press enter to create a shopping list!

Miroito:
## What type of PR is this?
- feature (kind of)

This has been bothering me for a while... Pressing enter now creates the shopping list instead of having to click the create button. I hate using my mouse to do anything :)

No issue is related to this as far as I'm aware.

This initial implementation is near-sighted. Maybe it makes more sense to add it to the BaseDialog component if such behavior is more generally desired. A decision I leave to the reviewer.

## Testing
No testing added

## Release Notes

```
It is now possible to press enter to create a shopping list instead of having to click the create button.
```

### DIFF
--- a/frontend/components/global/BaseDialog.vue
+++ b/frontend/components/global/BaseDialog.vue
@@ -7,6 +7,10 @@
       :width="width"
       :content-class="top ? 'top-dialog' : undefined"
       :fullscreen="$vuetify.breakpoint.xsOnly"
+      @keydown.enter="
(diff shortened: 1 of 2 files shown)


hay-kot:
> Maybe it makes more sense to add it to the BaseDialog component if such behavior is more generally desired. A decision I leave to the reviewer.

I think that's the correct thing to do. I could see `enter` bound to the submit/confirm action and `esc` bound to the cancel/exit option.

PR #1661, opened 2022-09-21
Refactor webhook events

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- cleanup

_(REQUIRED)_

This takes the previous webhook implementation and unifies it with our new event bus API. There are no functional changes, although since we're using our uniform API the webhook content changed to conform to the `Event` schema. It also allows us to easily implement webhooks for other scheduled events.

_(REQUIRED)_

Unsure if there's a particular issue open for this, but I know it's on the release roadmap somewhere (discussed in #1574)

There are two changes I made that I'm unsure about; they work fine, but I wonder if there's a better way:

1. I modified the `EventBusService` to be executable outside of the FastAPI context. This is required since the app triggers scheduled events without using FastAPI. To accommodate this, I made the `BackgroundTasks` param optional, and added a "FastAPI-friendly" method that preserves dependency injection. One issue this causes is that since webhooks can't access background tasks, the `publish` method is called directly. If this is an issue, we either need to trigger scheduled events through FastAPI (not sure if this is possible) or use a different service for handling background tasks that isn't dependent on FastAPI.
2. Since our event bus service is tied to a group, and the scheduled service is not, I needed a way to trigger the service for all groups. I _think_ the best way to do this is to query the database for all groups, then instantiate `EventBusService` for each of these groups, then let the service handle the filtering, so this is what I did in the `post_group_webhooks` method. If we don't like this database call, we need to decouple the event bus service from the group id, which I'm not sure we want to do.

Some outstanding things to take care of after this PR is accepted:

1. On the frontend there is a "test webhook" button that does nothing. It didn't do anything before this PR either. We should implement this. I created a super rudimentary test route to re-fire all webhooks, but we need something more granular. I'm happy to tackle that after this is accepted (the implementation depends on this being finalized).
2. Currently _all_ meal plans are sent in this webhook. I _think_ this is how it behaved previously, I'm pretty sure I didn't change the content of the webhook (just the format). We should probably limit this to a reasonable range (otherwise as time goes on the webhook data will continue to grow).

## Testing

https://webhook.site/ is amazing for testing webhooks. I used this a lot when refactoring the Apprise events, too.

## Release Notes

_(REQUIRED)_

```release-note
changed mealplan webhook format to aid integration triaging
```

### DIFF
--- a/mealie/services/event_bus_service/event_bus_service.py
+++ b/mealie/services/event_bus_service/event_bus_service.py
@@ -2,6 +2,7 @@ from typing import Optional
 
 from fastapi import BackgroundTasks, Depends
 from pydantic import UUID4
+from sqlalchemy.orm.session import Session
 
 from mealie.core.config import get_app_settings
 from mealie.db.db_setup import generate_session
@@ -35,10 +36,15 @@ class EventSource:
 
 
 class EventBusService:
-    def __init__(self, bg: BackgroundTasks, session=Depends(generate_session)) -> None:
+    def __init__(
+        self, bg: Optional[BackgroundTasks] = None, session: Optional[Session] = None, group_id: UUID4 | None = None
+    ) -> None:
+        if not session:
+            session = next(generate_session())
+
         self.bg = bg
         self.session = session
-        self.group_id: UUID4 | None = None
+        self.group_id = group_id
 
         self.listeners: list[EventListenerBase] = [AppriseEventListener(self.session, self.group_id)]
 
@@ -59,10 +65,18 @@ class EventBusService:
             document_data=document_data,
         )
 
-        self.bg.add_task(self.publish_event, event=event)
+        if self.bg:
+            self.bg.add_task(self.publish_event, event=event)
+
+        else:
+            self.publish_event(event)
 
     def publish_event(self, event: Event) -> None:
         """Publishes the event to all listeners"""
         for listener in self.listeners:
             if subscribers := listener.get_subscribers(event):
                 listener.publish_to_subscribers(event, subscribers)
+
+    @staticmethod
+    def create(bg: BackgroundTasks, session=Depends(generate_session), group_id: UUID4 | None = None):
+        return EventBusService(bg, session, group_id)
--- a/mealie/services/event_bus_service/event_bus_service.py
+++ b/mealie/services/event_bus_service/event_bus_service.py
@@ -84,6 +84,6 @@ class EventBusService:
             if subscribers := listener.get_subscribers(event):
                 listener.publish_to_subscribers(event, subscribers)
 
-    @staticmethod
-    def create(bg: BackgroundTasks, session=Depends(generate_session), group_id: UUID4 | None = None):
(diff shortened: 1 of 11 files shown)


michael-genson:
Backend tests should work; I need a "throw cache at the wall" button :)

mealie/services/event_bus_service/event_bus_service.py
```diff
@@ -59,10 +72,18 @@ def dispatch(
...
 
     def publish_event(self, event: Event) -> None:
         """Publishes the event to all listeners"""
         for listener in self.listeners:
             if subscribers := listener.get_subscribers(event):
                 listener.publish_to_subscribers(event, subscribers)
+
+    @staticmethod
+    def create(bg: BackgroundTasks, session=Depends(generate_session), group_id: UUID4 | None = None):
+        return EventBusService(bg, session, group_id)
```
hay-kot: This should be a classmethod

mealie/services/scheduler/tasks/post_webhooks.py
```diff
@@ -1,54 +1,65 @@
...
 
-    last_ran = datetime.now(timezone.utc)
+    end_dt = datetime.now(timezone.utc)
+    if start_dt is None:
+        # set the webhook query bounds to start at the last time the service ran
+        start_dt = last_ran
 
-    repos = get_repositories(session)
+    # update the last ran time so we continue here next time the service runs
+    last_ran = end_dt
```
hay-kot: I think this could be simplified into something like this...

```python
last_ran = end_dt = datetime.now(timezone.utc)
stard_dt = start_dt or last_ran
```

Mostly personal preference, ¯\_(ツ)_/¯

hay-kot (changes requested):
Looks good to me, few small nit-picky things I'd like to see changed if you've got time. 

Probably worth noting that nothing we're doing here is really scalable beyond a few dozen users, so I don't know how much optimization we need to do around that area. Really the only way we're going to make any significant gains in perf is by running a queue with Redis or similar. And if I add another container to this application I'm fearful of the community backlash lol.

michael-genson:
Sounds good! I'll see if I can finish this up tonight, if not tomorrow.

>Probably worth noting that nothing we're doing here is really scalable beyond a few dozen users, so I don't know how much optimization we need to do around that area.

Hah, fair. My inner optimizer likes to lash out any time I look at a database funny.

mealie/services/event_bus_service/event_bus_service.py
michael-genson: TIL how class methods work; they've been a bit of a mystery to me. Thanks!

mealie/services/scheduler/tasks/post_webhooks.py
michael-genson: Ah yeah I like this much better. This also lets me set start _then_ end, which is also nice

michael-genson:
I was browsing issues and forgot that we discussed this in #1463. Adding this comment to link these (we may also be able to close that task).

PR #1663, opened 2022-09-23
add trivy image scanning

firefly2442:
## What type of PR is this?

- feature
- devops

This adds support for [Trivy container scanning as a Github Action](https://github.com/aquasecurity/trivy-action).  This can help with identifying security issues in underlying software, libraries, dependencies, etc.

N/A

## Testing

Review the Github Security tab and look at the `Code Scanning` section.

## Release Notes

```release-note
None
```

### DIFF
--- a/.github/workflows/partial-trivy-backend-container-scanning.yml
+++ b/.github/workflows/partial-trivy-backend-container-scanning.yml
@@ -1,0 +1,31 @@
+name: Trivy Backend Container Scanning
+
+on:
+  workflow_call:
+
+jobs:
+  build:
+    name: Build and Scan Backend Container
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: true
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+
+      - name: Build Dockerfile
+        run: |
+          docker build -t mealie .
+
+      - name: Run Trivy vulnerability scanner
+        uses: aquasecurity/trivy-action@master
+        with:
+          ignore-unfixed: true
+          image-ref: "mealie"
+          format: "sarif"
+          output: "trivy-results.sarif"
+
+      - name: Upload Trivy scan results to GitHub Security tab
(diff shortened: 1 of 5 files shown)


hay-kot:
Thanks! This looks like a good addition for security scanning. Could you implement this as a partial workflow and add it only to the PR reviews? You can see the `partial-backend.yml` workflow as an example on how to implement this. Basically you set the `on` value as below

```yml

on:
    workflow_call:
```

And then include it in the pull-request.yml workflow. Making it dependent on the backend/frontend tests helps reduce unnecessary compute resources when the build fails anyways.

hay-kot (changes requested):
Thanks! This looks like a good addition for security scanning. Could you implement this as a partial workflow and add it only to the PR reviews? You can see the `partial-backend.yml` workflow as an example on how to implement this. Basically you set the `on` value as below

```yml

on:
    workflow_call:
```

And then include it in the pull-request.yml workflow. Making it dependent on the backend/frontend tests helps reduce unnecessary compute resources when the build fails anyways. 

---

Sorry about the double comment!

firefly2442:
I think I restructured it as a partial as you requested but I'm not sure why the `Dockerfile` didn't build in the Github Action listed in this PR.  It seems to build fine on my local machine.

firefly2442:
Also it looks like you've split into two services now, `./Dockerfile` and `./frontend/Dockerfile`.  I don't think Trivy can run both and upload both results files.  If you're wanting to keep the separation of services and scan each we might need to add another partial action.  Are you intending to keep them separate in the future?

hay-kot:
> I don't think Trivy can run both and upload both results files. If you're wanting to keep the separation of services and scan each we might need to add another partial action. Are you intending to keep them separate in the future?

Yes, we'll be using both docker files for the foreseeable future.

firefly2442:
@hay-kot Apologies for the delay.  I believe I've refactored to scan both the frontend and backend Dockerfiles.  Take a look and let me know what you think.  Cheers.

PR #1676, opened 2022-09-26
fix/Add tesseract bin to container

Miroito:
## What type of PR is this?

- bug

The recently merged OCR capability is unusable because the docker container is missing the Tesseract binary. 
I have 0 experience with docker but I think this might fix it.

N/A

## Testing

N/A

## Release Notes

N/A

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -34,6 +34,7 @@ RUN apt-get update \
     build-essential \
     libpq-dev \
     libwebp-dev \
+    tesseract-ocr-all \
     # LDAP Dependencies
     libsasl2-dev libldap2-dev libssl-dev \
     gnupg gnupg2 gnupg1 \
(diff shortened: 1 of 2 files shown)


hay-kot:
Did you try building this container and checking if it worked with the new dependency?

Miroito:
> Did you try building this container and checking if it worked with the new dependency?

I've tried running `make docker-prod` but I get an error down the line that I did not get the time to troubleshoot

_edit_: turns out dns resolution was failing for a token required to build the image... Deactivating my vpn enforcing certain dns servers did the trick but I have a new error now when poetry tries to install dependencies...

Miroito:
I made the docker image build by upgrading the version of poetry. With the previous version I ran into https://github.com/python-poetry/poetry/issues/4210. (fixes #1691)
I have no idea how the current docker image successfully built with this problem existing... 

I could run make docker-prod and this solution worked. I still have some doubts because somehow the data persisted through fresh builds even though I removed the images and ran system prune after each... Hopefully this is enough now.

_More notes_:
I've looked a bit deeper into the api Dockerfile and I don't really understand what this is supposed to accomplish?
`RUN useradd -u 911 -U -d $MEALIE_HOME -s /bin/bash abc \`
    `&& usermod -G users abc \`

PR #1677, opened 2022-09-26
Default null recipe quantities to zero when adding to shopping list

### LINKED ISSUE #1465: [v1.0.0b] - Internal server error when adding some recipes to the shopping list
### First Check

I have a recipe that contains the following ingredient:
quantity: not defined
unit: q.b. (which means "just enough", commonly found on italian recipes)
ingredient: arugola

If I try to add this recipe to the shopping list I get error 500. The reason is the undefined quantity. Setting the quantity to 1 allows me to add the recipe to the shopping list.

Given that it is allow to leave a quantity undefined, the shopping list should not give an error. Otherwise, the quantity must be made a required field when adding the recipes.

version: beta3

### Deployment

Docker (Linux)

### Deployment Details

_No response_

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- bug

_(REQUIRED)_

When a recipe item is added to a shopping list, if that recipe item has a null quantity (`quantity is None`) it's set to zero.
More specifically: `quantity = quantity or 0`. The recipe schema allows this, but the recipe reference schema does not.

_(REQUIRED)_

resolves #1465

## Release Notes

_(REQUIRED)_

```release-note
fixed server error when a recipe is added to a shopping list and one or more ingredients has a null quantity
```


michael-genson:
The backend server test gods frown upon me today. Tests should run fine, I didn't change much

mealie/schema/group/group_shopping_list.py
```diff
@@ -15,12 +15,7 @@
 
 class ShoppingListItemRecipeRef(MealieModel):
     recipe_id: UUID4
-    recipe_quantity: NoneFloat
-
-    # some recipes have a null quantity, so we default to 0
-    @validator("recipe_quantity")
-    def default_null_quantity(cls, v):
-        return v or 0
+    recipe_quantity: NoneFloat = 0
```
michael-genson, line 18: I used the validator in case None wasn't an acceptable value in shopping list recipe refs, so it would convert None to 0. In case you're wondering why I did it this way haha. But if that doesn't matter this works too

PR #1685, opened 2022-09-28
feat: recipe timeline backend api

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- feature

_(REQUIRED)_

This PR is one major step towards achieving FR #1451. It creates a new `RecipeTimelineEvent` table for the purpose of tracking arbitrary "events" in a Recipe's lifecycle. The focus of this PR is to implement the backend framework, rather than fully utilize it within the application (which will be in a future PR). The features of this PR include:
- The new table
- Schema definitions
- API routes
- A simple "Recipe Created" timeline event, which is created immediately after a recipe is created

_(REQUIRED)_

Works towards #1451

For the most part I think the changes speak for themselves, with the goal of serving #1451, but there are a few things I want to point out:
- You were probably going to do this anyway, but check out the new table definition and let me know if you see any issues with it. Some specific things:
  - I'm pretty sure I did this right, but check out the parent/child relationships and foreign keys with recipes and users. I don't _think_ we need to generate a composite key using the `recipe_id`, but let me know
  - the intent of the `image` column is to enable an upload on the frontend so users can share pictures of their recipes when they make them. Obviously more work needs to be done on the frontend for uploading images, but I figure I can put the column there now
  - the intent of the `event_type` enum is so we can stylize timeline events on the frontend. I figure we can have a switch statement that adds a special icon for different event types
- This is more of an observation, but I ran into trouble originally when the `timestamp` column was named `event_dt`. The issue was in the backup script: because `event_dt` wasn't in the hardcoded list of column names to convert to a `datetime` object, the backup failed (since SQLite doesn't work with datetimes represented as strings). I was _going_ to add `event_dt` to the hardcoded list, but then I saw `timestamp` in there and figured that's a better name anyway. I feel like there should be a better way to handle that conversion other than a hardcoded list, but I have no clue what that better way is, if there is one ¯\\\_(ツ)_/¯

## Testing

Various test cases were added for recipe timeline events.

## Release Notes

_(REQUIRED)_

```release-note
added backend API for recipe timeline events
```


alembic/versions/2022-08-05-17.07.07_089bfa50d0ed_add_is_ocr_recipe_column_to_recipes.py
```diff
@@ -1,7 +1,7 @@
 """Add is_ocr_recipe column to recipes
 
 Revision ID: 089bfa50d0ed
-Revises: f30cf048c228
+Revises: 188374910655
```
michael-genson, line 4: FYI this is just a change to the comment, not the actual migration, looks like I forgot to edit it in #1619

alembic/versions/2022-08-29-13.57.40_44e8d670719d_add_extras_to_shopping_lists_list_items_.py
```diff
@@ -1,7 +1,7 @@
 """add extras to shopping lists, list items, and ingredient foods
 
 Revision ID: 44e8d670719d
-Revises: 188374910655
+Revises: 089bfa50d0ed
```
michael-genson, line 4: Same here, only edited the comment

michael-genson:
These workflows hate me. Tests ran fine for me locally, but I'm using SQLite not Postgres. I don't think there will be any issues on Postgres but we should check

mealie/db/models/recipe/recipe_timeline.py
```diff
@@ -0,0 +1,45 @@
...
+
+from .._model_base import BaseMixins, SqlAlchemyBase
+from .._model_utils import auto_init
+from .._model_utils.guid import GUID
+
+
+class TimelineEventType(enum.Enum):
+    system = "system"
+    info = "info"
+    comment = "comment"
```
hay-kot: I think I'd prefer to implement these in the schemas folder, and leave the database column as a generic string and let Pydantic handle the validation. That's more consistent with everything else we've done up to this point. Might not be the best, but I'd rather them be consistent.

hay-kot (commented):
Overall looks pretty good, just that Enum thing I think needs to be changed.

 I did some minor cleanup to fix a few inconsistency with the most recenter merges and run the code generator. As a result I _did_ rebase and force push your branch. 

Don't hate me :(

mealie/db/models/recipe/recipe_timeline.py
michael-genson: Yeah, makes sense! I was back and forth on whether to make it an enum or string

michael-genson:
Haha no problem at all on rebasing, I'm just happy to see it cleaner

PR #1703, opened 2022-10-05
Allow to overwrite data dir using environtment variable

1848:
Mealie tries to create its data dir in **/app/data**.
This is not the place where you want stuff like this if you dont use docker.

This PR introduces another environment variable called **DATA_DIR** which overwrites this location if set.

It is needed while installing (e.g.)
`
PRODUCTION=1 DATA_DIR=/var/lib/mealie/app/data python mealie/db/init_db.py
`
and when running the application (e.g.)
`
PRODUCTION=1 DATA_DIR=/var/lib/mealie/app/data uvicorn mealie.app:app --host 0.0.0.0 --port 9000
`

### DIFF
--- a/mealie/core/config.py
+++ b/mealie/core/config.py
@@ -18,16 +18,17 @@ ENV = BASE_DIR.joinpath(".env")
 dotenv.load_dotenv(ENV)
 PRODUCTION = os.getenv("PRODUCTION", "True").lower() in ["true", "1"]
 TESTING = os.getenv("TESTING", "False").lower() in ["true", "1"]
+DATA_DIR = os.getenv("DATA_DIR")
 
 
 def determine_data_dir() -> Path:
-    global PRODUCTION, TESTING, BASE_DIR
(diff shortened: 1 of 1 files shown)


mealie/core/config.py
```diff
@@ -18,16 +18,17 @@
...
+DATA_DIR = os.getenv("DATA_DIR")
 
 
 def determine_data_dir() -> Path:
-    global PRODUCTION, TESTING, BASE_DIR
+    global PRODUCTION, TESTING, BASE_DIR, DATA_DIR
 
     if TESTING:
-        return BASE_DIR.joinpath("tests/.temp")
+        return BASE_DIR.joinpath(DATA_DIR if DATA_DIR else "tests/.temp")
```
hay-kot, line 28: Is there an instance where you would want to override the testing folder?

1848: I dont have a usecase for testing but I guess it does not hurt when you allow to override this for testing too.

PR #1716, opened 2022-10-09
Fix min-original version of image not saved when the original image is already a webp 

### LINKED ISSUE #1713: [v1.0.0b] - Manually imported images not showing on "All recipes" page
### First Check

Some manually imported images not showing on "All recipes" page, after clicking into the recipe, it will show as normal on the individual recipe page, this shows that there is an image tied to the recipe. 

Copied from submit bug:
**Details**
Version: v1.0.0beta-4
Build: 13850cda1f9cb02a8c788030ba6cfa96bba33571
Application Mode: Production
Demo Status: Not Demo
API Port: 9000
API Docs: Enabled
Database Type: sqlite
Recipe Scraper Version: 14.11.0

**Checks**
Secure Site: Yes
Server Side Base URL: Yes
LDAP Ready: No
Email Configured: No
Docker Volumes: success

### Deployment

Docker (Linux), Unraid

### Deployment Details

Unraid using docker. Deployed V1 with separate api/frontend.

Miroito:
## What type of PR is this?
- bug
When a webp file uploaded as a recipe image goes through the minifying process, the min-original version of the image is skipped because the extension is already webp.
This behaviour is not wanted.

Fixes #1713 

I removed the condition skipping webp images.
Overall, the code of this process looks messy to me, I don't understand why the minify function is so unnecessarily complex.

Additionally, this change will prevent this bug to happen in the future but **will not** fix the damage that is already done. e.g. Missing min-orignal.webp images will still be missing.

## Testing

No further testing written.

## Release Notes

```
Fixed bug where uploading a webp image as a recipe's image will not create the thumbnail sized version. 
```

### DIFF
--- a/mealie/pkgs/img/minify.py
+++ b/mealie/pkgs/img/minify.py
@@ -65,9 +65,6 @@ class PillowMinifier(ABCMinifier):
         Converts an image to the webp format in-place. The original image is not
         removed By default, the quality is set to 100.
         """
-        if image_file.suffix == WEBP:
-            return image_file
-
         img = Image.open(image_file)
 
         dest = dest or image_file.with_suffix(WEBP)


hay-kot:
From #1713

> I understand this comment I made is irritating, I did not want to go into the details because this issue is not the right place for it. Hence this vague statement I made.
>
> If you want to discuss these concerns a place should be made for it.

If you have comments about that section of code, this would be the place to provide those.

Miroito:
Ok here is my review of this module:

- Some things are unused:
  - static.py defines the NOT_WEBP variable which I don't see referenced anywhere else.
  - ABCMinifier.get_image_sizes is never used and contains only a log entry.
- Some arguments are set to None by default but their type isn't Optional[x] of X | None.
- A lot of unnecessary info log statements about images that have been "minifed", or not.
- The miniature image receives the same treatment as the original image, it is only converted to webp. Only their names differ. Is that intended? Basically, if the original image is not already a webp, both will always be the exact same as far as I understand. 
- minify takes an argument called image_file which is a Path. I would rename it to image_path. Also given that the only reference to minify is in the `RecipeDataService.write_data` function and there it is given a variable named image_path.
- The crop_center method image_file argument should take an Image.Image type, Image is the PIL module, not a type.
- The success variable is there to prevent deleting the original image in case the function ran into some issue. It is set to true if any of the original, min, or tiny conversions succeed. It should probably be true only if all have succeeded. (Keeping in mind that not all the options will be asked for each time).
-  The minify parameter force, is never set to anything else than its default (True). It only is important to set when a recipe's image is updated. I don't see a scenario where force has a meaning. In all cases, the images should all represent the same original, regardless if by some chance one of the versions already existed.
- If the image_file argument given to minify is already an image called original.webp, it doesn't hurt to overwrite it to enforce the quality to be 70 instead of whatever the original was. That is, if overwriting a webp image with a lower quality than 100 does still write an image with a quality of 70 and not less. I'm not completely sure about that part.
- The init.py file says that this module should not contain any Mealie-specific funcitonality. I believe that the hardcoded image file names is a Mealie-specific functionality.


As to code in itself:
```
        if not force and min_dest.exists() and tiny_dest.exists() and org_dest.exists():
            self._logger.info(f"{image_file.name} already minified")
            return
```
Is useless given the current implementation, it only serves the purpose of printing this log. If purge was set to true, the original image will not be deleted contrary to what the class advertises.


```
        if self._opts.original:
            if not force and org_dest.exists():
                self._logger.info(f"{image_file.name} already minified")
            else:
                PillowMinifier.to_webp(image_file, org_dest, quality=70)
                success = True

        if self._opts.miniature:
            if not force and min_dest.exists():
                self._logger.info(f"{image_file.name} already minified")
            else:
                PillowMinifier.to_webp(image_file, min_dest, quality=70)
                self._logger.info(f"{image_file.name} minified")
                success = True

        if self._opts.tiny:
            if not force and tiny_dest.exists():
                self._logger.info(f"{image_file.name} already minified")
            else:
                img = Image.open(image_file)
                tiny_image = PillowMinifier.crop_center(img)
                tiny_image.save(tiny_dest, FORMAT, quality=70)
                self._logger.info("Tiny image saved")
                success = True
```
This could use a rewrite, making the code less confusing. All of the redundant conditions make it seem more complex than it really is.

That is pretty much it. The point is not to blame anybody, just to maintain code that is easily readable and that every party understands what this module's role is.

Let me know what you think.

hay-kot:
I agree with pretty much everything you've mentioned. I appreciate the detailed and thoughtful response. I think most of this has to do with the various roles this code has played. It's been moved around and refactored many many times over the last few years. A rewrite if warranted at this point

Is that something you want to take on or should I create an issue for these changes?

Miroito:
> Is that something you want to take on or should I create an issue for these changes?

It can be a group effort, but what should be done first before any actual change to the code, is to define what requirements this module should fulfill. 
An issue can be created to have a clear definition of what the module should do. When everything becomes clear, the code just writes itself.

PR #1730, opened 2022-10-16
Feature: Live Shopping List Updates

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- feature

_(REQUIRED)_

When a user is viewing a shopping list, it will poll the backend for changes every 5 seconds. This allows multiple devices/users to view the same shopping list in sync.

_(REQUIRED)_

Resolves #1715

## Testing

I made sure that this properly tracks changes across multiple tabs, and I monitored the backend to confirm once all tabs have been closed the polling stops.

## Release Notes

_(REQUIRED)_

```release-note
Added live updates to shopping lists across all users and devices
```

### DIFF
--- a/frontend/pages/shopping-lists/_id.vue
+++ b/frontend/pages/shopping-lists/_id.vue
@@ -239,7 +239,36 @@ export default defineComponent({
     }
 
     // constantly polls for changes
-    const pollTimer: ReturnType<typeof setInterval> = setInterval(() => { refresh() }, 5000);
+    async function pollForChanges() {
+      try {
+        await refresh();
+
+        if (shoppingList.value) {
+          attempts = 0;
+          return;
+        }
+
+        // if the refresh was unsuccessful, the shopping list will be null, so we increment the attempt counter
+        attempts ++;
+      }
+
+      catch (error) {
+        attempts ++;
+      }
+
+      // if we hit too many errors, stop polling
+      if (attempts >= maxAttempts) {
+        clearInterval(pollTimer);
+      }
+    }
+
+    // start polling
+    const pollFrequency = 5000;
+
+    let attempts = 0;
+    const maxAttempts = 3;
+
+    const pollTimer: ReturnType<typeof setInterval> = setInterval(() => { pollForChanges() }, pollFrequency);
     onUnmounted(() => {
       clearInterval(pollTimer);
     });
--- a/frontend/pages/shopping-lists/_id.vue
+++ b/frontend/pages/shopping-lists/_id.vue
@@ -188,6 +188,7 @@
 import draggable from "vuedraggable";
 
 import { defineComponent, useAsync, useRoute, computed, ref, watch } from "@nuxtjs/composition-api";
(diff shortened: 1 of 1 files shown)


frontend/pages/shopping-lists/_id.vue
```diff
@@ -188,6 +188,7 @@
 import draggable from "vuedraggable";
 
 import { defineComponent, useAsync, useRoute, computed, ref, watch } from "@nuxtjs/composition-api";
+import { onUnmounted } from "vue-demi";
```
hay-kot: This ins't the correct import 'vue-demi' is only in development. I think all of this needs to come from the Nuxt composition API. I thought I had a linter rule for this, but maybe I missed this package.

frontend/pages/shopping-lists/_id.vue
```diff
@@ -238,6 +239,12 @@ export default defineComponent({
       shoppingList.value = await fetchShoppingList();
     }
 
+    // constantly polls for changes
+    const pollTimer: ReturnType<typeof setInterval> = setInterval(() => { refresh() }, 5000);
+    onUnmounted(() => {
+      clearInterval(pollTimer);
+    });
+
```
hay-kot: Looks good overall, can we wrap this in a Try/Catch and maybe cancel the interval if we make a few bad requests? If you went offline I wouldn't want it to just constantly keep polling. 

Also, should we consider a maximum amount of time to poll for? If I  left my browser open would I just want it to keep polling forever?

frontend/pages/shopping-lists/_id.vue
michael-genson: Ah okay, I wondered what that was but I saw it somewhere else. Will fix, thanks!

frontend/pages/shopping-lists/_id.vue
michael-genson: > can we wrap this in a Try/Catch

For sure!

> should we consider a maximum amount of time to poll for?

Good point, any thoughts on how long? Maybe like 2 hours? I'm thinking that's the upper limit of how long it takes me to shop at one store. Otherwise, is there a way to capture focus? (like if there's no activity for 30 minutes or if it's not the active tab or something)

frontend/pages/shopping-lists/_id.vue
hay-kot: You're absolutely right, it's all over the place. No idea how any of that is working in production... Will look at that in a PR shortly.

frontend/pages/shopping-lists/_id.vue
hay-kot: There's the `isIdle` composable we could use

https://vueuse.org/core/useidle/#useidle

If we use that we could set a more aggressive timeout, like 5 minutes of idle and we stop polling, and re-trigger when idle is false again?

michael-genson: Oh perfect that's exactly what I was looking for. Then yeah I think 5 minutes makes sense! If not we can always adjust later

PR #1750, opened 2022-10-22
Feature: Duplicate recipes

xarantolus:
## What type of PR is this?

- feature

This PR adds a button to the recipe view that allows duplicating a recipe, as suggested in #1198.

This is the button:

grafik

It opens a dialog for the new recipe name, prefilled with the old name. The text input is focused automatically:

grafik

Upon confirmation, a request to the `/api/recipes/{slug}/duplicate` API endpoint is made to duplicate the recipe (with the custom new name). After duplication, the user is redirected to the newly created recipe.

Fixes #1198

There are some things I'm not entirely sure about:
* I implemented the duplication logic in `mealie/routes/recipe/recipe_crud_routes.py`. Is this OK or should it go into `mealie/services/recipe/recipe_service.py` or another place?
* The duplication logic does not copy e.g. comments
* The duplication logic does copy the data directory of the old recipe and updates links to assets in steps. If the old recipe is deleted the new recipe (especially images in steps) is not affected
* Did I forget any important recipe data?
* How can I add tests for the duplication part?

Another thing to note is that I only added translations for en-US and de-DE.

## Testing
I have tested this change manually, with a focus on making sure that the deletion of the old recipe does not affect the newly created one and vice-versa.

## Release Notes

```release-note
Add duplicate button in recipe view
```


hay-kot:
> I implemented the duplication logic in mealie/routes/recipe/recipe_crud_routes.py. Is this OK or should it go into mealie/services/recipe/recipe_service.py or another place?

I think it probably makes sense to move this into the recipe_service.py. I think that's where most of the "work" should happen for this stuff. 

> The duplication logic does not copy e.g. comments

👍 

> The duplication logic does copy the data directory of the old recipe and updates links to assets in steps. If the old recipe is deleted the new recipe (especially images in steps) is not affected

Are you sure that all the related nested SQL fields are linked to unique items. Like if I edit an ingredient on Recipe A, does that change Recipe A Copy? We probably want a tests that ensures this doesn't happen for the nested fields.

> Did I forget any important recipe data?

LGTM

> How can I add tests for the duplication part?

We use automated testing in the `tests` directory. Checkout the `user_recipe_tests` directory for some examples of how we're running the tests.

hay-kot:
Also this implementing looks really really good. Thanks for taking the time to submit this PR! 🚀

xarantolus:
Thank you for the review :) I wasn't aware of the ingredients being linked, so they likely don't get copied correctly as of now. I'll look into that and the other points you mentioned 👍

xarantolus:
I have moved the logic to the recipe service. Copied ingredients are just assigned a new ID, that way they don't appear to get affected when editing them in the new recipe. I checked for each attribute of recipe if it can be copied as-is or if it needs extra logic.

I also added a test that makes sure different editing scenarios are handled correctly.

hay-kot:
This looks awesome, thanks so much for the implementation. I pulled it down and did a rebase + fix a few code generation issues. Stoked for this feature!

PR #1777, opened 2022-10-26
1775 LDAP login and user creation fix

### LINKED ISSUE #1775: [v1.0.0b] - LDAP login/user creation not possible due to an incorrect if statement in the code
### First Check

When using LDAP, it is not possible to login or create an user due to an incorrect if statement. 

In the [security.py](https://github.com/hay-kot/mealie/blob/mealie-next/mealie/core/security/security.py) code, on line 83, a search for the user dn and attributes is done within LDAP, which returns an array with the matches. Afterward, an if statement check the result of the search query. However, this if statement has an incorrect negation in it. 

The current logic checks if a results is found, if that is the case, false will be returned which equals an incorrect login. The `not` should be removed from this if statement. 

Current code:
```    
    # Search "username" against "cn" attribute for Linux, "sAMAccountName" attribute
    # for Windows and "mail" attribute for email addresses. The "mail" attribute is
    # required to obtain the user's DN for the LDAP_ADMIN_FILTER.
    user_entry = conn.search_s(
        settings.LDAP_BASE_DN,
        ldap.SCOPE_SUBTREE,
        f"(&(objectClass=user)(|(cn={username})(sAMAccountName={username})(mail={username})))",
        ["name", "mail"],
    )
    if not user_entry:
        user_dn, user_attr = user_entry[0]
    else:
        return False
```

The code should look like this:
```    
    # Search "username" against "cn" attribute for Linux, "sAMAccountName" attribute
    # for Windows and "mail" attribute for email addresses. The "mail" attribute is
    # required to obtain the user's DN for the LDAP_ADMIN_FILTER.
    user_entry = conn.search_s(
        settings.LDAP_BASE_DN,
        ldap.SCOPE_SUBTREE,
        f"(&(objectClass=user)(|(cn={username})(sAMAccountName={username})(mail={username})))",
        ["name", "mail"],
    )
    if user_entry:
        user_dn, user_attr = user_entry[0]
    else:
        return False
```

After the above fix, login is possible.

However, if a user is created, the user has the binary values within the name and email fields, instead of the string values. 

The following code creates the user.
```
        user = db.users.create(
            {
                "username": username,
                "password": "LDAP",
                "full_name": user_attr["name"][0],
                "email": user_attr["mail"][0],
                "admin": False,
            },
        )
```

The problem is the user_attr values are binary instead of strings. When looking at the [python-ldap documentation](https://www.python-ldap.org/en/python-ldap-3.4.3/bytes_mode.html) this is to be expected. In the documentation it states the following:
 _Attribute values, on the other hand, MAY contain any type of data, including text. To know what type of data is represented, python-ldap would need access to the schema, which is not always available (nor always correct). Thus, attribute values are always treated as bytes. Encoding/decoding to other formats – text, images, etc. – is left to the caller._

This means the user attributes (name and mail) need to be decoded to string. 

### Deployment

Docker (Linux)

### Deployment Details

Docker install using a portainer stack on an unraid system.

ERLKDev:
## What type of PR is this?
- bug

This PR fixes two bugs in the code:

- LDAP login not possible due to a programming error (wrong if statement)

- Fullname and email adres in newly created users using LDAP had the binary values instead of the string values of the corresponding attributes

The full description of the bugs is available in #1775 

Fixes #1775


ERLKDev:
In #1487 the unit test has been changed from (see [comment](https://github.com/hay-kot/mealie/pull/1487#issuecomment-1242991147) for an explanation):
```
    assert result is not False
    assert result.username == user
```

to 

```
assert result is False
```

The result should however not be false, since an array with an empty tuple is still a result since it is not an empty array, so the search was "successful" and an user is created. I will revert the asserts back in the unit test.

hay-kot:
FYI, I tried to run formatting to get this through, but the test is still failing. You can run both of these locally using the make file.

hay-kot:
This good to go now?

mealie/core/security/security.py
```diff
@@ -54,57 +54,77 @@ def user_from_ldap(db: AllRepositories, username: str, password: str) -> Private
...
         conn.set_option(ldap.OPT_X_TLS_NEWCTX, 0)
-    user = db.users.get_one(username, "email", any_case=True)
 
-    if not settings.LDAP_BIND_TEMPLATE:
-        return False
+    user = db.users.get_one(username, "email", any_case=True)
 
     if not user:
-        user_bind = settings.LDAP_BIND_TEMPLATE.format(username)
+        user_bind = "cn={}, {}".format(username, settings.LDAP_BASE_DN)
```
flocke, line 69: This might break some LDAP setups, I suggest to leave the `LDAP_BIND_TEMPLATE` as it is.
For example on my server I need `uid={},...`, not `cn={},...`.

Zackptg5:
@ERLKDev, I attempted to build with this but got: `NameError: name 'Optional' is not defined` from settings.py. Optional needs imported at the top of it: `from typing import Optional`

ERLKDev:
> This good to go now?

Sorry for the late response, have been quite busy last month and kind of forgot about it.

Next week I have time to finish the PR.

michael-genson:
FYI in Python >= 3.10 you can use `str | None` rather than `Optional[str]` which doesn't require an import, and looks a bit nicer

mealie/core/security/security.py
jado: Did you get LDAP  to work with v1? It seems that you are using FreeIPA.
Mine was working with v0.5, but v1 broke it.
I am trying

```
LDAP_BIND_TEMPLATE=uid={},cn=users,cn=accounts,dc=...,dc=...
LDAP_BASE_DN=cn=users,cn=accounts,dc=...,dc=...
```

flocke: I'm not using FreeIPA, I use my own custom setup with OpenLDAP. But yes, it worked with v1.

hay-kot:
Merged in #2107

Issue #1397, opened 2022-06-16
[v1.0.0b] - Rating is discarded if recipe is edited upon creating


epheterson:
### First Check

- [X] This is not a feature request
- [X] I added a very descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Mealie documentation, with the integrated search.
- [X] I already read the docs and didn't find an answer.

### What is the issue you are experiencing?

When creating a new recipe via URL, if you also edit that recipe (adding tags, categories, etc...) then the rating is discarded. If you do not edit the recipe, the rating does stick.

These steps reproduce every time:

1. Import new recipe via URL
2. On initial recipe presentation, add rating
3. Click Edit, add tags, categories, etc...
4. Click Save

Rating will not have saved.

It also reproduces if you add the rating during the add step, i.e.
1. Import new recipe via URL
2. On initial recipe presentation, click Edit, add tags, categories, etc...
3. Add rating
4. Click Save

### Deployment

Docker (Synology)

### Deployment Details

```
**Details**
Version: v1.0.0beta-3
Build: 56eb0bca7142715984c8028c4c881f086e979564
Application Mode: Production
Demo Status: Not Demo
API Port: 9000
API Docs: Enabled
Database Type: sqlite
Recipe Scraper Version: 14.3.0

**Checks**
Secure Site: Yes
Server Side Base URL: Yes
LDAP Ready: No
Email Configured: No
Docker Volumes: success
```

epheterson:
As a workaround, while adding recipes, this flow preserves the rating:

1. Import new recipe via URL
2. On initial recipe presentation, click Edit, add tags, categories, etc...
3. Click Save
4. Add Rating

michael-genson:
I can reproduce this at the API level; the `POST` route for a new recipe doesn't save the rating. `PATCH` does. I'll see if I can figure out why.

michael-genson:
Ignore my last comment, I see the problem.

When you set the rating, the frontend immediately sends a `PATCH` request to the API to update it. However, when you save a recipe in the frontend, it sends a `PUT` request with the _old_ value. In the case of this specific issue, the `PUT` is overwriting the user's rating with null (since there is no previous rating).

This is repeatable at any stage, even for existing recipes:
1. open a recipe for editing
2. change the rating (e.g. 4 --> 5)
3. save the recipe
4. refresh the page and observe that the rating didn't change (e.g. it's still 4)

PR #1801, opened 2022-11-03
Feature: "I Made This" Dialog

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- feature

_(REQUIRED)_

This adds a "Last Made" property to recipes and a new "I Made This" dialog on the recipe page. It creates a new event on the recipe event timeline and updates the recipe's `lastMade` timestamp to today (unless the user chooses a specific date).

The last made date is also displayed on the recipe page, and recipes can be sorted by this property.
image
*Works on both landscape and non-landscape view, and does not appear on the public recipe page. The date format is localized*

image
*If the recipe has never been made*

image
*The comment isn't viewable in the UI yet, pending a recipe timeline view*

_(REQUIRED)_

Works towards #1451. This utilizes the timeline API, but does not implement the timeline frontend.

Originally I had this in the recipe Action/Context menus, but I moved it to the recipe header/title instead since I think it works better there. It also made manipulating the last made date significantly easier.

I wasn't sure how to add new text localizations, so at the moment everything is just in English. If we want to add localizations in this PR I'm happy to do so, if you could help point me in the right direction. Not too much text was added, and it was only added in two places (the new component and the sort menu) so it should be pretty easy.

Also, to avoid the plugin conflict we talked about on discord, I modified the timeline event API to use `event_message` instead of `message`. This ended up being the easiest solution by far, since the plugin directly accesses the `POST` response. I also needed a new database migration for this PR anyway.

## Testing

I ran through the following scenarios:
1. Verify the dialog reactively updates the "last made" date without a refresh
2. Verify if the last made date is updated, then the user edits and saves the recipe without leaving the page, then the last made date is still the new date and not the old one (i.e. checking for the issue in #1397)
3. Verify the dialog can't be accessed when the recipe is being viewed publicly (i.e. there is no user)

## Release Notes

_(REQUIRED)_

```release-note
added tracking for when a recipe was last made ("I made this")
added sorting recipes by last made date
```


alembic/versions/2022-11-03-13.10.24_1923519381ad_renamed_timeline_event_message_and_.py
```diff
@@ -0,0 +1,31 @@
...
+revision = "1923519381ad"
+down_revision = "2ea7a807915c"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    # we need to rename the timeline event message field due to a frontend plugin conflict
+    op.alter_column("recipe_timeline_events", "message", new_column_name="event_message")
```
hay-kot: alter_column isn't supported in SQLite. If it's just a name conflict can we rewrite it using a pydantic alias?

hay-kot (changes requested):
Looks good to me overall. Just need to take another look at the Migration and see if we can use a Pydantic Alias to avoid altering the column since SQLite doesn't support that feature which will put is in a weird state.

alembic/versions/2022-11-03-13.10.24_1923519381ad_renamed_timeline_event_message_and_.py
michael-genson: Oh fun. Yeah I can do that too!

michael-genson:
@hay-kot should be good now :)

Issue #1241, opened 2022-05-18
[v1.0.0b] [Task] - Color Code Tags/Categories


hay-kot:
### What is the problem this task addresses?

 - https://github.com/hay-kot/mealie/discussions/1220

TLDR: Add color options for Tags and Categories for displaying when in chip format like on the recipe pages 

### Proposed/Possible Solution(s)?

The MultiPurposeLabels components does nearly the same thing that is requested here, most of the logic can be utilized to determine text color and default colors with the caveat that the existing behavior should be kept in that the default color for the chips should be the accent of the theme.

dorj222:
if no one is working on this issue, may I claim this issue?

bencondemi:
I'd like to work on this issue, can I claim it?

michael-genson:
You're certainly welcome to work on it! FYI, a previous PR was pretty close to completing it, in case you wanted to use it as a reference, though things have changed drastically since it was opened/closed: https://github.com/mealie-recipes/mealie/pull/1828

bencondemi:
Thanks. How have they changed drastically?

michael-genson:
That PR is over two years old!

bencondemi:
I understand that, I was hoping you had a specific idea of what has changed since then (but I'm assuming you will say everything?)

michael-genson:
No not really, you could do a `git diff` on the branches but it's not really worth the effort. If you're going to reference the old PR, I'd suggest looking at a high level what it does, rather than the specifics. The general mechanics of the app haven't changed too much

PR #1828, opened 2022-11-15
Fix/issue#1241

dorj222:
## What type of PR is this?

_(REQUIRED)_

- feature

_(REQUIRED)_

This PR introduces a new functionality to add color options for Tags and Categories for displaying in chip format. This feature was suggested in #1241. 

New Create Category:

Screen Shot 2022-11-15 at 2 32 05 PM

New Category List View: 

Screen Shot 2022-11-15 at 2 33 14 PM

_(REQUIRED)_

#1241

## Testing

I manually and thoroughly tested the web app on Chrome and Mozilla Firefox browsers.

## Release Notes

_(REQUIRED)_

```release-note
implemented new functionality to add color options for Tags and Categories
```


alembic/versions/2022-02-21-19.56.24_6b0f5f32d602_initial_tables.py
```diff
@@ -278,6 +279,7 @@ def upgrade():
         sa.Column("group_id", mealie.db.migration_types.GUID(), nullable=False),
         sa.Column("name", sa.String(), nullable=False),
         sa.Column("slug", sa.String(), nullable=False),
+        sa.Column("color", sa.String(), nullable=False),
         sa.ForeignKeyConstraint(
             ["group_id"],
```
hay-kot: These files are readonly. You can't update an existing migration, you need to generate a new one.

hay-kot (changes requested):
Everything looks good but the data migration. Once you get that generated we can merge this. Thanks!

alembic/versions/2022-02-21-19.56.24_6b0f5f32d602_initial_tables.py
dorj222: Thanks a lot for your feedback! I have one question. How do you generate these read-only files?

michael-genson: You'll want to generate a migration using [alembic](https://alembic.sqlalchemy.org/en/latest/autogenerate.html).

Here's an example:
```alembic revision --autogenerate -m "add color to tags and categories"```

(that might be all you have to do for this migration, you can adjust the message)

Your new auto-generated file will have a new `Revision ID` at the top, which you'll want to put in the alembic test file (if you try running pytest it will fail on the alembic test if you don't update it)

Specifically `tests\unit_tests\services_tests\backup_v2_tests\test_alchemy_exporter` (replace the `version_num` with your new `Revision ID`)

dorj222: hi there! thanks a lot for your response. I generated a new migration using alembic and replaced the version_num with the new Revision ID.

hay-kot: @dorj222, CI is still failing. I believe there have been some migrations since you've started this PR. Migrations _need_ to happen in a specific order so you will likely need to rebase the PR and get all the changes from the latest branch and update your migration revision information to reflect the correct order.

hay-kot (changes requested):
-> https://github.com/hay-kot/mealie/pull/1828#discussion_r1064182051

alembic/versions/2022-02-21-19.56.24_6b0f5f32d602_initial_tables.py
dorj222: ah, I see. Thanks for the feedback!

Issue #1829, opened 2022-11-15
[v1.0.0b] - Recipe Description text overflows on a card view 


dorj222:
### First Check

- [X] This is not a feature request
- [X] I added a very descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Mealie documentation, with the integrated search.
- [X] I already read the docs and didn't find an answer.

### What is the issue you are experiencing?

The description text overflows on a card view if the text length is longer than 500 characters. 

Screenshot:

![Screen Shot 2022-11-15 at 4 23 45 PM](https://user-images.githubusercontent.com/26856199/201957717-74256f5a-4625-485c-aea2-9bfd9f7e0d5a.png)

### Deployment

Other

### Deployment Details

**Details**
Version: v1.0.0beta-5
Build: unknown
Application Mode: Development
Demo Status: Not Demo
API Port: 9000
API Docs: Enabled
Database Type: sqlite
Recipe Scraper Version: 14.20.0

**Checks**
Secure Site: Yes
Server Side Base URL: No
LDAP Ready: No
Email Configured: No
Docker Volumes: success

PR #1830, opened 2022-11-15
fix the text overflow on RecipeCard

dorj222:
## What type of PR is this?

_(REQUIRED)_

- bug

_(REQUIRED)_

This PR fixes the recipe description text overflow issue when the text length is longer than 500 characters. If the description text is longer than 350 characters, only the first 350 characters will be shown on a card view.

New Card View with the long description:

Screen Shot 2022-11-15 at 4 42 24 PM

#1829

_(REQUIRED)_

## Release Notes

_(REQUIRED)_

```release-note
 the recipe description text overflow issue on a card view has been fixed
```

### DIFF
--- a/frontend/components/Domain/Recipe/RecipeCard.vue
+++ b/frontend/components/Domain/Recipe/RecipeCard.vue
@@ -19,7 +19,7 @@
           <v-expand-transition v-if="description">
             <div v-if="hover" class="d-flex transition-fast-in-fast-out secondary v-card--reveal" style="height: 100%">
               <v-card-text class="v-card--text-show white--text">
-                {{ description }}
+                {{ description.length < 350 ? description : description.substring(0,350) + "..." }}
               </v-card-text>
             </div>
           </v-expand-transition>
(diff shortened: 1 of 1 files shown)


michael-genson:
I was curious if this issue persisted on the condensed (mobile) view, and it looks like it's solved for using CSS:
```css
.headerClass {
  white-space: nowrap;
  word-break: normal;
  overflow: hidden;
  text-overflow: ellipsis;
}
```

Then I checked out why the normal card didn't do this, but then I noticed that it does, just only for the name. Can you try implementing this with a CSS class similar to the mobile view?

Something like this:
```html
<v-card-text class="v-card--text-show white--text">
  <div class="headerClass">
    {{ description }}
  </div>
</v-card-text>
```

But probably with a new class to show more than a single line

dorj222:
Hello, thank you so much for your feedback! I added a wrapper class that utilizes the [WebKit-line-clamp](https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-line-clamp) property, and now at most first 8 lines of the description will be displayed, and the remaining overflow text will be hidden. 

The screenshot example: 
Screen Shot 2022-11-22 at 4 25 15 PM

PR #1831, opened 2022-11-15
Feat/recipe timeline event UI

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- bug
- feature

_(REQUIRED)_

This implements a new UI for recipe timeline events, and fixes a few bugs and visual glitches with the new "I Made This" functionality (#1801). With these changes we more-or-less complete #1451, save for implementing image handling (this PR is big enough already, I'll add that later).

To open the new UI, there's a new badge:
image

Here's what it looks like on desktop:
image

And on mobile we use the dense view:
image

If you are the "owner" of the event (your `userId` matches the event's `userId`") and the `eventType` is not `system`, then there's a context menu that lets you edit or delete the event:
image
image
image

If no events are on the recipe, you get a generic message instead. Note that this shouldn't happen with new recipes, as they always create an event since that API was implemented. To trigger this I had to delete the default event through the API.
image

The already-implemented "I Made This" UI rendered horribly on mobile, so that has been updated.
Desktop:
image

Mobile:
image

Lastly, I made all of this text localizable.

_(REQUIRED)_

Completes FR #1451

Ooh boy did I learn a lot about Vue and JS working on this. Happy to fix/refactor anything that doesn't make sense or can look better. I tried to mirror the existing components as best as possible.

Did I implement the localizations correctly? Will the JSON be automatically updated next time the Crowdin code runs? I mirrored a different PR I found that added new localizations, and I think I did it right.

In a few places I had to do some cheeky stuff to convert between local datetimes and UTC datetimes. What I did definitely _works_, but my Python brain had trouble wrapping my head around how JS handles datetimes. It's worth reviewing and seeing if I cooked too much pasta with the datetime stuff, specifically in `RecipeLastMade.vue` and `RecipeDialogTimeline.vue`.

## Testing

I tested the following functionalities:
- Using "I Made This" to generate new events
- Using the API to generate other events
- Editing and deleting existing events via the context menu (and confirming the changes are reflected without a refresh)

I tested the following elements:
- Desktop rendering (various screen sizes)
- Mobile rendering  (various screen sizes)
- Having a lot of events
- Having few events
- Having no events

## Release Notes

_(REQUIRED)_

```release-note
added new recipe timeline interface
```


frontend/components/Domain/Recipe/RecipeDialogTimeline.vue
```diff
@@ -0,0 +1,279 @@
...
+      useAsync(async () => {
+        const { response } = await api.recipes.deleteTimelineEvent(props.slug, timelineEvents.value[index].id);
+        if (response?.status !== 200) {
+          alert.error(i18n.t("events.something-went-wrong") as string);
+          return;
+        }
+
+        timelineEvents.value.splice(index, 1);
+        alert.success(i18n.t("events.event-deleted") as string);
+      }, useAsyncKey());
```
hay-kot: Is there a reason you're using useAsync Here? That should only be necessary for top level statements.

frontend/components/Domain/Recipe/RecipeDialogTimeline.vue
```diff
@@ -0,0 +1,279 @@
...
+
+      useAsync(async () => {
+        const { response } = await api.recipes.updateTimelineEvent(props.slug, event.id, payload);
+        if (response?.status !== 200) {
+          alert.error(i18n.t("events.something-went-wrong") as string);
+          return;
+        }
+
+        alert.success(i18n.t("events.event-updated") as string);
+      }, useAsyncKey());
```
hay-kot: Is there a reason you're using useAsync Here? That should only be necessary for top level statements.

frontend/components/Domain/Recipe/RecipeDialogTimeline.vue
```diff
@@ -0,0 +1,279 @@
...
+            </v-row>
+          </v-timeline-item>
+        </v-timeline>
+      </v-card>
+      <v-card v-else>
+        <v-card-text class="justify-center pa-1">
+          {{ $t("recipe.timeline-is-empty") }}
+        </v-card-text>
+      </v-card>
+    </div>
```
hay-kot: A lot of this seems like duplicated code with a few attributes changed if mobile or not. Could we consolidate these a bit and use a computed property that returns and object that has the attributes we need for mobile/desktop?

Something like this?

```typescript

const attrs = computed(() => {
    if (mobile) {
       return { card: { dense: true } /* other stuff */ }
    }

    return { card: { dense: false } /* other stuff */ }
})

```

hay-kot (changes requested):
A few small things in the frontend but looks good overall. Given the previous merged, does anything here need to change?

michael-genson:
> A few small things in the frontend but looks good overall. Given the previous merged, does anything here need to change?

Thanks! I'll take a look at your comments - looking briefly they make sense.
Regarding other merges I don't think anything needs to change but I'll merge and run through everything and confirm

frontend/components/Domain/Recipe/RecipeDialogTimeline.vue
michael-genson: Oh duh, my Python brain assumed all functions were invoked synchronously, so I had to `await` outside of async. Instead I'll just make the function async. Thanks!

frontend/components/Domain/Recipe/RecipeDialogTimeline.vue
michael-genson: The differences between the two are more coupled than just attributes, but really that just means I need to clean it up more haha. Will work on this!

michael-genson:
I ended out changing the mobile view a bit to more closely match the desktop view, and I think it looks better:
image

I replaced most of the different attributes using an `attrs` object like you suggested, but I had to use some ternaries for other attributes (like `col`)

michael-genson:
@hay-kot this should be ready :)

PR #1836, opened 2022-11-16
Filtering special characters during automatic linking of ingredients …

### LINKED ISSUE #1803: [v1.0.0b] - Automatic "Link Ingredients" does not work when special characters are next to the word
### First Check

When using the "link ingredients" automatic feature, it only works for an exact match, which is to be expected. However, any special character like `,` en `**` (used to make ingredients bold in the instructions) will also break the feature.
Example:
`Add **sugar**, Mascarpone, cheese`
Only cheese will automatically get selected as a food. `sugar` has `**` next to it, `Mascarpone` has `,` next to it.

Note that with ```Add **sugar**, Mascarpone, cheese.```, `cheese` will not be linked, because of `.`

It could be that a food item has some special characters included, that should also be matched. But most typical sentence characters (, . and *) should probably be ignored, improving the workings of this feature.

### Deployment

Docker (Linux)

### Deployment Details

_No response_

jenscalaerts:
…to instructions

Used a unicode group to have a set of all unicode punctuation marks

## What type of PR is this?

- bug

_(REQUIRED)_
The pull request aims to improve 

Fixes #1803 

## Testing
Create recipe with ingredients:
* 2 eggs
* 3 potatoes
* 5 carrots
* 2 steaks
Create an instruction with containing words matching ingredients, who have punctuation marks or formatting characters attached to them. For example, 
```
add the **eggs**  to the 
potatoes, then mix
```
Press "Link Ingredients" on this instruction then press the auto button.
eggs and potatoes should be selected. 
## Release Notes

_(REQUIRED)_
```
Fixed bug where steps were not automatically linked to ingredients when special characters are attached to the word.
```


hay-kot (changes requested):
Could you pull this function into a composable and provide a unit test for it? Other than that it looks great!

jenscalaerts:
I have added filtering for 2 letter words in the instruction. They seemed to generate a large amount of false positives (in Dutch). This seems like a good idea because most languages have few 2 letter ingredients. It also replaces a large part of the blacklist, which has no internationalization.

hay-kot:
Excellent, thanks so much for following up with tests!

hay-kot (approved):

PR #1858, opened 2022-11-25
Tweaked Recipe Sort Behavior

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- feature

_(REQUIRED)_

Tweaks some of the behavior for sorting recipes on the frontend. It changes the default sort direction for certain `orderBy` attrs (e.g. `rating` will default to descending, thus showing the highest-rated recipes first).

It also filters out null values for certain sorts (e.g. `lastMade` will only show recipes that actually have a value for `lastMade` and not just stick the nulls at the end).

_(REQUIRED)_

N/A

I had to implement a slightly-dirty workaround for filtering out null datetimes for the `lastMade` date. I added a note to revisit the filter API to properly implement `IS NULL` and `IS NOT NULL`. In the meantime, this implementation will handle `<> null` as `>= 0001-1-1` strictly when addressing dates/datetimes.

## Testing

Manually via the frontend.

## Release Notes

_(REQUIRED)_

```release-note
improved recipe sorting to only show relevant recipes when sorting by certain attributes (such as rating)
```

### DIFF
--- a/frontend/components/Domain/Recipe/RecipeCardSection.vue
+++ b/frontend/components/Domain/Recipe/RecipeCardSection.vue
@@ -243,7 +243,10 @@ export default defineComponent({
         cookbook.value,
         category.value,
         tag.value,
-        tool.value
+        tool.value,
+
+        // filter out recipes that have a null value for the property we're sorting by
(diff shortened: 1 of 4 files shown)


hay-kot:
LGTM 🚀 🚀 🚀

PR #1915, opened 2022-12-23
Convert scraper to use async

### LINKED ISSUE #1256: [v1.0.0b] [Task] - Convert scraper to use async

Currently the implementation of the scraper and endpoint are using a synchronous method. This limits scalability for concurrent scrapers. 

With the new release of recipe-scrapers, you can now directly pass html to the scraper to get data. We should utilize and async `requests` alternative to get the html and feed it to the scraper. This would also allow us to fetch images in an async manner and speed up the overall processing of a recipe.

fleshgolem:
## What type of PR is this?

- feature

As mentioned in #1256 most of the scraper works synchronously which is not as efficient as can be. This PR changes the "create-url"-routes to run asynchronously from the ground up and changes downloading to use an httpx AsyncClient. This should help alleviate the existing issues and also make it easier to add other asynchronous code in parsers in the future 

Fixes #1256

## Testing

Used `test_recipe_parser` (which has unrelated issues, for which i added a workaround), `test_create_by_url` and `test_bulk_import` to test all changed functionality. Also ran the frontend locally and imported all the test recipes manually

There are 2 other unrelating tests that are failing right now

## Release Notes

```release-note
Improve parser efficiency by running downloads asynchronously
```


hay-kot:
Thanks for cleaning up those conflict! Cursory overview looks good overall. Will give it a deep look this weekend and get it merged. Thanks for this one!

tests/utils/recipe_data.py
```diff
@@ -38,14 +38,15 @@ def get_recipe_test_cases():
...
+            # For some reason, parsing instructions from cookpad seems to be broken in recipe-parsers right now
+            num_steps=0,
         ),
         RecipeSiteTestCase(
             url="https://www.greatbritishchefs.com/recipes/jam-roly-poly-recipe",
             html="jam-roly-poly-with-custard.html",
             html_file=test_data.html_jam_roly_poly_with_custard,
             expected_slug="jam-roly-poly-with-custard",
-            num_ingredients=13,
+            num_ingredients=11,
```
hay-kot: Any idea what's going on with these? They seem to be fine in the other branches, but I don't see anything that was changed that would have caused this problem?

mealie/services/recipe/recipe_data_service.py
```diff
@@ -13,29 +12,32 @@
...
+        for response in responses:
 
-                len_int = int(response.headers.get("Content-Length", 0))
-                if len_int > largest_len:
-                    largest_url = response.url
-                    largest_len = len_int
+            len_int = int(response.headers.get("Content-Length", 0))
+            if len_int > largest_len:
+                largest_url = str(response.url)
+                largest_len = len_int
```
hay-kot: can you add a type annotation for the responses declaration? That way we know that you're casting a URL type to a str later on.

hay-kot (changes requested):
This looks pretty good overall. I do think we hit a regression somewhere along the lines that's causing some different behavior.

When using the nightly to scrape this url

```
https://www.rezeptwelt.de/backen-herzhaft-rezepte/schinken-kaese-waffeln-ohne-viel-schnickschnack/4j0bkiig-94d4d-106529-cfcd2-is97x2ml
```

I get 7 ingredients, and 1 instruction, however that's not the case when I pulled down this branch. I don't get any of the ingredients. 

I also think that we need to update the `test_create_by_url` to make assertions based on the length of ingredients and instructions

```python
    recipe = api_client.get(api_routes.recipes_slug(recipe_data.expected_slug), headers=unique_user.token)

    assert recipe.status_code == 200

    recipe_dict: dict = json.loads(recipe.text)

    assert recipe_dict["slug"] == recipe_data.expected_slug
    assert len(recipe_dict["recipeInstructions"]) == recipe_data.num_steps
    assert len(recipe_dict["recipeIngredient"]) == recipe_data.num_ingredients
```

That will give us a more accurate idea of when we have a regressions vs a problem with an outside website.

**Can you check that URL and add those assertions to the `test_create_by_url` test?**

https://github.com/hay-kot/mealie/blob/9919698cb8a4f7753a5b5bfb437aa64ea7821bdc/tests/integration_tests/user_recipe_tests/test_recipe_crud.py#L58

mealie/services/recipe/recipe_data_service.py
fleshgolem: Yeah, sure I can take a look at both

fleshgolem:
> I also think that we need to update the test_create_by_url to make assertions based on the length of ingredients and instructions

I added this. This does conflict with the changes in recipe_data i made for test_recipe_parser before, so i reverted those, since this case will run on every test run. This means that test_recipe_parser will start failing again if anyone uses it for any future features, but I really dont now how to reconcile this right now...

 > can you add a type annotation for the responses declaration? That way we know that you're casting a URL type to a str later on.

Done

> I get 7 ingredients, and 1 instruction, however that's not the case when I pulled down this branch. I don't get any of the ingredients.

I really cant reproduce this. This URL and page is already part of the test cases and it works for me without issues even with the added checks from above (as do the automated github test runs apparently). I also imported the url manually from the frontend and that worked fine as well.
I did merge all changes from mealie-next and  did a poetry install before working on this again. I am really not sure what could differ anywhere to cause this issue, but it is of course a problem. Are you getting any logged exceptions anywhere that could help

hay-kot:
> I added this. This does conflict with the changes in recipe_data i made for test_recipe_parser before, so i reverted those, since this case will run on every test run. This means that test_recipe_parser will start failing again if anyone uses it for any future features, but I really dont now how to reconcile this right now...

I think that's okay...? I'm more concerned with what happens once we get the data, so I'm okay merging with those tests broken. Ideally we'd spin up some kind of server and isolate those tests so they wouldn't have to hit the external internet. 

> I really cant reproduce this. This URL and page is already part of the test cases and it works for me without issues even with the added checks from above (as do the automated github test runs apparently). I also imported the url manually from the frontend and that worked fine as well.

I may be a region issue. Sometimes I just can't get sites to work that other people say work fine. 🤷 

---

That said, I think this is good to merge and we can create an issue to track those failing tests. Does that sound good?

fleshgolem:
Yeah sure, that works for me

PR #1998, opened 2023-01-07
Fix example postgres docker-compose setup in docs

fleshgolem:
Add a local volume to postgres container so changes get persisted between restarts

## What type of PR is this?

- documentation

As it turned out on the discord yesterday, the example docker-compose.yml for a postgres setup is actually fairly misleading which can easily lead to frustrations: The postgres container does not get any volume attached, so it will never persist any data between restarts.

This quick change should alleviate that issue for inexperienced users (even though they probably shouldnt use postgres to begin with)

AFAIK no issue has been filed for this

## Release Notes

```release-note
* Improved sample postgres setup in installation docs
```


hay-kot:
> The postgres container does not get any volume attached, so it will never persist any data between restarts.

Postgres uses named volumes defined in the Dockerfile so the data should be persistent by default without an explicit declaration. Though I think this is a good addition since explicit is better than relying implicit behavior.

Thanks for the PR!

docs/docs/documentation/getting-started/installation/postgres.md
```diff
@@ -55,13 +55,17 @@ services:
     container_name: postgres
     image: postgres
     restart: always
+    volumes:
+      - mealie-pgdata:/app/data/
```
hay-kot: Should this be?

```yaml
- mealie-pgdata:/var/lib/postgresql
```

fleshgolem: Yeah, of course, now i didnt pay attention myself...
Will fix that in a moment

PR #2021, opened 2023-01-09
Refactor Shopping List API

### LINKED ISSUE #2068: [v1.0.0b] - Make quantity optional for shopping list
### First Check

When adding unparsed ingredients from a recipe to a list through the "Add to shopping list" feature, the items appear with a default quantity `1`, because the quantity is already a part of the ingredient text. This makes the list a bit hard to read.

Ideally all the recipes would be already parsed for quantity, but since that step is currently optional this is the default behavior when adding items to a list.

Maybe a checkbox to hide the quantity would be a good short-term workaround for this?

image

### Deployment

Docker (Linux)

### Deployment Details

_No response_

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- bug
- cleanup
- feature

_(REQUIRED)_

This PR re-writes the shopping list items API to make it significantly more efficient and knock out a ton of weird edge-case bugs. It also adds a few quality of life improvements for developers.

Schema changes:
- tidied up the list item schema so type hinting is more accurate (removed some redundant fields and removed unions)
- added a validator to populate a new "display" property on `ShoppingListItemOut` which does all the food + unit calculations
- added a validator to populate a list item's label with the food's label on the backend (if the food has a label and the item does not)

Service changes:
- all requests check for merges with each other (in bulk ops) and with existing items. This means a create request may result in an update operation, and not a create operation. This also means an update request can result in a delete operation instead of an update operation. In bulk operations these scenarios can be mixed
- checked items never merge
- if an item is checked off, its recipe references are deleted
- if all items in a recipe are checked off, the recipe reference at the list level is deleted

API changes:
- refactored each bulk operation to only perform one or two database calls (besides queries)
- changed API response to include a collection of created items, updated items, and deleted items, since all operations check for merges now

Frontend changes:
- removed old formatting util and reference the new display property from the backend
- fixed a bug with re-ordering items (completely unrelated to the backend changes, it exists in nightly)

_(REQUIRED)_

My notepad++ doc of accumulated edge cases... :)

Resolves #2068

The main thing I'm not entirely sure about is the changes I made to the generic repository: I added "update many" and "delete many" functions, which I _think_ are as efficient as reasonably possible, without messing with combining multiple bulk ops in the same transaction.

One issue I ran into is trying to provide an arbitrary `match_value` like in the `update_one` function. The reason for this issue is because I couldn't figure out how to use `IN` with the `filter_by`  method, so I just hardcoded filtering by the `id` attribute (using `filter` instead of `filter_by`). Realistically I don't know why you'd _ever_ want to match by anything other than id in an update many scenario, but it is a limitation I added. For the same reason the limitation exists in the new "delete many" function, too.

## Testing

I'm pretty sure the vast majority of the additions come from pytest. I also played around in the frontend for a while to make sure that's all working properly, but I barely changed anything there.

## Release Notes

_(REQUIRED)_
```release-note
improved shopping list stability
```


michael-genson:
Regarding the recipe reference, I had to switch the `NoneFloat` back to a `float` (see https://github.com/hay-kot/mealie/pull/1677) because the database set that field to `nullable=False`

I also added a test for it this time

PR #2025, opened 2023-01-11
Upload recipe step images from mobile devices

xarantolus:
## What type of PR is this?

- feature

This adds a button for uploading images in the recipe step dropdown, as not all mobile devices can drag and drop a file into the web page. This allows just taking a quick picture and adding it to the recipe from your phone without having to copy the image to a PC. It reuses the existing upload logic.

https://user-images.githubusercontent.com/32465636/211751894-a96bf984-6b81-4522-b679-7e8aab693dad.mp4

While #885 is already implemented, [this comment](https://github.com/hay-kot/mealie/discussions/885#discussioncomment-4452620) mentioned that it's currently not possible to use from a mobile browser.

This is currently a draft PR because I want to add some kind of loading animation. As you can see in the demo, the upload takes 3 seconds and doesn't display anything during that time, which is a bit confusing. If you have any suggestions for this please feel free to add them :)

## Testing
I tested this manually in chromium- and firefox-based mobile as well as desktop browsers. I don't have any iOS devices, so I could **not** test this in Safari.

## Release Notes

```release-note
Add a button that allows uploading recipe step images in mobile browsers
```

### DIFF
--- a/frontend/components/Domain/Recipe/RecipePage/RecipePageParts/RecipePageInstructions.vue
+++ b/frontend/components/Domain/Recipe/RecipePage/RecipePageParts/RecipePageInstructions.vue
@@ -174,6 +174,8 @@
                 </v-fade-transition>
               </v-card-title>
 
+              <v-progress-linear v-if="isEditForm && loadingStates[index]" :active="true" :indeterminate="true" />
+
               <!-- Content -->
               <DropZone @drop="(f) => handleImageDrop(index, f)">
                 <v-card-text
@@ -553,6 +555,8 @@ export default defineComponent({
       }
     });
 
+    const loadingStates = ref<{ [key: number]: boolean }>({});
+
     async function handleImageDrop(index: number, files: File[]) {
       if (!files) {
         return;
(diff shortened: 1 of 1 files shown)


xarantolus:
I have now added a progress bar, but I'm not super happy with how it looks on mobile devices. On desktop it's right above the the text input box of the step the image was dragged into, but on mobile the progress bar is always shown at the top of the page. This looks like another page is loading instead of just that one recipe step. Setting properties like [`absolute`](https://vuetifyjs.com/en/api/v-progress-linear/) to false on the progress bar did not appear to help, or at least I couldn't find the right combination.

xarantolus:
Now that I've tried it out a bit more I kind of think it's OK to have the progress bar like this on mobile, but would be happy to know others opinions on that :)
The mobile view looks as in the previous video, on desktop the progress bar can be seen here (with a simulated slower connection):


Recording 2023-01-25 at 09 54 04


If you want to I can also remove the progress bar again, it was more of an experiment that didn't really work out on mobile 🤔

hay-kot:
I like it, looks good!

PR #2029, opened 2023-01-13
doc(single-container): add hint on the exposed ports

bentolor:
## What type of PR is this?

_(REQUIRED)_

- documentation

Be explicit on the single container usage.

none


hay-kot:
I think we could just add port bindings to the docker-compose file? That might be a little more clear for most users.

bentolor:
Hi @hay-kot ! Thanks for your quick feedback. I kept my text note and added the port bindings. I felt the purpose of those ports are not self-explanatory. 

That's okay for you?

PR #2040, opened 2023-01-16
Add SSO auth support

knrdl:
This PR adds SSO support. The workflow looks like this:

1. user wants to open mealie, sends a request
2. therefore the RP (reverse proxy) forwards the request to the IDP (identity provider)
3. IDP handles user auth and provides user info via http headers to RP (user info can also be empty, if desired)
4. RP forwards the request (with additional headers) to mealie
5. Mealie frontend gets application info from endpoint /api/app/about on load. It contains a new bool field "sso_login_available" which is true if sso is enabled in settings and at least the "Remote-User" header is present
    * if sso_login_available=false the regular login form is presented to the user
    * if sso_login_available=true the frontend fires an authentication request (without username or password) to get a session token from the backend. The backend checks whether it is fulfilling a SSO login or credentials must be checked.
7. user gets logged in and can use mealie
8. to logout a custom logout link to the IDP can be provided.

**This PR is mostly based on the work of @tribut in https://github.com/hay-kot/mealie/pull/1622**
However there are some notable differences:
1. in the version by @tribut SSO was exclusive (either SSO xor internal auth/ldap). This PR allows internal/ldap as fallback
2. if the user is not logged in there is no redirect to the IDP (identity provider). The reverse proxy could handle this initial redirect instead if desired
3. instead of only "Remote-User" also the headers "Remote-Name" and "Remote-Email" are supported
4. no duplication of config parameter between frontend and backend

Please have  look!


tribut:
Thanks for picking this up!

I currently don't have time to test this, but afair the reason why the redirect was necessary was due to mealie being a PWA, the browser will not always request the login page from the proxy. Have you tested if having a redirect in the proxy works reliably?

knrdl:
Good point @tribut, I haven't thought about this edge case.
Assuming the following: Website is installed as PWA and user is logged in. The mealie session is valid but the IDP session is expired. The next http request will result in a 401 by the reverse proxy. So maybe redirect to the IDP URL on every 401 response is a solution? or am i missing something?

mealie/core/security/security.py
```diff
@@ -108,16 +109,34 @@ def user_from_ldap(db: AllRepositories, username: str, password: str) -> Private
...
-def authenticate_user(session, email: str, password: str) -> PrivateUser | bool:
+def authenticate_user_sso(session, username: str, email: str | None, name: str | None) -> PrivateUser:
+    db = get_repositories(session)
+    user = db.users.get_one(username, "username", any_case=True)
+    if not user:
+        user = db.users.create(
+            {
+                "username": username,
+                "password": "SSO",
+                # Fill the next two values with something unique and vaguely
```
namelivia, line 120: Could this be a security risk? Could you log in through the login form using "SSO" for this user? In that would be the case, maybe a randomly generated password could add some security.

mealie/routes/users/crud.py
```diff
@@ -61,7 +61,8 @@ def get_logged_in_user(self):
     @user_router.put("/password")
     def update_password(self, password_change: ChangePassword):
         """Resets the User Password"""
-        if not verify_password(password_change.current_password, self.user.password):
+        # when logged in via SSO, do not check old password
+        if self.user.password != "SSO" and not verify_password(password_change.current_password, self.user.password):
```
namelivia, line 65: Could reusing the password field with a specific value be a problem in the future? It is giving the password field two meanings, maybe a boolean value for the user could be a safer option.

namelivia:
Nice! It would be very nice to have this in Mealie, if you need any kind of help testing or whatever is needed to help you push this forward let me know.

mealie/routes/users/crud.py
knrdl: > Could reusing the password field with a specific value be a problem in the future? It is giving the password field two meanings, maybe a boolean value for the user could be a safer option.

Passwords are not stored as plaintext. They are always hashed, and only the hash value gets persisted. Currently bcrypt is in use as hashing algorithm, see https://github.com/hay-kot/mealie/blob/39012adcc1b431c59ff9d83f75100971ffa540e0/mealie/core/security/hasher.py#L27
The hash values have a fixed size which is always longer than 3 characters (e.g. "SSO"). So the value "SSO" is an invalid hash value. Hashing always produces valid hash values. So there is no chance of overlapping values.

TL;DR no security problem here

knrdl:
> Nice! It would be very nice to have this in Mealie, if you need any kind of help testing or whatever is needed to help you push this forward let me know.

Implementing 99% of this feature is easy and done by this pull request. But the last 1% are nasty and non trivial. That's why I marked the PR as draft.

As I chose Tandoor Recipes instead of Mealie for now, I am currently not pushing this any further. Maybe others can help.

namelivia:
> > Nice! It would be very nice to have this in Mealie, if you need any kind of help testing or whatever is needed to help you push this forward let me know.
> 
> Implementing 99% of this feature is easy and done by this pull request. But the last 1% are nasty and non trivial. That's why I marked the PR as draft.
> 
> As I chose Tandoor Recipes instead of Mealie for now, I am currently not pushing this any further. Maybe others can help.

I see, well, thanks for your work!

aroberts:
@knrdl can you please list the issues you currently see that aren't handled by PR (that last 1%), so someone else can pick up the work?

knrdl:
Sure, the PWA/ServiceWorker aspect in conjunction with the Reverse Proxy is not fully handled or tested yet:
* The SSO relevant endpoints (app_about, get_token) must not be cached by the SW
* When the user is offline and the session expires, a redirect to the IDP cannot take place. Otherwise a Server Not Found / Timeout or similar page would be shown.
* If the IDP session expires but the Mealie session is still valid, 401(or 403) api responses will be encountered. These should be handled by a logout. The action triggered by the user will then be lost(?)
* If SSO login is mandatory (the only allowed auth method) the reverse proxy must enforce this. But in a PWA setup the reverse proxy is not involved at all. Therefore the fallback described by  @tribut  in https://github.com/hay-kot/mealie/pull/2040#issuecomment-1386178859 could be applied by introducing an addition frontend env var like `SSO_MANDATORY_LOGIN_URL`.  So mandatory SSO would not be initiated by the RP but Mealie instead.

These are fewer problems than I thought there remained. If I find the time maybe I will take another look into it. Or I forgot some of the edge cases, who knows :)

aroberts:
thanks for the rundown!

morgan-dgk:
I'd be keen to pick this up and see if I can get this over the line with the previous work done! I've taken note of the outstanding work mentioned by the author above and will see how far I can get.

PR #2042, opened 2023-01-17
better cooking on phone with landscape

jlssmt:
## What type of PR is this?
- feature

cook mode needs to link ingredients.
a simpler cook mode without maintenance is col-sm split 6 instead of 12 to get the side-by-side view also on phones in landscape like on iPad.
the `overflow: scroll` is added so you can scroll left and right independently (ingredients keep showing while scrolling the instructions)

see:
image

--

scrolling can be removed if wished.

## Testing

chrome dev tools
NOT on phone. sorry.


hay-kot:
I don't thinks this is an improvement over the current implementation. <s> It's basically unusable on mobile as the view is so small on side-by-side. </s> I was incorrect it doesn't apply at smaller screens

I think we'd be better off improving the auto-linking of ingredients to steps OR provide a slide up menu to show ingredients in the event no ingredients are linked instead of trying to manage these different types of views for different devices

jlssmt:
It is not a replacement of the current Cooking Mode (which I also find good) but an addition. In my opinion, the landscape is completely useless on mobile without the PR. I didn't come up with the idea either, but took it from other applications that have implemented exactly such a landscape view for mobile phones (e.g. chefkoch app). And it should work on all small devices (it's col-sm-6)

PR #2044, opened 2023-01-18
Localize hard-coded texts

sephrat:
## What type of PR is this?

- feature

Transform hard-coded texts into localizable texts

- In this PR, I will focus on end-user localization (admin pages may come later in another PR).
- Some work still needs to be done on the backend, but I'd like to postpone it to another PR
- Work still needed on integrating Vuetify locale setter ✋ help wanted
  - For now, I've implemented a dirty hack to set vuetify language as early as possible  (see af67eb5cad59e2c02452490cae8f615135f4a7ba)
  - The setter won't work because of the window reload. The language does not seem to persist after reload

## Release Notes

NONE


sephrat:
Ready for review now!

hay-kot:
Strange why the linters picking up those errors now, Merging anyways. We'll deal with them in another PR since I know those are only called within setup it shouldn't be an issue.

PR #2050, opened 2023-01-22
Feat/timeline event for mealplans

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- feature

_(REQUIRED)_

This PR makes it so when a meal plan is scheduled for _today_ (as a daily scheduled task), a recipe timeline event is created and the `last_made` property on the recipe is updated (when applicable).

Because timeline events require a user, I had to add the user id to the meal plan table (as a foreign key). I kept it optional in the schema for compatibility, but when meal plans are created they now add the user automatically.

_(REQUIRED)_

Resolves discussion #2037

I used batch ops in the migration and manually named the foreign key relationship [per the alembic docs](https://alembic.sqlalchemy.org/en/latest/batch.html#dropping-unnamed-or-named-foreign-key-constraints) ([and this SO thread](https://stackoverflow.com/questions/30378233/sqlite-lack-of-alter-support-alembic-migration-failing-because-of-this-solutio)). Seems like this is the right way to do it, but it's not something I've come across before so let me know if there's a better way.

The solution says to mark `render_as_batch=True` but all that does is change how alembic auto-generates the migration file, so I didn't actually change our `env.py` config since I don't know if it has other implications (I changed it, generated the file, then changed it back). Might be worth changing permanently, but the workaround is easy enough for development purposes.

## Testing

Added a bunch of backend tests. I didn't test the frontend because of that frontend dev issue mentioned in Discord, but this is purely backend anyway.

## Release Notes

_(REQUIRED)_

```release-note
meal plans now create recipe timeline events and update the recipe's last made date
```


michael-genson:
Looks like the backend failures were only the network requests for that one recipe site that's having issues. If I find time I'll get a PR in to get rid of those requests. Hopefully by the time you get to this PR you can just re-run the workflow and tests will pass. Locally I've seen it work a few times (but I definitely want to get it to work here because I don't have postgres set up locally)

PR #2054, opened 2023-01-23
Show the group name in the AppHeader

RealFoxie:
## What type of PR is this?

- feature

Instead of showing "Mealie" on every instance of mealie, it will instead show the group name of the user. This makes the page more personal and makes it easier to distinguish groups when you have multiple accounts.

I am not yet as familiar with the code, so I'm not sure if some kind of null check would be needed.

## Release Notes

_(REQUIRED)_

```
Show the group name in the header instead of Mealie
```

### DIFF
--- a/frontend/components/Layout/AppHeader.vue
+++ b/frontend/components/Layout/AppHeader.vue
@@ -8,7 +8,7 @@
     </router-link>
 
     <div btn class="pl-2">
-      <v-toolbar-title style="cursor: pointer" @click="$router.push('/')"> Mealie </v-toolbar-title>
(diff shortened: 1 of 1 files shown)


zierbeek:
maybe a toggle would be nice or as an env variable so it's not obligatory?

RealFoxie:
I'd think that something like this was always supposed to be done (instead of having it show "Mealie" on release too) and having a toggle is mostly to avoid the change.
But I might be wrong, and then an env variable would make sense.

hay-kot:
Hey! Thanks for the PR.

I don't think this belongs in the Header. I'd support adding this as a sub-heading to the sidebar though

CleanShot 2023-01-24 at 13 12 48@2x

PR #2062, opened 2023-01-27
Fix/all recipes performance

fleshgolem:
## What type of PR is this?

- bug

Calling /api/recipes with `loadFood: false` (or any other endpoint that lists recipes) would lead to an interesting case, where all the recipe ingredients are not eager loaded any more but pydantic would still try to fill the `recipe_ingredient` field of the response model in orm mode, i.e. sqlalchemy will generate a new lazy-loading ingredient query for every single recipe. I assume this issue gets even worse, if you use ingredient foods/units, but I have not tested this any further. 
I created a new response model that explicitly has the ingredients and removed them from the normal recipe summary, so only /api/recipes with loadFood enabled should ever try to load them now and all other endpoints that list recipes should behave the correct way.

Additionally - as discussed on discord - I also moved all validation from RecipeSummary to Recipe, since they are never needed for output validation. They are still needed on Recipe, otherwise all the migration tests will fail (and I assume the migrations wont work anymore)

This hopefully should improve performance for loading recipe pages considerably, especially for slower devices that really seem to have problems here

No explicit issue exists AFAIK, but it has been raised on discord a few times

## Testing

Ran all automated tests and also tried listing everything on the locally running frontend

## Release Notes

```release-note
* Improved recipe listing performance
```

### DIFF
--- a/mealie/repos/repository_recipes.py
+++ b/mealie/repos/repository_recipes.py
@@ -156,6 +156,7 @@ class RepositoryRecipes(RepositoryGeneric[Recipe, RecipeModel]):
 
         if load_food:
             args.append(joinedload(RecipeModel.recipe_ingredient).options(joinedload(RecipeIngredient.food)))
+            args.append(joinedload(RecipeModel.recipe_ingredient).options(joinedload(RecipeIngredient.unit)))
             item_class = RecipeSummaryWithIngredients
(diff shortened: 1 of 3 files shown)


hay-kot:
So good. Thank you! 🚀

PR #2070, opened 2023-01-28
Feature/sqlalchemy 2.0

fleshgolem:
## What type of PR is this?

- cleanup

Well... these changes allow migration to sqlalchemy 2.0, that part should be obvious
Truth be told, only very few of these changes are needed to actually make it work, neither the new mapping declarations nor the 2.0 style querying are needed (at least of right now) for the app to still work.
However

1) session.query is considered legacy now and this should hopefully future-proof everything enough for that to never become a problem
2) while using core select might seem unfamiliar at first, from personal experience you get used to it rather quickly and it has the added benefit of being closer to writing actual sql queries
3) the new mapping style should hopefully improve typing support at least to some degree
4) if there is ever any plan to move to async sqlalchemy this should hopefully make that task easier since in those cases core select querying is required

So in the end, IMO this is probably as good as a time as any to move everything to the new APIs

The errors in #2061 i guess

If you are not comfortable with all these changes right now and just want sqlalchemy 2.0 to work, all necessary changes are in 07e14b26f9be4b0ffd80fb0e7e16644a6f542477 in the files `db_setup.py`, `model_base.py` and `auto_init.py`

## Testing

I made sure all automated tests still work the same as before. I also tried to test as many things as I can with the locally running frontend. This app has a lot of features however and I would assume that I missed to test something somewhere, so it might not be the worst idea to have another round of manual testing before merging this

## Release Notes

```release-note
NONE
```


hay-kot:
I tried to resolve the conflicts over in https://github.com/hay-kot/mealie/pull/2096 but I exploded something :(

Will try to fix later this week. 

--- 

Amazing work in the PR really stoked you went through this. I was dreading it

hay-kot:
Changes incorporated in #2096  - Thanks so much for all the work on this

PR #2109, opened 2023-02-09
Update docker-compose.yml

rocket357:
ERROR: The Compose file './docker-compose.yml' is invalid because: services.mealie.environment.WEB_GUNICORN contains true, which is an invalid type, it should be a string, number, or a null

## What type of PR is this?

- bug

ERROR: The Compose file './docker-compose.yml' is invalid because:
services.mealie.environment.WEB_GUNICORN contains true, which is an invalid type, it should be a string, number, or a null

N/A - Should I open an issue first, or is jumping straight to PR sufficient?

None, simple fix.

## Testing

```
docker01:~# docker version
Client:
 Version:           20.10.16
 API version:       1.41
 Go version:        go1.18.2
 Git commit:        aa7e414fdcb23a66e8fabbef0a560ef1769eace5
 Built:             Sun May 15 14:51:52 2022
 OS/Arch:           linux/amd64
 Context:           default
 Experimental:      true

Server:
 Engine:
  Version:          20.10.16
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.18.2
  Git commit:       f756502055d2e36a84f2068e6620bea5ecf09058
  Built:            Sun May 15 14:46:21 2022
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          v1.6.6
  GitCommit:        10c12954828e7c7c9b6e0ea9b0c02b01407d3ae1
 runc:
  Version:          1.1.2
  GitCommit:        a916309fff0f838eb94e928713dbc3c0d0ac7aa4
 docker-init:
  Version:          0.19.0
  GitCommit:
```

## Release Notes

N/A

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -65,7 +65,7 @@ services:
 
       # =====================================
       # Web Concurrency
-      WEB_GUNICORN: true
+      WEB_GUNICORN: "true"
       WORKERS_PER_CORE: 0.5
       MAX_WORKERS: 1
       WEB_CONCURRENCY: 1
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -53,6 +53,7 @@ services:
       - 9092:9000
     environment:
       ALLOW_SIGNUP: "false"
+      BASE_URL: "http://localhost:8080" # changeme
 
       DB_ENGINE: sqlite # Optional: 'sqlite', 'postgres'
       # =====================================
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -72,13 +72,13 @@ services:
 
       # =====================================
       # Email Configuration
-      # SMTP_HOST=
-      # SMTP_PORT=587
-      # SMTP_FROM_NAME=Mealie
-      # SMTP_AUTH_STRATEGY=TLS # Options: 'TLS', 'SSL', 'NONE'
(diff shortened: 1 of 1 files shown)


rocket357:
Updated the PR to fix a few issues I ran into on enabling SMTP as well.  If I need to split anything off (since it was a basic bugfix to begin with), please let me know!  =)

docker/docker-compose.yml
```diff
@@ -65,20 +66,20 @@ services:
 
       # =====================================
       # Web Concurrency
-      WEB_GUNICORN: true
+      WEB_GUNICORN: "true"
```
hay-kot, line 67: This is the only change I think is warranted. 

This docker-compose file is meant for end users so the annotations aren't super useful. For end-users you should be using the documentation. This also doesn't seem to be a problem on newer versions of docker-compose? 🤷 

- https://nightly.mealie.io/documentation/getting-started/installation/postgres/

Configuration is also documented 

- https://nightly.mealie.io/documentation/getting-started/installation/backend-config/
- https://nightly.mealie.io/documentation/getting-started/installation/frontend-config/

hay-kot (changes requested):
Can you remove the rest of the changes, or let me know why we need to make these changes if it's no intended for end users?

rocket357:
Yep, I got a little over-zealous making changes =)

I'll revert the other two commits.  Thanks!

PR #2139, opened 2023-02-16
Fix: recipe scraper image cleaning

### LINKED ISSUE #2087: [v1.0.0b5] - Issue Scraping from Delish
### First Check

Fails to find recipe ingredients on the following page https://www.delish.com/cooking/recipe-ideas/a19677130/instant-pot-chicken-and-rice-recipe/

### Deployment

Docker (Linux)

### Deployment Details

_No response_

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- bug

_(REQUIRED)_

The recipe scraper wasn't accounting for nested dictionaries of image urls, as mentioned in #2087. I found that the scraper actually wasn't cleaning images at all, so I enabled that.

_(REQUIRED)_

resolves #2087

## Release Notes

_(REQUIRED)_

```release-note
recipe scraper has a higher chance of finding an image when scraping
```

### DIFF
--- a/mealie/services/scraper/cleaner.py
+++ b/mealie/services/scraper/cleaner.py
@@ -47,7 +47,7 @@ def clean(recipe_data: dict, url=None) -> dict:
     recipe_data["recipeYield"] = clean_yield(recipe_data.get("recipeYield"))
     recipe_data["recipeIngredient"] = clean_ingredients(recipe_data.get("recipeIngredient", []))
     recipe_data["recipeInstructions"] = clean_instructions(recipe_data.get("recipeInstructions", []))
-    recipe_data["image"] = clean_image(recipe_data.get("image"))
+    recipe_data["image"] = clean_image(recipe_data.get("image"))[0]
     recipe_data["slug"] = slugify(recipe_data.get("name", ""))
     recipe_data["orgURL"] = url
 
@@ -77,7 +77,7 @@ def clean_string(text: str | list | int) -> str:
     return cleaned_text
 
 
-def clean_image(image: str | list | dict | None = None, default="no image") -> str:
+def clean_image(image: str | list | dict | None = None, default: str = "no image") -> list[str]:
     """
     image attempts to parse the image field from a recipe and return a string. Currenty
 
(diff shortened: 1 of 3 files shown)


mealie/services/scraper/cleaner.py
```diff
@@ -95,15 +96,20 @@ def clean_image(image: str | list | dict | None = None, default="no image") -> s
...
 
-    match image:  # noqa - match statement not supported
+    match image:
         case str(image):
             return image
+        case {"url": str(url)}:
+            return url
         case list(image):
-            return image[0]
-        case {"url": str(image)}:
```
fleshgolem, line 103: Is it correct for this case to be completely removed though?

michael-genson: I just moved it slightly down so it's easier to read, it's not gone

fleshgolem: Right, it seems like i am a bit blind

mealie/services/scraper/cleaner.py
```diff
@@ -95,15 +96,20 @@ def clean_image(image: str | list | dict | None = None, default="no image") -> s
...
-            return image
-        case _:
-            raise TypeError(f"Unexpected type for image: {type(image)}, {image}")
+            for image_data in image:
+                match image_data:
+                    case str(image_data):
+                        return image_data
```
fleshgolem: I am not a 100% sure about this, but as far as i remember this will ultimately get used by the recipe_data_service to retrieve the image. That service however also accepts a list of image-urls and will use that to determine the largest one, so it would make sense here to not just return the first hit but the full list of urls so the best one can decided on further down the chain

hay-kot: @fleshgolem I was also looking at this and confused on why this returns a string. We do have logic that fetches the largest image so I'm wondering if this should _really_ just return a list to begin with? 

@michael-genson I'm also not sure we need to nested match statement here - this is sort of what I was thinking

```python
match image:  # noqa - match statement not supported
    case str(image):
        return [image]
    case [str(_), *_]:
        return image
    case [{"url": str(_)}, *_]:
        return [x["url"] for x in image]
    case {"url": str(image)}:
        return [image]
    case _:
        raise TypeError(f"Unexpected type for image: {type(image)}, {image}")
```

That would catch all cases we're looking at now right? This returns the list, but could also just return the first resulting match as needed. I don't think we need recursion like I initially thought though.

michael-genson: LGTM. I'll update it and make sure it passes the original issue's scrape

michael-genson:
Per discussion I refactored `clean_image` to return a list of strings, rather than a single string. At the final step (cleaning before loading data into the database) we select the first image in the list as _the_ recipe image. Since `clean_image` always returns a non-empty list (using a default string), this will never produce an error.

There's something to be said about the image tag being cleaned twice (@fleshgolem raised this in discord), but I wasn't comfortable dropping it in either place since we're mixing two different workflows (scraping and migrations) and I don't think both workflows use the cleaning functions the same way.

PR #2140, opened 2023-02-16
Fix: Allow Last Made to be Updated on Locked Recipes

### LINKED ISSUE #2089: [v1.0.0b] - Permission denied when saving "I Made This" on locked recipe
### First Check

Permission denied when saving "I Made This" on locked recipe.  Entry is added to timeline but "Last Made" time is not updated

```
02/03/2023 4:05:01 PM
INFO:     172.20.0.2:34988 - "POST /api/recipes/quick-pickled-red-onions/timeline/events HTTP/1.1" 201 Created
02/03/2023 4:05:01 PM
ERROR: 03-Feb-23 16:05:01 	Permission Denied on recipe controller action
```

### Deployment

Docker (Linux)

### Deployment Details

_No response_

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- bug

_(REQUIRED)_

This allows certain fields to bypass the locked state of a recipe in a patch request. All fields in the patch request _must_ be explicitly allowed (e.g. a request with `name` and `last_made` will fail with a 403 error). At the moment this only applies to `last_made`.

_(REQUIRED)_

resolves #2089

## Testing

pytest

## Release Notes

_(REQUIRED)_

```release-note
fixed bug with locked recipes that prevented users from using the "I Made This" button
```

### DIFF
--- a/frontend/components/Domain/Recipe/RecipeLastMade.vue
+++ b/frontend/components/Domain/Recipe/RecipeLastMade.vue
@@ -126,8 +126,7 @@ export default defineComponent({
 
       // we also update the recipe's last made value
       if (!props.value || newTimelineEvent.value.timestamp > props.value) {
-        const payload = {lastMade: newTimelineEvent.value.timestamp};
-        actions.push(userApi.recipes.patchOne(props.recipeSlug, payload));
+        actions.push(userApi.recipes.updateLastMade(props.recipeSlug,  newTimelineEvent.value.timestamp));
 
         // update recipe in parent so the user can see it
         // we remove the trailing "Z" since this is how the API returns it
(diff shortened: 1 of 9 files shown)


michael-genson:
Apparently `last_made` wasn't being backed up properly, so the restore fails. I'm surprised no one flagged this yet. Should be good now though

hay-kot (commented):
PR looks good overall. 

Would it make more sense to just use a dedicated endpoint like the favorites endpoint? That way we don't expose ourselves to any issues down the line and it simplifies the permissions model here as well. 

Would we use this for anything else that you could think of?

michael-genson:
> Would it make more sense to just use a dedicated endpoint like the favorites endpoint? That way we don't expose ourselves to any issues down the line and it simplifies the permissions model here as well.
> 
> Would we use this for anything else that you could think of?

Yeah I think a dedicated endpoint works fine. We can always refactor if another use case comes up

PR #2143, opened 2023-02-17
Feature: Add "Authentication Method" to allow existing users to sign in with LDAP

cmintey:
## What type of PR is this?

- documentation
- feature

### Overview

This PR introduces a new `auth_method` to the `users` table so that an administrator can set the method of authentication for their users. The current authentication methods are 
- AuthMethod.MEALIE (`Mealie`, *default*)
- AuthMethod.LDAP (`LDAP`)

### Features

- When a new user is created via LDAP, the `auth_method` is set to `LDAP`.
- Existing users can be updated in the admin panel to use `LDAP` as their authentication.
- Users with `LDAP` authentication method will not be able to update or reset their password as that should be handled via the LDAP provider itself

### Migration

- adds `auth_method` column to the user table
- default `auth_method` is `Mealie`
- users with a password of "LDAP" will be automatically set up with `auth_method` as `LDAP`

## Open Items
- [ ] If a user created by LDAP is changed over to use `Mealie` auth, their password will be "LDAP" (plaintext). How should we handle that?
- [x] Should users be able to be created in the admin panel with `LDAP` as auth method? Doesn't make a lot of sense since they can just log in and be created automatically.

## Screenshots

<details>
<summary>User Management Screen</summary>
User management screen
</details>

<details>
<summary>User Edit Screen</summary>
User management screen
</details>

<details>
<summary>User Creation Screen</summary>
User management screen
</details>

## Testing

- [x] Test `authenticate_user` method with the new `auth_method` set for a test user
- [x] Test create and update user actions
- [x] Test the password reset service

## Release Notes

```release-note
Adds authentication method field to allow existing users to be switched over to LDAP auth
```


hay-kot:
> If a user created by LDAP is changed over to use Mealie auth, their password will be "LDAP" (plaintext). How should we handle that?

The "correct" thing to do would be to prompt them on first login to reset their password. We don't currently have any infrastructure setup for anything that, so probably not the best. I don't have any good ideas... I'll have to think on this for a bit

> Should users be able to be created in the admin panel with LDAP as auth method? Doesn't make a lot of sense since they can just log in and be created automatically.

I would say let's not. Simplifies the implementation.

cmintey:
> > If a user created by LDAP is changed over to use Mealie auth, their password will be "LDAP" (plaintext). How should we handle that?
> 
> The "correct" thing to do would be to prompt them on first login to reset their password. We don't currently have any infrastructure setup for anything that, so probably not the best. I don't have any good ideas... I'll have to think on this for a bit

One solution (maybe the easiest) is just to have it documented that if this happens, the user will need to go through the reset password flow.

We could also include an option on the admin panel to generate a password reset link for a user. This could be useful for folks who don't have SMTP set up on their server, which makes the password reset functionality useless, currently.

I do think that long-term, it would be better to have that prompt-on-initial-login flow.

hay-kot:
> One solution (maybe the easiest) is just to have it documented that if this happens, the user will need to go through the reset password flow.

I agree. If we can document that somewhere and mark it as a temporary solution until we do proper password reset on login, that would be sufficient. 

> We could also include an option on the admin panel to generate a password reset link for a user. This could be useful for folks who don't have SMTP set up on their server, which makes the password reset functionality useless, currently.

I think that would be ideal, I would say most users don't have SMTP enabled. I think that would be a fairly easy endpoint to implement, let me know if you want to do that as apart of this PR or in the future

> I do think that long-term, it would be better to have that prompt-on-initial-login flow.

Tracking in

- https://github.com/hay-kot/mealie/issues/2166

cmintey:
> If we can document that somewhere and mark it as a temporary solution until we do proper password reset on login, that would be sufficient.

I've added the documentation in a new LDAP page that you'll need to reset your password.

> I think that would be ideal, I would say most users don't have SMTP enabled. I think that would be a fairly easy endpoint to implement, let me know if you want to do that as apart of this PR or in the future

I'll look into this soon. I think it should probably be in a different PR, that way we can get this main functionality merged in now.

cmintey:
Changed the create user screen to be more like the password field on the update user screen.

The field now shows up, but is readonly and disabled.

image

PR #2145, opened 2023-02-18
Feature: Shopping List Item Pagination Route

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- feature

_(REQUIRED)_

Adds a pagination route for shopping list items. This is particularly useful to query for only unchecked items, i.e. `checked=false` (which is generally what you want in an external app).

_(REQUIRED)_

N/A

## Testing

Pytest

## Release Notes

_(REQUIRED)_

```release-note
added API route for querying shopping list items
```

### DIFF
--- a/mealie/schema/group/group_shopping_list.py
+++ b/mealie/schema/group/group_shopping_list.py
@@ -177,6 +177,10 @@ class ShoppingListItemsCollectionOut(MealieModel):
     deleted_items: list[ShoppingListItemOut] = []
 
 
+class ShoppingListItemPagination(PaginationBase):
+    items: list[ShoppingListItemOut]
+
+
 class ShoppingListCreate(MealieModel):
     name: str | None = None
     extras: dict | None = {}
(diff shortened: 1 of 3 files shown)


mealie/schema/group/group_shopping_list.py
```diff
@@ -177,6 +177,10 @@ class ShoppingListItemsCollectionOut(MealieModel):
     deleted_items: list[ShoppingListItemOut] = []
 
 
+class ShoppingListItemPagination(PaginationBase):
```
fleshgolem, line 180: There really is no need for this, since PaginationBase is generic to begin with and this subclass doesnt add anything in particular. You can just use `PaginationBase[ShoppingListItemOut] ` instead

michael-genson: Yeah the only reason I added it is because we have them for all other `get_all` routes, so I wanted to keep it consistent with the rest of the code base

PR #2162, opened 2023-02-21
Fix: Print Preferences Menu Missing

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- bug

_(REQUIRED)_

I think a merge conflict removed the print preferences from the context menu. This restores the menu, and also fixes some console errors due to bad props.

_(REQUIRED)_

N/A

## Release Notes

_(REQUIRED)_

```release-note
NONE
```


hay-kot:
> I think a merge conflict removed the print preferences from the context menu. This restores the menu, and also fixes some console errors due to bad props.

:( My bad... Thanks for fixing my mistakes!

PR #2163, opened 2023-02-21
fix performance issues on /api/foods

fleshgolem:
# What type of PR is this?

- bug

As mentioned in discord, the /api/foods endpoint can be pretty slow if you have a larger number of foods and especially if you use the extras feature for foods. I identified two causes for this:

1) Any repository that just uses RepositoryGeneric's page_all functionality without modifications can not make use of eager-loading and if there are relations used in the response (i.e. label and extras in this case), they will all be lazy-loaded for any given entry
2) IngredientFood's getter_dict override was kinda terrible, because the line `**GetterDict(name_orm)` would cause pydantic to try and resolve ALL model relationships, even if they are not used in the response, causing even more lazy-loading

To improve on this, I added a query_options function to RepositoryGeneric, so every sub-repository can specify its own loading options, if needed. Additionally I rewrote the getter_dict override so it just does the thing it is supposed to do and not cause any more unnecessary loading

No issue filed

## Testing

Ran all the tests and tried everything manually

## Release Notes

```release-note
* Improve performance for loading foods on search page
```

### DIFF
--- a/mealie/repos/repository_foods.py
+++ b/mealie/repos/repository_foods.py
@@ -1,5 +1,7 @@
 from pydantic import UUID4
 from sqlalchemy import select
+from sqlalchemy.orm import joinedload
+from sqlalchemy.orm.interfaces import LoaderOption
 
 from mealie.db.models.recipe.ingredient import IngredientFoodModel
 from mealie.schema.recipe.recipe_ingredient import IngredientFood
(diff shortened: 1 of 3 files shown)


hay-kot:
> To improve on this, I added a query_options function to RepositoryGeneric, so every sub-repository can specify its own loading options, if needed.

Very slick. Love this!

PR #2171, opened 2023-02-24
Feature: add password reset token endpoint to the admin panel

cmintey:
## What type of PR is this?

- feature

This PR adds a password reset link in the admin panel for each user. This link will allow the administrator to generate a password reset link for a user so as not to require the use of SMTP as discussed [here](https://github.com/hay-kot/mealie/pull/2143#issuecomment-1442254353)

None

## Testing

Tested visually that the button shows up and works.
Manually tested that the endpoint could only be called with an admin JWT.

## Release Notes

```release-note
admins can now generate password reset links without setting up SMTP
```

## Screenshots
image

After url generation, the copy button shows up
image

### DIFF
--- a/frontend/pages/admin/manage/users/_id.vue
+++ b/frontend/pages/admin/manage/users/_id.vue
@@ -27,6 +27,13 @@
             label="User Group"
             :rules="[validators.required]"
           ></v-select>
+          <div class="d-flex py-2 pr-2">
+            <BaseButton type="button" :loading="generatingToken" create @click.prevent="handlePasswordReset">
+              Generate Password Reset Link
+            </BaseButton>
+            <AppButtonCopy v-if="resetUrl" :copy-text="resetUrl"></AppButtonCopy>
+          </div>
+
           <AutoForm v-model="user" :items="userForm" update-mode />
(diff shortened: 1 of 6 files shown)


fleshgolem (dismissed):

frontend/pages/admin/manage/users/_id.vue
```diff
@@ -27,6 +27,13 @@
             label="User Group"
             :rules="[validators.required]"
           ></v-select>
+          <div class="d-flex py-2 pr-2">
+            <BaseButton type="button" :loading="generatingToken" create @click.prevent="handlePasswordReset">
+              Generate Password Reset Link
```
hay-kot: Can you add this as a translatable string in the en-US.json file? Probably here

https://github.com/hay-kot/mealie/blob/refactor/global-exception-handling/frontend/lang/messages/en-US.json#L698

hay-kot (changes requested):

frontend/pages/admin/manage/users/_id.vue
cmintey: Sure can

PR #2199, opened 2023-02-27
LDAP Improvements and E2E testing

### LINKED ISSUE #2185: [v1.0.0b] - LDAP server requiring STARTTLS causes error
### First Check

My LDAP server requires STARTTLS which causes Mealie to throw an exception when trying to authenticate any user:

`ldap.CONFIDENTIALITY_REQUIRED: {'msgtype': 97, 'msgid': 1, 'result': 13, 'desc': 'Confidentiality required', 'ctrls': [], 'info': 'TLS confidentiality required'}`

Quickly looking through the code, I don't see any reference to a `.start_tls_s()` call by the LDAP client.

### Deployment

Docker (Linux)

### Deployment Details

_No response_

### LINKED ISSUE #2218: [v1.0.0b] - LDAP_QUERY_BIND and LDAP_QUERY_PASSWORD shouldn't be "required"
### First Check

My LDAP server allows for anonymous binding to search for users so I do not need to supply a username/password for the user search to work. Leaving them unspecified (which defaults to `None`) works perfectly fine. However, the admin dashboard shows a warning for LDAP configuration:

`Not all LDAP Values are configured. This can be ignored if you are not using LDAP Authentication.`

`python-ldap` [supports passing None for both username/password](https://www.python-ldap.org/en/python-ldap-3.3.0/reference/ldap.html#ldap.LDAPObject.simple_bind_s) which, I'm guessing, is just an anonymous bind under the hood. [Mealie explicitly checks if the values are `None`](https://github.com/hay-kot/mealie/blob/mealie-next/mealie/core/settings/settings.py#L149-L150) which is what is throwing the error/warning.

### Deployment

Docker (Linux)

### Deployment Details

_No response_

### LINKED ISSUE #2228: [v1.0.0b] - LDAP_USER_FILTER is confusingly named
### First Check

I'll be upfront and say this is probably 60% user error and 40% bug.

Typically, things like `LDAP_USER_FILTER` is used to filter all available LDAP users down to those eligible to log into a particular service. Some software offers to do a simple `memberOf` check against a group or let's you specify the full LDAP filter string.

The [Mealie documentation ](https://github.com/hay-kot/mealie/blob/mealie-next/docs/docs/documentation/getting-started/installation/backend-config.md) says `LDAP_USER_FILTER` is "The LDAP search filter to find users" and then provides an example LDAP string that is clobbered by markdown syntax. The `LDAP_ADMIN_FILTER` behavior matches what I described above.

What I found, and is admittedly documented in the clobbered example, is that `LDAP_USER_FILTER` is actually not a filter but rather is used to find a particular user. This is how this caused me problems:

1. I initially created a new LDAP group and put my user as the sole member.
2. I set the `LDAP_USER_FILTER` to do essentially a membership check on that group.
3. Tried out Mealie and sold my wife on trying it
4. Put her user in the new group
5. I continued to use/log into Mealie without problem
6. When she finally tried logging in, she got a password incorrect message.

What I'm nearly certain is happening is that, due to my incorrect usage of `LDAP_USER_FILTER`, the filter was returning both of our users, my user was listed first, the [code actually just picks the first item returned by the LDAP filter](https://github.com/hay-kot/mealie/blob/mealie-next/mealie/core/security/security.py#L77-L90) (which happened to be my user), I was able to continue to log in, and she was not as her credentials were attempted against my user.

### Deployment

Docker (Linux)

### Deployment Details

_No response_

cmintey:
## What type of PR is this?

- documentation
- bug/feature

- Introduces a new optional environment variable to enable STARTTLS with LDAP.
- Introduces support for anonymous bind by setting `LDAP_QUERY_BIND` and `LDAP_QUERY_PASSWORD` env variables to `None`
- Make `id_attribute` and `mail_attribute` be implied in `LDAP_USER_FILTER`. This now makes `LDAP_USER_FILTER` optional. This change maintains compatibility with previous changes.

Fixes #2185 
Fixes #2218
Fixes #2228

This PR also introduces end-to-end tests for LDAP as discussed [here](https://github.com/hay-kot/mealie/pull/2107#issuecomment-1426857429). Instead of GLAuth, I found this [awesome docker image](https://hub.docker.com/r/rroemhild/test-openldap) that comes pre-baked with users as well as support for STARTTLS (which GLAuth did not support).

## Testing

- Ensure existing tests still pass.
- Created E2E tests for ldap with a openldap docker image. **These tests will only be run in GitHub Actions.**

## Release Notes

```release-note
support for STARTTLS and anonymous bind with LDAP authentication
```


.github/workflows/partial-backend.yml
```diff
@@ -82,5 +88,17 @@ jobs:
         env:
           DB_ENGINE: ${{ matrix.Database }}
           POSTGRES_SERVER: localhost
+          LDAP_AUTH_ENABLED: True
+          LDAP_SERVER_URL: ldap://localhost:10389
```
swails, line 92:
```suggestion
          LDAP_SERVER_URL: ldap://ldap-service:10389
```

I think you can set it to the name of the service.

.github/workflows/partial-backend.yml
```diff
@@ -27,6 +27,12 @@ jobs:
         options: --health-cmd pg_isready --health-interval 10s --health-timeout 5s --health-retries 5
         ports:
           - 5432:5432
+      ldap:
```
swails, line 30:
```suggestion
      ldap-service:
```

Just suggesting that the name match my suggestion on line 92.

.github/workflows/partial-backend.yml
```diff
@@ -27,6 +27,12 @@ jobs:
...
+        image: rroemhild/test-openldap
+        ports:
+          - 10389:10389
+          - 10636:10636
```
swails, line 34:
```suggestion
```

You shouldn't have to expose any ports if you use the service name (similar to how you can do in docker-compose).

tests/integration_tests/user_tests/test_user_login.py
```diff
@@ -57,3 +58,192 @@ def test_user_lockout_after_bad_attemps(api_client: TestClient, unique_user: Tes
     user_service = UserService(database)
     user = database.users.get_one(unique_user.user_id)
     user_service.unlock_user(user)
+
+
+@pytest.mark.skipif(not os.environ.get("GITHUB_ACTIONS", False), reason="requires ldap service in github actions")
```
swails, line 63: This makes it impossible to test locally.  Is there a reason you can't set it to skip unless `LDAP_SERVER_URL` is set in the environment?

swails (commented):
Very cool!

tests/integration_tests/user_tests/test_user_login.py
cmintey: The reason it's like that and not just checking for the LDAP url variable is because the tests rely on the specific docker image that is spun up during the tests. So just setting your LDAP url, the tests would likely fail since it wouldn't be the correct LDAP url and wouldn't have the correct users the tests expect.

You can still test it locally by setting `GITHUB_ACTIONS=true` in your environment. Then you'd need to spin up the docker image and point your LDAP environment variables to that docker image. Same process as in the GitHub workflow.

.github/workflows/partial-backend.yml
cmintey: The reason I used localhost instead of the service name is mainly to keep it consistent with how the Postgres service is being used above

hay-kot:
> This PR also introduces end-to-end tests for LDAP as discussed https://github.com/hay-kot/mealie/pull/2107#issuecomment-1426857429. Instead of GLAuth, I found this [awesome docker image](https://hub.docker.com/r/rroemhild/test-openldap) that comes pre-baked with users as well as support for STARTTLS (which GLAuth did not support).

Amazing. I really really really really appreciate all the work you've been putting into the LDAP integration lately. It's super impressive!

PR #2212, opened 2023-03-06
Feature: Migrate from Copy Me That

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- feature

_(REQUIRED)_

Implements migration from [Copy Me That](https://www.copymethat.com/). While in there I also changed a few other things:
1. Made all of the text on the migration page translatable
2. Slightly refactored the migration base to use the recipe service, which does a few extra things (namely it creates the "created" timeline event)

image

_(REQUIRED)_

Closes #2202

I had to add beautiful soup to our dependencies, but we were already using it in pytest (some other package used it as a dependency already).

While testing the recipe set @BrandonG777 provided on discord (over 800 recipes) it took foreeeever to finish. We may want to consider optimizing this to use bulk methods. It might break some things though (namely the image processing) so it's more than just a method change. It also may not be worth it since this is an unusually large number of recipes, and users rarely perform this action more than once. Mostly just a pain for testing.

## Testing

Added pytest, also did some end-to-end testing using the application. Results are nice:
image

We also grab the ratings when available, and we set the last made date to today if the recipe is marked as "made":
image

We also grab notes when available:
image

## Release Notes

_(REQUIRED)_

```release-note
added migration support for Copy Me That
```


mealie/services/migrations/copymethat.py
```diff
@@ -0,0 +1,123 @@
...
+        with tempfile.TemporaryDirectory() as tmpdir:
+            with zipfile.ZipFile(self.archive) as zip_file:
+                zip_file.extractall(tmpdir)
+
+            source_dir = Path(tmpdir)
+
+            recipes_as_dicts: list[dict] = []
+            for recipes_data_file in source_dir.glob("*.html"):
+                with open(recipes_data_file, encoding="utf-8") as f:
+                    soup = BeautifulSoup(f, "html.parser")
```
fleshgolem: If speed is a problem, wouldnt it be better to use lxml as the parser here, because its usuall a lot faster?
Its an implicit part of the dependencies any way already, so it shouldnt really cost anything to add it

michael-genson: Ah good call, didn't realize it was considerably faster. I think most of the slowdown is because of the database stuff, but it certainly wouldn't hurt to switch parsers for some free speed

PR #2260, opened 2023-03-16
Remove all sqlalchemy lazy-loading from app

fleshgolem:
## What type of PR is this?

- cleanup

Well, here I am doing the thing again... I mentioned it before, but I am of the pretty firm opinion that lazy-loading is a huge performance trap and generally evil, so I tried to scour the whole app of all occurences.
My previous approach of repos defining their loader options proved to not be fine-grained enough because of all the schema overrides, so all schemas can now specify the required joins themselves... after all they should know best.
There are some cases here that will generate huge queries now, but these should still be preferrable over hundreds of automatically fired small ones. Still, it might be worthwhile to check in the future if all that related data is needed in all those cases, at least they are laid pretty bare now

None

## Testing

Made sure that all tests run and tested whole app manually

## Release Notes

```release-note
* General performance improvements
```


fleshgolem:
Well, apparently this totally tanks performance of the postgres test runner on github, so it is probably a wise idea to not merge this yet

fleshgolem:
So, I think I know what is happening here. I suppose the service container that runs the pg database in the actions gets way less CPU/RAM than the actual test runner does, which causes the pg tests to tank while the sqlite tests run totally fine
I can reproduce this behaviour in a docker container with extremely low specs. This also just makes sense, because these kinds of nested joins definitely would increase memory pressure by a bit.
Since we have to assume that some users will probably try to run this on a potato and also having to wait 20 minutes for a test result is really not acceptable, I will try to find a solution that goes a bit easier on all the joining while still not completely going back to the old ways

fleshgolem:
Alright, i replaced quite a bunch of joinedloads with selectinload to find a better balance between number of queries generated and complexity of them and everything seems to run fine now, so i think this should be safe to merge

PR #2288, opened 2023-03-24
[Fix] Filter out the LDAP entries which do not have a DN

### LINKED ISSUE #2279: [v1.0.0b] - LDAP Not functional with AD on current nightly (3/23/2023)
### First Check

As of latest nightly with changes to the LDAP_USER_FILTER I am seeing the new error that there are multiple users coming back.

```
ERROR: 23-Mar-23 15:33:54 [LDAP] Multiple users found with the provided user filter
```

When using the same filters via `Get-ADUser -LDAPFilter 'expression'` in powershell I am only sing a single result.

I would expect that if there where multiple responses it would at least take the first response in the set and attempt a bind to that.

It is also unclear if this was a breaking change to how this field worked before merging https://github.com/hay-kot/mealie/pull/2199 based on the discussion here https://github.com/hay-kot/mealie/issues/2228

### Deployment

Other

### Deployment Details

Kubernetes using containerd on linux.

Application works flawlessly outside of raised issue and worked for a few weeks before pulling latest nightly.

cmintey:
## What type of PR is this?

- fix

- Filter the entries returned from the LDAP search to those which have a non-null DN. This addresses the issues faced with Active Directory returning forest results
- Sets `OPT_REFERRALS` on the connection object instead of the global object as per [this post](https://stackoverflow.com/questions/140439/authenticating-against-active-directory-using-python-ldap)
- Add some debug logging to help with, well, debugging

fixes #2279

## Release Notes

```release-note
NONE
```

### DIFF
--- a/mealie/core/security/security.py
+++ b/mealie/core/security/security.py
@@ -59,8 +59,8 @@ def user_from_ldap(db: AllRepositories, username: str, password: str) -> Private
     if settings.LDAP_TLS_INSECURE:
         ldap.set_option(ldap.OPT_X_TLS_REQUIRE_CERT, ldap.OPT_X_TLS_NEVER)
 
-    ldap.set_option(ldap.OPT_PROTOCOL_VERSION, 3)
     conn = ldap.initialize(settings.LDAP_SERVER_URL)
+    conn.set_option(ldap.OPT_PROTOCOL_VERSION, 3)
     conn.set_option(ldap.OPT_REFERRALS, 0)
 
     if settings.LDAP_TLS_CACERTFILE:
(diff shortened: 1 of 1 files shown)


mealie/core/security/security.py
```diff
@@ -111,16 +112,18 @@ def user_from_ldap(db: AllRepositories, username: str, password: str) -> Private
         return False
 
     if len(user_entry) > 1:
-        conn.unbind_s()
-        logger.error("[LDAP] Multiple users found with the provided user filter")
-        return False
+        logger.warning("[LDAP] Multiple users found with the provided user filter. Using the first one")
+        logger.debug(f"[LDAP] The following users were returned: {user_entry}")
 
     user_dn, user_attr = user_entry[0]
```
LargoUsagi, line 123: Intead of just using the first result every time, I would suggest something along these lines.

```python
    if len(user_entry) > 1:
        logger.warning("[LDAP] Multiple users found with the provided user filter")

        has_id_att = False
        for entry in user_entry:
            dn, attr = entry
            if LDAP_ID_ATTRIBUTE in attr:
                has_id_att = True
                user_dn, user_attr = user_entry[0]
                logger.debug(f"[LDAP] The following users has id attribute: {entry}")
                break

        if not has_id_att:
            logger.error("[LDAP] Multiple users found with the provided attribute id")
            return False
    else:
        user_dn, user_attr = user_entry[0]
```

This should add a level of safety around MS changing the order of the responses in future versions of AD.

LargoUsagi (dismissed):
Overall this will work and looks good, the code snippet I supplied could be tested and improve the solution.

mealie/core/security/security.py
cmintey: I see what you're saying. Another alternative I present, we can just filter out the `user_entry` by `user_dn not None`. Because ultimately, I need the `user_dn` for the bind. So something like this could do the trick

```py
    # we only want the entries that have a dn
    user_entry = [(dn, attr) for dn, attr in user_entry if dn]

    if len(user_entry) > 1:
        logger.warning("[LDAP] Multiple users found with the provided user filter")
        logger.debug(f"[LDAP] The following entries were returned: {user_entry}")
        conn.unbind_s()
        return False;
```

LargoUsagi: That is a bit more eloquent and would filter out the AD forests as they do not have a user_dn and then grabbing the first result after filtering would suffice safely.

LargoUsagi:
Does the pipeline automatically push a new container image after each PR is merged?

If so I will test this once its accepted into the main branch.

michael-genson:
> Does the pipeline automatically push a new container image after each PR is merged?

Yes, new PRs (after being merged) will trigger a new nightly build automatically. It usually takes a few hours after merge (if you're in the Mealie Discord there is a notification in the "nightly" channel)

LargoUsagi (approved):
LGTM and resolves the feedback from the earlier commits.

LargoUsagi (approved):
Tested changes against my AD server and new code works as expected.

PR #2357, opened 2023-04-23
Fix for broken frontend development

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- bug

_(REQUIRED)_

_(REQUIRED)_

Discussed in Discord. Not my fix, thanks @fleshgolem!

## Testing

I played around in the frontend and everything seems to be working right. I didn't update nuxt-vite, only the postcss and ufo packages, and it seems to be working fine.

## Release Notes

_(REQUIRED)_

```release-note
NONE
```

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -54,10 +54,10 @@
     "vitest": "^0.29.0"
   },
   "resolutions": {
+    "@nuxtjs/i18n/**/ufo": "0.7.9",
     "vue-template-compiler": "2.7.14",
     "vue-demi": "^0.13.11",
     "postcss-preset-env": "^7.0.0",
-    "typescript": "^4.9.5",
-    "ufo": "0.7.9"
+    "typescript": "^4.9.5"
   }
 }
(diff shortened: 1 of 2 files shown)


fleshgolem:
I think this should really use the form where only the sub-dependency of that specific dependency is pinned, otherwise we might run into serious issues down the line, because e.g. some other packages that depend on ufo > 1.0 (there are some) will stop working correctly

This would be `"@nuxtjs/i18n/**/ufo": "0.7.9"` for ufo and it should probably be doable as well for postcss-preset-env as well, but i dont know from the top of my head anymore what required what here. I can look it up in the evening if you want though

michael-genson:
> I think this should really use the form where only the sub-dependency of that specific dependency is pinned

TIL that's a thing. I'm primarily a Python dev, lol. I'll try pinning it for just nuxtjs, and if that doesn't work figure out what else needs it. Thanks!

michael-genson:
Pinning to just nuxtjs/i18n seems to still do the trick, without pinning it for postcss. I definitely encourage grabbing it yourself and confirming though

PR #2369, opened 2023-05-01
fix: navigating back after importing a recipe by file reimports

### LINKED ISSUE #2267: [v1.0.0b] - back button restarts import
### First Check

when you import a url and then use the back button (whether phone or desktop, and whether "stay in edit mode" is checked or not). 
the import will be started again. 

IMHO the import of the same url should not be disabled in general. 

### Deployment

Docker (Linux)

### Deployment Details

_No response_

jenscalaerts:
## What type of PR is this?
- bug

Removes automatic import when navigating back to the previous page after importing a recipe by url

Fixes #2267 

## Testing
- Open the application and navigate to /recipe/create/url
- import a recipe by url
- press the back button
- the recipe should not be imported again 

then do the same with stay in edit mode enabled

## Release Notes

_(REQUIRED)_

```release-note
Fixed a bug where recipes were imported again when pressing the back button.
```

### DIFF
--- a/frontend/pages/recipe/create/url.vue
+++ b/frontend/pages/recipe/create/url.vue
@@ -69,8 +69,7 @@ import {
   ref,
   useRouter,
   computed,
-  useRoute,
-  onMounted,
+  useRoute
 } from "@nuxtjs/composition-api";
 import { AxiosResponse } from "axios";
 import { useUserApi } from "~/composables/api";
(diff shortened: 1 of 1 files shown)


hay-kot:
Doesn't this break bookmarklets?

jenscalaerts:
Looking at the nightly documentations, I can't find a reference to how the bookmarklet should work in nightly .
I assume, the bookmarklet redirects to  /recipe/create/url?recipe_import_url=https://my.url.com

That being so, it looks like it will break this to a point. As it is, it will navigate to the url import page and not submit. 

I will update the pr adding a property to the which specifically represents the textfield state. This way it should be possible to differentiate between a bookmarklet redirect and the user navigating back.

hay-kot:
> Looking at the nightly documentations, I can't find a reference to how the bookmarklet should work in nightly .
I assume, the bookmarklet redirects to /recipe/create/url?recipe_import_url=https://my.url.com/

That's exactly right. The bookmarklet grabs the URL from the current page and sends it to Mealie through the query parameter

PR #2400, opened 2023-05-20
Feature: LDAP attribute validation

cmintey:
## What type of PR is this?

- cleanup
- feature

This PR adds validation to the user attributes we get from the LDAP server. Previously if an attribute was missing, we could catch the error and just silently return. Now we'll log that error and tell the user which attribute is missing so that it's easier to debug.

During this PR I also realized that we are updating the user admin flag every time an LDAP user logs in, even if there was no change. So now we only update the admin flag if it actually changes.

And finally, the LDAP related code was getting to be a lot, so it's now been moved into it's own file for better maintainability.

None

## Testing

Ran all tests to make sure they still work

## Release Notes

```release-note
better logging for LDAP user attribute validations
```

### DIFF
--- a/mealie/core/security/ldap.py
+++ b/mealie/core/security/ldap.py
@@ -10,7 +10,7 @@ from mealie.schema.user.user import PrivateUser
 logger = root_logger.get_logger("security.ldap")
 
 
-def search_user(conn: LDAPObject, username: str):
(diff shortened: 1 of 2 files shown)


cmintey:
bump

hay-kot:
Love this, thanks so much!

PR #2416, opened 2023-06-08
IP logging fix

### LINKED ISSUE #2407: [Bug] Docker Caddy IP Forwarding Headers
Continuation of the original issue detailed in this PR: #2365 

When attempting to use a fail2ban regex filter on the "Incorrect username or password from *" log, the log always shows the internal IP of the front-end docker container.

From @hay-kot
This is likely due to a mis-configuration on the internal proxy that we use here https://github.com/hay-kot/mealie/blob/mealie-next/docker/frontend.Caddyfile
Probably just need to adjust it to forward the headers from the original request.

From @jecorn 
...But the caddy docs suggest that it should be easy to do add the relevant IP forwarding....

Issue created at the request of @jecorn

jecorn:
## What type of PR is this?

- bug

IP addresses were logged in mealie.log, but the front and backend weren't correctly set up to get the real IP address through the docker gateway.

Fixes #2407
Fixes a poetry breaking 404 error in the dev container that was already fixed in the production containers https://github.com/python-poetry/poetry/issues/6676
Fixes a bash bug in the docker entry scripts that prevented gunicorn from ever running, sets WEB_UNICORN in the front/backend and omni docker-compose files to false (per docs).

`auth.py` checks for the existence of an x-forwarded-for header because caddy and gunicorn/uvicorn handle IP address forwarding differently. gunicorn/uvicorn reset the apparent client IP based on what they find in a trusted x-forwarding-for header and then don't seem to pass along x-forwarded-for. Caddy passes along x-forwarded-for without resetting the client IP. Also, it seemed safer to take a trusted x-forwarded-for header if it exists and fall back to the apparent IP if not.

gunicorn/uvucorn are now set up to only trust IP forwarding headers from the gateway in the default route. Which is the docker host if inside a container. This setting could be opened up more if it ends up being too restrictive. But the option allows only individual IPs or completely open.

The x-forwarded-for header needs to be set by the outer-most proxy. **Both** Caddy and nuxt are set up to proxy, with Caddy outward-facing for the split front/backend image and nuxt outward-facing for the omni image (no Caddy in omni). So x-forward headers need to be set by both Caddy and nuxt, then auth.py takes the canonically true client per the x-forward-for standard.

All of the above only works on linux, since Mac/Windows docker don't correctly pass real IPs (see message thread on Discord)

## Testing

Passes all automated tests.
I built production docker images (both separated front/backend and omni) then manually tested finding the correct IP both through frontend authentication and backend authentication when coming in to these containers from a different IP. 

## Release Notes

```
Correctly forward external IP into docker container for logging
Respect gunicorn vs unicorn setting in docker compose files
```

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -26,7 +26,7 @@ ENV PYTHONUNBUFFERED=1 \
 # prepend poetry and venv to path
 ENV PATH="$POETRY_HOME/bin:$PATH"
 
-RUN curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/install-poetry.py | python -
+RUN curl -sSL https://install.python-poetry.org | python3 -
 # RUN poetry config virtualenvs.create false
 
 RUN apt-get update \
--- a/docker/api.Dockerfile
+++ b/docker/api.Dockerfile
@@ -71,6 +71,7 @@ ENV GIT_COMMIT_HASH=$COMMIT
 RUN apt-get update \
     && apt-get install --no-install-recommends -y \
     gosu \
+    iproute2 \
     tesseract-ocr-all \
     libldap-2.4-2 \
     && apt-get autoremove \
--- a/docker/api.entry.sh
+++ b/docker/api.entry.sh
@@ -51,7 +51,7 @@ GUNICORN_PORT=${API_PORT:-9000}
 
 # Start API
 hostip=`/sbin/ip route|awk '/default/ { print $3 }'`
-if [ "$WEB_GUNICORN" == 'true' ]; then
+if [ "$WEB_GUNICORN" = 'true' ]; then
     echo "Starting Gunicorn"
     gunicorn mealie.app:app -b 0.0.0.0:$GUNICORN_PORT --forwarded-allow-ips=$hostip -k uvicorn.workers.UvicornWorker -c /app/gunicorn_conf.py --preload
 else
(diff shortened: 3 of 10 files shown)


jnortell (approved):

jnortell:
Anyway I  can implement this in my setup manually while we wait for code review?

jnortell (approved):

jecorn:
I wouldn’t recommend trying to mix/match a bare metal instance pulled from GitHub with something you have already working through docker. 

@hay-kot I know you’re super busy moving house these days. But is there anything else you need from me for this PR?

jnortell:
Anyway to get this pushed to the `[api-nightly]` docker image?

jecorn:
> Anyway to get this pushed to the `[api-nightly]` docker image?

Images are built automatically every day. Should be available tomorrow. Let me know how it works for you. I’m happy to revisit if something seems off.

jnortell:
> > Anyway to get this pushed to the `[api-nightly]` docker image?
> 
> Images are built automatically every day. Should be available tomorrow. Let me know how it works for you. I’m happy to revisit if something seems off.

Awesome! Thank you so much for your continued support! I will report back once the image updates.

jnortell:
> > Anyway to get this pushed to the `[api-nightly]` docker image?
> 
> Images are built automatically every day. Should be available tomorrow. Let me know how it works for you. I’m happy to revisit if something seems off.

Still no update 😢
The most recent push was the frontend-nightly image two days ago which I do not believe contains this PR because the IP logging still has the false behavior.

hay-kot:
@jnortell 

There was a problem with the build, fixed with #2429 

See here for the latest builds progress https://github.com/hay-kot/mealie/actions/runs/5396228883 or monitor the discord release feed for when the image gets published.

PR #2429, opened 2023-06-26
Change libldap-2.4-2 to libldap-common in docker

cmintey:
## What type of PR is this?

- bug

`libldap-2.4-2` package is no longer able to resolve causing docker builds to fail. Using the suggested `libldap-common`, the build succeeds.

N/A

## Testing

Tested the change manually that ldap login was still working.

## Release Notes

```release-note
NONE
```


hay-kot:
The hero we don't deserve ❤️

PR #2438, opened 2023-07-03
Feature: Migrate from Tandoor

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- cleanup
- feature

_(REQUIRED)_

Adds support for Tandoor migrations (see https://github.com/hay-kot/mealie/issues/2427). Also refactors the display property implemented in https://github.com/hay-kot/mealie/pull/2021 to apply to the `RecipeIngredient` class.

_(REQUIRED)_

N/A - discussed in https://github.com/hay-kot/mealie/issues/2427. The actual issue in that discussion is fixed by https://github.com/hay-kot/mealie/pull/2434

Since Tandoor provides structured data similar to Mealie (food, unit, qty) I needed to reuse the display property implemented in https://github.com/hay-kot/mealie/pull/2021. To make it easier to reuse I refactored all ingredient-like objects to use a new `RecipeIngredientBase` class which contains the display parsing logic.

To do this, I had to add the `disable_amount` and `is_food`  to the base, which are effectively the same thing (but ingredients use `disable_amount` and shopping lists use `is_food`). There's probably room to simplify this, but it would involve removing one of the fields which would require refactoring either recipes or shopping lists, which is outside the scope of this PR (honestly this refactor is too, but that would be _even more_ out of scope...). I also noticed that the recipe's `disable_amount` setting doesn't propagate down to its ingredients, so I overrode the Recipe `__init__` and `from_orm` to populate it correctly based on the recipe's settings (and shopping list items use `is_food` so no changes were needed there):
```python
...
@classmethod
def from_orm(cls, obj):
    recipe = super().from_orm(obj)
    recipe.__post_init__()
    return recipe

def __init__(self, **kwargs) -> None:
    super().__init__(**kwargs)
    self.__post_init__()

def __post_init__(self) -> None:
    # the ingredient disable_amount property is unreliable,
    # so we set it here and recalculate the display property
    disable_amount = self.settings.disable_amount if self.settings else True
    for ingredient in self.recipe_ingredient:
        ingredient.disable_amount = disable_amount
        ingredient.is_food = not ingredient.disable_amount
        ingredient.display = ingredient._format_display()
```

## Testing

Imported the sample 500 or so recipes from https://github.com/hay-kot/mealie/issues/2427 and they look pretty good (see some example screenshots there). Manually checked a dozen or so, including some edge cases (e.g. missing data). 100% of those recipes imported with no errors. I also added sample data to the migration tests.

## Release Notes

_(REQUIRED)_

```release-note
added support for Tandoor migrations
```


hay-kot:
So. Good.

PR #2444, opened 2023-07-07
Feature: Recipe Timeline Images

### LINKED ISSUE #2450: [v1.0.0b] - Editing recipe timeline events changes the date to current date
### First Check

Editing recipe timeline events (whether created manually or via the Meal Planner) result in the event date being changed to the current date.

1. Manual event created on recipe with date 2023-07-01 
   image
2. Edit event
   image
3. Event date changed to edit date (today, 2023-07-09)
   image

### Deployment

Docker (Linux)

### Deployment Details

Version: omni-nightly
Build: 4af9eec89dd0b309ebea752d715add3fe0980b3d

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- bug
- feature

_(REQUIRED)_

This adds the ability to upload pictures along with your timeline events. The backend routes have been added, and the frontend uses an updated "I Made This" form to include an image. The timeline has also been updated to fetch the images when given. 

This fixes a few bugs too, both on the frontend and the backend. One issue I ran into in production is that when you create a new event (click "I Made This") the new event doesn't show up until you refresh the page. I think we used to refresh the timeline every time it opens, then we disabled that (I think I did that because we weren't paginating yet, so everything loaded all the time). But, now that we're paginating events, I think it's safe to restore the original functionality. It's a little slower, but it's much nicer for the "I Made This" -> _Look at timeline_ flow which I know I do a lot.

The new form:
image

Mobile:
image

The new timeline items (cat optional):
image

Mobile:
image

With the recipe on the global timeline:
image
image

The images are auto-sized with a max height to avoid any obnoxious rendering (note the sideways picture is just the tall picture rotated in paint.net, Mealie didn't do that):
image
image

_(REQUIRED)_

Fixes https://github.com/hay-kot/mealie/issues/2450 

There are a couple of things that definitely need to be checked out before merging this PR:

The backend route/service for uploading images: I mirrored the "update recipe image" route more or less exactly, save for the few necessary changes. I created a subdirectory in the recipe's image directory called "timeline" and each timeline event has a sub directory (similar to how recipes each have their own directory by id).

I used the recipe's minifying service to generate three images for the event image (original, small, tiny), but I don't know if this is necessary. Should I eliminate the extra images and just use the original? Or, if keeping all three, should I build additional support to use the smaller images in certain scenarios? Any guidance here is welcome.

~~Side note - I just realized if timeline events are deleted, their images aren't. I'll fix that in this PR and cross this out.~~ fixed

---

While testing the image upload, often I would click the submit button, and when the frontend sends the image to the backend, it just... hangs... forever? Like the function never finishes in the frontend, and if I close the backend it says "waiting for connections to close" until I kill it, and Chrome says the request isn't completed.

I _think_ this is a development environment issue with uvicorn or something, because I can reproduce with recipe image uploads, albeit far less consistently (try changing the recipe image 20 times, and you'll probably see it, unless it's an issue with my machine), but I can't reproduce in my production instance. Can you see if I'm doing anything wrong with the image upload or connection handling? I don't think I'm doing anything particularly weird, but it happens _very often_ with this upload feature.

Maybe it's just an issue with my machine? I have no idea. It's very inconsistent (sometimes requests work just fine, some hang forever). If you can successfully create 5 timeline events with images in a row, it's something messed up on my end for sure.

## Testing

Pytest, manual frontend testing

## Release Notes

_(REQUIRED)_

```release-note
added full image support for timeline events
added image uploads to the I Made This dialog
```


michael-genson:
I think that test is flaky, I've seen it fail before. I didn't touch anything on the recipe search. If you run the tests again they will probably pass.

PR #2470, opened 2023-07-28
Prevent All Changes To Default User In Demo Mode

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- feature

_(REQUIRED)_

If the `IS_DEMO` env var is set, all changes to the default user don't go through. The user service still returns the "updated" user, they just aren't actually updated.

_(REQUIRED)_

Probably 50% of discord messages.

## Testing

Tested manually using `export IS_DEMO=true && make backend`

## Release Notes

_(REQUIRED)_

```release-note
NONE
```

### DIFF
--- a/mealie/repos/repository_users.py
+++ b/mealie/repos/repository_users.py
@@ -5,15 +5,24 @@ from pydantic import UUID4
 from sqlalchemy import select
 
 from mealie.assets import users as users_assets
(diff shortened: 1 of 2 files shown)


hay-kot:
Not what am I supposed to do on discord???

This is amazing thank you.

PR #2351, opened 2023-04-18
Reworking search: tokenization, handling of quoted literal search, and postgres fuzziness

jecorn:
## What type of PR is this?

- feature
- bug (one tiny one)

## What this PR does / why we need it:

The current search is based on exact phrases entered into the search box. This behaviour is similar to quoting a search string in a search engine (though diacritics are removed and matches are case insensitive). For example, searching for `beans pinto` does not find `pinto beans`, `egg over easy` does not find  and `eggs over easy`, etc.. This is convenient to code, but leads to some unexpected behaviour for users where "easy" search hits are missed.

This PR adds four improvements to search:
1. Search is by default now tokenized based on whitespace, so that search strings are separated and hits can be found from matches independent of intervening words and word order. For example, `pinto beet burgers` matches both `pinto burgers` and `beet pinto burgers`
2. Special characters except quotes (' and ") are removed from the search string and replaced by whitespace, which is then used for tokenization. So `slow-cooker carrots` is tokenized for search to `slow` and `cooker` and `carrots`
3. Tokenized search (including special character removal) can be overridden using quotes (either ' or "). Quoted searches are searched for literally and can be mixed and matched with non-quoted searches. For example, `"slow-cooker" beans casserole 'with paprika'` is tokenized to a search for `slow-cooker`, `beans`, `casserole` and `with paprika`.
4. On a postgres backend, fuzzy searching is now default using word similarity trigrams with GIN indexing of the `recipe.name_normalized`, `recipe.description_normalized`, `recipe_ingredient.note_normalized` and `recipe_ingredient.original_text_normalized` columns. Trigram searching avoids the need to define a language for the database, as would otherwise be needed for full-text search stemming and stop word removal. Fuzziness can therefore work on mixed-language databases. The fuzziness is calibrated to try to return hits if a search term has 1 or two mismatches (depending on word length) while keeping false positives low. Fuzzy search orders hits by trigram similarity to the recipe name. Fuzzy search is incompatible with quoted literal search, so adding quoted strings to search automatically falls back to token search.

## Which issue(s) this PR fixes:

(fixes an un-numbered bug in RecipeIngredientModel.note_normalized triggers)
Fixes #2325
Addresses #2335 

## Special notes for your reviewer:

[...]

PR #2472, opened 2023-07-29
Feature: Generalize Search to Other Models

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- feature

_(REQUIRED)_

This generalizes the recipe search from #2351 to work for any model. After these changes, adding search support to a model is as easy as:
1. adding a list of strings of searchable fieldnames to the searchable model
2. adding the `search` param into the model's `get_all` route

_(REQUIRED)_

N/A

This PR does _not_ add normalized fields to the newly-supported models, so diacritics and symbols will decrease search accuracy. In the future I want to re-work how the recipe parser finds food/unit matches (using this search API) and that PR should add normalization. I figure it's okay to not normalize in this PR since often it's a less common issue, and fuzzy matching will work around it anyway (if using postgres). We'll probably also want to index those fields (similar to recipes).

If you think this PR _should_ normalize search fields, I'll happily add them, I just didn't want this PR to become too much of a headache (since normalized fields will add migrations) for little benefit (the frontend isn't using this API yet).

## Testing

I refactored the recipe search tests to be parametrized (instead of one large test) and added a generic set of tests for units (since recipes use a more complex filter, including ingredients). They probably don't _need_ to both be there, but they're good tests, so I think it's fine.

I also tweaked the tests to allow extra results (e.g. if only one recipe is expected to match, and two match, that's fine, as long as the first one is the one we wanted) to avoid those annoying test failures we've seen in other PRs. I also sandboxed them using their own user/group to avoid other tests leaking in.

## Release Notes

_(REQUIRED)_
```release-note
added APIs to search for foods, labels, units, categories, tags, and tools
```


jecorn:
This is awesome! And the test refactor is super appreciated. Trigram matching can be very slow on columns without an GIN index, so like you said it’s probably a good idea to index anything that will be searched.

michael-genson:
Thanks! I pretty much just copy pasted what you worked on and reorganized things a bit. And yeah, my thought process here is to add a quick win for expanding the search, then we can normalize/index anything we use on the frontend.

Unless we want to do that upfront now, I just wasn't sure

hay-kot (approved):

PR #2482, opened 2023-08-06
Feature: Improve Public URL Readability

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- feature

_(REQUIRED)_

The current public URLs are kinda gross:
`https://demo.mealie.io/explore/recipes/49279644-6d42-4f5f-80c1-c2e07e599974/easy-30-minute-homemade-chicken-tortilla-soup`

So this PR makes them look like this:
`https://demo.mealie.io/explore/recipes/home/easy-30-minute-homemade-chicken-tortilla-soup`

This is done by adding a slug to groups (which appears in the URL). Similar to the recipe id/slug implementation in https://github.com/mealie-recipes/mealie/pull/2265, you can pass either an id or a slug, so old public URLs still work, we just generate URLs with slugs now.

_(REQUIRED)_

N/A

I originally intended to use the group name instead of adding a slug, but unfortunately while group names are unique, they are case sensitive in most databases (including both sqlite and postgres, by default), so "Home" and "home" are both independently valid group names. This is unfortunate because, while URLs definitely _can_ be case-sensitive, [it's easy to run into issues with that](https://stackoverflow.com/questions/50298340/chrome-changes-url-case).

Plus, subjectively, I like how it looks better:
`https://demo.mealie.io/explore/recipes/home/easy-30-minute-homemade-chicken-tortilla-soup`
`https://demo.mealie.io/explore/recipes/Home/easy-30-minute-homemade-chicken-tortilla-soup`

## Testing

Updated pytest and also manually tested the new public route as well as the old public route (both work).

## Release Notes

_(REQUIRED)_

```release-note
made public URLs prettier
```


michael-genson:
Hopefully https://github.com/mealie-recipes/mealie/pull/2472 fixes the flaky search test

hay-kot:
Amazing work!

PR #2501, opened 2023-08-14
Feature: Display Shopping List Item Recipe Refs

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- bug
- feature

_(REQUIRED)_

This adds shopping list recipe references to the frontend. This allows users to see:
1. If an item comes from a recipe
2. If an item comes from multiple recipes, how much came from each
3. If an item is prepared multiple ways (i.e. has different notes) which recipe calls for which way

By default, when you open your shopping list, nothing looks different except for the new pot icon:
image

(note that items with no recipes don't have that icon)
When you click the pot icon, it shows you the item's recipes (note the tooltip on hover, too):
image

The card that shows the recipe is a modified version of the list's recipe references (with new props to configure), so it functions the same (e.g. clicking on it brings you to the recipe). A few things to notice:

If an item only has one recipe reference, the "extra" info is redundant, so it isn't displayed:
image

Only when there is more than one ref is it displayed:
image

The refs wrap on smaller screens:
image

_(REQUIRED)_

This PR also fixes a bug most people would never see - there was a bug in the note merge logic if you had 3 or more recipes that shared a single item, and at least two of them shared a note, _unless_ the first n-1 items were the ones that shared the note. I never noticed in testing a while back, but it happened a lot while testing this.

## Testing

Manually :)

## Release Notes

_(REQUIRED)_

```release-note
added recipe references to individual shopping list items
```


hay-kot:
Need to re-do alembic migration I think

michael-genson:
Should be good now!

hay-kot (dismissed):

hay-kot (approved):

hay-kot (approved):

PR #2502, opened 2023-08-15
feat: improve readability of ingredients list

hugo-vrijswijk:
## What type of PR is this?

- feature

Makes the ingredients list more readable by making the quantity more distinct with a highlight colour and the notes slightly smaller and opaque. This creates a better visual separation between each part of the ingredient.

image

None

## Testing

Added some unit tests for the parseIngredientText, including for some existing functionality. 

## Release Notes

```release-note
Improve readability of ingredients list
```


hugo-vrijswijk:
@hay-kot  could you give this a review? 🙏

Imrhien:
That's a nice way of doing it. I'd also suggest making the Unit the alternative colour as well.

This change should also affect the shopping list.

hugo-vrijswijk:
> That's a nice way of doing it. I'd also suggest making the Unit the alternative colour as well.
> 
> This change should also affect the shopping list.

I've made the unit the accent colour now as well.

The shopping list also uses this component:

image

michael-genson:
I almost think the accent color should be inverted. I've seen it that way on a few recipe sites where they bold the food and I think it looks nice:

https://www.savorytooth.com/instant-pot-fish-stew/

One other consideration - if there is no food, is the note still dimmed? I'm on my phone so it's hard to read through the logic in the PR otherwise I'd look myself. I have a few recipes where I'll omit the food and just put the food in the note (if, for instance, the food is super obscure)

Love this though, I think it makes the ingredients pop a lot more. Much easier to read

hay-kot (changes requested):
Code looks good. 

I'd prefer if we dropped the blue coloring of the text and instead just bold text or something a little more subtle. 

Along with what others have said, I'm assuming this all works in light mode as well?

hugo-vrijswijk:
I agree having the food bold does look clear, if perhaps slightly less visually appealing (to me). I've pushed that change:

image


>  if there is no food, is the note still dimmed?

If there's no food, the note will act as the 'main' part, like the food does if it is present. E.g. here is an ingredient with quantity 1 and a note (and nothing else):

image

I agree, the code isn't super readable to read through, especially on the web. I didn't want to refactor the parse logic too much as there weren't any tests, and I didn't want to break stuff I didn't understand either 😅.

> I'm assuming this all works in light mode as well?

Yes, especially now with it only being bold, it works the same:

image

hay-kot (approved):

hugo-vrijswijk:
@hay-kot I think you need to approve the workflow again as I merged the target branch

PR #2507, opened 2023-08-16
Feature: Improved recipeYield Parsing For Fractions and Decimals

### LINKED ISSUE #2506: [BUG] - Recipe servings only uses first numbers for scale. Ignores decimals/fractions/etc
### First Check

The Regex looking at serving size for scaling up/down only looks at the firs number(s).
It ignores decimals (4.5 will double to 8.5).
When using less than 1 serving (ie, something that only makes 0.5 cups), the 0 is used, so doubling the recipe will result in it still being 0.5.
This also messes things up when a recipe that starts out as (for example) 4 servings. By manually setting the scale to 1.2, it correctly updates to 4.8.
However then the regex limitation kicks back in, and doubling that makes it 8.8 servings.
(And weirdly enough halving that makes it 4.800000000000001. So... It does do something with the decimals)

### Steps to Reproduce

When using less than 1 serving (ie, something that only makes 0.5 cups), the 0 is used, so doubling the recipe will result in it still being 0.5.
This also messes things up when a recipe that starts out as (for example) 4 servings. By manually setting the scale to 1.2, it correctly updates to 4.8.
However then the regex limitation kicks back in, and doubling that makes it 8.8 servings.
(And weirdly enough halving that makes it 4.800000000000001. So... It does do something with the decimals)

N/A

### Mealie Version

omni-nightly

### Deployment

Docker (Linux)

_No response_

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- feature

_(REQUIRED)_

Currently, when trying to parse a recipe's yield, we only account for integers (e.g. "3 servings") but not fractions or decimals. This leads to behavior that seems really weird if you don't understand the implementation:
2023-08-15_20h51_13
2023-08-15_20h49_37
2023-08-15_20h50_28

While these might not look super weird, they're all incorrect (the serving counter is incremented by 1, but in reality the recipe ingredient amounts were doubled).

This PR adds a ton of parsing logic using magic regex to account for mixed fractions, regular fractions, and decimals (and, of course, whole numbers still work):
2023-08-15_20h53_15
2023-08-15_20h53_55
2023-08-15_20h54_23

In addition, I added some rounding logic to avoid weird floating point precision errors. Decimals are rounded to 3 decimal places, and fractions are limited the same way ingredients are limited:
image
image

image
image

_(REQUIRED)_

Fixes https://github.com/mealie-recipes/mealie/issues/2506

Parsing strings sucks.

## Testing

The above gifs are pretty much my tests.

## Release Notes

_(REQUIRED)_

```release-note
added support for fractional and decimal recipe yield scaling
```


JvdMaat:
The only thing I could mention (and that's me being very pedantic) is that in the ingredients list if I enter 0.66, it updates it to 2/3, whereas your implementation allows the fraction to stand in decimal format.

michael-genson:
> in the ingredients list if I enter 0.66, it updates it to 2/3

That's only true if you marked your unit as a fraction. If you uncheck the fraction option on the unit, it stays as a decimal (I don't know _why_ you'd do this, but hey, I'm sure plenty of people prefer it that way).

Since we don't have a similar checkbox for recipe yield, I figured we could just infer based on how you inputted it. Since it preserves whatever the original format is, I figure it's fine.

JvdMaat:
That actually makes a lot of sense.

hay-kot (changes requested):
I think it's probably worth pulling this scaling regex/function into a composable and adding some unit tests

Would you be up for adding those?

michael-genson:
Let me know if I didn't do this right, first time with the composable and tests.
Good news - the tests exposed some bugs with the regex

hay-kot (approved):

PR #2423, opened 2023-06-17
UI/UX improvements

p0lycarpio:
## What type of PR is this?

_(REQUIRED)_

- bug
- cleanup

## What this PR does / why we need it:

This PR aims to improve the user interface for smaller screens (tablet and smartphone) but also for readability and comfort. Several aspects are fixed and modified in the commits.

## Which issue(s) this PR fixes:

- 5c0c3bb9d081684630fc7e054571cebb1c954ede Fixes #2348

## Special notes for your reviewer:

- 65046c1e4d3c9371e223d7a9b808da8baeefd7ff Improves login page bottom links readibility on small screens. Disable _onhover_ (three dots buttons) on on small screens for usability.
- 8d54c5c7c2c7ab93c24b0ba34769af72a833731e Adds the untranslated string "Keep Screen Awake" on recipe page.
- db3c8f4610b4c171734878e30fbf3de18d4b2d1c Improves the Timeline page by "flattening" the cards for simplicity

## Release Notes

_(REQUIRED)_

```release-note
Fix context menus closing on tablets/smartphones
Improve the accessibility and readability of the login page and timeline
Moving an untranslated string from the recipes page
```

PR #2519, opened 2023-08-23
Fix/recipe timeline UI glitches

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- bug

_(REQUIRED)_

https://github.com/mealie-recipes/mealie/pull/2423 broke the recipe timeline, it won't open anymore. I'm guessing it used to work fine, but because of some combination of merge changes it broke. Some of the merges also caused info to be hidden (e.g. the event image and the message text).

In addition, a few minor graphical issues appeared on dark mode, so I tweaked those a bit.

_(REQUIRED)_

N/A

Isn't merging fun?

## Testing

Manual testing

## Release Notes

_(REQUIRED)_

```release-note
NONE
```

### DIFF
--- a/frontend/components/Domain/Recipe/RecipeCardMobile.vue
+++ b/frontend/components/Domain/Recipe/RecipeCardMobile.vue
@@ -17,7 +17,7 @@
           :image-version="image"
         />
       </v-img>
-      <v-list-item three-line class="px-0">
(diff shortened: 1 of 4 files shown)


p0lycarpio (commented):
Indeed, the recipe timeline was broken during my merge and I didn't think to test it... Thanks for the fixes!

PR #2535, opened 2023-09-01
feat: Improved Ingredient Matching

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- feature

_(REQUIRED)_

This improves the ingredient parsers by leveraging the fuzzy search API. I've also added the ~~fuzzywuzzy~~ rapidfuzz package to choose the best result from the search, increasing the likelihood that it's right.

There are no UX changes, it's just been tacked on to the parser workflow. After the parser does its thing, it runs `find_ingredient_match` to try to match against the food/unit stores using the search API.

The results are pretty good even only using the brute force parser and sqlite. Here's an example before the changes, using my personal food/unit store (which includes seeded data and then some):
image

And here are the same ingredients with the new changes:
image

And keep in mind this is sqlite (no db-side fuzzy matching) and the brute parser, so we're working with pretty meh data. It should work even better with postgres and/or the NLP parser.

This PR also restores the auto-suggestion to create new units (it was halfway built, I just reused the logic for creating new foods) and makes it more responsive (since the matching is being done server-side now I could remove a lot of that code from the frontend, and safely update the food/unit stores without comparing names since we have ids).

_(REQUIRED)_

N/A, but depending on how well it works we could potentially work towards automating ingredient matching.

Still need to do a few things before this is ready:
- [x] Make sure the tests pass with postgres
- [x] See if any frontend changes are needed since we're matching on the backend
- [x] Play around with a few more recipes to make sure the thresholds are where they should be

One part of this PR that probably looks weird is the added Python-side fuzzy matching (using fuzzy wuzzy). ~~While the postgres trigram matching works well for sorting searches based on their similarity to the source string~~ see below, I moved to using this for both postgres and sqlite, in sqlite we're using "LIKE" sorting, which... kinda sucks. When the search is human-triggered (e.g. you're searching for a recipe and choosing the correct one) it doesn't matter because the right one is probably there. When you have to select one result, it really needs to be accurate. One instance of this I ran into was the food "onion"; I had "green onion" and "frozen pearl onions" as well as "onion" in my food store. Per the sqlite logic, the string "onion" (which is converted to "%onion%" when searching) is equally close to all of those.

Edit: looks like it's helpful in postgres too; some tests failed before I enabled it for postgres.

To solve this, after getting a maximum of 10 close matches from the DB search, I fuzzy-compare the original string against all results and choose the closest match that way. It uses the Levenshtein distance, implemented in C, so it's pretty fast (and we're comparing a max of 10 items since the search is capped):
```python
best_match: IngredientUnit | None = None
best_match_score: int = 0

for item in response.items:
    score = fuzz.ratio(unit.name, item.name)
    if score == 100:
        return item
    elif score > best_match_score and score >= self.unit_fuzzy_match_threshold:
        best_match = item
        best_match_score = score

return best_match
```

Note I had to put a threshold in there because some matches were really, really bad (e.g. "salt to taste" matched with "baby potatoes").

One avenue this may open up in the future is adding suggestions - e.g. if I auto-selected "onion", but I returned "green onion" from the search, I could provide that to the user and say "maybe you wanted green onion?". With this approach that would be easy to implement on the backend, not sure about the frontend.

## Testing

Added a bunch of tests and tested the brute force parser with sqlite locally a lot.

## Release Notes

_(REQUIRED)_

```release-note
improved the recipe ingredient parser
```


michael-genson:
I played around a bit with the NLP + sqlite with the default (seeded) food store and the threshold I had set was too lenient. I made it more aggressive (85% match) and didn't seem to get any false matches. If you've got any weird recipes it's worth trying out and seeing how the threshold is.

I left units at 50% since those are a lot more static, and they're smaller words so 1-2 characters being off is a huge hit to accuracy

jecorn:
Super cool! I’m so eager to start using this!

 A few minor thoughts:

You probably want to switch to thefuzz instead of fuzzywuzzy. But even better might be RapidFuzz. I’m a big fan of RapidFuzz. The benchmarks on the RapidFuzz GitHub page are eye-opening. https://github.com/seatgeek/fuzzywuzzy https://github.com/seatgeek/thefuzz

https://pypi.org/project/rapidfuzz/
https://github.com/maxbachmann/RapidFuzz


For even more speed, maybe test process.extract and process.extractOne instead of iterating in python.

michael-genson:
> thefuzz instead of fuzzywuzzy

Huh, had no idea they changed it

> even better might be RapidFuzz

Sure, I can take a look!

michael-genson:
Can't argue with a drop-in replacement that's faster ¯\\\_(ツ)_/¯

lindell:
Improvements in the ingredient matching are great! Trying it out, however, did not really give me the results I expected.

1) Units seem to not match on abbreviations. Example:
image
image

2) Small plural vs singular changes in the ingredient name are not watched.
image
Here the difference is "salladslök" vs "salladslök**ar**".

Number 2 seems to be because of this the database querying not returning it in the first place. Since this should be the case for most potential matches, including all plural matches when the database ingredient is in singular.  ("Onion**s**" is not included in Onion etc.)
https://github.com/mealie-recipes/mealie/pull/2535/files#diff-052302a82b68fba85945faddcbab71b1ff8c3be8445f2eac26f9542b8e0ad070R75

Replacing the query with fetching all ingredients instead:
```py
        query = PaginationQuery(page=1, per_page=-1)
        response = self.foods.page_all(query)
```

Makes it work exactly as expected!

How expensive is the fuzzing really? (especially with the RapidFuzz 🏎️) Fetching all (group) ingredients is something already expected for the frontend, so it might be OK here?

michael-genson:
> Units seem to not match on abbreviations

Ah you're right, it's because while the initial search matches abbreviations, the second refining step only looks at the name. ~~Should be an easy fix~~ fixed!

> Small plural vs singular changes in the ingredient name are not watched

I'm guessing you're using sqlite. We use token matching when searching the database because sqlite doesn't support fuzzy matching, so a simple query would be, for instance: `SELECT * WHERE name ILIKE '%cups%`. So if your unit is "cup" it won't be found because it doesn't contain "cups". With a postgres database, since we use proper fuzzy matching, it will be found.

> How expensive is the fuzzing really? (especially with the RapidFuzz 🏎️) Fetching all (group) ingredients is something already expected for the frontend, so it might be OK here?

@hay-kot, do you have any thoughts on that? Basically when parsing ingredients we'd fetch all units and foods and then fuzzy match against the full food/unit stores, bypassing database searching. I can try benchmarking it but I'm curious on your input.

mealie/services/parser_services/ingredient_parser.py
```diff
@@ -30,19 +67,70 @@ def parse_one(self, ingredient_string: str) -> ParsedIngredient:
...
+        choices_by_name = {item.name: item for item in response.items}
+        fuzz_result = process.extractOne(food.name, choices_by_name.keys(), scorer=fuzz.ratio)
+        if fuzz_result is None:
+            return None
+
+        choice_name, score, _ = fuzz_result
+        if score < self.food_fuzzy_match_threshold:
+            return None
+        else:
+            return choices_by_name[choice_name]
```
lindell: I think this should be done before the fuzz searching? Causing a O(1) find if the item exist with an exact match


```py
        choices_by_name = {item.name: item for item in response.items}
        if choice_name in choices_by_name:
                return choices_by_name [choice_name]
        fuzz_result = process.extractOne(food.name, choices_by_name.keys(), scorer=fuzz.ratio)
```

lindell:
The unit abbreviations works as expected now! 😄 

It seems the fuzzing takes around two orders of magnitude less time than the fetching of ingredients (on my dev environment). But a propper benchmarking should probably be made. Fetching all ingredients will also allow for reuse between ingredients, and only one DB query has to be made per parsing. I do think that will be way faster for most setups.
image

How many ingredients per group can we expect to be an upper limit? I tried to populate my test setup with 5000 foods. And the parsing times look like this:
Time DB: 0.13456988334655762
Time fuzz: 0.0005047321319580078
(sqlite)

lindell:
After running `VACUUM;` on the sqlite DB file, the fetching of all foods decreased **significantly**. Like 25x the speed. Not sure if it would have automatically auto-vaccumed it or not. The parsing was done right after I inserted the new foods.

This is still the result for 5000 food items.
Time DB: 0.00564265251159668
Time fuzz: 1.7642974853515625e-05

michael-genson:
Based on your benchmarks it seems to make sense to use pure fuzzing. I've gone ahead and updated this to pull the full food/unit stores and fuzz from there. Still want to get @hay-kot's input on that first before committing to it, though.

If we go this route, I could also remove the changes made to improve food/unit searching (indexing, normalization, etc.) although it might be worth it to leave it in anyway.

hay-kot:
> This is still the result for 5000 food items.
> Time DB: 0.00564265251159668
> Time fuzz: 1.7642974853515625e-05

Were there any considerations in benchmarking to consider memory or CPU? Some people run this on a potato, so if the fuzzer is optimized for heavy memory or heavy cpu workloads, I'd be interested to see what these benchmarks look like when operating in a constrained environment.

lindell:
I could try to do some real profiling to figure out memory, but let's first see if you are convinced by some worst-case numbers 😉 

Assuming the 5000 food items per group I used in my previous test is an upper limit (at least when running on a 🥔):

Memory:
The fuzzing algorithm should scale the amount of memory needed based on the number of items, but on the length of the strings. This should be very small and is in either case the same as for the previous implementation where a maximum of 10 where returned.

The actual food items returned are what should be pretty much all consumed memory. But it is the same as what is done by Mealie, just browsing recipes. Since it fetches all foods.
image
Even if not taking that into consideration, looking at the returned JSON (but Python obviously stores the data more efficiently), it returns **<200kb which should be an upper limit**.

CPU:
Assuming the fuzzing is CPU-bound, and that my machine is **100x faster** than what Mealie should be able to run on. The fuzzing should take 1.7642974853515625e-05 * 100 = **1ms** per ingredient.

hay-kot:
@lindell LGTM. We can always ship and see if anyone screams, thanks for all the info!

@michael-genson - is this ready to go then?

michael-genson:
@hay-kot should be good now, had to resolve some merge conflicts. I left in the food/unit database indexing/normalization, even though we aren't using it anymore since we switched to pure fuzzing. If you want me to remove that I can, otherwise it's good as-is

hay-kot (approved):

PR #2541, opened 2023-09-09
Fix #2534 - Add clearable ratings

### LINKED ISSUE #2534: [BUG] - Star Rating cant remove all stars
### First Check

You can add stars and update stars.  
Screen Shot 2023-08-31 at 09 30 35
Screen Shot 2023-08-31 at 09 30 52
Screen Shot 2023-08-31 at 11 22 30
Screen Shot 2023-08-31 at 09 34 58
 
You can not remove all stars.  If you give it a one Star you cannot remove it.  Unless you go in the JSON file and delete the RATING line from the file.  Then save and refresh the browser. 

### Steps to Reproduce

Pick any recipe give it one star.
The click the same star to remove 
nothing happens.

I also don't see an X or anything that has a tool-tip click to remove all stars.

none found.

### Mealie Version

Version
nightly
Build
408ca88cb2013210d3a3bfee26a16bf47195915a
DB Postgres

### Deployment

Docker (Linux)

_No response_

Grygon:
## What type of PR is this?

- bug

Allow clearing ratings

Fixes #2534 

## Testing

Added rating, refreshed to validate. Cleared rating, refreshed to validate.

## Release Notes

```
Ratings can now be removed by tapping the already-selected rating a 2nd time.
```


michael-genson:
Looks good! The only thing I'm not huge on is that it changes the value to "0" rather than null/undefined, which might be confusing (we'll have some recipes with a null value and some with a zero value, even though they mean the same thing). It also could be an issue if we ever decide to implement a true 0 rating.

Can you modify the `updateRating` function to convert 0 to undefined? Something like:
```javascript
function updateRating(val: number) {
  if (val === 0) {
    val = undefined;
  }
  ...
}
```

I'm not 100% sure the backend will actually update it to undefined/null so we might need to tweak the backend logic

hay-kot (changes requested):
See @michael-genson commnets, thanks for the PR definitely will have some happy users with this one!

Grygon:
Looking into it! It does look like the CRUD API doesn't like accepting it to be `undefined`.

From some testing it'll accept `null`, so I'll swap out the types for ratings to make it `number | undefined | null` and use that!

frontend/lib/api/types/recipe.ts
```diff
@@ -206,7 +206,7 @@ export interface Recipe {
   recipeCategory?: RecipeCategory[];
   tags?: RecipeTag[];
   tools?: RecipeTool[];
-  rating?: number;
+  rating?: number | null;
```
michael-genson, line 209: I don't think you need this because it's already optional. These types are auto-generated from the backend anyway

michael-genson (changes requested):
Looks good, thanks! Just a small change with the type definition

frontend/lib/api/types/recipe.ts
Grygon: Typescript doesn't like accepting `null` for an optional type:
image

I missed that they're just generated though so I'll go tweak that instead.

Grygon: I haven't been able to get the generation to work--I get this error after running `make generate`:
`Module Error: json2ts must be installed. Instructions can be found here: https://www.npmjs.com/package/json-schema-to-typescript`
Any method I try to install that doesn't resolve it.

Given [this comment](https://github.com/mealie-recipes/mealie/pull/526#issuecomment-862478853), are we _actually_ using the generation method or is it still a little hacky? Even once we get auto-generated types to work I suspect there will be problems explicitly having it generate a type that accepts `null`, so I'm perfectly happy adjusting the type in `recipe.ts` if we aren't actually using it as a generated type.

michael-genson: Ah that's funny, I've been using it a bunch. That's fine then, if it's a problem we can fix it later. Not really in the scope of this PR to figure that out anyway. Good work!

michael-genson (approved):

hay-kot (approved):

PR #2553, opened 2023-09-16
use score_cutoff for fuzzy string matching

maxbachmann:

Use `score_cutoff` parameter instead of manually performing the check afterwards. This requires less code and is potentially faster. The library is able to perform some optimizations based on the `score_cutoff` / currently best score.

It's cool to see my library being of use to mealie. I am personally using mealie for quite a while already :)


michael-genson (approved):
This looks great, thanks so much!

PR #2555, opened 2023-09-18
fix: CSS rendering bug, again

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- bug

_(REQUIRED)_

I don't know what the bug is. I don't know why this fix used to work but doesn't work anymore. I don't know why making the timeout longer makes it work again. But it works fine now I guess ¯\\\_(ツ)_\/¯

Before:
image

After:
image

Before:
image

After:
image

etc.

_(REQUIRED)_

N/A

Hopefully this actually works. I tried both 1ms and 10ms and they didn't work. 100ms seems to work

¯\\\_(ツ)\_\/¯ ¯\\\_(ツ)\_\/¯ ¯\\\_(ツ)\_\/¯ ¯\\\_(ツ)\_\/¯ ¯\\\_(ツ)\_\/¯

## Testing

Deployed prod locally and messed with the timeout until it worked again.

## Release Notes

_(REQUIRED)_

```release-note
NONE
```


hay-kot:
😆 

CleanShot 2023-09-19 at 12 06 55@2x

PR #2572, opened 2023-09-25
fix: removed unused node in docker image

lindell:
## What type of PR is this?

- bug
- cleanup

Node (and yarn) is not needed in the production image. Remove them and save ~100mb of image size.

## Release Notes

```release-note
Remove Node from the production docker image.
```


lindell:
Fixed merge conflict.

Issue #2596, opened 2023-10-06
Add header authentication


PhilippMundhenk:
To avoid having to tackle the more complex OpenID Connect stuff, reverse proxy/header authentication could be used instead / as a stop gap. Basically you run the application behind a reverse proxy, and the proxy authenticates the user and adds `Remote-User` and `Remote-Groups` (if applicable) headers to the request to tell the app which user is logged in, the app often has a "trusted proxies" configuration which prevents other sources from using this to bypass auth.

_Originally posted by @m00nwtchr in https://github.com/mealie-recipes/mealie/issues/1197#issuecomment-1280044865_

PR #2598, opened 2023-10-06
first shot at Remote-User support

PhilippMundhenk:
## What type of PR is this?
- feature

Currently, LDAP or internal authentication is supported. Most people will be using Mealie behind a reverse proxy and already have some form of (single sign-on) authentication in place. While OIDC is a very clean solution, for many setups, the use of the Remote-User header is sufficient. This is a first try of implementing this.

I don't have a running development environment, but this is rather the try to get things started instead of just coming empty handed with only a feature request. I will need some help with the implementation.

Addresses #2596 

This is just a first shot, I will need some help with the implementation.

## Testing

No tests at all. This is just a first shot, I will need some help with the implementation.

## Release Notes

```release-note
Support for HTTP header authentication with Remote-User header.
```

### DIFF
--- a/mealie/core/security/security.py
+++ b/mealie/core/security/security.py
@@ -11,7 +11,8 @@ from mealie.core.security.hasher import get_hasher
 from mealie.db.models.users.users import AuthMethod
 from mealie.repos.all_repositories import get_repositories
 from mealie.schema.user import PrivateUser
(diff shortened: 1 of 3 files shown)


PhilippMundhenk:
Ok, I tried testing this on my running instance, but I can't seem to get through the layers of frameworks here. A little out of my depth.

PR #2602, opened 2023-10-07
Fixed "remember me" with long TOKEN_TIME

lindell:
## What type of PR is this?

- bug
- feature

I wanted to increase the login time to something close to infinity, by increasing the TOKEN_TIME value. Opted for one year. But when trying it out, the token was only set for 14 days. Realized that the logic is
If remember_me:
   set to 14 days
else
   set to the TOKEN_TIME value.

This makes sense when using the default value (or TOKEN_TIME < 14 days). But it's a bit silly that using "Remember me" when logging in decreases the login time.

## Release Notes

```release-note
Fixed "remember me" with long TOKEN_TIME
```


mealie/core/security/security.py
```diff
@@ -22,11 +22,10 @@ class UserLockedOut(Exception):
     ...
 
 
-def create_access_token(data: dict, expires_delta: timedelta | None = None) -> str:
+def create_access_token(data: dict, expires_delta: timedelta) -> str:
     settings = get_app_settings()
 
     to_encode = data.copy()
-    expires_delta = expires_delta or timedelta(hours=settings.TOKEN_TIME)
```
michael-genson: I think if you leave this method alone it should achieve the same thing you're looking for without breaking other functionality, since in the instance you fixed it always passes an expires_delta

michael-genson:
This looks good, thanks! I was wondering why I kept getting logged out sooner than I thought. I left a comment which should fix the failing test

lindell:
A, I now see that there was one additional place (in the same file) that needed the argument to be set. 

But I don't see the comment you are referring to
> I left a comment which should fix the failing test

michael-genson:
You should see it on GitHub as a review right above my comment, but here's a screenshot just in case:
image

lindell:
The review is still pending :)
image

michael-genson:
LGTM. My only thought is that now the "remember me" toggle doesn't really do anything when you manually set the token time higher than 14 days, so maybe we need to rethink the logic there, _but_ either way this is still a net improvement over what we have currently

michael-genson (approved):

JvdMaat:
ROFL. So you're saying if I had just stopped clicking Remember Me, the token time would have worked better this whole time?
Remember Me should still have functionality though...
If you do not check it, it should not cache the login. (It should NOT use the token time then. Since you do not want to be remembered on that device. Maybe you're logging in on a public device)
If you do check it, it caches for either 14 days or TOKEN TIME.

hay-kot:
Could we just make it a multiple? So Remember Me is Token Time * {SOME_VALUE} ?

JvdMaat:
Honestly from an end user perspective. Checking the Remember Me box means you want your session to be cached (either the default 14 days, or TOKEN_TIME hours).
When it is not checked, you should be asked for credentials every time. You do not want to be remembered.

PR #2632, opened 2023-10-14
feat: Remove Explore URLs and make the normal URLs public

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- bug
- cleanup
- feature

_(REQUIRED)_

This removes the frontend explore pages and adapts the "private" pages to allow public access (unless things are marked as private, of course).

To accomplish this, all pages (with a few exceptions, see below) now have the group slug in their URL. So, for example:
- This URL is invalid: `https://my-mealie-domain.com/recipe/pasta-fagioli`
- This URL is valid: `https://my-mealie-domain.com/home/recipe/pasta-fagioli`

I'm not _too_ horrible of a person, so any time you reach the 404 page and you're logged in, we try to redirect you to `.../${groupSlug}/${whereverYouWereTryingToGo}` in an attempt to preserve existing links. This usually works, though there are bound to be edgecases. In the above example, for instance, the invalid URL will redirect you to the valid URL (if you're logged-in, otherwise we don't know which group to put you in).

The only non-group-slug-dependent pages are the following:
- all admin pages (since you manage app-wide things there)
- the registration page (because there was no need to change it)
- forgot-password, login, and reset-password pages (since we'll know where to put you once you log in)
- the main index page (although it will re-direct you immediately based on your logged-in state)

As a side-effect, all of the public URL copying (on recipes and on your user profile page) has been removed, since it's useless now. The recipe share functionality is still there, though, since that allows private access.

While I'm already messing with the sidebar, I noticed the cookbook secondary header wasn't working (hence https://github.com/mealie-recipes/mealie/discussions/2679) so I got that working again. I also conditionally hid it if you don't have any cookbooks yet.

---

As an added bonus to this already-too-big-for-my-own-good PR, I tweaked the `loggedIn` refs throughout the app to check if the logged-in user belongs to the group. This way a logged-in user can stay logged-in while visiting other groups (you can't really do anything special, you just aren't forced to log out to view other groups now).

This leads to some weird scenarios if you try manually adjusting your URL (e.g. if you go to `.../not-my-group-slug/shopping-lists` you will be successful, but it's still going to show you _your_ group's shopping lists). Fixing that would probably take 10x more work than the work done on this PR, since the app was originally built with the assumption that you're always in your own group, but there is no way through the UI to get to this weird state (that I know of, anyway), which I think is adequate.

~~Another option is to instead keep user and group pages within their own spaces~~ <-- I did this. I moved cookbooks and the recipe timeline under `.../${groupSlug}/recipe/...` to accommodate this.

_(REQUIRED)_

Closes discussion https://github.com/mealie-recipes/mealie/discussions/2679

Tons of things were changed to get this to work, but it's not _too_ much of a mess. Thankfully all the lower-level stuff didn't need much/any modification. I had to do some funky stuff with the routes/layouts to inject your group slug when you log in (since the login page is group-agnostic). It probably needs a few eyes on it to make sure nothing too jank is happening.

## Testing

Unfortunately, this PR touches pretty much every page and most of the main components, _and_ none of it affects anything with test coverage, so all testing is manual. I did a lot of poking around and global searching in VSC for patterns likely to break, but there are bound to be edge cases.

## Release Notes

_(REQUIRED)_

```release-note
migrated all pages to include the group name in the url
migrated all publicly-accessible pages to use the same urls as the private pages (if public access is enabled)
moved cookbook and timeline page URLs
users can stay logged-in while visiting other groups, but are treated as a "public" user
removed all explore pages
```


michael-genson:
Never mind, I went with the second option below. Keeping this comment for reference.

~~I added a few things to this PR, as noted above in my edited comment, but I wanted to specifically mention one thing regarding the user and group pages:~~

~~These changes lead to some weird scenarios if you try manually adjusting your URL (e.g. if you go to `.../not-my-group-slug/group/data` you will be successful, but it's still going to show you your group's data, not `not-my-group-slug`'s data). Fixing that would probably take 10x more work than the work done on this PR, since the app was originally built with the assumption that you're always in your own group, but there is no way through the UI to get to this weird state (that I know of, anyway), which I think is adequate.~~

~~Another option is to instead keep user and group pages within their own spaces, and not put the group slug in the URL, but this breaks the favorites page (and any page referencing recipe components), so we'd have to modify all of those components to work without a group slug in the URL (which I'm willing to do if you think that's a better idea than how it is currently).~~

hay-kot (commented):
Looks pretty good, I do have a couple of questions. 

1) I'm skeptical that a top level path variable isn't going to bite us in the ass. How would you feel about nesting the _groupSlug paths under `/u/` or `/g/` or something? That way it's name spaced and the routing rules become a little more clear? Then we could remove the `/recipe/` from the path and make them shorter. So `/g/{group-slug}/{recipe-slug}` ? 
2) Do the shoping-lists need to be under this path? Could those and meal-plans live under the root path since we aren't expecting those to be shared? 
3) I think this change is significant enough to warrant another RC round, so once this gets merged I'll publish another RC and then we'll only make small changes before the v1 release.

michael-genson:
> nesting the _groupSlug paths under /u/ or /g/ or something

Love that idea, that removes some of the ambiguity that already causes some weird behavior. So basically anything currently under `/groupSlug/...` would move to `/g/groupSlug/...`. We'd have to keep "recipe" in the URL though because there's also other things at that level (`cookbooks`, `recipes/tags`, etc). We could shorten it to `/r` instead of `/recipe` though, if you think that's better?

> Do the shoping-lists need to be under this path?

Nope, makes sense, I can move those back to being under the root path

michael-genson:
I changed the routes as discussed above (I went with `/r/` to replace `/recipes/` since we're kinda doing the same thing with groups and `/g/`). I need to play around with it more just to confirm nothing broke, but it looks good. I also updated the 404 logic to migrate `/recipe/...` links to `/r/...`

I realized the metadata injection was broken in prod since the FastAPI SPA was manually constructing those routes. I'm _pretty_ sure I fixed it, but I need to verify with a prod docker build (hopefully I'll have time later tonight, if not tomorrow, unless you get the chance to confirm)

michael-genson:
Not sure if I'm just doing something wrong or I broke something but running the docker compose locally doesn't work. I'm getting: `No such file or directory: '/spa/static/index.html'` (which shouldn't happen since that file should be there)

hay-kot (approved):
Pushed a fix for the build issue. I confirmed the sudo SSR is working. Can you double check the SSR and then merge? 

Super stoked to push this one out 🚀

hay-kot (approved):

dathbe:
So what does this now make the public url for viewing all recipes?  I had a shortcut on my iPhone for that page, which no longer works, and it seems this is the reason.  Is there still a link to view all recipes, or does public access only work for specific recipes like the example above?

On a somewhat related note, the old link was a redirect, and the redirected-to url was not publicly accessible directly, so I had to take the janky step of turning off my internet connectivity to save the public url to my iPhone home screen.  Is that something that has changed with this update?  If not, I can submit a feature request.

I find that it's nice to be able to share your catalogue with family and friends without requiring them to log in.  But more selfishly, it's just far easier to have the public read-only link quickly accessible from my phone for those times when your hands are covered in food.

michael-genson:
> So what does this now make the public url for viewing all recipes?

The same URL you use to browse while logged-in. There's no special URL anymore. So, if you navigate to your list of recipes, copy that link, log out, and try going to that link again. You'll see all your recipes, but you won't be logged-in.

dathbe:
Thanks.  I can confirm that using `https://myurl.com/g/home` or `https://myurl.com/g/home?orderBy=created_at` or  `https://myurl.com/g/home?orderBy=name&orderDirection=asc` and various other combinations does indeed allow you to view the recipes without logging in.

It's still creating an issue with iOS shortcuts, though, because when I try to add that page to my home screen it instead creates a link to `https://myurl.com`, which just redirects to the login screen.  I assume this is some quirk of how iOS/mobile Safari deal with creating shortcuts and not a problem per se with Mealie, but I wanted to flag it because it is probably something we can create an easy workaround to.  The trick of creating the shortcut while offline isn't working now.

nkringle:
> Thanks. I can confirm that using `https://myurl.com/g/home` or `https://myurl.com/g/home?orderBy=created_at` or `https://myurl.com/g/home?orderBy=name&orderDirection=asc` and various other combinations does indeed allow you to view the recipes without logging in.
> 
> It's still creating an issue with iOS shortcuts, though, because when I try to add that page to my home screen it instead creates a link to `https://myurl.com`, which just redirects to the login screen. I assume this is some quirk of how iOS/mobile Safari deal with creating shortcuts and not a problem per se with Mealie, but I wanted to flag it because it is probably something we can create an easy workaround to. The trick of creating the shortcut while offline isn't working now.

There is a workaround that you could use today by creating a redirect rule if you have mealie behind a load balancer/reverse proxy. See the POC on this [feature request](https://github.com/mealie-recipes/mealie/discussions/2751).

michael-genson:
> when I try to add that page to my home screen it instead creates a link to https://myurl.com

I'm not familiar with iOS shortcuts, but yeah, that seems weird

> There is a workaround that you could use today by creating a redirect rule if you have mealie behind a load balancer/reverse proxy

FYI I have a PR in for this now: https://github.com/mealie-recipes/mealie/pull/2772

boc-the-git:
Re the iOS shortcuts issue, can you provide specific detail for how that is being setup? The ideal would be a screen recording of start to finish. That might allow us to understand it better.

(I'm wondering if it might be using the PWA or something)

dathbe:
Sure.  
1) Open mobile Safari and navigate to `https://myurl.com/g/home`



2) Page will load as expected and user will be logged out.




3) Click the share icon at the bottom middle of the screen to bring up the share menu.



4) Click "Add to Home Screen," which brings up the shortcut dialogue but, curiously, only includes the root domain without the necessary "g/home".  (In Apple's wisdom of always trying to dumb things down, the url is not editable.)

boc-the-git:
Annoying you can't edit the url.

Are you comfortable building a Dev instance? I wonder if this is the cause of the url being what it is: https://github.com/mealie-recipes/mealie/blob/mealie-next/frontend/nuxt.config.js#L336

I'm not sure we realistically have much scope to change that though.

michael-genson:
> I'm not sure we realistically have much scope to change that though.

https://github.com/mealie-recipes/mealie/pull/2772 should at least fix this for users only using a single group (or who want all public users to be directed to the default group). This would effectively make the "default group" _also_ the "iOS Shortcut" group.

No idea if such a thing exists to have multiple start urls.

dathbe:
Yeah, I don't know enough about iOS shortcuts to know how it works.  But on RC1.1, you could get it to work using the offline method I described above.  And yes, 2772 should do what I'm looking for.  It's one extra click if you actually do want to log in, but that's fine for my purposes.  I'll wait for it to get merged.

I'm sure I could figure it out, but I've never built a dev instance.

PR #2639, opened 2023-10-15
fix: 500 error when sending unauthorized requests

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- bug

_(REQUIRED)_

Currently, if you send a request to a private endpoint without a token, a server error occurs. This is due to the token being mistyped as `str` rather than `str | None`, leading to an uncaught `AttributeError`.

_(REQUIRED)_

N/A, caught on Discord.

To be honest I'm surprised it took this long to notice. I tried to follow the git blame out of curiosity and couldn't figure out how long this issue's been around. Seems like a while, possibly.

## Testing

pytest


hay-kot (approved):
FWIW, i think FastAPI changed how they handle request errors. I feel like I've been seeing strange behavior in response errors after an upgrade.

michael-genson:
Huh, weird

PR #2674, opened 2023-10-21
feat: plural foods and units, and aliases

### LINKED ISSUE #1242: [v1.0.0b] [Task] - Support for Plural and Alternate Names for Ingredients and Units

Currently the recommendation algorithm for suggesting matches for the NLP and Brute parsers are implemented by exact match. This makes it difficult for some users to properly parse and import ingredients through the interface. 

### Related Discissions

- https://github.com/hay-kot/mealie/discussions/1219
- https://github.com/hay-kot/mealie/discussions/1185

So far I think there are a few things that need to be implemented for this to work for everyone

### Allow User Input for Aliases

We'd first need to capture potential aliases for Units and Ingredients through the UI and potentially in the seed data, though this may warrant moving the seed-data into it's own repository for greater exposure. These Alises need to be index in a way that can easily be searched on the backend. likely using their own database table. 

### Generate Mappings for Parser on the Fly

Currently the parser will look for abbreviations and replace them with the full text options (at least the NLP) to aid in uniform parsing. This is especially important for the NLP parser. Instead of this process, we could utilize the Alias database table and perform a rewrite of the text to find all the aliased values in the string and replace them with the root value. I think this approach would work for the units because those are fairly uncommon words to be used, but foods may be another story.

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- feature

_(REQUIRED)_

This PR adds plural names and aliases to foods and units (and a plural abbreviation for units). When parsing, the parser takes all names/abbreviations/aliases into account when matching against an input.

In addition, as you'd expect, foods/units conditionally render their plural counterparts, depending on the quantity and such:

- 1 cup onion
- 3 cups onions
- 0 onions (the zero is omitted)
- 3/4 cup onion
- 2.5 cups onions

If a food or unit is missing its plural counterpart (i.e. everyone's data prior to this getting merged) it falls back to singular (i.e. on migration nothing will look any different).

---

The data management pages have been updated to accommodate these changes, along with some helpful hint text:

image

image

image

image

_(REQUIRED)_

Closes https://github.com/mealie-recipes/mealie/issues/1242

Technically this adds several hundred tests... apparently you can chain `pytest.mark.parametrize`... lol. However none of them do anything expensive (such as database access) so they zoom.

## Testing

Pytest for backend, updated composable for frontend (and manually tested).


michael-genson:
Originally I was going to wait until https://github.com/mealie-recipes/mealie/pull/2632 was merged, but the data pages got reverted anyway so there shouldn't be any merge conflicts either way

hay-kot (approved):
LGTM

PR #2676, opened 2023-10-22
Update import-recipe-bookmarklet.md

philbog:
Adding optional settings for using keywords and remaining in edit mode to the sample javascript

## What type of PR is this?

_(REQUIRED)_

- documentation

_(REQUIRED)_

  The javascript in the doc didn't have parameters for the other options on the URL submission page. I added both a parameter for including tags from the recipe, and remaining in edit mode, and changed the URL to include those parameters.

None

## Testing
I tested the updated javascript with the bookmarklet generator, and it worked


boc-the-git (approved):
Thanks!

PR #2684, opened 2023-10-25
fix: Mealplanner Date Off By One (Daylight Savings)

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- bug

_(REQUIRED)_

As mentioned in Discord, the mealplanner has two November 5ths (easily reproducible on any instance):
image

I realized that November 5th is when daylight savings time ends, every developer's _favorite_ time of year. Peeking at the code, I noticed there were two entries for November 5th: one at 0:00, and one at 23:00.

This fix tweaks the date calculation to increment by a day (using JS magic), rather than 24 * 60 * 60 * 1000 ms:
```js
return Array.from(Array(numDays).keys()).map(
  (i) => {
    const date = new Date(weekRange.value.start.getTime());
    date.setDate(date.getDate() + i);
    return date;
  }
);
```

image

_(REQUIRED)_

N/A, Discord

## Testing

Locally. I checked some other red flag dates too just to make sure (e.g. next year's daylight savings).

### DIFF
--- a/frontend/pages/group/mealplan/planner.vue
+++ b/frontend/pages/group/mealplan/planner.vue
@@ -79,7 +79,6 @@ export default defineComponent({
       });
 
       if (sorted.length === 2) {
-        console.log(parseYYYYMMDD(sorted[0]));
         return {
           start: parseYYYYMMDD(sorted[0]),
           end: parseYYYYMMDD(sorted[1]),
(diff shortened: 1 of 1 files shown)


boc-the-git (approved):
Looks good

PR #2687, opened 2023-10-26
fix: translate hardcoded strings

p0lycarpio:
## What type of PR is this?

_(REQUIRED)_

- bug
- cleanup

_(REQUIRED)_

Some strings are hard-coded directly into components/pages. The aim of this PR is to replace them in favor of i18n translations.

I added the missing strings in `en-US.json`.
They will need to be added/synced in other locales.


boc-the-git (dismissed):
This looks good. Just need to resolve conflicts.

boc-the-git (approved):
Thanks @p0lycarpio!

PR #2690, opened 2023-10-28
docs: Fix en-GB translations for backups

boc-the-git:
## What type of PR is this?

- cleanup
- documentation

Fix doco

None.

Ironically the correct en-GB "translation" is just the same as en-US for this.

### DIFF
--- a/frontend/lang/messages/en-GB.json
+++ b/frontend/lang/messages/en-GB.json
@@ -577,7 +577,7 @@
       "import-summary": "Import Summary",
       "partial-backup": "Partial Backup",
       "unable-to-delete-backup": "Unable to Delete Backup.",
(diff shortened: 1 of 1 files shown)


michael-genson:
I don't think changes here will work, I'm pretty sure they have to be done through Crowdin:
https://crowdin.com/project/mealie

boc-the-git:
Fair enough! I was uncertain, but figured as they end up here that it would be sufficient. I'll look to make the changes in crowdin when time allows.

michael-genson:
iirc Crowdin will replace any manual changes made (other than en-US changes)

boc-the-git:
I've just made the change on Crowdin, so I'll close this PR.

FYI @hay-kot, it shows that you added the erroneous translation 22 days ago. Not sure if my change here will undo something you were working on..
image

PR #2707, opened 2023-11-07
Fix broken link to ingredient parser

mgrinspan:
## What type of PR is this?

- bug

The "PARSE" button that appears after enabling ingredient amounts currently 404s. This PR fixes the link.


michael-genson (approved):
Thanks!

PR #2709, opened 2023-11-08
feat: Add sort by total time to sort menu

jaredcat:
## What type of PR is this?

- feature

* Allows sorting of recipes by `Total Time`

Screenshot 2023-11-07 at 8 18 53 PM

N/A

## Testing

1. Added various recipes with different Total Time values, including some without any value
2. Went to explore page and change the sort option

### DIFF
--- a/frontend/components/Domain/Recipe/RecipeExplorerPage.vue
+++ b/frontend/components/Domain/Recipe/RecipeExplorerPage.vue
@@ -323,6 +323,11 @@ export default defineComponent({
         name: i18n.tc("general.random"),
         value: "random",
       },
+      {
+        icon: $globals.icons.timer,
+        name: i18n.tc("recipe.total-time"),
+        value: "total_time",
+      },
     ];
 
     onMounted(() => {


jaredcat:
This might be better if we add this value to the RecipieCard itself but wanted to check with maintainers if you would like this.
Example: 
Screenshot 2023-11-07 at 8 09 59 PM


If so, also let me know about placement: should it be on a different line on it's own? Same line with some styling? etc
Let me know, I can work this into this PR or anther one.

michael-genson:
For this to work I think we need to somehow measure the duration - simply sorting by it won't account for minutes, hours, etc. We could add [a calculated field to the database](https://docs.sqlalchemy.org/en/20/orm/extensions/hybrid.html) that parses the total duration into some common unit (probably seconds).

In terms of display, I think the existing `RecipeTimeCard` component would work, though you'd probably want add a configuration option to only display total time rather than all times

boc-the-git:
I'd add that some level of configurability would be nice if possible (which might just be multiple sort options, but factoring into display changes might be more work).. For instance prep time might be more important than total time - e.g 10min prep, 50min cook might be preferable to 30min prep and 15min cook, given prep might be hands on and cook hands off.
Just a thought.

jaredcat:
@michael-genson ah, yea i was just assuming people put in some consistent numbering but now realize that parsing from websites is just a string.
What's your opinion on a good way to handle this given theres no guarantee users will enter things correctly? Best I can think is change the UI to have distinct: hour, min, and seconds boxes instead of a text box.

@boc-the-git I was thinking about that too, wasn't sure if we want to clutter the menu with 3 options. Easy enough to add though.

michael-genson:
The biggest problem I have with changing the UI to only accept h/m/s is that we lose the ability to parse most/all recipe times. It removes utility from users who want everything to "just work". Another option would be to have a toggle similar to foods where you enable/disable structured times.

@hay-kot, any thoughts on this?

jaredcat:
> we lose the ability to parse most/all recipe times

I mean i would update this to include converting the parsed times into structured data but I like the idea of having a toggle.

hay-kot:
Ehhhhh. I'm skeptical here that we could bolt on anything useful that wouldn't result in a lot of poor UX and user complaints. 

IMO, if we wanted to do this we would first need to determine how we're going to store time durations and then move all existing time durations fields to that format. Since I doubt we could do this with 100% correctness, I'm not sure sure how to handle this. 

I don't want to store a text field and a parsed field for duration where the user just edits the text field and we try to parse it out into a time duration. 

---

Which is really to say, I think we've (read: me) dug a bug enough whole here that idk how to reasonable get out of it without pissing a lot of users off.

michael-genson:
> if we wanted to do this we would first need to determine how we're going to store time durations

Not saying whether or not we _should_ do this, but _if_ we do this, I recommend the ISO 8601 duration standard:
https://en.wikipedia.org/wiki/ISO_8601#Durations

e.g. PT1H10M is 1 hour and 10 minutes. Then let i18n handle the rest

PR #2710, opened 2023-11-08
feat: Add total time to recipe cards

jaredcat:
## What type of PR is this?
- feature

Adds a convenient way to see how long recipes will take to make on the explore page.

Screenshot 2023-11-08 at 12 32 23 AM
Screenshot 2023-11-08 at 12 35 57 AM

Idea from: https://github.com/mealie-recipes/mealie/pull/2709
Seems like a good idea to have this in place first.

I've never used Vuejs before.

## Testing

Added recipes and changed the layout size and tested the toggle switch

### DIFF
--- a/frontend/components/Domain/Recipe/RecipeCard.vue
+++ b/frontend/components/Domain/Recipe/RecipeCard.vue
@@ -31,7 +31,9 @@
             {{ name }}
           </div>
         </v-card-title>
-
+        <v-card-text v-if="totalTime && displayTotalTimes" class="px-1 pt-1 pb-0">
+          <RecipeTimeCard :total-time="totalTime"/>
+        </v-card-text>
         <slot name="actions">
           <v-card-actions class="px-1">
             <RecipeFavoriteBadge v-if="isOwnGroup" class="absolute" :slug="slug" show-always />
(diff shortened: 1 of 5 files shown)


jaredcat:
Additional question: Do we want switches for `prep time` and `cook time`?

michael-genson:
LGTM, need to dive in and see if there are any edge cases but looks like you covered them.

In general I like that the cards show just basic info on them, so I don't know if we want to add the other times. When I'm looking at my list of recipes, I don't care so much about how much prep/cook time there is, but I might be looking for something that doesn't take too long, so the total time can be useful.

Curious to hear anyone else's input on adding other times.

hay-kot:
Couple Notes

The spacing is off on the card for the time, it should be evenly distributed between 

CleanShot 2023-11-20 at 11 49 48@2x

I don't think adding a toggle for this makes sense, either we should display it or we shouldn't. 

I don't think reusing UI component here makes a lot of sense, it takes up a lot of real-estate on an otherwise dense view. If we're going to add more things to the card, I think it should probably be refactored so that the star rating is just a single star and condense the `time`, `favorite` and `ratings` into one row

`<clock-icon> 1:30  <heart-icon (outlined or not)>  3 <star-icon>` something vaguely like this? That might require making the star icon non-clickable.  

I also think that we should just include one time, so total is probably fine.

PR #2717, opened 2023-11-13
feat: Edit existing mealplan meals

JackBailey:
## What type of PR is this?

- feature

To change a meal plan item, from one to another you have to currently delete and re-add the item. This gets a bit tedious, especially if you're using a note rather than a recipe.

It's just a simple click on the item, and that'll open up an altered version of the "create meal" dialogue, pre-filled with the existing meal information. Only graphical changes are the text for the title and confirm button for the dialogue.

## Testing

I ran through editing the date, type, note or recipe and ran the `make backend-all` command.


frontend/lang/messages/ja-JP.json
boc-the-git, line 1: I can see various language definitions that are modified by this PR (putting this comment against Japan, chosen at random).
I'm not an expert on it, but I'd expect you to only add the definition to English (specifically en-US), rather than also putting it on these other languages.

I fully accept I could be wrong.

@JackBailey is there something that led you to modifying these specific language files?

@michael-genson any thoughts?

JackBailey: Thank you for taking the time to review my PR @boc-the-git.

I added them to non-english languages as it looks like all  of the other languages that haven't been fully translated have the English value. I wasn't sure how the translation process works and if it'd be added to the other languages automatically later not. 

I _did_ mean to add it to all of the languages, not sure why only some were changed, I probably searched the files wrongly. If this isn't the way to do it I can edit this PR so the translation is only for en-US and en-GB (Or add the english value for the rest of the languages), just let me know.

boc-the-git: I think @michael-genson might be more familiar, so let's give him a chance to comment. I _think_ you only need to add the new message to en-US and the rest is sorted out for you.
Probably an area where documentation can be improved (because obviously I don't know the answers either!)

Kuchenpirat: adding the key to the en-US file is enough. Crowdin will pick the new keys up and manage it from there.

JackBailey: Thanks - I've removed the changes to the other language files

frontend/components/global/BaseDialog.vue
```diff
@@ -178,6 +178,7 @@ export default defineComponent({
     close() {
       this.dialog = false;
       this.logDeprecatedProp("close");
+      this.$emit("closed");
```
boc-the-git: This emits `closed`, but in edit.vue you've got `@close="resetDialog()"`.. I'm assuming there's an issue caused by the discrepancy in close vs closed?

frontend/pages/group/mealplan/planner/edit.vue
```diff
@@ -300,19 +311,42 @@ export default defineComponent({
       text: "",
       recipeId: undefined as string | undefined,
       entryType: "dinner" as PlanEntryType,
+      existing: false,
+      id: 0,
+      groupId: ""
```
boc-the-git, line 316: Can you explain the usecase for `groupId` please? (in general, as it doesn't exist in the program prior to your changes)

boc-the-git (commented):
I've added some comments to try and plug gaps in my knowledge.
I think I'd need to test this briefly to be in a position to approve this - not certain when I'll get to that.

frontend/components/global/BaseDialog.vue
JackBailey: It was just a typo, but it looks like there's already "close" emitted so the above change is unneeded.

frontend/pages/group/mealplan/planner/edit.vue
JackBailey: Both id and groupId are unneeded here, was making sure there weren't any type errors but works and lints fine without them.

boc-the-git (approved):
LGTM.
Sorry it took me so long to get back to this @JackBailey. I've given it a brief test as well, and I must say, I like the experience!

PR #2727, opened 2023-11-16
feat: Implement github actions cache, for docker build

boc-the-git:
## What type of PR is this?

- cleanup
- feature

Adds caching to the docker build, giving an incremental speed improvement. Testing section has key info.

None

This is an incremental improvement.. definitely more work to be done here in future I'd think.
We need to get this in use and then see what more we can improve.

## Testing

Using my fork, I performed a number of builds. Represented in the image below. The general pattern I'd read about happened.. first build post-implementation ([1] to [2]) saw a longer run time, then subsequent runs were less ([3]). Having removed the caching change, run time started to creep back up ([4]).

image

I'm not keen to make any huge claims on what benefit this will deliver in practice but the testing showed roughly a ~9% reduction in build time.

### DIFF
--- a/.github/workflows/partial-builder.yml
+++ b/.github/workflows/partial-builder.yml
@@ -46,3 +46,6 @@ jobs:
           tags: ghcr.io/${{ github.repository }}:${{ inputs.tag }}
           build-args: |
             COMMIT=${{ github.sha }}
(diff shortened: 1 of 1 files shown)


michael-genson (approved):
This looks good to me! We'll see how it pans out as we start to use it on more PRs

PR #2746, opened 2023-11-21
feat: Manage data - Improve delete prompt

Kuchenpirat:
## What type of PR is this?

- (mini) feature

Shows which item will get deleted in the Confirmation Prompt. 

image

None

## Testing

Did some manual testing. As this does not touch any of the actual functionality it **should** not break anything


Kuchenpirat:
Have another version with the icon of the item in the front, but that might be a bit too playful.

image

boc-the-git (approved):
Looks good to me, nice QOL improvement

PR #2770, opened 2023-11-28
fix: Missing Title and Metadata

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- bug

_(REQUIRED)_

This restores the recipe tab title (again) and fixes the open graph metadata by using Beautiful Soup to parse the HTML before serving it.

_(REQUIRED)_

N/A, discussed in Discord

## Testing

Pytest. I can't get a local prod instance running to verify this actually works, though I think the tests are comprehensive. If whoever reviews this PR can build a prod instance and verify the metadata appears as it should, that'd be great. My goal with the metadata fixes is to restore the social share previews (e.g. sharing a link on Discord).


frontend/pages/g/_groupSlug/r/_slug/index.vue
```diff
@@ -32,28 +32,28 @@ export default defineComponent({
       const api = usePublicExploreApi(groupSlug.value);
       recipe = useAsync(async () => {
         const { data, error } = await api.explore.recipes.getOne(slug);
-
         if (error) {
           console.error("error loading recipe -> ", error);
           router.push(`/g/${groupSlug.value}`);
         }
 
+        setMeta(data);
```
Kuchenpirat: the metadata will not be set when the recipe belongs to the group of the user. So the page title only changes to the recipe name for non logged in users

michael-genson: Ah you're right, I must've been logged out while testing this. Thanks!
The reason it's in the branch like this is because the metadata wasn't getting set after the async call (so the title was never set). Just fixed it for the logged-in branch using a watcher

Kuchenpirat: why have the watcher in the if _isOwnGroup_ case and not just put it after the if-else statement. Could delete the `setMeta(data)` in the else block

michael-genson: Huh, for some reason I thought that didn't work because the owner block replaces the variable (rather than updating the ref data), but I tested it and it does work, so never mind

michael-genson:
Kuchenpirat was able to build it locally and test the social preview and it's still seeing the old tags, so I'm guessing adding them server side isn't enough. Will investigate

michael-genson:
Made some tweaks to fix the recipe share link and Kuchenpirat confirmed everything's working now. Images are only showing up on some social apps (e.g. Discord), but I think that has something to do with the image size. I'll investigate and try to open another PR, but this PR successfully restores the busted metadata.

Kuchenpirat:
just to add some screenshots:

left side: Discord: where the image of the recipe works
right side: Signal wich apparently needs to have a smaller image size provided

image


update:
just tried a few additional message services. Also works on WhatsApp, Instagram and Telegram so signal seems to be an outlier

Kuchenpirat:
only think that should be noted is that the browser does currently not recieve the updated metadata so here the meta tags will show the default mealie metadata. 

But this is also the current behaviour for mealie.

Kuchenpirat (approved):
Did some more testing and had a good look through the code. LGTM 

Besides Signal every platform i tested it on worked flawlessly.

PR #2772, opened 2023-11-29
feat: Redirect Logged Out Users to Default Group, If It's Public

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- feature

_(REQUIRED)_

This adds the default group slug to the app info route so we can use it to automatically redirect logged-out users to the default group's landing page. The route only returns the default group slug if the group exists, _and_ isn't private (that way the existence of a private group isn't exposed).

_(REQUIRED)_

Closes https://github.com/mealie-recipes/mealie/discussions/2751

## Testing

Added backend tests, tested the frontend manually.


hay-kot:
I'm skeptical that we need this if it can be accomplished with a reverse proxy. Would it instead make sense to add a guide to do this with the proxy? 

I'm willing to be convinced otherwise.

michael-genson:
The only reason I thought we might want to just make this a feature is that it makes single-group, fully-public instances more convenient, and make them feel like a groupless site (it just works). That, and I don't think it takes away from instances that do use groups (they already need to use the correct links, and could also just set the default group private to disable the redirect).

That said, I personally don't care either way, because I just use our group link (and have no issue with adding a reverse proxy if I change my mind). So I leave that up to you :)

nkringle:
> The only reason I thought we might want to just make this a feature is that it makes single-group, fully-public instances more convenient, and make them feel like a groupless site (it just works). That, and I don't think it takes away from instances that do use groups (they already need to use the correct links, and could also just set the default group private to disable the redirect).
> 
> That said, I personally don't care either way, because I just use our group link (and have no issue with adding a reverse proxy if I change my mind). So I leave that up to you :)

As the person who requested this feature (and has a 'good enough' implementation with my reverse proxy), I also don't care either way. I'd suggest that if it is left to a user to implement a reverse proxy, there be good documentation on a way to do that. It may also make sense to make it part of the migration guide from v0.5.

Kuchenpirat:
Do we want to implement this, or just direct users to the reverse-proxy method?

Either way, we should either close or merge this.

Norsagir:
Hi, I opened a discussion about this and Kuchenpirat redirected me to this PR.
I think this is a cool solution for the users who do not want to use a reverse proxy as their instance is offline and they dont not want to go through many configurations to get this working.
In my opinion, if this PR does not introduce bugs or security problems, it should be implemented in order to simplify the use case where Mealie is offline.

EDIT : By the way, how do you use reverse proxy to do that? I'm using Synology reverse proxy.

hay-kot (approved):

PR #2779, opened 2023-12-03
Link to GitHub registry instead of DockerHub in the docs

dmarcoux:
## What type of PR is this?

- documentation

When going to https://hub.docker.com/r/hkotel/mealie, we see this message:

> These images are no longer regularly updated. Please see the Github registry for the most up to date images.
>
> We plan to re-publish new images to docker-hub some time after a stable release is released.

I believe pointing users directly to the GitHub registry should be better.


Kuchenpirat:
Good catch 👍

Kuchenpirat (approved):

PR #2784, opened 2023-12-03
chore: restore latest tag

hay-kot:
## What type of PR is this?

_(REQUIRED)_

- chore

_(REQUIRED)_

Updates workflows to 

1. Publish latest tag
2. Publish to both ghcr and dockerhub

Ne because

1. Some systems don't allow you to pull images easily from other registries besides dockerhub
2. With the up-coming stable release, this will allow us to have a latest tag to recommend to users
4. People big mad about us not having a latest tag.

## Testing

I don't think there's really a way to test these without just pushing them to production :(

### DIFF
--- a/.github/workflows/partial-builder.yml
+++ b/.github/workflows/partial-builder.yml
@@ -4,6 +4,9 @@ on:
   workflow_call:
     inputs:
       tag:
+        required: true
+        type: string
+      tags:
         required: false
         type: string
     secrets:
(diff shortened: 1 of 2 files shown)


Kuchenpirat:
i see no problems with this, should we just go full steam ahead and open the flood gates? 🚀

michael-genson:
+1 to pushing and dealing with it later if it's a problem. I don't see a cleaner path forward

michael-genson (approved):

PR #2790, opened 2023-12-05
chore: remove GitHub Actions deprecated command

stumpylog:
## What type of PR is this?

_(REQUIRED)_

- cleanup

_(REQUIRED)_

GitHub has depracated usage of the `echo ::set-output` style commands.  See https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/.  Though they work for the moment, they should be updated to the new format.

While in these files, the action versions were generally behind, and were updated.  The "breaking" change is just requiring Node 20 to run, which does not affect those using a GitHub hosted runner.

_(REQUIRED)_

## Testing

Running the action is all that is required.


.github/workflows/partial-frontend.yml
```diff
@@ -9,20 +9,20 @@ jobs:
 
     steps:
       - name: Checkout 🛎
-        uses: actions/checkout@master
+        uses: actions/checkout@v4
 
       - name: Setup node env 🏗
-        uses: actions/setup-node@v3.7.0
```
stumpylog, line 15: I wasn't sure if there was a reason these version were locked more strictly.  I chose to follow that format, but they could be updated to be less restrictive (and require less work in the future)

boc-the-git (approved):
LGTM, thanks! I'd been meaning to do this myself but hadn't got to it yet.

.github/workflows/partial-frontend.yml
boc-the-git: @hay-kot I've approved the PR, but not yet merged; do you have any concerns with this change in particular, setup-node from 3.7.0->4.0.0?
[Relevant "releases" page](https://github.com/actions/setup-node/releases)

hay-kot: I usually just skim the release notes and push it to see if anything breaks.

PR #2793, opened 2023-12-06
feat: Implement security scans using CodeQL

boc-the-git:
## What type of PR is this?

- cleanup
- feature

Adds code security scanning, both in terms of regular scans against mealie-next, and on PRs.
For PRs, it doesn't reject them when they have the same issues as mealie-next, only new issues introduced on that PR.

Initially this is going to identify a number of issues (10 it looks like) that we'll need to progressively work through and either resolve or choose to ignore. We get our main benefit from this moving forward.

Note the initial CodeQL implementation is effectively the default supplied by github - I haven't tried to do anything custom/wacky.

None (that I know of).

## Testing

I initially merged the CodeQL change to my mealie-next and saw it identify 10 issues. I then made the change to scan PRs and threw in an intentionally "bad" commit to see that it failed the test and rejected the PR, as shown here:
image


michael-genson:
LGTM, though I wonder how it compares to Trivy

boc-the-git:
Good point @michael-genson, I was overlooking Trivy. I added this because GitHub doesn't consider us to have any security scanning on the repo (see the security tab)

I don't know if there's a minor modification to Trivy that could achieve that, or maybe GitHub don't consider it good enough, or something else. Factually which is better, I'm not qualified to say at this time (will do some reading).

boc-the-git:
So the key difference as I can see it, is CodeQL runs an analysis on the actual code base, where Trivy builds the container and then scans within the container.
Coming at it from two different angles.

hay-kot (approved):

PR #2803, opened 2023-12-08
fix: Prevent Creating Groups With No Name

### LINKED ISSUE #2800: [BUG] - Lack of Group Naming Validation
### First Check

Mealie allows users to create groups without assigning them a name. 

Once created it is not possible to move users into the group, because for user creation or moving into a new group the group name is required, but an empty string will not be accepted. 

### Steps to Reproduce

Via the Gui: 
- visit `/admin/manage/groups` page
- Click **Create** Button
- Click Create without doing anything

Via the API
- post a request to `/api/admin/groups` with the content `{ "name": "" }`
- get this response:
```
{
  "name": "",
  "id": "8944b068-71e6-40ed-9855-037621d76fd0",
  "slug": "",
  "categories": [],
  "webhooks": [],
  "users": [],
  "preferences": null
}
```

Just a 201

[12/08/23 09:30:39] INFO     127.0.0.1:0 - "POST /api/admin/groups HTTP/1.1" 201 

### Mealie Version

Issue can be reproduced in my dev container and the demo

### Deployment

Docker (Linux), Docker (Windows)

_No response_

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- bug

_(REQUIRED)_

This makes it so groups can't have an empty name (i.e. `name == ""`). Prior to this PR an empty group name is disallowed in some places, but not at group creation time.

I also added a db fix to check if a group with no name already exists (and rename it if it does).

_(REQUIRED)_

Fixes #2800

Horses with no name are still allowed.

## Testing

I tested the db fix manually (disabled the validator, created a group with no name, restarted the db, observed the fix). I'm not sure why the frontend allows you to submit the form with an empty name (it's set as required) but at least now it won't work (which I also tested, just to be sure).


mealie/db/fixes/fix_group_with_no_name.py
```diff
@@ -0,0 +1,51 @@
...
+
+            counter = i + offset
+            try:
+                _do_fix(session, group, counter)
+                break
+            except IntegrityError:
+                session.rollback()
+                attempts += 1
+                offset += 1
+                continue
```
hay-kot, line 50: Instead of a integer offset, could we just use part of a UUID? That way it's sudo random and you have less chances of collision. 

Or we just use a UUID for the name. Either way I'm good with

mealie/schema/user/user.py
```diff
@@ -66,6 +66,14 @@ class GroupBase(MealieModel):
...
         orm_mode = True
 
+    @validator("name")
+    def not_null(cls, v: str):
+        v = v.strip()
+        if not v:
+            raise ValueError("name cannot be empty")
+
+        return v
+
```
hay-kot: Probably easier to do this with a constrained type

https://docs.pydantic.dev/1.10/usage/types/#constrained-types

tldr:

```python
class Model(BaseModel):
    upper_str: constr(to_upper=True)
    lower_str: constr(to_lower=True)
    short_str: constr(min_length=2, max_length=10)
    regex_str: constr(regex=r'^apple (pie|tart|sandwich)$')
    strip_str: constr(strip_whitespace=True)
```

hay-kot (changes requested):

mealie/db/fixes/fix_group_with_no_name.py
michael-genson: Oh duh, yeah I'll just use the id as the name, thanks!

mealie/schema/user/user.py
michael-genson: Huh, that's a nice feature, TIL

hay-kot (approved):

PR #2806, opened 2023-12-09
Fix: Add 'loading' message to settings page

boc-the-git:
## What type of PR is this?

- bug
- cleanup

Currently when loading the settings, there is just _nothing_ while waiting for the app info to load. The populates a `Loading...` message.

None.

## Notes for Reviewer

The git diff looks to make it fairly hard to compare the actual scope of the change, which was only a few lines. It could be easier to look at the old and new version side by side, or in a better comparison tool.

## Testing

Mealie Settings Loading

### DIFF
--- a/frontend/lang/messages/en-US.json
+++ b/frontend/lang/messages/en-US.json
@@ -114,6 +114,7 @@
     "json": "JSON",
     "keyword": "Keyword",
     "link-copied": "Link Copied",
+    "loading": "Loading",
     "loading-events": "Loading Events",
     "loading-recipe": "Loading recipe...",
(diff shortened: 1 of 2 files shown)


michael-genson:
A couple things:

- Can we add some y padding to bottom to center it in the card? It looks a little crunched.
- Alternatively you might want to use the `AppLoader` component which gives a nice animation. For example, on the global timeline page:
2023-12-09_09h05_55

`<AppLoader :loading="loading" :waiting-text="$tc('general.loading-events')" />`

boc-the-git:
Good suggestion, I've implemented the AppLoader.

michael-genson (approved):
LGTM!

PR #2817, opened 2023-12-11
fix: Dev Settings and Nuxt Hot Reload

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- bug

_(REQUIRED)_

For some reason my dev container's frontend broke - polling for hot-reloading was constantly failing, resulting in an endless refresh loop. There were some other weird VSCode automatic updates which I applied to this PR. For the hot reloading, this thread tipped me off:

https://stackoverflow.com/a/75818603

Just forwarding port 24678 fixes it ¯\\\_(ツ)_/¯

_(REQUIRED)_

N/A

## Testing

Rebuilt my dev container and manually verified it works now


boc-the-git (approved):
Can't hurt any users.
Let's get it merged and go from there.

PR #2820, opened 2023-12-11
fix: Prevent Duplicate Recipes When Hitting Back Button

### LINKED ISSUE #2267: [v1.0.0b] - back button restarts import

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- bug

_(REQUIRED)_

As detailed in https://github.com/mealie-recipes/mealie/issues/2267 - if a recipe is created via URL, then the user hits the back button, Mealie scrapes the URL again (and creates a duplicate recipe). This clears the query param from the browser history to prevent this.

_(REQUIRED)_

Fixes https://github.com/mealie-recipes/mealie/issues/2267

## Testing

Manually

### DIFF
--- a/frontend/pages/g/_groupSlug/r/create/url.vue
+++ b/frontend/pages/g/_groupSlug/r/create/url.vue
@@ -103,7 +103,11 @@ export default defineComponent({
       if (refreshTags) {
         tags.actions.refresh();
       }
-      router.push(`/g/${groupSlug.value}/r/${response.data}?edit=${edit.toString()}`);
+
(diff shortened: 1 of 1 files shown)


Kuchenpirat (approved):
This will certainly come handy 👍
LGTM

PR #2832, opened 2023-12-12
feat: RTL Support for RTL Languages (Hebrew, Arabic)

### LINKED ISSUE #2452: [v1.0.0b] - Hebrew language isn't correctly displayed from Right To Left (RTL)
### First Check

Hebrew (and Arabic, and few other languages) are written from Right To Left (`RTL`).
When the language is set to Hebrew, the strings should be aligned to the right. Currently they are aligned to the left, as if they are English strings, which looks cumbersome in Hebrew.

In order to demonstrate the situation, here is how English looks like when aligned to the right:
```
 English
  should
      be
 aligned
      to
the left
```
So, it's the same problem, only the opposite.

I'm not sure if it's relevant, but maybe https://stackoverflow.com/questions/51683353/vue-js-need-to-switch-direction-rtl-to-ltr-and-vice-versa-without-reload  or https://stackoverflow.com/questions/64410664/change-direction-with-selecting-language-in-vuejs  can help (just note that this code check only for Arabic, it should be checked for Hebrew and Farsi (if supported) as well.

Also, there is something wrong with the combination of letters and numbers/punctuations. It *might* be solved by solving the right-alignment issue first. If not, it can be solved by using a standard Bi-Direction library (`BiDi`) such as https://pypi.org/project/python-bidi/

Thanks

### Deployment

Docker (Linux)

### Deployment Details

_No response_

### LINKED ISSUE #2469: [v1.0.0b] - Recipe is printed only on half page
### First Check

When I print my recipe (as PDF) it's printed over two pages, each one of them taking only half of the page's width.
The recipe's .json (zipped) is attached.
[rvglk-shl-tl.zip](https://github.com/hay-kot/mealie/files/12184230/rvglk-shl-tl.zip)

### Deployment

Docker (Linux)

### Deployment Details

_No response_

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- feature

_(REQUIRED)_

Some languages (such as Hebrew) are read right to left. This has a bunch of formatting considerations as well (text should be right aligned, etc.). Currently, while we have translations for these languages, we don't change the formatting.

Thankfully, Vuetify supports this pretty simply with a `rtl` boolean property (which is False by default). This PR adds some metadata to our locales to include the text direction (and sets `rtl` to `true` if the direction is `rtl`). I've also updated the codegen to preserve this.

_(REQUIRED)_

Fixes https://github.com/mealie-recipes/mealie/issues/2452
Tentatively fixes https://github.com/mealie-recipes/mealie/issues/2469, I'll close this with a note to re-open if not fixed

@hay-kot, if possible, can you confirm the codegen works? I can't really test it without the Crowdin API Key. I'm pretty confident it works, though.

Interestingly, it reverses a lot of the layout too (which I guess makes sense for RTL readers). This breaks a few things, but nothing major. We'll want to iron some of these out at some point.

image

## Testing

Poked around the frontend switching between LTR and RTL langs.


michael-genson:
If any of our supported languages are read RTL besides Hebrew or Arabic, let me know, I can add the metadata. I'm pretty sure those are the only two, but I'm no language expert.

boc-the-git (approved):

DeepSpace2:
Looks like we got a regression?

using v3.0.1:

image

michael-genson:
Please file a new bug report and we can take a look!

Issue #2698, opened 2023-10-31
Import of ZIP files not possible


FrankSchoene:
### First Check

- [X] This is not a feature request
- [X] I added a very descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Mealie documentation, with the integrated search.
- [X] I already read the docs and didn't find an answer.
- [X] This issue can be replicated on the demo site (https://demo.mealie.io/)

### What is the issue you are experiencing?

Its not possible to import zip-files. Neither for migration nor a single recipe.

### Steps to Reproduce

Wherever you get to the import dialog, you fail. Looks like "submit" is doing just nothing.

### Please provide relevent logs

nothing to log because nothing is happening.

### Mealie Version

**Details**
Version: nightly
Build: 4fc1ec7663f22dd5367c64d1f6009b9abdedac83
Application Mode: Production
Demo Status: Not Demo
API Port: 9000
API Docs: Enabled
Database Type: postgres
Recipe Scraper Version: 14.51.0

**Checks**
Secure Site: No
Server Side Base URL: Yes
LDAP Ready: No
Email Configured: Yes
Docker Volumes: success

### Deployment

Docker (Synology)

brunoelgrande:
Hi
I've got the same problem here.  My collection of recipes comes from **Paprika V2**, so I thought it was the bug (V3 is the most recent and probably the one supported by Mealie ?)

If I refresh the page, I can see that the import is in progress, but nothing change or complete many hours post import. 

![image](https://github.com/mealie-recipes/mealie/assets/90357412/f915f7f2-a2c5-4832-a880-11c74327f0e1)


Nothing in the logs neither. 

**Details**
Version: v1.0.0-RC1.1
Build: 
Mode de l’application: Réalisation
Mode démo: Non
Port de l’API: 9000
Documentation de l’API: Activé
Type de base de données: postgres
Version du Scraper de recette: 14.42.0

**Checks**
Site sécurisé: No
URL de base côté serveur: Yes
Prêt pour LDAP: No
Email Configured: Yes
Docker Volumes: success

michael-genson:
Can you guys try upgrading to the latest nightly and see if the behavior persists? We fixed some issues regarding backups and migrations since RC1.1.

Otherwise, can you please post your migration data that's failing? Migrations are working for most folks so your data might contain something we haven't seen previously.

brunoelgrande:
[...]

PR #2834, opened 2023-12-12
fix: Paprika Migration Error

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- bug

_(REQUIRED)_

Currently, our Paprika migrator fails if the recipe ingredients or instructions are null (we try to do `None.split(...)`). This adds a check for that.

On top of that, I added a try/except around the migrator func, so if there are any other edge-cases we should be able to get around them.

_(REQUIRED)_

Fixes one issue in https://github.com/mealie-recipes/mealie/issues/2698, but I'm still investigating the other

## Testing

Using the data in https://github.com/mealie-recipes/mealie/issues/2698


Kuchenpirat (approved):
LGTM

PR #1891, opened 2022-12-11
Intial OIDC login

TomCranitch:
## What type of PR is this?

- documentation
- feature

## What this PR does / why we need it:

This PR adds OIDC login.

## Which issue(s) this PR fixes:

This has been mentioned a few times (e.g. #1197)

## Special notes for your reviewer:

 - I'm not too sure what kind of tests would be good here (if any)
 - I've added authlib as it can use JWKS to verify the OIDC JWT
 - I'm not confident that what I've done in index.vue is the best approach, but I needed a way to override the user that is returned from the OIDC provider with the mealie user

## Testing

Have tested by hand using authentik.

## Release Notes

```release-note
Adds OIDC login for SSO.
```

## Remain todos

 - Documentation
 - Testing?? (I'm looking for suggestions here)
 - Linting

PR #2860, opened 2023-12-18
Login with OAuth via OpenID Connect (OIDC)

cmintey:
## What type of PR is this?

- feature

This PR adds the popularly requested login via OpenID Connect (OIDC), specifically using the Authorization Code Flow with PCKE. The work in this PR was originally based off of #1891, but has been modified to work with the switch over to the frontend being a SPA. 

This PR comes complete with end-to-end tests for testing the login flow using OIDC. These tests make use of a mock oidc server docker container and Playwright to run the UI tests.

Since we are close to a v1 release, this could maybe wait until after the release so as not to push it back. While I don't think there are any breaking changes with this, various configurations should be tested by users to make sure it covers as many use cases as possible.

## Testing

New E2E tests and manual


hay-kot:
> This PR comes complete with end-to-end tests for testing the login flow using OIDC. These tests make use of a mock oidc server docker container and Playwright to run the UI tests.

🔥 🔥 🔥 

---

Couple of questions from an idiot.

- Why OIDC over something like SSO via Forward Auth Headers? 
- If move to OIDC, should we still support LDAP (long term) i.e. is there any reasons why LDAP users couldn't easily switch to OIDC implementation?
- I've been looking at https://github.com/markbates/goth for another project as an "auth abstraction". They have this model of a Provider/Session interface and they use that to support multiple authentication models. You've look at the Auth code probably more than anyone else. Do you think there's a benefit here in spending some time to abstract the Username/Password, LDAP, and now OIDC into a more generic Provider interface? 
- Are you interested in being added as a maintainer for Mealie?

cmintey:
> Why OIDC over something like SSO via Forward Auth Headers?

Mainly because I have more experience with OAuth and OIDC. From my understanding, Forward Auth Headers are great for adding authentication to an application that otherwise doesn't have authentication. I've also seen that popular providers like Authelia, Authentik and Keycloak offer both Forward Auth and OIDC, so if someone is using one of those providers, it's probably easier to set up auth via OIDC instead.

Finally, because its pretty much "built-in" with [nuxt-auth](https://auth.nuxtjs.org/schemes/openidconnect/). Obviously I ended up having to create a custom provider because we serve the frontend as a SPA, most of the work is handled by that package, which is great.

If Header Auth is something the community wants, I know there was some great work done in #2040, which I could certainly take a look at also getting over the finish line.

> If move to OIDC, should we still support LDAP (long term) i.e. is there any reasons why LDAP users couldn't easily switch to OIDC implementation?

I think it is still worthwhile to support both methods. I can see someone wanting to just have an LDAP server without setting up an authentication provider. Especially if all their self-hosted apps all support LDAP (which many do), there would really be no need. I think as long as LDAP is used and doesn't become a chore to maintain, we should keep it.

> I've been looking at https://github.com/markbates/goth for another project as an "auth abstraction". They have this model of a Provider/Session interface and they use that to support multiple authentication models. You've look at the Auth code probably more than anyone else. Do you think there's a benefit here in spending some time to abstract the Username/Password, LDAP, and now OIDC into a more generic Provider interface?

I think it could be worth looking into. I would like to see if the OIDC implementation could be cleaned up a bit still. One thing I dislike is that now, our token verification is different between OIDC vs credential and LDAP. I should have some time next week to look into potential solutions

> Are you interested in being added as a maintainer for Mealie?

Maybe. I got a new job about halfway through the year, so I've been much more busy than I was earlier in the year when I was fixing LDAP issues. Let me think on it some more

sevmonster:
Is there a plan for RP-initiated SSO, or has such a feature been implemented? I for example would not be using any other authentication so login page is not really necessary. Though, there should probably be a backdoor link (e.g. Nextcloud uses `/login?direct=1` to bypass SSO) for emergencies in that case.

cmintey:
Are you asking for functionality to automatically redirect you to your identity provider and bypass the login screen? That is something I would like to add still.

sevmonster:
> Are you asking for functionality to automatically redirect you to your identity provider and bypass the login screen?

Yes, exactly. Since all users in my environment will use SSO, the RP (Relying Party) should be able to automatically initiate SSO, i.e. redirect to the IdP (Identity Provider).

mealie/core/security/providers/ldap_provider.py
```diff
@@ -0,0 +1,179 @@
...
+        search_filter = "(&(|({id_attribute}={input})({mail_attribute}={input})){filter})".format(
+            id_attribute=settings.LDAP_ID_ATTRIBUTE,
+            mail_attribute=settings.LDAP_MAIL_ATTRIBUTE,
+            input=self.request_data.username,
+            filter=user_filter,
+        )
+
+        user_entry: list[tuple[str, dict[str, list[bytes]]]] | None = None
+        try:
+            logger.debug(f"[LDAP] Starting search with filter: {search_filter}")
```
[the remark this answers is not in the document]
cmintey: These are false positives. The `search_filter` does not contain a password, only the username/email

mealie/core/security/providers/ldap_provider.py
```diff
@@ -0,0 +1,179 @@
...
+from sqlalchemy.orm.session import Session
+
+from mealie.core import root_logger
+from mealie.core.config import get_app_settings
+from mealie.core.security.providers.credentials_provider import CredentialsProvider
+from mealie.db.models.users.users import AuthMethod
+from mealie.repos.all_repositories import get_repositories
+from mealie.schema.user.user import PrivateUser
+
+logger = root_logger.get_logger("ldap_provider")
```
hay-kot: Can we initialize this within the class?

mealie/core/security/providers/openid_provider.py
```diff
@@ -0,0 +1,101 @@
...
+import requests
+from authlib.jose import JsonWebKey, JsonWebToken, JWTClaims, KeySet
+
+from mealie.core import root_logger
+from mealie.core.config import get_app_settings
+from mealie.core.security.providers.auth_provider import AuthProvider
+from mealie.db.models.users.users import AuthMethod
+from mealie.repos.all_repositories import get_repositories
+
+logger = root_logger.get_logger("openid_provider")
```
hay-kot: Can we initialize this within the class?

mealie/scripts/change_password.py
```diff
@@ -18,22 +20,32 @@ def main():
 
         if not user:
             logger.error("no user found")
-            exit(1)
+            sys.exit(1)
```
hay-kot, line 23: nice catch!

hay-kot (dismissed):
This looks fantastic! Super stoked to get this one merged. I'll approve it, but I want at least another maintainer to weigh in before we merge. 

You may now be everyones favorite contributor 😄

mealie/core/security/providers/ldap_provider.py
cmintey: Absolutely, done!

hay-kot:
Hey @cmintey we're good to merge this, could you fix the merge conflict? It should be an easy rebase/rebuild of the docs page. 

I tried to do it locally, but it said I had 85 changes to push and I was too scared to totally bork your branch :D

cmintey:
Hey @hay-kot, sorry for the delay, I was on vacation. I'll get this branch up to date.

I've been using this branch for a few weeks and did find a bug with the JWT not expiring leading to basically never expiring sessions. I should probably fix this bug before merging

cmintey:
Alright, I think I've squashed the bug now. This is ready for a review again

PR #2906, opened 2024-01-04
Update Recipe Organizer Delete Prompt Text

### LINKED ISSUE #2875: [BUG] - "Are you sure you want to delete this X?" popup text is grammatically incorrect when deleting tags and categories
### First Check

When deleting a tag or category, the text of the confirmation pop-up reads a bit odd. For example, when trying to delete a tag named "vegetarian":

image

I'm asked if I want to delete a vegetarian!

This appears to be due to the "confirm-delete-generic` being used improperly:

https://github.com/mealie-recipes/mealie/blob/342fa6128e0cbd18b1b6b3a44e14514c564e3a1e/frontend/lang/messages/en-US.json#L87

Instead of setting `${name}` to the name of the item, it should instead be set to "tag" or "category". Such that the text correctly reads "Are you sure you want to delete this tag?"

### Steps to Reproduce

1) Create a tag or category with name X.
2) Attempt to delete it.
3) See confirmation pop-up stating "Are you sure you want to delete this X?", which is grammatically incorrect.

N/A

### Mealie Version

v1.0.0-RC2 - c9852daf0e213dc4cadf97971c7787cf6fc66dfc

### Deployment

Docker (Linux)

_No response_

Kuchenpirat:
## What type of PR is this?

- bug
- cleanup

As described in #2875 delete prompts might create some funny delete prompts ala "Are you sure you want to delete this vegetarian?" This will change out the text to "Are you sure you want to delete this", move the item to be deleted into its own row and adds the Organizer Type to the Dialog Title. 

<details> 
<summary>Screenshots</summary>
<table>
  <tr>
    <th>Before</th>
    <th>After</th>
  </tr>
  <tr>
    <td></td>
    <td></td>
  </tr>
</table>
</details>

This will also fix an issue where the label delete prompt in the manage data page would put the label in the same row as the delete confirmation text, which was mitigated by wraping it with v-row. This didn't merrit its own pr so it included it in this one. 

Fixes #2875 

## Testing

Manual testing


frontend/components/Domain/Recipe/RecipeOrganizerPage.vue
```diff
@@ -5,12 +5,15 @@
...
       v-model="dialogs.delete"
-      :title="$t('general.delete-with-name', { name: deleteTarget.name })"
+      :title="$t('general.delete-with-name', { name: $t(translationKey) })"
       color="error"
       :icon="$globals.icons.alertCircle"
       @confirm="deleteOne()"
     >
-      <v-card-text> {{ $t("general.confirm-delete-generic-with-name", { name: deleteTarget.name }) }} </v-card-text>
+      <v-card-text>
+        {{ $t("general.confirm-delete-generic") }}
```
ekcom: Specify what is being deleted in dialog text and better HTML

```suggestion
<p>{{ $t("general.confirm-delete-generic-with-name", { name: $t(translationKey) }) }}</p>
```

ekcom (approved):
Looks much better :)

Tested locally myself.

frontend/components/Domain/Recipe/RecipeOrganizerPage.vue
boc-the-git: @Kuchenpirat when you get a chance, please have a look at this :)

Kuchenpirat: First of all thanks @ekcom for the feedback. Was bussy during the weekend, so my reply comes in a bit late. 

The only issue i can see with this is that Tag would be capitalized in this case, which is the reason i did not implement it this way. 

So either:
- we say capitalized is ok, and we go foreward with the solution proposed by @ekcom 
- Add a translation string specifially for _confirm-delete-tag_ ... which would be the cleanest
- keep it generic with _confirm-delete-generic_

Looking at it i like the proposed implementation more than leaving it all out, so im plementing it for now, but happy to change to either of the above.

boc-the-git:
From a code perspective I don't see any issue, but do you have updated screenshots of the after?

Kuchenpirat:
updated the PR description

boc-the-git (approved):
LGTM. Let's not let perfect be the enemy of good.

Nice one @Kuchenpirat, good feedback @ekcom 👍

PR #2914, opened 2024-01-06
fix: Warn on external navigation during editing

### LINKED ISSUE #2912: [BUG] - Warn when leaving edit page
### First Check

When navigating away via an internal link from the recipe edit screen and changes have been made, the user is prompted with: `You have unsaved changes. Do you want to save before leaving? Okay to save, Cancel to discard changes.`

However, when leaving the page in a different manner (e..g, refreshing the page, closing the tab), changes are lost without warning. There is no prompt to save changes before leaving.

I am considering this a bug since the current behavior is not consistent with the behavior when leaving the edit page to an internal site.

### Steps to Reproduce

1) Navigate to the edit page of a recipe
    - E.g., https://demo.mealie.io/g/home/r/crispy-fried-gnocchi?edit=true
2) Change something in the recipe
    - E.g., change the title to `Crispy Fried Gnocchi`
3) Reload the page
4) Observe that changes are lost without warning

N/A

### Mealie Version

nightly

### Deployment

Docker (Linux), Other (please specify below)

Reproducible on demo site

ekcom:
## What type of PR is this?

_(REQUIRED)_

- bug

- Provide a last-second warning when navigating
- Covers navigation which the current navigation on unsaved changes alert
    does not handle (before this, only internal redirection gave the alert)

Fixes #2912


ekcom:
Note: this currently will display the warning message (e.g., "Reload site? Changes you made may not be saved.") on the edit page even if no changes were made.

I am not sure how to efficiently detect any changes. Can't check `JSON.stringify(props.recipe) === JSON.stringify(originalRecipe.value)` during the `beforeunload` event, so would probably need an event listener for any changes in the form. This may not be worth the performance cost.

This feature still works and is ready.

frontend/composables/use-navigation-warning.ts
boc-the-git, line 1: These console.log messages seem a bit strange/blunt to me. Is there any precedent for these messages?

ekcom: Ope, that was a debugging log that I forgot to remove :)

boc-the-git (approved):
LGTM. Tested it a bit and does what it says on the tin. Nice work!

I do agree with what you've suggested, that there's some inconsistency with some of the other internal navigation/warnings (e.g. refreshing the page (new warning) is different to navigating to the timeline (old warning)), but importantly this does work and will stop people losing data.

Further enhancements are absolutely welcome! :)

PR #2919, opened 2024-01-07
feat: Add "id" to notes so they can be a hyperlink target

boc-the-git:
## What type of PR is this?

- feature

Make it so notes can be hyperlinked to.

None.

This implementation works well, but is not perfect. There are other limitations in the wider Mealie application that mean you can't write something like `<a href="#note2">Note 2</a>`, as the hyperlink will go to `http://localhost:3000/#note2` despite you being on for example `http://localhost:3000/g/home/r/asian-beef-bowls` to begin with. 

Where this implementation actually works though, is if you just specify the full URL (e.g. `<a href="http://localhost:3000/g/home/r/asian-beef-bowls#note2">Note 2</a>`) you _will_ get linked to the correct place.

So there's definitely future enhancements to be added for this, but I don't think it should stop us from pushing ahead with this PR. I see no downside, and plenty of upside, even if it's not perfect.

## Testing

Manual. Documented a bit on Discord (and informing the above comments).


michael-genson (approved):
I agree we should investigate a better way to enable anchor link navigation, but I also agree that this PR is a quick win with no real downsides. Thanks!

PR #2922, opened 2024-01-07
feat: Add a Create Bookmarklet feature

mcnamamj:
## What type of PR is this?

- feature

Today, users who want to have a bookmarklet to ingest recipes from a desktop browser need to web search to find the [documentation](https://nightly.mealie.io/documentation/community-guide/import-recipe-bookmarklet/) to manually create a bookmarklet, which requires more technical skill than my mom can handle.  (Sorry mom.)

Ideally, users should be able to:

- Self-discover inapp about the bookmarklet feature
- Have URL parameters like protocol, domain, and port pre-populated
- Use familiar UI options to configure how the bookmarklet can work for them
- Be provided instructions on how to install the bookmarklet

This PR implements the following changes

- [x] Within `r/create.vue`, add a new menu option for the subpage switch to a new local `bookmarklet` route
- [x] Implements `r/create/bookmarklet.vue` using the UI from `r/create/url.vue` and the logic from the existing documentation's [Import Bookmarklet](https://nightly.mealie.io/documentation/community-guide/import-recipe-bookmarklet/) page.  This allows users to decide if they want to automatically import keywords as tags or to stay in edit mode.
  - [x] Import `~/composables/use-copy` to enable an `@click` listener on the `v-textarea#bookmarkletResult` JS code that, when able, puts the bookmarklet code into the user's clipboard.  `use-copy` imports `use-toast`, so users will see an appropriate success or failure toast depending on whether writing to the clipboard succeeded.
- [x] Updates `messages/en-US.json` with 4 new strings
- [x] Updates the documentation website with new documentation for this feature

During testing on Safari, after using the bookmarklet to import a recipe, I would see the RecipePage and as soon as it finished loading the page would switch to a 404 with an uncaught error exception (NotAllowedError).  I traced this to the Safari Wake Lock API's security policy.  I believe Safari was being more restrictive and not allowing a newly opened tab to establish the lock w/o user interaction.  Therefore I also made this change:
- [x] Add a try/catch to `RecipePage.vue` around `@vueuse/core.useWakeLock.request()` within our implementation of `lockScreen()`.

I didn't create an issue.

No special notes.

## Testing

Tested on:
- [x] Mac Chrome (120.0.6099.199)
- [x] Mac Safari (17.2.1)
- [x] Mac Vivaldi (6.5.3206.50)
- [x] Mac Firefox (121.0)

For each browser above, the following tests are performed:
- [x] Confirm navigation works:
  - [x] Able to directly navigate to the page
  - [x] Able to toggle between ./url and ./bookmarklet
  - [x] URL parameters are updated on check/uncheck for both options
  - [x] URL parameters are respected in setting the checkboxes on page draw
- [x] Rendering and layout: 
  - [x] Widescreen on load
  - [x] Narrow on load
  - [x] Resize between
- [x] Bookmarklet generation:
  - [x] Correctly sets the JS with the protocol, domain, subdomain, port, and groupSlug
  - [x] Live respects the checkboxes
  - [x] Copy works
- [x] Bookmarklet works:
  - [x] Opens in a new tab
  - [x] Existing tab's URL is unchanged
  - [x] Mealie gets a properly encoded URL with protocol, domain, path, and URL parameters
  - [x] Mealie correctly follows the settings defined during bookmarklet generation (auto keywords/auto edit)

### DIFF
--- a/frontend/pages/g/_groupSlug/r/create/bookmarklet.vue
+++ b/frontend/pages/g/_groupSlug/r/create/bookmarklet.vue
@@ -1,34 +1,29 @@
 <template>
     <div>
-      <v-form ref="domUrlForm" @submit.prevent="createBookmarklet(importKeywordsAsTags, stayInEditMode)">
+      <v-form ref="domUrlForm">
         <div>
           <v-card-title class="headline"> {{ $t('recipe.create-bookmarklet') }} </v-card-title>
           <v-card-text>
             {{ $t('recipe.create-bookmarklet-description') }}
-            <v-checkbox v-model="importKeywordsAsTags" hide-details :label="$t('recipe.import-original-keywords-as-tags')" />
-            <v-checkbox v-model="stayInEditMode" hide-details :label="$t('recipe.stay-in-edit-mode')" />
+            <v-checkbox v-model="importKeywordsAsTags" validate-on-blur hide-details :label="$t('recipe.import-original-keywords-as-tags')" />
+            <v-checkbox v-model="stayInEditMode" validate-on-blur hide-details :label="$t('recipe.stay-in-edit-mode')" />
           </v-card-text>
-          <v-card-actions class="justify-center">
-            <div style="width: 250px">
-              <BaseButton rounded block type="submit" :loading="loading" />
-            </div>
-          </v-card-actions>
+            <v-textarea
+            v-model="bookmarkletResult"
+            :label="$t('recipe.create-bookmarklet-result')"
+            :prepend-inner-icon="$globals.icons.tags"
+            filled
+            rows="2"
+            class="rounded-lg mt-2"
+            rounded
+            :hint="$t('recipe.create-bookmarklet-hint')"
+            persistent-hint
+            :readonly="true"
+            @click="copyBookmarklet"
+          ></v-textarea>
+
         </div>
       </v-form>
-      <v-expand-transition>
-        <v-alert v-show="error" color="error" class="mt-6 white--text">
-          <v-card-title class="ma-0 pa-0">
-            <v-icon left color="white" x-large> {{ $globals.icons.robot }} </v-icon>
-            {{ $t("new-recipe.error-title") }}
-          </v-card-title>
-          <v-divider class="my-3 mx-2"></v-divider>
-
-          <p>
-            {{ $t("new-recipe.error-details") }}
-          </p>
-
-        </v-alert>
-      </v-expand-transition>
     </div>
   </template>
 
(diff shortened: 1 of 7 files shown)


mcnamamj:
https://github.com/mealie-recipes/mealie/assets/1053961/cfa0d19a-7a61-4264-94ef-3176e19c58f3

Kuchenpirat:
Hey thanks for the cool feature idea. I like it a lot.

However without having taken a look in the code yet, I believe dedicating an extra field in the overflow button might be excessive. 
This is considering that a user would typically utilize this feature once, and only revisit it when changing browsers or the address of their mealie instance. 

I have not yet thougt of a better option, but maybe you have already something in mind having worked on this.

frontend/pages/g/_groupSlug/r/create/bookmarklet.vue
```diff
@@ -0,0 +1,118 @@
...
+      });
+
+      const stayInEditMode = computed({
+        get() {
+          return route.value.query.edit === "1";
+        },
+        set(v: boolean) {
+          router.replace({ query: { ...route.value.query, edit: v ? "1" : "0" } });
+        },
+      });
```
Kuchenpirat: As those are not needed in the route, you can just implement them as normal variables
```suggestion
      const importKeywordsAsTags = ref<boolean>(false);
      const stayInEditMode = ref<boolean>(false);
```

frontend/pages/g/_groupSlug/r/create/bookmarklet.vue
```diff
@@ -0,0 +1,118 @@
...
+        }
+
+        // Now that we found the third slash, remove it and everything after
+        url = url.substring(0, position);
+
+        // window.history.replaceState is appended to fix Vivaldi bug
+        // https://forum.vivaldi.net/topic/31409/bookmarklets-replaces-the-url-in-the-address-bar/25?lang=en-US&page=2
+        return "javascript:" + encodeURIComponent(`(function(){var dest="${url}/g/${groupSlug.value}/r/create/url?use_keywords=${importKeywordsAsTagAsNumber.value}&edit=${stayInEditModeAsNumber.value}&recipe_import_url="+encodeURIComponent(document.URL);window.open(dest,"_blank");window.history.replaceState({},"",location.href);})()`) + ";"
+      }
+      });
```
Kuchenpirat: the easier way to do this would be to use the URL origin of the page
Furthermore you can just cast the booleans (eg. importKeywordsAsTag) as Number without having to add a additional variables
```suggestion
      const bookmarkletResult: ComputedRef<string> = computed(() => {
        // get the origin of the current page
        const url = new URL(document.URL).origin;

        // window.history.replaceState is appended to fix Vivaldi bug
        // https://forum.vivaldi.net/topic/31409/bookmarklets-replaces-the-url-in-the-address-bar/25?lang=en-US&page=2
        return "javascript:" + encodeURIComponent(`(function(){var dest="${url}/g/${groupSlug.value}/r/create/url?use_keywords=${Number(importKeywordsAsTags.value)}&edit=${Number(stayInEditMode.value)}&recipe_import_url="+encodeURIComponent(document.URL);window.open(dest,"_blank");window.history.replaceState({},"",location.href);})()`) + ";"
      });
```

frontend/pages/g/_groupSlug/r/create/bookmarklet.vue
Kuchenpirat, line 1: There are some unused imports to clean up and you return a few unused variables

Kuchenpirat (commented):
A few suggestions to shorten and simplify some things.

mcnamamj:
Thank you for your review and suggestions!  I learned a bit from reviewing your changes.  I appreciate your time.

> I believe dedicating an extra field in the overflow button might be excessive. This is considering that a user would typically utilize this feature once, and only revisit it when changing browsers or the address of their mealie instance.

Totally true.  Although the same is probably true for Import with Zip and Bulk URL Import.
 
> I have not yet thought of a better option, but maybe you have already something in mind having worked on this.

I really should have made a note in the top issue: I'm very much not married to this patch in any way, and agree the suggested spot in the overflow isn't ideal (additionally because it's not obviously intuitive to explore what the overflow has).

In leu of the overflow, maybe just adding text to the `v-text` description of the Import URL page?

Replacing:
> Scrape a recipe by url. Provide the url for the site you want to scrape, and Mealie will attempt to scrape the recipe from that site and add it to your collection.

with
> Scrape a recipe by url. Provide the url for the site you want to scrape, and Mealie will attempt to scrape the recipe from that site and add it to your collection.  For quicker access to this feature, you can [create a bookmarklet](https://example.com) (<-- would link to the new view in this PR) for a desktop web browser or [add a tool](https://nightly.mealie.io/documentation/community-guide/import-recipe-bookmarklet/) to import recipes from your mobile web browser.

Kuchenpirat:
Hey, do you still want to pursue this? Sorry for taking a long time to come back to this.

If we would want to continue, my idea about where to place this would be alongside the "Looking for Migrations?" link on the create page. And move the create screen to a dialog. That way we would not need an extra page.

Kuchenpirat:
Since progress here seems to have died down, i will close this PR. 
Feel free to reopen if you want to continue working on this. If you have any questions feel free to reach out.

Issue #2904, opened 2024-01-04
[BUG] - Database migration (unique constraints) fails with foreign key violation


hbh7:
### First Check

- [X] This is not a feature request.
- [X] I added a very descriptive title to this issue (title field is above this).
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Mealie documentation, with the integrated search.
- [X] I already read the docs and didn't find an answer.
- [ ] This issue can be replicated on the demo site (https://demo.mealie.io/).

### What is the issue you are experiencing?

I've been roughly following the latest versions for the last few months and at some point recently, I encountered a similar issue that was fixed by #2859, but now it revealed another error for me. This doesn't prevent the app from starting, but it doesn't let me open recipes and throws an error during startup. 

### Steps to Reproduce

1. Have existing installation from around the time when the single container switch was made
2. At some point (can't remember and apparently didn't notice for a while, sorry), notice that there's an error opening recipes.
3. Attempt using nightly container (last updated today), notice the error changed after some new fixes, open issue after finding no similar issues. 

### Please provide relevant logs
Startup error: 
```
mealie    | INFO: 03-Jan-24 20:03:00    Database connection established.
mealie    | INFO: 03-Jan-24 20:03:00    Context impl PostgresqlImpl.
mealie    | INFO: 03-Jan-24 20:03:00    Will assume transactional DDL.
mealie    | INFO: 03-Jan-24 20:03:00    Migration needed. Performing migration...
mealie    | INFO: 03-Jan-24 20:03:00    Context impl PostgresqlImpl.                                                                                                                         mealie    | INFO: 03-Jan-24 20:03:00    Will assume transactional DDL.
mealie    | INFO: 03-Jan-24 20:03:00    Running upgrade bcfdad6b7355 -> 0341b154f79a, added normalized unit and food names
mealie    | INFO: 03-Jan-24 20:03:01    Running upgrade 0341b154f79a -> dded3119c1fe, added unique constraints
mealie    | Traceback (most recent call last):
mealie    |   File "/opt/pysetup/.venv/lib/python3.10/site-packages/sqlalchemy/engine/base.py", line 1965, in _exec_single_context
mealie    |     self.dialect.do_execute(
mealie    |   File "/opt/pysetup/.venv/lib/python3.10/site-packages/sqlalchemy/engine/default.py", line 921, in do_execute
mealie    |     cursor.execute(statement, parameters)
[...]

PR #2859, opened 2023-12-17
fix: More Backup Restore Fixes

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- bug

## What this PR does / why we need it:

_(REQUIRED)_

This fixes more backup-restore-related issues by replacing some of the Alembic ORM stuff with raw SQL.

## Which issue(s) this PR fixes:

_(REQUIRED)_

Fixes https://github.com/mealie-recipes/mealie/issues/2857

## Special notes for your reviewer:

_(fill-in or delete this section)_

This (and other similar issues) seems to only be an issue when folks are restoring from an older versions (i.e. they make a backup, upgrade, then restore). When restoring on the same version as the backup it seems to be fine.

## Testing

_(fill-in or delete this section)_

Ran my ever-growing test suite of backups and they all restored successfully.

PR #2923, opened 2024-01-08
fix: Migration Failure On Postgres Due To Foreign Key Error

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- bug

_(REQUIRED)_

Adds some more commits to a troublesome migration. The migration is supposed to remove all references to the duplicate food/unit ids before deleting them, but it seems like the references aren't being deleted for some users.

This might be related to https://github.com/mealie-recipes/mealie/pull/2859, but I'm not sure since I can't reproduce using backups. Locally the migration works fine, and the test passes here (using both sqlite and postgres).

_(REQUIRED)_

Possible fix for https://github.com/mealie-recipes/mealie/issues/2904, need to keep issue open and confirm

## Testing

Pytest


boc-the-git (approved):
LGTM. Any concerns @hay-kot ?

Kuchenpirat (approved):
i don't see how this would break stuff (famous last words) since its just adding a bunch of session.commits and renaming a few files.

Issue #1106, opened 2022-03-28
[v1.0.0b] [Task] - Recipe Scraper Rules/Automations


hay-kot:
### What is the problem this task addresses?

Currently the scraper doesn't perform any "Smart" actions on scrape. It would be beneficial to create rules around the content of the recipe and apply them when/if it's successfully scraped.

### Proposed/Possible Solution(s)?

Example Rules:

1. If recipe text contains "dinner" apply "Dinner" tag
2. if recipe text containers "potatoes" apply "side" tag
3. Always apply "Haven't Cooked Tag"

Open to additional ideas of what could be done using this pattern of behavior

zierbeek:
If Matthias created this recipe, tag it Matthias

nkringle:
Something I manually do right now in v5, If the scraper does not pull in instructions or ingredients, I tag it as "Needs Fixing" or "Missing Data" so that I can circle back and fix it later.

wandt0n:
Could allow to specify external translation engine, e.g., URL that the recipe url is appended to. Possible workaround for [#1812], i.e., enables ingredient extraction on non-english recipes

PR #2930, opened 2024-01-09
feat: Apply default tag on recipe import (by URL)

boc-the-git:
## What type of PR is this?

- feature

The intent here is to allow for an "inbox" type of tag, that would get populated on every URL import and allows a user to review recipes that they haven't reviewed/checked for accuracy.
It might be that someone wants to go in and rename foods. Someone else might want to confirm nutrition looks right, etc.. 
Many reasons someone might want to confirm the end result of the parsing is exactly what they'd want.

This is just an enabler to make that easier to do - the tag gets added to all recipes imported by URL, bulk or single. It does not get added to other import types (e.g. zip, migration) as it's assumed those are already coming in in a clean state.

By default it's disabled, and requires the user to take action to enable it (in group settings).

None.

## Testing

Manual for the most part. Have added unit tests as well.


boc-the-git:
~Few minor issues to work through, will sort in the coming days and raise a new PR.~ Closing the PR might've been a bit dramatic, the effort to resolve type errors wasn't that great - though I really do need to be in bed!

boc-the-git:
I guess SQLite doesn't enforce the foreign key constraints in tests, but Postgres does. Will resolve hopefully tomorrow. Review/comments are welcome in the meantime should anyone wish - I'm expecting the subsequent changes will be local to the unit test only.
Solution is to actually create a tag, not just generate a uuid.

boc-the-git:
Have set this back to draft because I need to actually test the recent changes. Expecting it to go smoothly, but it won't get done today.

boc-the-git:
Tested and ready for review again :)

hay-kot:
We have this old issue/task

- https://github.com/mealie-recipes/mealie/issues/1106

That I think has some cross concerns here. This implementation is very specific, and I'm wondering if there is a better way to generalize the implementation to be more flexible. Otherwise if #1106 ever gets implemented most of this work would likely have to be redone/removed.

Having a db structure for Import Automations might be a good step? We could store something like this

```python
class ImportAutomation:
    rules: RegexAutomation | AlwaysAutomation | TagAutomation # json column maybe? 
    type: str
    groupd_id: UUID 
    order: int
```

Then we could have multiple rules that could apply in series on every import.

boc-the-git:
FWIW, I agree with the prior comment and it is my intention to at some point make this more generalised.. I have no timeline for when though.

PR #2951, opened 2024-01-10
Update pylintrc

cmintey:
## What type of PR is this?

- cleanup

Removes useless pylint options from disabled that were spamming VSCode with warning messages. Also changed some pylint options to be a little more reasonable. Now there are a lot less pylint issues, but there are still a ton. This will hopefully make it easier to identity actual code smells


hay-kot (dismissed):

michael-genson (approved):
had to fix the lockfile because of all the other dependency PRs, but it should work now

boc-the-git (approved):

PR #2959, opened 2024-01-10
deps: add frontend to renovate

Kuchenpirat:
## What type of PR is this?

- dependencies
- feature

This adds the frontend dependencies to the renovate bot. 

None. It just adds more work 

## Testing

tested locally. Frontend dependencies are discovered.


hay-kot (commented):
I had intentionally left out the frontend dependencies because

1. We are now using a unsupported version of Vue/Nuxt and upgrading is not a simple merge
2. We don't have sufficient frontend test coverage to give me confidence things won't break. 
3. Upgrading frontend dependencies with this process has resulted in many broken versions of mealie going out.

Worth discussing with the other maintainers. We could optionally only watch the critical packages that relate to security or sanitization like DOMPurify? 

🤷

Kuchenpirat:
Ok, so it's what i thought it would be 👍

Automatic updating via renovate certainly does not make much of a sense for most of our frontend dependencies. Is there a "notify me or display in dependency dashboard that there is a new version so i can work on it" with renovate?

That might be the right step for frontend deps then.

Kuchenpirat:
I grouped the frontend dependecies into two groups for minor/major deps and disabled pr creation by renovate. This would give us an overview in the dependency dashboard what can be upgraded. 

Below an image. It's not pretty, though i could not find another way to seperate the frontend deps from the backend ones and i would say it's at least functional. 

<details>
<summary>Screenshot</summary>

</details>

Kuchenpirat:
since frontend deps are a bit of a mess currently, i will be closing this and reopen, once we are in a better place 😊

PR #2980, opened 2024-01-11
fix: Bump FastAPI

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- bug

_(REQUIRED)_

Bumps FastAPI and fixes token auth when testing. This seems like a FastAPI `TestClient` bug.

_(REQUIRED)_

N/A - should close a bot PR (https://github.com/mealie-recipes/mealie/pull/2946)


michael-genson:
FastAPI broke something in 0.105.0, so we can't update until that's fixed:
https://github.com/tiangolo/fastapi/issues/10857

PR #2984, opened 2024-01-12
Add LOG_DIR env var to set log file directory

yodaldevoid:
If LOG_DIR is not set, the data directory is used to match the previous behavior.

I would add tests, but I am unsure what the best way of testing this feature is.

## What type of PR is this?

- documentation
- feature

In the old implementation, the mealie log file always was placed in the /app/data folder, preventing it from being separated from non-log application data. i.e. if a user wants to place it under /var/log

- Added `LOG_DIR` handling mirroring the existing `DATA_DIR` support, including adding `determine_log_dir` function
  - Fall back to `determine_data_dir` if `LOG_DIR` is not set
- Used `determine_log_dir` to set the root logger file path
- Added `mkdir` call to ensure `LOG_DIR` exists

Fixes https://github.com/mealie-recipes/mealie/discussions/2954 (discussion)

### DIFF
--- a/docs/docs/documentation/getting-started/installation/backend-config.md
+++ b/docs/docs/documentation/getting-started/installation/backend-config.md
@@ -15,6 +15,7 @@
 | API_DOCS      |         True          | Turns on/off access to the API documentation locally.                               |
 | TZ            |          UTC          | Must be set to get correct date/time on the server                                  |
 | ALLOW_SIGNUP  |         true          | Allow user sign-up without token                                                    |
+| LOG_DIR       |       /app/data       | Directory to place log files in                                                     |
 
 ### Security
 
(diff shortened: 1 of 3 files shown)


hay-kot:
While I can appreciate the want to stick a log file in a specific directory. I don't really want to merge this because I think it's a move in the wrong direction. 

IMO, we shouldn't even be logging to a file. We should just log to STDOUT/STDERR and let whomever is deploying the app decide what to do with the logs. 

--

I'll let the other maintainers weight in here though if they have any other opinions

yodaldevoid:
Bump for bot and visibility

hay-kot:
Hey! This kind of feature will be available once #3104 is merged.

- https://github.com/mealie-recipes/mealie/pull/3104

Closing this in favor of the more flexible implementation. Thanks!

PR #2988, opened 2024-01-14
fix: admin pages accessible by non admin users

### LINKED ISSUE #2927: [BUG] - Non-admin users can navigate directly to admin pages via URL (even though they can't do anything)
### First Check

If a non-admin user navigates directly to `.../admin/site-settings`, they can see the admin pages. They don't work (all the APIs throw 403 errors), but it can lead to some confusing behavior since the app assumes only admins can get there

### Steps to Reproduce

1) Log-in as a non-admin
2) Go to an admin-only page by typing in the URL, e.g. https://demo.mealie.io/admin/site-settings
3) Observe that things seem to work, but performing actions is broken

N/A, backend works as expected, throwing `403` errors

### Mealie Version

nightly

### Deployment

Docker (Synology)

_No response_

Kuchenpirat:
## What type of PR is this?

- bug

This reroutes users that are authenticated, but not admins back to the homepage if they acces one of the admin pages directly though an url (eg. `/admin/backups`).

- fixes #2927 

## Testing

Manual testing by going to the relevant pages as "bob" a non admin user.


michael-genson (approved):
Looks great, thanks! Love the use of middleware

frontend/middleware/admin-only.ts
```diff
@@ -0,0 +1,10 @@
+interface AuthRedirectParams {
+    $auth: any
+    redirect: (path: string) => void
+}
+export default function ({ $auth, redirect }: AuthRedirectParams) {
+    // If the user is not an admin redirect to the home page
+    if (!$auth.user.admin) {
```
michael-genson, line 7: I feel like this shouldn't work for public users (i.e. `$auth.user` is null) but it does (I just tested it). Maybe because it evaluated the `auth` middleware first?

Kuchenpirat: yeah, tested that too. 
Im having a thing in the back of my mind that because the auth middleware is defined in the nuxt config it will be evaluated first. 
I'll recheck that tomorrow.

michael-genson: Seems to work fine, I just found it weird that it worked

Kuchenpirat: by default they will be executed in [alphabetic order](https://nuxt.com/docs/guide/directory-structure/middleware#ordering-global-middleware)

PR #3001, opened 2024-01-17
feat: Use GitHub actions for Stale detection

boc-the-git:
## What type of PR is this?

- cleanup

We haven't had stalebot working for a while. This implementation gets rid of the old stale config, and introduces the official GitHub Action for stale issues/PRs: https://github.com/marketplace/actions/close-stale-issues

None.

You'll note debug mode is turned on - this means that it should take no action against anything until we remove that flag. [Relevant doco](https://github.com/marketplace/actions/close-stale-issues#debug-only)

I'll monitor what I can see in the Actions output as time allows, and raise a new PR in the next week or so to remove the flag, including any tweaks that seem worthwhile.

Also TBC whether there'd be permissions issues, but I think we fix forward with that once we see how it wants to act.

## Testing

Very little, per the special notes, the plan would be to implement this effectively dormant, and then tweak config as needed based on what we see "in production".


.github/stale.yml
```diff
@@ -1,21 +0,0 @@
...
-# Number of days of inactivity before a stale issue is closed
-daysUntilClose: 7
-# Issues with these labels will never be considered stale
-exemptLabels:
-  - pinned
-  - security
-  - early-stages
-  - "bug: confirmed"
-  - feedback
-  - task
```
hay-kot, line 12: Can you we port over as well, or re-evaluate? 

I don't want to auto close everything thats "stale". I think we'd really only need `pinned`, `security`, `bug: confirmed`, and `task` everything else could probably be considered a discussion.

Kuchenpirat: Look at row 14

hay-kot: Dammmmmmmmmmmmit.

Thank you!

hay-kot (approved):

Kuchenpirat:
only thing im not sure on here is whether 30 (+5) days for issues are enough to close them automatically. 
Tho at this point it should have an label that would keep it open, so im fine with the current implementation 👍

Kuchenpirat (approved):

boc-the-git:
Thanks for the comments guys. Definitely happy to review the 30+5, I'm not wedded to any particular value. Main focus with this initial PR was to get it implemented and see how it functions in dry run mode.

Issue #2807, opened 2023-12-09
[SCRAPER] - Units for "Servings", "Minutes" and "Hour" missing  in German localization


P9k:
### First Check

- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I have verified that this issue _is not_ related to the underlying library
[hhyrsev/recipe-scrapers](https://github.com/hhursev/recipe-scrapers) by **1)** checking
the [debugger](https://demo.mealie.io/g/home/r/create/debug) and data is returned, **2)**
verifying that there _are_ errors in the log related to application level code, or
**3)** verified that the site provides recipe data, or is otherwise supported by
[hhyrsev/recipe-scrapers](https://github.com/hhursev/recipe-scrapers)

- [X] This issue can be replicated on the demo site (https://demo.mealie.io/)

### Please provide 1-5 example URLs that are having errors

https://www.chefkoch.de/rezepte/1313551236082544/Baumkuchen.html

### Please provide your logs for the Mealie container `docker logs <container-id> > mealie.logs`

[2023-12-09-mealie.log](https://github.com/mealie-recipes/mealie/files/13624577/2023-12-09-mealie.log)

### Deployment

Docker (Linux)

P9k:
To give more context, when scraping from a German recipe site like chefkoch.de, the units for "Servings" as well as for the preparation/cooking times (in "Minutes" and "Hours") are not translated into German, while the rest of the recipe details are properly translated is (e.g. "Teelöffel" for "teaspoon").

- Instead of "serving/servings" it should say "Portion/Portionen" 
- Instead of "Minute/Minutes" it should say "Minute/Minuten"
- Insteadt of "Hour/Hours" it should say "Stunde/Stunden"

michael-genson:
The times using English words is definitely an issue with how we parse ISO 8601 durations:
https://github.com/mealie-recipes/mealie/blob/f8ad72ec3147320c191b1bb1284f06e345a708b0/mealie/services/scraper/cleaner.py#L408-L423

I'm not sure where the word "serving" comes from but it's probably a similar issue.

PR #3005, opened 2024-01-19
feat: Stale action tweaks

boc-the-git:
## What type of PR is this?

- cleanup
- feature

Further tweaks for the stale action, to see how they run in production. Debug mode is still on.
The main reason to actually merge this is because I saw the `bug: confirmed` label didn't successfully exclude https://github.com/mealie-recipes/mealie/issues/2807 from being marked as stale - hopefully this change solves that, but there's no clear documentation on the best path forward.

None.

Debug mode is still on.


Kuchenpirat (approved):
LGTM 
Maybe we want to decrease the `operations-per-run` if we go into production with this, but for debugging rate-limits seem to be an non issue

PR #3017, opened 2024-01-20
feat: Authenticated navigation and duplication for public recipes

hunterjm:
## What type of PR is this?

- bugfix
- feature

This pull request adds navigation for authenticated users to view other group's public recipes (building on top of #2525). It also enables duplicating recipes across different groups, so when a user is logged in and viewing another recipe, they can add it to their own collection in order to modify it.

| Authenticated - Own Group | Authenticated - Public Group | Not Authenticated |
|--------------------------------|---------------------------------|----------------------|
| image | image | image |

This PR also fixes some caching bugs in cookbooks, tags, etc. that were introduced in the frontend stores when switching between recipe listings for different groups when authenticated.

Lastly, to make it more useful, I modified the recipe duplication functionality introduced in #1750 in order to allow duplication of recipes from group to another. This cross-group duplication removes categories, tags, and tools from the duplicated recipe in order to simplify the implementation. This feature is enabled in the frontend by allowing logged in users to see the "duplicate" menu item when browsing a public recipe in addition to the "print" items.

image

This is a WIP towards my [comment](https://github.com/mealie-recipes/mealie/discussions/2134#discussioncomment-8187007) in #2134 

## Testing

Currently changes are only tested manually. I will go back and add test coverage if the general direction is approved.


hay-kot:
Hey! So sorry to take so long to get back to you. I truly thought that I had commented on this some time ago. 

I don't think we have any intention of merging a stop-gap solution like this over what is described here.

- https://github.com/mealie-recipes/mealie/discussions/2134

Duplicating recipes across groups isn't the sort of collaboration method we're looking to implement in Mealie, and this PR would likely just add work that needs to get undone when/if we get to implementing the improved collaboration with the "Household" style integration. 

Thanks for your work on this, and the discussion you added too, we appreciate it!

PR #2882, opened 2023-12-28
fix: re-enable change user and fix SIGTERM

hay-kot:
## What type of PR is this?

_(REQUIRED)_

- bug

## What this PR does / why we need it:

_(REQUIRED)_

- Revert a commented out function I did in a PR (and fixes command)
- Fixes SIGTERM listen (verified locally)

## Which issue(s) this PR fixes:

_(REQUIRED)_

Closes #2845
Closes #2723
Closes #2411 (probably)

## Special notes for your reviewer:

_(fill-in or delete this section)_

## Testing

_(fill-in or delete this section)_

PR #3030, opened 2024-01-22
fix: running the container with PUID=0 and PGID=0

xThaid:
## What type of PR is this?

_(REQUIRED)_

- bug

_(REQUIRED)_

I run the container with PUID=0 and PGID=0 and when updated to the recent release (V1.0.0), it couldn't start. I was getting a stream of logs like:
```
Switching to dedicated user
usermod: no changes
Switching to dedicated user
usermod: no changes
Switching to dedicated user
usermod: no changes
Switching to dedicated user
usermod: no changes
Switching to dedicated user
...
```
It is because the entrypoint script (running with id=0) tries to create a new user (with id=0) and then executes itself again as that user. Hence the infinite loop.

Regression caused by #2882 

_(REQUIRED)_

## Testing
Build the container and run it with PUID=0 and PGID=0


hay-kot:
Thanks for the PR! Why do you want to run as 0?

xThaid:
I use rootless Podman to run containers, so there's no need to change the user ID further.

renhardy:
My setup also requires having PUID=0 and PGID=0. This change would be appreciated.

renhardy (approved):

hay-kot (approved):
Tested locally, LGTM

PR #3037, opened 2024-01-23
feat: Updating "Label" Seed file to align with future seed logic changes

### LINKED ISSUE #1230: [v1.0.0b] [Task] - Link food labels to foods when importing seed data

Seed foods are missing link to the food labels. This is a shame because it means individual mealie admins have to re-organize foods themselves and it is very time consuming. 

One of the benefits of having foods linked to food labels is for the shopping list feature; Mealie is able to sort the list by food labels so that every food of the same type is grouped together (usually like in supermarkets).

Make it so that when the seed foods are created, they are linked with the right food label (also seeded data).

catduckgnaf:
## What type of PR is this?

- cleanup
- feature

-  The goal of this PR is to work towards having improved seed files and improved seed logic.  
- I took very detailed file from @hay-kot and organized all the labels and formatted JSON correctly
- much better seed files, that are aligned with together, with labels and foods. 
- some aliases are added
- plurals are added
 

- fix #1230

- I would like feedback on the label categories, to make sure they are appropriate. 
- I also have attached the updated "Master" Seed file. I would appreciate feedback on that as well as how to potentially implement the seed logic
- I added colors to the seed file. I have no opinion and they were picked at random.

## Testing

Validated JSON


Kuchenpirat:
1. So taking a look at the labels, i would say the categories are pretty good, can't think of any missing. Personally i would put the spices of "Herbs and Spices" into "Seasonings & Spice Blends" maybe just rename that "Spices and Seasonings". 

2. Currently the seeds are completely seperate from each other, so you would need to rewrite the seed logic to to first seed the labels and then seed the actual foods. The Frontend should also reflect that one is about to seed labels and foods at once i would say. But i think you should be able to reuse a lot of the existing code basis. 

3.  Regarding the colors i had mine organized to roughly represent something of the group. So "Meats" would be red, "Nuts & Seeds" a shade of brown and "Dairy & Egs" in white .... 
There are 30 categories which means there will be similarities between the colors, but i think it makes more sense than to pic a random color.

catduckgnaf:
> 1. So taking a look at the labels, i would say the categories are pretty good, can't think of any missing. Personally i would put the spices of "Herbs and Spices" into "Seasonings & Spice Blends" maybe just rename that "Spices and Seasonings".
> 
>     2. Currently the seeds are completely seperate from each other, so you would need to rewrite the seed logic to to first seed the labels and then seed the actual foods. The Frontend should also reflect that one is about to seed labels and foods at once i would say. But i think you should be able to reuse a lot of the existing code basis.
> 
>     3. Regarding the colors i had mine organized to roughly represent something of the group. So "Meats" would be red, "Nuts & Seeds" a shade of brown and "Dairy & Egs" in white ....
>        There are 30 categories which means there will be similarities between the colors, but i think it makes more sense than to pic a random color.

Thanks, for now I think simply updating the current seed files are the way to go. I plan to go through and add some color logic.
Also if you take a look, I added TONS of foods, with some aliases as well as labels. and plurals. So if someone imports the new labels, then the new foods, that should import the foods with the labels, the plurals, and aliases. 

My thoughts are to merge that, let things get translated (for everyone) before changing seed logic?

hay-kot:
My initial feedback is that I'm not sure that this is the right format when we consider translations

```json
    {
        "label": "Supplements & Extracts",
        "aliases": [],
        "description": "",
        "name": "lemon extract",
        "pluralName": "lemon extracts"
    }
```

Each entry is going to have to be translated so the labels will have the be translated multiple times in the translation UI. 

Would reworking the structure into a nested field work better? 

```
 {
  "label": "Some label"
  "foods": [
    ...foods
  ]
 }
```

Then we can manage all the seeds and stuff in one place. This also might make it easier to do seeding by language. 

---

> My thoughts are to merge that, let things get translated (for everyone) before changing seed logic?

I don't think this is really possible. we don't want to ship anything half implemented, and we _always_ want everything in the `mealie-next` branch to be release ready so we can push a release whenever we need to. I think the way to go is to get the new format and seeder figured out, and then migrate the existing translations into the new format, that way we can at least get a head start on the translations and those who already put in a ton of work don't lose their progress.

catduckgnaf:
@hay-kot I did it the way I did, because that matched how things looked both currently, and how an export looked. 
I feel your solution could become more complicated with nest structured. Would it be even more complicated to instead of it being nested, simply having 22 or so Seed Files for each food category? So essentially multiple files, however every food in that file belongs to the same label?

Kuchenpirat:
Hey, do you plan on continuing to work on this?
If yes and you have any questions feel free to reach out.

catduckgnaf:
I want too, some things came up. I'm concerned about my ability to do the bigger changes required, not having as much experience. Let me try to dive into again within the next few weeks

Kuchenpirat:
I'll go ahead an close this PR for now. 

Feel free to reopen if you come back to this.

Cameronwyatt:
Revisiting this... @hay-kot is this the [format you were thinking?](https://github.com/user-attachments/files/18268432/data.json) I took the foods generated by @catduckgnaf and moved to the format (I think), you suggested. If people are in alignment with using this format, I can take a crack at the necessary changes needed to seeders.py to parse the new file. Using this format, I believe both the labels and the food seeds could be parsed from the same file at the same time.

PR #3065, opened 2024-01-28
docs: Update example docker-compose files

boc-the-git:
## What type of PR is this?

- cleanup
- documentation

Add note to make it easier for people to find the latest version (making it less of an issue if we don't update the value in a timely manner post-release).

None.


Kuchenpirat (approved):
👍
We propably should build something that automatically updates the tag in the docs or maintaining that will be quite anoying.

boc-the-git:
> We propably should build something that automatically updates the tag in the docs or maintaining that will be quite anoying. 

Agreed. I might pursue that at some point

PR #3072, opened 2024-01-29
feat: On new release publish, update image tags in sample docker-compose files

boc-the-git:
## What type of PR is this?

- documentation
- feature

Modifies the release workflow to also update our documentation when a new release is published.

None.

## Testing

Via my fork, I created [this release](https://github.com/boc-the-git/mealie/releases/tag/vtest3), which triggered [this workflow](https://github.com/boc-the-git/mealie/actions/runs/7694934608/job/20966760514) (note that it's not the exact one I'm merging here, it's a modified version for testing purposes), which resulted in [this commit](https://github.com/boc-the-git/mealie/commit/f7fc8ee40d05379f1afe07b6db278110ea2ba5de).

The github workflow used for testing is screenshot, as I'll likely delete it before too long: 
image


Kuchenpirat (commented):
This will make this so much easier to keep the docs updated :)

Just to get what is going on here: 
This will update the files after the build release has finished. Will this then also trigger another nightly action? 
(which propably is fine, as those run pretty fast for doc changes)

boc-the-git:
> This will make this so much easier to keep the docs updated :)

🥳 🎉 

> This will update the files after the build release has finished.

Yes 👍 

> Will this then also trigger another nightly action? (which propably is fine, as those run pretty fast for doc changes)

No.
It didn't happen on my test and I was curious why (after your question). My belief is it's because we're committing directly against the mealie-next branch and not _pushing_ to it. [This](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#push) is the documentation for push, which we're using [here](https://github.com/mealie-recipes/mealie/blob/mealie-next/.github/workflows/nightly.yml#L4) to trigger the nightly.

I think _not_ building a new nightly is the most logical answer, because this is a change that doesn't impact the application in any way; it only changes the documentation site, which reads directly from the repo.

Kuchenpirat (approved):
Sound all good to me :)
This will certainly come handy in future releases 😊

PR #3084, opened 2024-01-30
fix: Update Group Slug When Updating Group

### LINKED ISSUE #2868: [BUG] - Updating group name does not update group slug
### First Check

When updating the group name (via group settings), the slug is not updated.
It is expected that the url (which uses the slug) is updated.
It would be nice that the old slug redirects to the new slug, but that might not be worth the effort (have to handle edge cases like changing another group to an old name).

### Steps to Reproduce

1) go to /admin/manage/groups
2) click on a group to update its settings
3) update the name
4) see that all links still use the old group name

/

### Mealie Version

Latest Release candidate

### Deployment

Docker (Linux)

_No response_

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- bug

_(REQUIRED)_

When you update the group name (e.g. through the UI) the slug isn't recalculated. Since the slug is used in the URL, this leads to weird-looking behavior (even though functionally everything still works fine).

_(REQUIRED)_

Fixes https://github.com/mealie-recipes/mealie/issues/2868

If your slug ends up being the same as an existing slug, we don't try to resolve the conflict, it just fails (as opposed to during create, where we add a counter to the name/slug). I think this is fine, since it's not preventing anything critical, it's just indicating that you should pick a different name (whereas during create time you can run into this a multitude of ways, including setup and migration, so resolution is important).

## Testing

Pytest for backend, manually for frontend

### DIFF
--- a/frontend/pages/admin/manage/groups/_id.vue
+++ b/frontend/pages/admin/manage/groups/_id.vue
@@ -79,6 +79,8 @@ export default defineComponent({
           window.location.reload();
         }
         group.value = data;
+      } else {
+        alert.error(i18n.tc("settings.settings-update-failed"));
       }
     }
 
--- a/frontend/pages/admin/manage/groups/_id.vue
+++ b/frontend/pages/admin/manage/groups/_id.vue
@@ -74,6 +74,10 @@ export default defineComponent({
 
       const { response, data } = await adminApi.groups.updateOne(group.value.id, group.value);
       if (response?.status === 200 && data) {
+        if (group.value.slug !== data.slug) {
+          // the slug updated, which invalidates the nav URLs
+          window.location.reload();
+        }
         group.value = data;
       }
     }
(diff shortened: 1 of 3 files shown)


Kuchenpirat (commented):
the only problem i see is when you try to use a name that is already used. 

It fails in the backend, but there is no indication at all why it is not working, or even that something isn't working

michael-genson:
Yeah we might want to raise an issue for that. Even before this PR that happens. At least in the meantime there's a log

frontend/pages/admin/manage/groups/_id.vue
```diff
@@ -74,6 +74,10 @@ export default defineComponent({
 
       const { response, data } = await adminApi.groups.updateOne(group.value.id, group.value);
       if (response?.status === 200 && data) {
+        if (group.value.slug !== data.slug) {
+          // the slug updated, which invalidates the nav URLs
+          window.location.reload();
+        }
         group.value = data;
       }
```
Kuchenpirat:
```suggestion
      if (response?.status === 200 && data) {
        if (group.value.slug !== data.slug) {
          // the slug updated, which invalidates the nav URLs
          window.location.reload();
        }
        group.value = data;
      }
      else {
        alert.error(i18n.tc("settings.settings-update-failed"));
      }
```

Maybe just add that, so the user has some sort of feedback

michael-genson:
Deal :)

Kuchenpirat (approved):
Deal 🤝

PR #3031, opened 2024-01-22
feat: Add Additional SMTP Headers to Decrease Spam Score

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- bug
- feature

## What this PR does / why we need it:

_(REQUIRED)_

Adds missing headers detailed in https://github.com/mealie-recipes/mealie/discussions/2993 (and https://github.com/mealie-recipes/mealie/issues/3028) to decrease our spam score

## Which issue(s) this PR fixes:

_(REQUIRED)_

Closes https://github.com/mealie-recipes/mealie/discussions/2993
Fixes https://github.com/mealie-recipes/mealie/issues/3028

PR #3094, opened 2024-02-01
Rectify email message ID, change multipart order

be-joeri:
## What type of PR is this?

- improvement

Small changes to the email sender updates done in https://github.com/mealie-recipes/mealie/pull/3031

- message ID was missing brackets (< >)
- the plain text part was added after the html part. The last part of a multipart message is considered best and is preferred, so if the mail client follows RFC 2046 (Gmail for example), it shows the text message instead of the HTML message.

No new issue was raised for this.

Fixes https://github.com/mealie-recipes/mealie/pull/3031

My first pull request ever, so please be gentle if I made a mistake in the flow somewhere :)

## Testing

Sent a test mail after the change, my spam filter no longer complains about the message ID and Gmail shows the html message instead of plain text.

### DIFF
--- a/mealie/services/email/email_senders.py
+++ b/mealie/services/email/email_senders.py
@@ -41,15 +41,15 @@ class Message:
         msg["From"] = f"{self.mail_from_name} <{self.mail_from_address}>"
         msg["To"] = to
         msg["Date"] = formatdate(localtime=True)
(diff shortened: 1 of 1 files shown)


Kuchenpirat (approved):
Hey, 
congrats to your first PR :)

It all looks great.

PR #3102, opened 2024-02-02
Allow overriding of some absolute paths using environment variables

litchipi:
## What type of PR is this?
- feature
> Small improvements for deployability on servers (specifically NixOS)

In the context of https://github.com/NixOS/nixpkgs/pull/278454 (adding this service as one of NixOS's pre-configured setting).
Because we need to be able to modify the paths, and because the source has to stay in read only, some hardcoded paths in the codebase are causing some troubles.

- Alembic config path is required because the we build the package without the `alembic` scripts in them, and won't be find as long as we don't point to its new location.

Allows to override some hardcoded paths using environment variables.

No issue on this repository

Meant for greater flexibility while not disrupting how things are working inside.

We can work with patches that modifies the code as we need, but it's always better to propagate upstream.

## Testing

Using the built package in NixOS, things still works well


hay-kot:
Why doesn't overriding "DATA_DIR" work currently? Isn't everything else immutable?

litchipi:
My bad, that's unnecessary :sweat_smile:

mealie/core/root_logger.py
```diff
@@ -9,7 +9,9 @@
 
 from .config import get_app_settings  # noqa E402
 
-LOGGER_FILE = DATA_DIR.joinpath("mealie.log")
+import os
```
mweinelt: Should probably go above `import sys` in L2.

mealie/db/init_db.py
```diff
@@ -87,7 +88,12 @@ def main():
             if max_retry == 0:
                 raise ConnectionError("Database connection failed - exiting application.")
 
-        alembic_cfg = Config(str(PROJECT_DIR / "alembic.ini"))
+        alembic_cfg_path = os.getenv("ALEMBIC_CONFIG_FPATH", default=str(PROJECT_DIR / "alembic.ini"))
```
mweinelt: Why `FPATH` here and `FILE` for the logger file?

litchipi:
Made the changes, also edited the PR message with the specific reasons why the changes are necessary in our cases

mealie/core/root_logger.py
```diff
@@ -9,7 +10,7 @@
 
 from .config import get_app_settings  # noqa E402
 
-LOGGER_FILE = DATA_DIR.joinpath("mealie.log")
+LOGGER_FILE = os.getenv("MEALIE_LOG_FILE", default=DATA_DIR.joinpath("mealie.log"))
```
hay-kot: Is this necessary if the log file is being written to the data directory? 

I have plans for a larger logging over-hall so the env variable will get removed once that happens, and you'll have much richer control over the logging implementation

Can you kick this can down the road until we get to that refactoring? Otherwise we'd break this configuration in the next release or two.

litchipi: No worries, we'll just use this change as a patch for now, and when the refactoring is done I'll adapt the setup :slightly_smiling_face:

litchipi:
Edited PR message to remove the logger part, as we'll not make it in this PR.

@hay-kot The `poetry run pytest` doesn't work on the branch `mealie-next`:
```
FAILED tests/unit_tests/services_tests/scheduler/tasks/test_post_webhook.py::test_get_scheduled_webhooks_filter_query - AssertionError: assert 0 == 4
```
So I cannot really merge as it's based on it. Ping me once this is resolved :slightly_smiling_face:

Kuchenpirat:
Seems this was just a flaky test.
Hadn't had one of those in a long time. 

All passed after a re-run.

litchipi:
@hay-kot I added a new change in exaclty the same spirit: `MODEL_PATH` for crfpp was absolute, I added an `os.getenv` which defaults to the absolute path.

Tell me if you think of other paths that are relative to the project's root and could benefit of it

litchipi:
(rebased over `mealie-next`)

hay-kot (approved):

PR #3135, opened 2024-02-09
dev: add docker-in-docker to dev container

Kuchenpirat:
## What type of PR is this?

- dev
- feature

This PR adds the ability to run docker commands from within the dev container console. 
This is needed to run `task docker:prod` to create  a production image and `task dev:services` to run a postgres db and mailpit

Uses the feature described here: https://github.com/devcontainers/features/tree/main/src/docker-in-docker

None

Rebuild of the dev container needed. 
[Docker-outside-of-Docker](https://github.com/devcontainers/features/blob/main/src/docker-outside-of-docker/README.md) would be also be an option which runs the container directly on the host. 

## Testing

build a local prod image and started the services.
Only tested on a Windows (win 11) machine.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -49,9 +49,6 @@
   "onCreateCommand": "sudo chown -R vscode:vscode /workspaces/mealie/frontend/node_modules && task setup",
   // Comment out to connect as root instead. More info: https://aka.ms/vscode-remote/containers/non-root.
   "remoteUser": "vscode",
-  // "features": {
-  //   "git": "latest"
-  // }
   "features": {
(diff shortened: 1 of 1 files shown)


.devcontainer/devcontainer.json
```diff
@@ -52,4 +52,9 @@
   // "features": {
```
boc-the-git: Minor.. Should probably remove the commented out features section, lines 52-54.
I'm pretty sure it exists only as an example of file structure, and is redundant after your change.

boc-the-git (changes requested):
Minor change, but will approve after

Kuchenpirat:
Have removed the comments.
Thanks 😊👍

boc-the-git (approved):

PR #3142, opened 2024-02-10
fix: password reset link not shown in admin pages

### LINKED ISSUE #3141: [BUG] - Cannot copy to clipboard over HTTP, so no way to access password reset link.
### First Check

Copy to clipboard buttons don't work when accessing Mealie via HTTP. Browser functionality is blamed for this, however I have no issues copying to clipboard over HTTP on any other web apps I run, such as Gitea and SnippetBox. This seems to only be an issue with Mealie.
This leads to the other issue: on the Admin User Management page the "Generate Password Reset Link" does not provide any visible link to copy, relying on the copy to clipboard button. This means there's no way for me to get the password reset link.

For the first part, Mealie should either support copying over HTTP, or make it clearer in the error message that copying over HTTP is not supported by Mealie rather than blaming the browser.
For the second part, the password reset link should appear somewhere on the page to allow for manual copying.

### Steps to Reproduce

1) Access mealie over a HTTP connection (not HTTPS)
2) Navigate to */admin/manage/users, and click on any user
3) Click the Generate Password Reset Link

Browser console logs this:
> Clipboard is currently not supported by your browser. Ensure you're on a secure (https) site.

Mealie container logs this:
> "GET /g/home/admin/manage/users/<user-id>/password-reset HTTP/1.1" 200 OK

### Mealie Version

v1.2.0
Build
0800a8d00a9fa154df40040304d46d4900d42e07 

### Deployment

Docker (Linux)

_No response_

Kuchenpirat:
## What type of PR is this?

- bug

Shows the password reset link in the GUI at the page `/admin/users/_id`. This is necessairy because if the users connects via http copyToClipboard is not availlable. Furthermore it makes it a bit clearer that a link has been generated instead of just adding a little copy button. 

Screenshots: 
|  | Before | After |
|--|--------|-------|
| not active | image | image |
| active | image | image |

Fixes #3141

None

## Testing
 Manual testing


boc-the-git (approved):
LGTM.

My understanding @Kuchenpirat is that despite the button wording "cancel", all that is happening is the URL is being hidden from the screen?
I'll allow you to merge after seeing this message 👍

Kuchenpirat:
Not exactly. It overwrites the reset link with an empty string. 

This is the same implementation we use for the invite link generation.

PR #3143, opened 2024-02-10
fix: Change release workflow to create a PR, rather than commit directly to branch

boc-the-git:
## What type of PR is this?

- bug
- cleanup
- documentation

I think you know by now..
Auto-updating our docs on new release publish. This implementation will create a PR with the change, rather than attempting to commit directly to the branch (which is restricted).

None.

## Testing

On my fork, [this actions run](https://github.com/boc-the-git/mealie/actions/runs/7854264175/job/21434807776), created [this PR](https://github.com/boc-the-git/mealie/pull/5) (and corresponding branch). All looks good.


.github/workflows/release.yml
```diff
@@ -60,11 +61,12 @@ jobs:
...
-          file-path: |
-            docs/docs/documentation/getting-started/installation/sqlite.md
-            docs/docs/documentation/getting-started/installation/postgres.md
-          commit-msg: "Change image tag, for release ${{ github.event.release.tag_name }}"
-          github-token: ${{ secrets.GITHUB_TOKEN }}
+          commit-message: "Change image tag, for release ${{ github.event.release.tag_name }}"
+          branch: "docs/newrelease-update-version-${{ github.event.release.tag_name }}"
+          delete-branch: true
+          base: mealie-next
+          title: "docs(auto): Change image tag, for release ${{ github.event.release.tag_name }}"
```
Kuchenpirat: LGTM 
Really nitpicky, but i would use _Update_ instead of _Change_. Will leave it up to you if you want to update (or change 😁) that, or just merge as is. 

```suggestion
          commit-message: "Update image tag, for release ${{ github.event.release.tag_name }}"
          branch: "docs/newrelease-update-version-${{ github.event.release.tag_name }}"
          delete-branch: true
          base: mealie-next
          title: "docs(auto): Update image tag, for release ${{ github.event.release.tag_name }}"
```

Kuchenpirat (dismissed):

.github/workflows/release.yml
boc-the-git: I had no preference between them to be honest, so went with the suggestion

boc-the-git: Looks like that gets rid of your approval, BTW

Kuchenpirat (approved):

PR #3145, opened 2024-02-10
fix: Taskfile Email Config

Kuchenpirat:
## What type of PR is this?

- bug

Mealie needs a SMTP_FROM_EMAIL to be able to send emails. This adds mealie@example.com as sender adress for testing purposes together with mailpit. 

None

## Testing

Tested sending an example email from the admin dashboard.


boc-the-git (approved):
LGTM

PR #3147, opened 2024-02-10
feat: email password reset link directly from admin dashboard

Kuchenpirat:
## What type of PR is this?

- feature

This adds the option to directly send an email with the password reset link to the email adress of the user. 

Most users will propably find the "Forgot Password" field on the login screen themselves, but some not so tech savy will just call their e.g. family member that hosts mealie and aks them to fix it. 

None

- This does currently not test whether email is enabled. But this does seem to be common for other buttons as well. 
- The token that is send per email is not the same as the one displayed in the dashboard, because the email function generates a new token. Both work. 

## Testing

Tested with the mailpit service. ~~Mealie does generate the link a bit false (port 8080 instead of 3000) but if the user has email configured correctly, this sould not fail as it uses the normal forgot-password api endpoint.~~


cmintey (approved):

boc-the-git:
@Kuchenpirat any chance you can add before and after images please?

boc-the-git:
The mail pit issue, does that relate just to the Dev container?

Kuchenpirat:
> The mail pit issue, does that relate just to the Dev container?

Yep, that is only mailpit. I just pushed another commit setting the BASE_URL in the taskfile which fixes that.

Kuchenpirat:
Screenshots: 

Before: 
image


After: 
image

boc-the-git (approved):
LGTM 🚀

PR #3157, opened 2024-02-11
fix: Pydantic Serialization Issues

michael-genson:
Well, it couldn't be _that_ easy to upgrade, could it?

## What type of PR is this?

_(REQUIRED)_

- bug

_(REQUIRED)_

In certain instances, Pydantic will bypass its `__init__` (mostly related to ORM). In a few instances we were overriding Pydantic's `__init__` to do some post-init validation. Thankfully, Pydantic V2 supports this out of the box with its `model_validator` when set to `mode="after"`.

I also noticed when adding recipes to the shopping list we were losing units for some reason. I'm not _quite_ sure what happened, but somewhere we were relying on Pydantic choosing the correct sub-model (`IngredientUnit` vs `CreateIngredientUnit`), and now it isn't anymore. So I fixed that by adding the id as an optional field so it always gets serialized when present.

_(REQUIRED)_

N/A

🙃

## Testing

Manually

### DIFF
--- a/mealie/schema/group/group_shopping_list.py
+++ b/mealie/schema/group/group_shopping_list.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 from datetime import datetime
 
-from pydantic import UUID4, ConfigDict, field_validator
+from pydantic import UUID4, ConfigDict, field_validator, model_validator
 from sqlalchemy.orm import joinedload, selectinload
 from sqlalchemy.orm.interfaces import LoaderOption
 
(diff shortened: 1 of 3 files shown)


mealie/schema/group/group_shopping_list.py
```diff
@@ -100,14 +100,15 @@ class ShoppingListItemOut(ShoppingListItemBase):
...
-        super().__init__(**kwargs)
-
+    @model_validator(mode="after")
+    def post_validate(self):
         # if we're missing a label, but the food has a label, use that as the label
         if (not self.label) and (self.food and self.food.label):
             self.label = self.food.label
             self.label_id = self.label.id
 
+        return self
```
michael-genson, line 110: Pydantic reassigns the instance for some reason, so yeah, we do actually have to `return self` here

hay-kot (approved):

PR #3160, opened 2024-02-12
fix: Pydantic Validation For Empty ID String

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- bug

_(REQUIRED)_

https://github.com/mealie-recipes/mealie/pull/3157 introduced a placeholder id field to fix an issue where Pydantic chose the wrong model in a type union. However, if it receives an empty string, instead of coercing that to `None`, it tries to coerce it to a `UUID` (which fails). This PR adds an explicit validator.

This issue presents itself when parsing a recipe, if you try to click the button to create a missing food/unit based on the parse result. I can't seem to find anywhere else where that happens (the data management page, for example, doesn't do that).

_(REQUIRED)_

Raised in Discord

I tried to poke around and see where else this might be an issue, but I couldn't find any other `id: UUID4 | None` situations that don't work as expected ¯\\\_(ツ)_/¯

## Testing

Manually

### DIFF
--- a/mealie/schema/recipe/recipe_ingredient.py
+++ b/mealie/schema/recipe/recipe_ingredient.py
@@ -37,6 +37,15 @@ class UnitFoodBase(MealieModel):
     description: str = ""
     extras: dict | None = {}
 
+    @field_validator("id", mode="before")
+    def convert_empty_id_to_none(cls, v):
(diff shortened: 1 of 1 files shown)


boc-the-git (approved):
LGTM 👍

PR #3163, opened 2024-02-12
fix: URL-Encode Postgres Password

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- bug

_(REQUIRED)_

There's a bug in Pydantic V2 that doesn't properly URL-encode Postgres passwords. They used to work fine in V1, for whatever reason. This specifically affects passwords with any of these characters, at a minimum: `,#/?` (see the linked ticket below).

~~This PR attempts to create the Postgres URL normally, and if it fails, attempts to URL parse it ourselves. I did it this way in case there's some unintended side-effect, I didn't want to bork any other users.~~ apparently using urllib.parse.quote_plus is the right way to do it: https://stackoverflow.com/questions/1423804/writing-a-connection-string-when-password-contains-special-characters

The related Pydantic bug: https://github.com/pydantic/pydantic/issues/8061

_(REQUIRED)_

N/A, raised in Discord

Hopefully this covers all use-cases, but worst-case users can choose passwords without problematic characters (though I fully understand why we want to avoid that, since postgres does support them).

## Testing

Added a test

### DIFF
--- a/mealie/core/settings/db_providers.py
+++ b/mealie/core/settings/db_providers.py
@@ -45,28 +45,15 @@ class PostgresProvider(AbstractDBProvider, BaseSettings):
     @property
     def db_url(self) -> str:
         host = f"{self.POSTGRES_SERVER}:{self.POSTGRES_PORT}"
-        try:
-            url = PostgresDsn.build(
+        return str(
+            PostgresDsn.build(
                 scheme="postgresql",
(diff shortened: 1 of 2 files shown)


boc-the-git:
LGTM, good stuff @michael-genson

boc-the-git (approved):

PR #3170, opened 2024-02-13
fix: Make Mealie Alpha Migrations More Fault Tolerant

### LINKED ISSUE #3073: [BUG] - JSON errors in a single recipe cause the entire Pre-v1 migration to fail
### First Check

When migrating from pre-v1, if one recipe has malformed JSON it causes the entire migration to fail. We should instead skip over the bad recipes.

This is due to the Mealie Alpha `_migrate` code here:
```python
for x in temp_path.rglob("**/recipes/**/[!.]*.json"):
  if (y := MigrationReaders.json(x)) is not None:  # this can fail
    ...
```

### Steps to Reproduce

1) Create a Mealie Alpha backup
2) Bork one recipe's JSON (e.g. delete the content of the file)
3) Attempt to run the migration

`JSONDecodeError: Expecting value: line 1 column 1 (char 0)`

### Mealie Version

Nightly, tested in the dev container

### Deployment

Docker (Linux)

_No response_

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- bug

_(REQUIRED)_

This adds some additional `try except` logic when migrating from Mealie alpha, so bad recipes don't cause the entire migration to fail. It also restores some broken report functionality (probably broken from the Pydantic upgrade, but who knows) and updates the migration tests to catch if it breaks again.

_(REQUIRED)_

Fixes https://github.com/mealie-recipes/mealie/issues/3073

## Testing

Added a test for bad data handling, and also updated the existing tests to verify the report entries are actually being generated


mealie/services/migrations/_migration_base.py
```diff
@@ -109,6 +111,7 @@ def _save_all_entries(self) -> None:
         if not is_success and not is_failure:
             self.report.status = ReportSummaryStatus.partial
 
+        self.report.entries = new_entries
```
michael-genson, line 114: I don't know why this used to work. Since we're saving the report again without updating the report's entry list, it should delete all the report entries that we just created. It's doing that _now_, but I don't know why it wasn't doing that before

boc-the-git (approved):
LGTM

PR #3196, opened 2024-02-19
fix: Bump SQLAlchemy

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- bug

_(REQUIRED)_

Bumps SQLAlchemy and fixes a few of our model definitions. I'm not really sure how these used to work, since they're clearly wrong.

_(REQUIRED)_

N/A


mealie/schema/group/group_shopping_list.py
```diff
@@ -225,10 +225,14 @@ class ShoppingListUpdate(ShoppingListSave):
 
 
 class ShoppingListOut(ShoppingListUpdate):
-    recipe_references: list[ShoppingListRecipeRefOut]
-    label_settings: list[ShoppingListMultiPurposeLabelOut]
+    recipe_references: list[ShoppingListRecipeRefOut] = []
+    label_settings: list[ShoppingListMultiPurposeLabelOut] = []
```
hay-kot, line 229: I don't think you can use the array like this? Won't it always be a reference to the same array?

mealie/db/models/group/shopping_list.py
```diff
@@ -124,15 +124,15 @@ class ShoppingList(SqlAlchemyBase, BaseMixins):
...
         ShoppingListItem,
         cascade="all, delete, delete-orphan",
         order_by="ShoppingListItem.position",
         collection_class=ordering_list("position"),
     )
 
-    recipe_references: Mapped[ShoppingListRecipeReference] = orm.relationship(
-        ShoppingListRecipeReference, cascade="all, delete, delete-orphan"
+    recipe_references: Mapped[list[ShoppingListRecipeReference]] = orm.relationship(
+        ShoppingListRecipeReference, cascade="all, delete, delete-orphan", uselist=True
```
hay-kot: how come uselist=True is added to some but not all of them?

mealie/schema/group/group_shopping_list.py
michael-genson: Normally that's true, but Pydantic does some special magic and makes it safe: https://stackoverflow.com/a/63808835

mealie/db/models/group/shopping_list.py
michael-genson: Let me remove those, originally I thought it made a difference but that was before I noticed the mapped attribute type was wrong. It's unnecessary when using `Mapped` on a collection: https://docs.sqlalchemy.org/en/20/orm/relationship_api.html#sqlalchemy.orm.relationship.params.uselist

mealie/schema/group/group_shopping_list.py
hay-kot: Thought this was in the models folder, ignore me! :D

boc-the-git (approved):
LGTM

Issue #1164, opened 2022-04-20
[v1.0.0b] [Task] - Initial Setup "Wizard"


hay-kot:
### What is the problem this task addresses?

Mealie currently uses default credentials and default initialization methods to setup the database. This can be confusing and lead to excess issues or users looking for help in discord. 

### Proposed/Possible Solution(s)?

Mealie should provide a "Wizard" like setup that, on first run, will require the user to do the setup requirements.

- [ ] Create new  superuser user (or update default)
- [ ] Create new group (or update default)
- [ ] Choose / set group settings. 
- [ ] Initialize / seed data items taking into account the users language
- [ ] ???

Open to feedback or ideas, leave a comment below with ideas.

stroodle96:
Would be nice to be able to configure email setup in the wizard so you don't have to use environment variables. And for setting timezones as well. 

And maybe setting language. So you could remove that from the sidebar menu.

fluaten:
> And maybe setting language. So you could remove that from the sidebar menu.

A **setting language** per user could be practical when the browser is set to another language than the one you want mealie translate in.

I have to set the language each time I log in.. Mainly because I clean my cookie and cache each time I close the browser.

hay-kot:
Closing, I think the current setup is sufficient and the work here is pretty significant.

PR #3204, opened 2024-02-20
feat: First Time Setup Wizard

### LINKED ISSUE #2166: [v1.0.0b] [Task] Prompt User For Action on Login (Reset Password)

Currently there is no way to force a user to do anything on initial login. Things like forced password resets are not a thing. 

The Mealie API should provide a framework to alert the frontend that a User must complete some "action" this should be generic enough to be reused for different things in the future but currently they would be for at least:

1. Password Resets / Force New Password
2. Initial Guided Setup for Mealie #1164 

I see this being implemented in 2 phases

1. API Setup, set the initial API that checks a User for required actions
2. Frontend redirect logic based on backend, this would likely be done through some kind of enum "actions". For example user get's payload with action `password-reset-required` which redirects the user to the reset password route. 

I have not looked much into how other projects implement this, so it may be worth taking some time to investigate how this kind of thing is normally done.

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- feature

_(REQUIRED)_

This PR adds a first time setup wizard. It was first proposed in https://github.com/mealie-recipes/mealie/issues/1164, but getting this ready before V1 would've been too much. Now that we're past V1, we're in a good place to add this sort of thing.

When a user logs-in as the default user (`changeme@example.com`), or they go to the root URL while logged-in with the default user (index), they are redirected to `/admin/setup`. Here they perform a few tasks:
1. Change their email, username, and password
2. Toggle public access
3. Toggle seed data

Once that's complete, they're given a list of places to check out (e.g. migrations).

Here's a full walkthrough of the new setup wizard:

2024-02-20_16h24_00

I think there's more that can be done here to help new setups, but as-is I'm pretty happy with how it turned out, and adding-on to it should be pretty straight-forward.

_(REQUIRED)_

N/A, idea comes from https://github.com/mealie-recipes/mealie/issues/1164

Fixes 2166

A few implementation notes:
- In order to avoid having the user re-type their password (since they're going to be using the default password anyway) I added a sort of workaround to the update password route on the backend. If the user trying to update their password is the default user, and they don't provide the current password, we assume they meant to provide the default password. I added test cases for this
- If the site is in demo mode, the setup page is skipped (as if you're not logging-in as the default user, even if you are)
- I refactored the new user registration form so I didn't have to duplicate it here. Hopefully I didn't make too much of a mess of things, but that was probably the hardest part of this PR, so take a look if you can. I did end-to-end testing to confirm registration still works as expected. I don't know if I used the form composable correctly, so definitely give that an extra look over

## Testing

Added a backend test for the update password change, and did end to end testing for new instances as well as registrations (since that draws from the same code)


mealie/routes/users/crud.py
```diff
@@ -70,7 +70,12 @@ def update_password(self, password_change: ChangePassword):
             raise HTTPException(
                 status.HTTP_400_BAD_REQUEST, ErrorResponse.respond(self.t("user.ldap-update-password-unavailable"))
             )
-        if not verify_password(password_change.current_password, self.user.password):
+        if (not password_change.current_password) and self.user.is_default_user:
+            # this might be from the setup wizard, which doesn't require the user's current password
+            password_change.current_password = self.settings._DEFAULT_PASSWORD
```
hay-kot: Couldn't anyone always change the default users password in this case? Because we're only checking that it's the default user and no current_password was provided. 

Could we just provide the default password as in the request body and leave this code unchanged? 

Maybe I'm missing something

hay-kot (changes requested):
Pulled down and tested, everything looks good to me besides my one comment about verifying the current password.

mealie/routes/users/crud.py
michael-genson: > Couldn't anyone always change the default users password in this case?

Yes, but we show the default password at login anyway so we can assume all visitors know the default username and password (until it's changed, because once the email changes it's no longer the default user)

> Could we just provide the default password as in the request body and leave this code unchanged?

...that's a much better solution, lol. I think I had a reason for not doing that but I can't remember what it is, so maybe I'm crazy

michael-genson:
It's lost in the merge commit but I removed the backend password stuff (I just let the OIDC PR overwrite the changes since there was a conflict there anyway)

hay-kot (approved):

PR #3032, opened 2024-01-22
feat: Add Ingredient Sections To Copied Ingredients

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- feature

## What this PR does / why we need it:

_(REQUIRED)_

When copying the recipe ingredients to your clipboard, the ingredient sections are dropped. This adds them in.

Before:
```
1 lb leg of lamb boneless, chopped into one inch cubes
1 tbsp cornstarch
1/2 tsp salt
1 1/2 tbsps olive oil
1/2 tbsp olive oil
1/2 tsp salt
3/4 cup onion diced
2 tsps garlic minced
2 tsps paprika
1 tsp cumin
1/4 tsp allspice
3/4 cup red wine (such as cabernet sauvignon)
1 1/2 cups beef broth
1 cup mushroom sliced
2 small yukon gold potatoes cut into 1 1/2 inch cubes
1 large carrot sliced
1 tbsp tomato paste
3 sprigs fresh thyme
1 bay leaf
1 tbsp cornstarch
```

After:
```
[For the Lamb]
1 lb leg of lamb boneless, chopped into one inch cubes
1 tbsp cornstarch
1/2 tsp salt
1 1/2 tbsps olive oil

[For the Veggies]
1/2 tbsp olive oil
1/2 tsp salt
3/4 cup onion diced
2 tsps garlic minced
2 tsps paprika
1 tsp cumin
1/4 tsp allspice

[For the Broth]
3/4 cup red wine (such as cabernet sauvignon)
1 1/2 cups beef broth
1 cup mushroom sliced
2 small yukon gold potatoes cut into 1 1/2 inch cubes
1 large carrot sliced
1 tbsp tomato paste
3 sprigs fresh thyme
1 bay leaf
1 tbsp cornstarch
```

## Which issue(s) this PR fixes:

_(REQUIRED)_

N/A, was going to raise an issue for it, but opened this PR instead

## Testing

_(fill-in or delete this section)_

Manually

PR #3226, opened 2024-02-27
feat: sort by labels in shopping list copy if labels toggled

### LINKED ISSUE #3227: [BUG] - Copying shoppings lists results in blank output
### First Check

When copying shopping list items (either as text or markdown), each line in the output is just a blank string. Having had a look at the `copyListItems` function, it uses the `display` property on each list item to populate the copy output. The display property seems to always come back as blank in the `nightly` build of Mealie but *not* in the `latest` image, so it must be a recent regression. Looks like a change has been made on the backend side which is breaking this display property.

### Steps to Reproduce

Go to the demo mealie site and log in. Go to [any shopping list](https://demo.mealie.io/shopping-lists/5ad01392-93aa-4fd9-939d-5fc7af1a4fc8) with some items and click "Copy as text". The output will be a set of blank lines equal to the number of items in the list.

No logs, UI only issue

### Mealie Version

Nightly, https://github.com/mealie-recipes/mealie/commit/5ac7645350ba580f212fbd1c0d29eb96c1317076

### Deployment

Docker (Linux)

Docker compose on Linux with Postgres DB

ollywelch:
## What type of PR is this?

- feature

This is a simple quality of life feature when copying shopping lists. Currently the lists always get copied in the order that is displayed when label sorting is off. So, when you toggle label sorting on, the copy function ends up copying in a different order to what is displayed. This change deals with the case when label sorting is on.

### Updates
- Updated the `copyListItems` function to unpack the values of the `itemsByLabel` mapping into the `items` to copy if label sorting is toggled on. Otherwise, it will have the same behaviour as before. 

Fixes https://github.com/mealie-recipes/mealie/issues/3227

Go easy, not touched Vue before now :)

Also not sure if I need an issue creating for this, can create one and amend my commit msg to include it if needed.

## Testing

Ran the UI, made a shopping list with labels, then checked the copy output with label sorting toggled on and off.


ollywelch:
Actually discovered another issue while testing this, looks like the copy function is using the "display" property of the shopping list item which is always coming back blank - this is only an issue in the nightly build and not latest so must be a recent API change

ollywelch:
Raised https://github.com/mealie-recipes/mealie/issues/3227, will see if I can also fix this in my PR

michael-genson:
I haven't had the chance to fully test it out yet, but generally looks good! I went ahead and fixed that `post_validate` issue by renaming all the `post_validate` methods (I probably should've done that originally).

I should be able to take a closer look at this later today

michael-genson:
Would it make sense to also add the labels into the copied shopping list? I feel like if you want to keep them in label order, you probably also want the labels. Something like:

[Produce]
apples
bananas

[Meat]
1 lb chicken thighs

etc. I think we should at least have it in the markdown copy mode, though I'd argue you'd want it in the plaintext one too. I'm thinking something like this: https://github.com/mealie-recipes/mealie/pull/3032

ollywelch:
Nice one with just renaming the methods, that's a bit neater than calling the same `post_validate` method from the super class :+1: 

With copying the headings, I think you're right. I've had to re-engineer it and not use the `copyMarkdownCheckList` function, rather just use the `copyPlain` function regardless and do the markdown formatting up front. But have tested and it works well

frontend/pages/shopping-lists/_id.vue
```diff
@@ -337,21 +337,46 @@ export default defineComponent({
...
+      const text: string[] = [];
 
-      if (!items) {
-        return;
+      if (preferences.value.viewByLabel) {
+        // if we're sorting by label, we want the copied text in subsections
+        Object.entries(itemsByLabel.value).forEach(([label, items]) => {
+
+          // add an appropriate heading for the label depending on the copy format
+          text.push(formatCopiedLabelHeading(copyType, label))
```
michael-genson, line 351: Can we add an empty line before each header? Unless it's the first header. So it renders like this:

```
[Label 1]
- [ ] item 1
- [ ] item 2

[Label 2]
- [ ] item 3
- [ ] item 4
```

But skip the first index, so something like:

```js
Object.entries(itemsByLabel.value).forEach(([label, items], idx) => {
  if (idx) {
    text.push("");
  }
  ...
}
```

michael-genson:
One small change and I think this is good to go. Thanks!

michael-genson (changes requested):
Sorry, I just realized I left this as pending!

frontend/pages/shopping-lists/_id.vue
ollywelch: Just made the update, I initially tried this yesterday but couldn't understand why I was getting two blank lines by pushing `"\n"` to the array :facepalm: 

Thanks for the guidance :smile:

michael-genson (approved):
Looks great, thank you!

PR #3232, opened 2024-02-28
feat: Recipe Instructions, add buttons for move to top or bottom

boc-the-git:
## What type of PR is this?

- feature

Adds buttons to the recipe instructions to "move to top" or "move to bottom". When you've got 20 steps in a recipe and need to move things around, drag and drop can get very tedious!
Also introduces some dividers into the menu.

Before:
image 

After:
image

None.

## Testing

Manual.


frontend/components/global/BaseButtonGroup.vue
```diff
@@ -10,9 +10,12 @@
...
         </template>
         <v-list dense>
-          <v-list-item v-for="(child, idx) in btn.children" :key="idx" dense @click="$emit(child.event)">
-            <v-list-item-title>{{ child.text }}</v-list-item-title>
-          </v-list-item>
+          <template v-for="(child, idx) in btn.children">
+            <v-list-item :key="idx" dense @click="$emit(child.event)">
+              <v-list-item-title>{{ child.text }}</v-list-item-title>
+            </v-list-item>
+            <v-divider v-if="child.divider" :key="`divider-${idx}`" class="my-1"></v-divider>
```
michael-genson, line 17: Would we be able to switch the component to put a divider on top, instead of on bottom? Only for consistency, because I know there are other components we have that behave that way

michael-genson (changes requested):

frontend/components/global/BaseButtonGroup.vue
boc-the-git: @michael-genson just confirming as mentioned on Discord that this is an area where we already don't have consistency. A couple examples where the divider is on the bottom (which I had used as my basis):

https://github.com/mealie-recipes/mealie/blob/a475afd570bddde1e1dda20f68942f794a80d96b/frontend/components/global/BaseOverflowButton.vue#L23

https://github.com/mealie-recipes/mealie/blob/a475afd570bddde1e1dda20f68942f794a80d96b/frontend/pages/register/index.vue#L227

michael-genson: Thanks for taking a look!

michael-genson (approved):

PR #3249, opened 2024-03-05
Fix broken docs links and nested Markdown list

eneiss:
## What type of PR is this?

- documentation

`/docs/docs/contributors/developers-guide/code-contributions.md`: 
- Fix Github Flow broken links: [current links](https://guides.github.com/introduction/flow/index.html) are no longer valid and redirect to Github's home page. I replaced them with what I believe is their [new version](https://docs.github.com/en/get-started/using-github/github-flow).
- Fix Facebook's Draft broken link: **I'm not too sure about this one** since the [current link](https://github.com/facebook/draft-js/blob/a9316a723f9e918afde44dea68b5f9f39b7d9b00/CONTRIBUTING.md) points at a specific version of the CONTRIBUTING.md of Facebook's draft-js repo, and this version is no longer available.
Instead, I chose to target the link to [the version on the main branch](https://github.com/facebookarchive/draft-js/blob/main/CONTRIBUTING.md) of the now-archived repository. Since it is now read-only, this shouldn't be an issue. If you believe I should do it differently, please let me know :)
- Fix a Mardown list that should've been nested according to the indentation in the source file (it also makes more sense this way), but is currently rendered at the top level instead.

No linked issue, let me know if I should create one!

## Testing

I tested the code locally (dev container) with `mkdocs serve` to make sure the links and nested list were rendering as expected.

### DIFF
--- a/docs/docs/contributors/developers-guide/code-contributions.md
+++ b/docs/docs/contributors/developers-guide/code-contributions.md
@@ -5,8 +5,8 @@
 ## We Develop with Github
 We use github to host code, to track issues and feature requests, as well as accept pull requests.
 
-## We Use [Github Flow](https://guides.github.com/introduction/flow/index.html), So All Code Changes Happen Through Pull Requests
-Pull requests are the best way to propose changes to the codebase (we use [Github Flow](https://guides.github.com/introduction/flow/index.html)). We actively welcome your pull requests:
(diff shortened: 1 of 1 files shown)


boc-the-git:
Thanks for your first PR @eneiss !

We're fine without an issue, so no worries there.

I can't actually look at the change properly now (and understand the markdown rules) but I should get to it at least within 48 hours, if someone doesn't see it sooner.

boc-the-git (approved):
LGTM

PR #3284, opened 2024-03-11
feat: Timeline Filters

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- feature

_(REQUIRED)_

Adds filters to the recipe timeline:
image

Filters are stored in user preferences so they're persisted. When one or more filters are selected, there's a badge on the button:
image

There are also a few miscellaneous improvements:
- Translated some previously un-translated text
- Replaced the message for zero timeline items to make more sense in all contexts
- Updated the Recipe Explore filters with a few tweaks I made to improve the timeline filters
- Extracted the recipe timeline enum data into a composable so I don't need to keep writing switch statements

_(REQUIRED)_

Closes https://github.com/mealie-recipes/mealie/discussions/3250

## Testing

Manually poking around


frontend/components/Domain/Recipe/RecipeExplorerPage.vue
```diff
@@ -67,12 +67,16 @@
               <v-list>
                 <v-list-item @click="toggleOrderDirection()">
                   <v-icon left>
-                    {{ $globals.icons.sort }}
+                    {{
+                      state.orderDirection === "asc" ?
+                      $globals.icons.sortAscending : $globals.icons.sortDescending
+                    }}
```
Kuchenpirat: The icons should be the other way around. 
Tho, i have to say, looking at this i have a kinda hard time making sense of the icon. Why is the ascending icon having an arrow down? 🧐 Especially because the calendar sorting icon used on the timeline does it the other way. 

```suggestion
                    {{
                      state.orderDirection === "asc" ?
                      $globals.icons.sortDescending : $globals.icons.sortAscending
                    }}
```

frontend/components/Domain/Recipe/RecipeTimeline.vue
```diff
@@ -3,9 +3,52 @@
...
+          <template #activator="{ on, attrs }">
+            <v-badge :content="filterBadgeCount" :value="filterBadgeCount" bordered overlap>
+              <v-btn fab small color="info" v-bind="attrs" v-on="on">
+                <v-icon> {{ $globals.icons.filter }} </v-icon>
+              </v-btn>
+            </v-badge>
+          </template>
+          <v-card>
+            <v-list>
+              <v-list-item @click="reverseSort">
```
Kuchenpirat, line 17: I am not sure i like putting the sorting functionality behind the filter icon too much. 
But i can understand that it propaply does not get used too much, especially for long time users. Seing stuff from like two years ago is quite the unusual use case.

frontend/components/Domain/Recipe/RecipeTimeline.vue
```diff
@@ -31,7 +74,7 @@
     </div>
     <v-card v-else-if="!loading">
```
Kuchenpirat: I would add some top margin here so top of the card does not lie on top of the divider. 

```suggestion
    <v-card v-else-if="!loading" class="mt-2">
```

Kuchenpirat (commented):
This looks pretty good 🚀

One bug i have found that if you deselect the filter Checkboxes quite quickly it cannot follow. (Image below)
The checkboxes show everything deselected but the badge shows only one filter, and that is also how the timelime behaves as events are still displayed. The severity of this might increase when the user uses a slower system/connection, or has a lot of timeline events. 

image

frontend/components/Domain/Recipe/RecipeExplorerPage.vue
michael-genson: Should they be the other way around? The way I understand it is that it's showing the current state, and you want to click on it to reverse the state. In which case it should match:

image

frontend/components/Domain/Recipe/RecipeTimeline.vue
michael-genson: I had it this way mostly to match what we have on the recipe explore page. But I could separate these out into two buttons instead of one. Thoughts on that, vs how we have the recipe explore page?

I do agree though, it's not a setting I imagine will change often

Kuchenpirat: I think until we have something like a date-range selector for the timeline it makes no sense having a dedicated button for this.

frontend/components/Domain/Recipe/RecipeExplorerPage.vue
Kuchenpirat: The current state is 'Descending' right? 

So having the descending icon beside 'Sort Ascending' is confusing (at least to me)

michael-genson: Oh FML I confused myself. You are correct, we are currently descending, so we should flip the icon. I was thinking we were ascending

Kuchenpirat: Those icons are absolutely confusing 😅

michael-genson:
Pretty sure the bug you saw with the checkbox being out of sync is fixed. Let me know if you're able to reproduce. I noticed it only got out of sync when clicking on the checkbox (vs clicking on the label) so try clicking on both, if you can

Kuchenpirat (approved):
Looks all good 👍
Can't reproduce the bug anymore. Neither by using the checkboxes, nor the labels themselves.

PR #3305, opened 2024-03-12
fix: Limit shopping list owners to current group

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- bug

_(REQUIRED)_

The shopping list owner added in https://github.com/mealie-recipes/mealie/pull/3213 adds a setting on the shopping list to update the owner. This API call actually uses the admin route erroneously, so only admins can update the owner. Furthermore, it shows users from _all_ groups, not just the current group.

This PR adds a new route to fetch only the users belonging to the current user's group, and loosens permissions so that any user can use this API (rather than only admins).

_(REQUIRED)_

N/A

We should probably update the frontend's user API to prevent this from happening; the user API shouldn't be able to accidentally access the admin API routes, but due to how we've structured the user controller on the backend (the non-admin routes use the same base URL) it's a bit messy when using our generic frontend client. Ideally these are two separate base URLs (e.g. `/users` and `/admin/users` but since this would introduce breaking changes it's probably best to just work around it.

## Testing

Added backend tests and verified proper behavior on the frontend using a non-admin user.


mealie/routes/users/crud.py
```diff
@@ -56,6 +58,18 @@ def delete_user(self, item_id: UUID4):
 
 @controller(user_router)
 class UserController(BaseUserController):
+    @user_router.get("/group-users", response_model=UserPagination)
+    def get_all_group_users(self, q: PaginationQuery = Depends(PaginationQuery)):
+        """Returns all users from the current group"""
+
+        response = self.repos.users.by_group(self.group_id).page_all(
+            pagination=q,
+            override=UserOut,
```
hay-kot: Should we use a different type here instead of UserOut? It contains many fields that may be sensitive?

michael-genson: Yeah that makes sense, I can add something like UserSummary

hay-kot (approved):

PR #3306, opened 2024-03-12
fix: Only call store APIs once

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- bug

_(REQUIRED)_

Currently our frontend stores (e.g. foods store) shares the store state globally so changes in one component effect another (e.g. if a food is created, it should appear on other components). What's not shared, however, is the loading variable. While in most situations this doesn't matter much, when initializing the store every single component makes an API call to populate the store data (and all of them overwrite each other).

This is most egregious on the recipe editor. If, for instance, you have 20 ingredients, you're making 20 API calls to the food and unit stores and serializing _the entire store_ 20 times. Here is a snapshot of API calls from simply clicking edit _once_ on a single recipe with 20 ingredients (I can't fit the whole thing in a reasonably-sized screenshot):

image

This PR adds a shared loading state so the API is only called once per store. Here is a snapshot of the same recipe with these changes (this time with every API call in the image):

image

_(REQUIRED)_

N/A

Interestingly, while we did have a "loading" variable already, it wasn't being set correctly (it was set outside of the async function so it toggles immediately back to `false`) and it wasn't ever being referenced, it was only ever declared and updated. However, even _if_ we fixed all of those things, it still wasn't shared, so each instance of the store had its own loading value.

## Testing

Other than the above screenshots, I went and touched all the other stores (e.g. the data management page) and confirmed it works as expected. I can't see any side-effects of a shared loading state since the store _always_ uses the default params (no filters, return all items without pagination). If we used variable params this wouldn't work without some modification.

### DIFF
--- a/frontend/composables/partials/use-actions-factory.ts
+++ b/frontend/composables/partials/use-actions-factory.ts
@@ -37,6 +37,7 @@ export function usePublicStoreActions<T extends BoundT>(
     loading.value = true;
     const allItems = useAsync(async () => {
       const { data } = await api.getAll(page, perPage, params);
+      loading.value = false;
 
       if (data && allRef) {
         allRef.value = data.items;
@@ -49,7 +50,6 @@ export function usePublicStoreActions<T extends BoundT>(
       }
     }, useAsyncKey());
 
-    loading.value = false;
     return allItems;
   }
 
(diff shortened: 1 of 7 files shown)


hay-kot (approved):
LGTM

PR #3311, opened 2024-03-14
fix: remove deprecated lifecycle and consolidate startup actions

hay-kot:
## What type of PR is this?

_(REQUIRED)_

- fix/chore

The original event style startups were deprecated in FastAPI. This removes the old style and adds the new style. 

As a bonus it also removes the calling of init_db on it's own and pulls it into the startup logic. I couldn't find a reason not to do this, but it could use another set of eyes to make sure I'm not missing anything here.

N/A - dependency issues. Fixes log errors

## Testing

- [x] Local startup testing
- [x] Tested via `docker:prod` command and observed migrations ran and were logged.


michael-genson (approved):
Agreed with merging init_db into the startup, LGTM

PR #3323, opened 2024-03-16
OIDC Docs Updates

cmintey:
## What type of PR is this?

- documentation

Cleans up and clarifies some wording around OIDC enviornment variables.
Removes the examples from the docs and adds a link to examples in the GitHub discussions to make it easier for other users to add/update/comment specific configurations


docs/mkdocs.yml
```diff
@@ -41,7 +41,8 @@ markdown_extensions:
       custom_checkbox: true
   - admonition
   - attr_list
-  - pymdownx.tabbed
+  - pymdownx.tabbed:
+      alternate_style: true
```
boc-the-git, line 45: @cmintey can you provide some information on what this change is intending to address please?

Other than that, this PR looks fine to me 👍

cmintey: When I first started the PR I wanted to have tabs in the document to display configuration for different versions and I found that the tabs were broken and we needed to add this configuration.

I've since removed the docs that used the tabs, but there are other usages of them that were also broken, so that should be fixed with this

boc-the-git (approved):

PR #3364, opened 2024-03-24
fix: Add 127.0.1.1 to exclude list

boc-the-git:
## What type of PR is this?

- bug
- cleanup

Also exclude 127.0.1.1 as a potentially unsafe address.

None.

## Testing

N/A


michael-genson (commented):
Should we just add a regex and include all `127..*` addresses?

boc-the-git:
Hayden's PR addresses this and more, so closing.

PR #3368, opened 2024-03-24
security: gh security recs

hay-kot:
## What type of PR is this?

_(REQUIRED)_

- security
- bug

Addresses several security concerns raised by external audit/review

1. All user HTTP requests now use a custom Transport class that applies a timeout and blocks known internal address spaces
2. Note that you can't simple whitelist specific IP's, you must resolve the domain name first and then check the address space of the resolved IP. This prevents attackers from creating a DNS entry that resolves to a local address space and using that url to attack a Mealie instance.

Also, note that these are mitigations, they are not 100% fixes. 

- Closes #3364

I've tested locally and all the automated testing should pass. If possible, I would like another reviewer to pull down these changes and attempt to 

1. Scrape a few recipes from different domains
2. Use the Scrape Image functionality to pull external images into Mealie from a few different domains.

## Testing

See Above

### DIFF
--- a/mealie/pkgs/safehttp/transport.py
+++ b/mealie/pkgs/safehttp/transport.py
@@ -46,8 +46,6 @@ class AsyncSafeTransport(httpx.AsyncBaseTransport):
         netloc = request.url.netloc.decode()
         if ":" in netloc:  # Either an IP, or a hostname:port combo
             netloc_parts = netloc.split(":")
-            if len(netloc_parts) > 1:  # netloc of username:password@hostname:port not supported
-                raise InvalidDomainError(f"unsupported request format: {request.url} -> {ip}")
 
             netloc = netloc_parts[0]
 
--- a/mealie/pkgs/safehttp/transport.py
+++ b/mealie/pkgs/safehttp/transport.py
@@ -1,10 +1,7 @@
 import ipaddress
 import socket
-from urllib.parse import urlparse
 
 import httpx
-
-__http_timeout = 15
 
 
 class ForcedTimeoutException(Exception):
@@ -23,14 +20,12 @@ class InvalidDomainError(Exception):
     ...
 
 
-def resolve_ip_with_socket(url):
+def resolve_ip_with_socket(domain_name: str):
     """
     Resolve the IP address of a given URL. If the URL is invalid,
     return None.
     """
     try:
-        parsed_url = urlparse(url)
-        domain_name = parsed_url.netloc
         ip_address = socket.gethostbyname(domain_name)
         return ip_address
     except (socket.gaierror, ValueError):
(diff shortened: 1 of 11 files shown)


boc-the-git (changes requested):
The code changes all look good.
In testing the happy path, everything went well (i.e. I can still import recipes and images from legitimate internet sites.

In testing the negative path (trying to access things on my internal network), things came unstuck.

This first series of images is when accessing my Vikunja instance via the URL/reverse proxy (note I installed ping in the dev container to confirm what it's resolving to):
image
image
image

Then for a further test, I popped the IP/port in directly, avoiding the reverse proxy and any DNS resolution:
image

@hay-kot I'm interested in whether you can replicate this scenario.

I dug a bit deeper, and wanted to get some evidence of what was happening, which this code change and error message shows.. effectively the IP is `None`, rather than `192.168.x.y` as I'd expect.
image

p0lycarpio:
Along the same lines as the `ALLOW_SIGNUP` variable which was set to `False`, I suggest setting all groups to private by default when they are created. This is an additional security measure and also when using Mealie for the first time the user would be redirected to the login/welcome page instead of the empty group.
What is your opinion ?

The change would be made in the [group_preferences.py](https://github.com/mealie-recipes/mealie/blob/7f596c653bdeaaee64efd5c6ec5f5f0a3108c2f4/mealie/schema/group/group_preferences.py#L9) schema. I don't know the impacts though.

boc-the-git:
@p0lycarpio do you want to raise a feature request for that (or discuss it on Discord)?
We wouldn't likely tackle that under this PR.

That's more a _privacy issue_ than a _security issue_, if you take my meaning.
I don't have strong feelings on it one way or the other; other people might.

hay-kot:
@boc-the-git should be good now. Thanks to some folks on discord it's 99% less bad too!

mealie/pkgs/safehttp/transport.py
```diff
@@ -0,0 +1,80 @@
...
+
+        # validate the request is not attempting to connect to a local IP
+        # This is a security measure to prevent SSRF attacks
+
+        ip: ipaddress.IPv4Address | ipaddress.IPv6Address | None = None
+
+        netloc = request.url.netloc.decode()
+        if ":" in netloc:  # Either an IP, or a hostname:port combo
+            netloc_parts = netloc.split(":")
+            if len(netloc_parts) > 1:  # netloc of username:password@hostname:port not supported
```
boc-the-git: The check here doesn't align with the comment.. should this be checking for `... > 2`?
I won't stop this delaying a merge, but will circle back to it, via Discord.

mealie/pkgs/safehttp/transport.py
```diff
@@ -0,0 +1,80 @@
...
+                raise InvalidDomainError(f"unsupported request format: {request.url} -> {ip}")
```
boc-the-git: The `ip` variable has never been set. Assume that's unintended.

mealie/pkgs/safehttp/transport.py
boc-the-git: `... > 1` will always be true, except in some weird scenario where `:` was the first or last character of `netloc` (and tbh I'm not certain what happens in that scenario)

mealie/pkgs/safehttp/transport.py
```diff
@@ -0,0 +1,80 @@
...
+
+            netloc = netloc_parts[0]
+
+            try:
+                ip = ipaddress.ip_address(netloc)
+            except ValueError:
+                if self._log:
+                    self._log.debug(f"failed to parse ip for {netloc=} falling back to domain resolution")
```
boc-the-git, line 56: I'm not familiar with how `{netloc=}` is being used here.. assuming the `=` is a typo

boc-the-git: From python.org..
image

mealie/pkgs/safehttp/transport.py
```diff
@@ -46,8 +46,6 @@ async def handle_async_request(self, request):
         netloc = request.url.netloc.decode()
         if ":" in netloc:  # Either an IP, or a hostname:port combo
             netloc_parts = netloc.split(":")
-            if len(netloc_parts) > 1:  # netloc of username:password@hostname:port not supported
```
boc-the-git: @hay-kot see this log.. showing that even when `user:pass` is included in the url, it was being stripped out before getting to this netloc variable. Thus, the check on parts is irrelevant here.

image

boc-the-git (approved):

boc-the-git:
@hay-kot this has my blessing, but note I had to make a fix. Please review, and merge/release as desired.

mealie/pkgs/safehttp/transport.py
tba-code: It was a typo.

PR #3395, opened 2024-03-30
feat: PostgresProvider - Add POSTGRES_URL_OVERRIDE

### LINKED ISSUE #3392: [BUG] - Failure to connect to Postgres over a unix socket
### First Check

I'm attempting to run mealie without using docker, and I'm trying to connect to my postgres database over a unix socket. To do so, I'm setting the following environment variables:

```
DB_ENGINE = "postgres";
POSTGRES_USER = "mealie";
POSTGRES_SERVER = "/run/postgresql";
POSTGRES_DB = "mealie";
```

When I start mealie, I find that it is attempting to connect with the wrong database name:

```
Mar 29 17:37:27 server init_db[2383102]: ERROR: 29-Mar-24 17:37:27         Error connecting to database: (psycopg2.OperationalError) connection to server on socket "/run/postgresql/.s.PGSQL.5432" failed: FATAL:  database "run/postgresql:5432/mealie" does not exist
Mar 29 17:37:27 server init_db[2383102]: (Background on this error at: https://sqlalche.me/e/20/e3q8)
Mar 29 17:37:27 server init_db[2383102]: ERROR: 29-Mar-24 17:37:27         Database connection failed. Retrying in 1 seconds...
```

My unix socket does indeed exist at `/run/postgresql/.s.PGSQL.5432`, however the database name "run/postgresql:5432/mealie" is not correct. It appears that this value isn't getting set correctly by the code, which I believe the relevant portion of is:

https://github.com/mealie-recipes/mealie/blob/79fb1fb2990ad5d102e54f0bc3b4f04c45b4ad1d/mealie/core/settings/db_providers.py#L36-L56

I also attempted to set `POSTGRES_SERVER` to `/run/postgresql/.s.PGSQL.5432` instead, and ended up with:

```
Mar 30 01:15:31 plonkie init_db[2968013]: ERROR: 30-Mar-24 01:15:31         Error connecting to database: (psycopg2.OperationalError) connection to server on socket "/run/postgresql/.s.PGSQL.5432" failed: FATAL:  database "run/postgresql/.s.PGSQL.5432:5432/mealie" does not exist
```

...if that gives any hints as to how the DB name is being derived.

### Steps to Reproduce

1) Set up a postgres database with a database name, available over a unix socket.
2) Specify the appropriate environment variables and start mealie.
3) Observe that it fails to connect to the database.

See above.

### Mealie Version

1.2.0

### Deployment

Other (please specify below)

I'm using the [NixOS mealie package](https://github.com/NixOS/nixpkgs/blob/nixos-unstable/pkgs/by-name/me/mealie/package.nix#L153), which just passes environment variables directly through without modification. Hence I believe this is an upstream bug, rather than one in the NixOS package.

tba-code:
## What type of PR is this?

- bug
- cleanup
- documentation
- feature

- Added POSTGRES_URL to PostgresProvider as an override for the individual POSTGRES_* variables currently in use.
- PostgresProvider.db_url no longer rebuilds the db url on every call.
- Updated the documentation to reflect the new variable and it's usage.
- The override can be used to specify a UNIX socket for postgres.

Fixes #3392 

## Testing

1). Installed postgres locally into a mealie devcontainer.
2). Created a database for mealie
3). Tested multiple values of POSTGRES_URL (omitted, specified using default config, specified using file socket). using task dev:services and task py:postgres for more complete testing.

### DIFF
--- a/docs/docs/documentation/getting-started/installation/backend-config.md
+++ b/docs/docs/documentation/getting-started/installation/backend-config.md
@@ -35,7 +35,7 @@
 | POSTGRES_SERVER   | postgres | Postgres database server address |
 | POSTGRES_PORT     |   5432   | Postgres database port           |
 | POSTGRES_DB       |  mealie  | Postgres database name           |
-| POSTGRES_URL_OVERRIDE |   None   | Optional Postgres URL override to use instead of POSTGRES_* variables. |
+| POSTGRES_URL_OVERRIDE |   None   | Optional Postgres URL override to use instead of POSTGRES_* variables |
 
 ### Email
 
--- a/docs/docs/documentation/getting-started/installation/backend-config.md
+++ b/docs/docs/documentation/getting-started/installation/backend-config.md
@@ -33,7 +33,7 @@
 | POSTGRES_SERVER   | postgres | Postgres database server address |
 | POSTGRES_PORT     |   5432   | Postgres database port           |
 | POSTGRES_DB       |  mealie  | Postgres database name           |
-| POSTGRES_URL      |   None   | Postgres URL override. Must be used with DB_ENGINE: postgres |
+| POSTGRES_URL_OVERRIDE |   None   | Postgres URL override. Must be used with DB_ENGINE: postgres |
 
 ### Email
 
--- a/docs/docs/documentation/getting-started/installation/backend-config.md
+++ b/docs/docs/documentation/getting-started/installation/backend-config.md
@@ -33,7 +33,7 @@
 | POSTGRES_SERVER   | postgres | Postgres database server address |
 | POSTGRES_PORT     |   5432   | Postgres database port           |
 | POSTGRES_DB       |  mealie  | Postgres database name           |
-| POSTGRES_URL_OVERRIDE |   None   | Postgres URL override. Must be used with DB_ENGINE: postgres |
+| POSTGRES_URL_OVERRIDE |   None   | Optional Postgres URL override to use instead of POSTGRES_* variables. Must be used in conjunction with DB_ENGINE: postgres to have an effect. |
 
 ### Email
 
--- a/docs/docs/documentation/getting-started/installation/backend-config.md
+++ b/docs/docs/documentation/getting-started/installation/backend-config.md
@@ -33,6 +33,7 @@
 | POSTGRES_SERVER   | postgres | Postgres database server address |
 | POSTGRES_PORT     |   5432   | Postgres database port           |
 | POSTGRES_DB       |  mealie  | Postgres database name           |
+| POSTGRES_URL      |   None   | Postgres URL override. Must be used with DB_ENGINE: postgres |
 
 ### Email
 
--- a/docs/docs/documentation/getting-started/installation/backend-config.md
+++ b/docs/docs/documentation/getting-started/installation/backend-config.md
@@ -35,7 +35,7 @@
 | POSTGRES_SERVER   | postgres | Postgres database server address |
 | POSTGRES_PORT     |   5432   | Postgres database port           |
 | POSTGRES_DB       |  mealie  | Postgres database name           |
-| POSTGRES_URL_OVERRIDE |   None   | Optional Postgres URL override to use instead of POSTGRES_* variables. Must be used in conjunction with DB_ENGINE: postgres to have an effect. |
+| POSTGRES_URL_OVERRIDE |   None   | Optional Postgres URL override to use instead of POSTGRES_* variables. |
 
 ### Email
 
--- a/docs/docs/documentation/getting-started/installation/postgres.md
+++ b/docs/docs/documentation/getting-started/installation/postgres.md
@@ -26,16 +26,14 @@ services:
       PGID: 1000
       MAX_WORKERS: 1
       WEB_CONCURRENCY: 1
-      BASE_URL: http://localhost # (4)
+      BASE_URL: https://mealie.yourdomain.com
       # Database Settings
       DB_ENGINE: postgres
-      POSTGRES_URL: 'postgres://mealie:mealie@postgres/mealie' # (5)
-      # The following variables are supported if you do not specify POSTGRES_URL
-      # POSTGRES_USER: mealie
-      # POSTGRES_PASSWORD: mealie
-      # POSTGRES_SERVER: postgres
-      # POSTGRES_PORT: 5432
-      # POSTGRES_DB: mealie
+      POSTGRES_USER: mealie
+      POSTGRES_PASSWORD: mealie
+      POSTGRES_SERVER: postgres
+      POSTGRES_PORT: 5432
+      POSTGRES_DB: mealie
     depends_on:
       postgres:
         condition: service_healthy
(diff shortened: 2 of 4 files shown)


docs/docs/documentation/getting-started/installation/postgres.md
```diff
@@ -5,40 +5,41 @@ PostgreSQL might be considered if you need to support many concurrent users. In
 **For Environment Variable Configuration, see** [Backend Configuration](./backend-config.md)
 
 ```yaml
----
```
boc-the-git, line 8: Is this an accident?

docs/docs/documentation/getting-started/installation/postgres.md
```diff
@@ -5,40 +5,41 @@ PostgreSQL might be considered if you need to support many concurrent users. In
...
+      # Set Backend ENV Variables Here
+      ALLOW_SIGNUP: true
+      PUID: 1000
+      PGID: 1000
+      MAX_WORKERS: 1
+      WEB_CONCURRENCY: 1
+      BASE_URL: http://localhost # (4)
+      # Database Settings
+      DB_ENGINE: postgres
+      POSTGRES_URL: 'postgres://mealie:mealie@postgres/mealie' # (5)
```
boc-the-git: If default is none, we probably shouldn't be setting one in the sample file I'd think

docs/docs/documentation/getting-started/installation/postgres.md
```diff
@@ -5,40 +5,41 @@ PostgreSQL might be considered if you need to support many concurrent users. In
...
-      - MAX_WORKERS=1
-      - WEB_CONCURRENCY=1
-      - BASE_URL=https://mealie.yourdomain.com
+      # Set Backend ENV Variables Here
+      ALLOW_SIGNUP: true
+      PUID: 1000
+      PGID: 1000
+      MAX_WORKERS: 1
+      WEB_CONCURRENCY: 1
+      BASE_URL: http://localhost # (4)
```
boc-the-git: Why replace the old value? It would seem more clear to someone who doesn't know how it should be configured (no concerns re adding the note)

docs/docs/documentation/getting-started/installation/postgres.md
tba-code: No. It isn't required for valid yaml and most example compose files do not use it. Does it break something by removing it?

docs/docs/documentation/getting-started/installation/postgres.md
tba-code: Because of how the refactor was done, it ends up being that value anyway. The old values were unchanged, it's just now the function returns self.DATABASE_URL always. It's initially none so the function only needs to build it once, and from then on can just reference it for the lifecycle of the PostgresProvider instance. The idea of listing the new format in the example is that it works for more use cases - see the linked issue for one such example.

docs/docs/documentation/getting-started/installation/postgres.md
tba-code: The goal was to have a working compose directly from the example instead of needing tweaks, with appropriate context added to inform people of how to change it if they plan to use a domain. I did keep the original value in the note with an explanation, and figured it was sufficient.

hay-kot: This change is largely unnecessary and is just a preference. I don't think it should be included in the PR

docs/docs/documentation/getting-started/installation/postgres.md
```diff
@@ -5,40 +5,41 @@ PostgreSQL might be considered if you need to support many concurrent users. In
...
+      # Database Settings
+      DB_ENGINE: postgres
+      POSTGRES_URL: 'postgres://mealie:mealie@postgres/mealie' # (5)
+      # The following variables are supported if you do not specify POSTGRES_URL
+      # POSTGRES_USER: mealie
+      # POSTGRES_PASSWORD: mealie
+      # POSTGRES_SERVER: postgres
+      # POSTGRES_PORT: 5432
+      # POSTGRES_DB: mealie
```
hay-kot: The default here should be what it was before. POSTGRES_URL isn't the recommended approach, 99% of users should use the variables themselves. 

I suggest we remove POSTGRES_URL from the examples entirely and just leave it in the env table for documentation

mealie/core/settings/db_providers.py
```diff
@@ -39,22 +39,27 @@ class PostgresProvider(AbstractDBProvider, BaseSettings):
     POSTGRES_SERVER: str = "postgres"
     POSTGRES_PORT: str = "5432"
     POSTGRES_DB: str = "mealie"
+    POSTGRES_URL: str = ""
```
hay-kot: I'd suggest that we rename this to POSTGRES_URL_OVERRIDE and nix the re-assigning of POSTGRES_URL in the property. POSTGRES_URL_OVERRIDE is a little more clear in that you're bypassing any validation on inputs. 

It's also weird for a property to set a value that's not just used as a cache IMO and it's only called once or twice so there's no real performance gain to be had here.

docs/docs/documentation/getting-started/installation/sqlite.md
```diff
@@ -10,11 +10,11 @@ SQLite is a popular, open source, self-contained, zero-configuration database th
 
 ```yaml
 ---
-version: "3.7"
```
hay-kot, line 13: Why was the version removed?

tba-code: The version property has been deprecated for some time, and newer versions of compose now output a warning. Compose v1 reached EOL status years ago.

mealie/core/settings/db_providers.py
tba-code: Alright, I will take another look at this shortly.

docs/docs/documentation/getting-started/installation/postgres.md
tba-code: Will do. I'll make patches shortly.

docs/docs/documentation/getting-started/installation/sqlite.md
hay-kot: Yeesh. Good to know thanks!

hay-kot (dismissed):
LGTM, will defer to @boc-the-git for merge

docs/docs/documentation/getting-started/installation/postgres.md
boc-the-git: FWIW, I don't care one way or the other about the `---` syntax, but I asked the question because the other sample file didn't have the change. I like to think _including_ `---` is _technically correct_, but I'd never actually write it myself.
No concerns, now that you've removed it from the other file too.

tba-code: Yeah, I didn't realize this until much later, sorry. I made sure to correct it when I noticed. :)

docs/docs/documentation/getting-started/installation/backend-config.md
```diff
@@ -33,6 +33,7 @@
 | POSTGRES_SERVER   | postgres | Postgres database server address |
 | POSTGRES_PORT     |   5432   | Postgres database port           |
 | POSTGRES_DB       |  mealie  | Postgres database name           |
+| POSTGRES_URL_OVERRIDE |   None   | Postgres URL override. Must be used with DB_ENGINE: postgres |
```
boc-the-git: I don't like the wording "Must" here - it implies it's mandatory when in fact it's optional.

boc-the-git (changes requested):
One doco change needed, but otherwise happy for this to be merged - even ahead of the security release if that's how the timing works out. I don't see this as being a risky change given it's a new env var being introduced.

Oh, and I'm crowd sourcing an opinion to the change of how timezone is set in the sample files.. that's on Discord.

docs/docs/documentation/getting-started/installation/backend-config.md
tba-code: It's not optional if using the variable. You must define that or it will not work because the provider is created based on DB_ENGINE.

boc-the-git: Yea, I understand the message you're trying to communicate, but I don't agree that the words achieve that. Otherwise that same message should be on all the `POSTGRES_x` env vars?

Something that talks to what it replaces would make more sense, e.g. (don't quote me, I've given this very little thought!!) "Postgres URL override. Used in place of POSTGRES_x/POSTGRES_y/... if you want to specify the full connection URI"

tba-code: This is what I came up with: "Optional Postgres URL override to use instead of POSTGRES_* variables. Must be used in conjunction with DB_ENGINE: postgres to have an effect."

boc-the-git: It's absolutely an improvement, I just still find this a weird addition `Must be used in conjunction with DB_ENGINE: postgres to have an effect.`, per previous comments.

tba-code: I'd be nervous to not add something similar as this is a quirk of the implementation.

The URL includes the dialect already, so its quite weird to have it specific to postgres if I am honest here.
If you don't specify it must still be used in conjunction with DB_ENGINE: postgres, someone may think that specifying DB_ENGINE is redundant, and remove it, only to find the instance has gone back to sqlite.

boc-the-git: Pretend this new variable doesn't exist..
If a user does not specify `DB_ENGINE: postgres`, they get a sqlite DB already, right?
It looks that way to me, based on https://github.com/mealie-recipes/mealie/blob/mealie-next/mealie/core/settings/settings.py#L215

I'm all for the docs being improved, but adding this exact message, to this _one_ environment variable, adds more confusion rather than simplifying things.

tba-code: Changed.

boc-the-git (approved):
👍

PR #3424, opened 2024-04-04
fix: properly escape postgres password

### LINKED ISSUE #3418: Migration needed. Performing migration...
Using Mealie on unRAID. Mealer was working perfecftly fine last week, now I'm seeing:

ValueError: invalid interpolation syntax in 'postgresql://me:password@192.168.1.174:5432/postgres' at position 22

ERROR:    Application startup failed. Exiting.
Switching to dedicated user

  User uid:    99
     User gid:    100

INFO: 03-Apr-24 20:25:54        start: database initialization
INFO: 03-Apr-24 20:25:54        Database connection established.
INFO: 03-Apr-24 20:25:54        Context impl PostgresqlImpl.
INFO: 03-Apr-24 20:25:54        Will assume transactional DDL
INFO: 03-Apr-24 20:25:54        Migration needed. Performing migration...

and it does not perform a migration or do anything further. Mealie does not load and container is auto-stopped.

### LINKED ISSUE #3423: [BUG] - Migration from V1.3.2  to V1.4.0
### First Check

After migrate my container from v1.3.2 to V1.4.0, service doesn't boot. Sql error on data migration
Back to V1.3.2 successfully...

### Steps to Reproduce

Container is installed in kubernetes cluster
Update manifest (container tag: v1.4.0)
Worker download image
Worker start image
Crash
Worker rollback
Service up (v1.3.2)

```
  File "/app/mealie/db/init_db.py", line 119, in <module>
    main()
  File "/app/mealie/db/init_db.py", line 101, in main
    command.upgrade(alembic_cfg, "head")
  File "/opt/pysetup/.venv/lib/python3.10/site-packages/alembic/command.py", line 403, in upgrade
    script.run_env()
  File "/opt/pysetup/.venv/lib/python3.10/site-packages/alembic/script/base.py", line 583, in run_env
    util.load_python_file(self.dir, "env.py")
  File "/opt/pysetup/.venv/lib/python3.10/site-packages/alembic/util/pyfiles.py", line 95, in load_python_file
    module = load_module_py(module_id, path)
  File "/opt/pysetup/.venv/lib/python3.10/site-packages/alembic/util/pyfiles.py", line 113, in load_module_py
    spec.loader.exec_module(module)  # type: ignore
  File "<frozen importlib._bootstrap_external>", line 883, in exec_module
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
  File "/app/alembic/env.py", line 25, in <module>
    config.set_main_option("sqlalchemy.url", settings.DB_URL)
  File "/opt/pysetup/.venv/lib/python3.10/site-packages/alembic/config.py", line 272, in set_main_option
    self.set_section_option(self.config_ini_section, name, value)
  File "/opt/pysetup/.venv/lib/python3.10/site-packages/alembic/config.py", line 299, in set_section_option
    self.file_config.set(section, name, value)
  File "/usr/local/lib/python3.10/configparser.py", line 1206, in set
    super().set(section, option, value)
  File "/usr/local/lib/python3.10/configparser.py", line 896, in set
    value = self._interpolation.before_set(self, section, option,
  File "/usr/local/lib/python3.10/configparser.py", line 403, in before_set
    raise ValueError("invalid interpolation syntax in %r at "
ValueError: invalid interpolation syntax in 'postgresql://xxxxxx@xxxxxxxxxx:xxx/xxxxxxxxxxx' at position 26

```

### Mealie Version

hkotel/mealie:v1.4.0-arm64

### Deployment

Docker (Linux)

Posgresql V16.2-1

tba-code:
## What type of PR is this?

- bug

- Properly escapes postgres db urls for alembic
- POSTGRES_URL_OVERRIDE now escapes the password portion

Fixes #3423 
Fixes #3418 

The string manipipulation to find the password in POSTGRES_URL_OVERRIDE is quite ugly but I didn't know of a better way to handle it. Any suggestions for improvements would be appreciated.

## Testing

- Spun up the dev services postgres with the password ```P@ssword!@#$%%^^&&**()+;'"'<>?{}[]```
- Tried to launch the backend.

I was able to launch the backend from both POSTGRES_URL_OVERRIDE and POSTGRES_* variables using the same password.


michael-genson (changes requested):
Would you be able to add tests for this? Specifically the password you provided, as well as any other edge cases you can think of. You can parametrize `test_pg_connection_url_encode_password` in `test_config.py`. Something like:

```python
@pytest.mark.parametrize(
  "input, expected_output",
  [("input_1", "output_1"), ("input_2", "output_2"), ...],
)
def test_pg_connection_url_encode_password(input: str, expected_output: str, monkeypatch):
  ...
```

Additionally, does this properly handle cases where users have already escaped their passwords? (i.e. they have already inserted double % signs). Can we add a test for this as well?

tba-code:
Testing added.

alembic/env.py
```diff
@@ -22,7 +22,7 @@
 
 # Set DB url from config
 settings = get_app_settings()
-config.set_main_option("sqlalchemy.url", settings.DB_URL)
+config.set_main_option("sqlalchemy.url", settings.DB_URL.replace("%", "%%"))
```
michael-genson, line 29: Won't this mess up users who have already manually escaped their passwords? e.g. if my password is `123%456` and I manually escaped it with `123%%456` won't this turn it into `123%%%%456`?

michael-genson (changes requested):
Generally looks good, just one more bit of feedback and I think this is good to go

alembic/env.py
tba-code: Note that the PostgresProvider.db_url will always safely quote the password. So that means there will only ever be a single % in a row. as % turns into %25.

Users just provide the actual password in their configs. no escaping or special sequences needed. Rather, if you use those things, your config will break.

There is a reason many guides steer people away from using special characters in db passwords and instead opt for 32-64 character alphanumeric strings.

michael-genson:
Looks good, though I'd prefer to have @hay-kot look at it before signing off in case I'm missing anything

hay-kot (dismissed):
Made a small backwards compatibility change so that users who have manually escaped their passwords shouldn't have to update their config again. 

@tba-code let me know if there's any issue with that, otherwise I can merge.

tba-code:
@hay-kot we cannot unquote passwords first as it may break others. this was intentionally added then dropped in a previous commit in this PR

tba-code:
Breaking some configs is unavoidable for this one. Now that the underlying issue is fixed though, there is no longer a reason for people to try and escape any values in the configs. Would rather break the few that did rather than have an obscure bug down the line. I suggest we make a patch note for this.

hay-kot (approved):

michael-genson:
LGTM. Agreed that we're not going to be able to account for 100% of users, so this is good enough. Thankfully while it's a pain to have to fix your password one more time, it's not like we're breaking setups or breaking data, so it's a just-one-more-fix kind of breaking change

michael-genson (approved):

PR #3437, opened 2024-04-07
chore: remove black

hay-kot:
## What type of PR is this?

_(REQUIRED)_

- chore

Removes a dependency and replaces it with a faster alternative built into a tool we're already using.

N/A

Note that we previously used black in python for generated code, now we just run py:fmt task after we generate code. Everything should still be formatted nicely and we have one less dependency :D

### To Do

- [x] Remove VSCode Extension recomendations

## Testing

CI / manual workflow tasks


michael-genson:
On mobile so can't test, sorry if this is a dumb question - does this still work on alembic migrations? I see it was removed from alembic hooks, which is fine as long as it's caught on precommit, though it would be nice if it were automatically applied on alembic migrations

Regarding the VSCode extensions: not sure how much this matters as long as black is removed and precommit still catches formatting issues. Honestly I'd be happy just removing every formatting extension we have save for ruff/mypy/precommit. Ruff replaces flake8, pylance, isort, etc. Fewer tools is better IMO as long as the code is still formatted appropriately and it's easy for new devs to fix formating issues before opening PRs

Taskfile.yml
```diff
@@ -146,6 +147,12 @@ tasks:
     cmds:
       - poetry run python mealie/app.py
 
+  py:migrate:
+    desc: generates a new migration file e.g. task py:migrate:generate "add new column"
+    cmds:
+      - poetry run alembic revision --autogenerate -m "{{ .CLI_ARGS }}"
+      - task: py:format
+
```
hay-kot, line 155: @michael-genson should auto-format alembic migrations and make it easier to run them.

michael-genson (approved):

PR #3453, opened 2024-04-11
feat: Update Shopping List Timestamp on List Item Update

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- feature

_(REQUIRED)_

This PR listens for list item events (create, update, delete) and updates the parent list so the `update_at` property is changed.

_(REQUIRED)_

Fixes https://github.com/mealie-recipes/mealie/discussions/3449

The `SessionBuffer` class is used so the shopping list is only updated once per commit (otherwise a bulk update of 10 items would update the shopping list 10 times). The use of the `ContextVar` is to prevent race conditions in the FastAPI ASGI context (similar issue to having a global variable shared by multiple threads).

## Testing

Pytest


boc-the-git (approved):
LGTM 🚀

PR #3459, opened 2024-04-12
fix: Ratings UI and Filter Issues

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- bug

_(REQUIRED)_

This PR fixes two issues with the new user ratings:

`RecipeRating` Component
---
There are various quirks with setting ratings when the group rating matches the desired user rating For instance, if the group rating is 4, and you want to set your user rating to 4, the UI interprets this as you setting it to 0.

Now we use two `v-rating` components to achieve the same effect, which greatly simplifies the logic (and removes the same-rating edge case).

Sort Recipes By Rating
---
The rating sort on the home page can be jank depending on database-specific handling (e.g. I think postgres defaults to "nulls first" and sqlite defaults to "nulls last", or something like that) so I made the null handling explicitly nulls last (unless overridden with a query param).

I also coalesced 0 values to null, since currently they are equivalent (you can't rate something 0, you can only remove your rating).

_(REQUIRED)_

N/A

## Testing

Manual frontend testing


Kuchenpirat (approved):
Looks all good. 
This seems way less convoluted than the way we used to do it before 👍

michael-genson:
Yeah my original implementation sucked, haha

PR #3471, opened 2024-04-16
feat: Group Shopping List Items By Food

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- feature

_(REQUIRED)_

This adds additional group/sort logic to shopping list items. Currently items are:
- sorted by label (when toggled)
- merged together (when food and unit match)
- manually sortable

This adds additional sort logic to group items that are identical, but otherwise unmergeable. i.e. when two list items are on a list and they share a food, but can't be merged (e.g. due to differing units) they'll be put next to each other.

For example, consider the following list:
- 3 cloves garlic
- 10 apples
- 1 head garlic

This PR will re-order the list to look like this, before showing it to the user:
- 3 cloves garlic
- 1 head garlic
- 10 apples

Here's an example:
2024-04-16_11h08_17

A few considerations were made to accommodate this:
- if label sort is enabled, an item's label is always respected, even if foods are identical (though this is a weird scenario)
- if an item isn't a food, its note value is used for comparison instead
- if a user manually re-orders the list by dragging an item, the user's order is respected while the page is open. As soon as the page is refreshed/re-visited, items will be grouped again (this avoids awkward scenarios where a user drags an item and the list re-organizes itself after 1-2 seconds)

_(REQUIRED)_

N/A

When grouping items, we arbitrarily prefer the first item in the list with that group. So, for instance, if an item at position 1 is an apple, and the item at position 30 is an apple, we put all apples at position 1. There's no perfect solution to this, so I chose to just use the first one since it's easy to implement and remains reasonably consistent. Positions are re-written after grouping, so this scenario isn't that common anyway.

## Testing

Poking around a lot


Kuchenpirat (approved):
Looks all good to me. Making the shoping list even better 🚀🚀

boc-the-git:
Who's this guy buying _bunches_ of pumpkins!?! 😨

PR #3474, opened 2024-04-16
fix: make groups private by default

p0lycarpio:
## What type of PR is this?

_(REQUIRED)_

- fix

As mentionned in https://github.com/mealie-recipes/mealie/pull/3368#issuecomment-2024924163, I suggest to change group visibility by default to "private". This is an additional security measure and also when using Mealie for the first time the user would be redirected to the login/welcome page instead of the empty group.

This does not affect groups already created and I have not seen any impacts in the frontend.

N/A

## Testing
- manual testing via interface (during initial configuration and on the management groups page)
- pytest runs successfully

### DIFF
--- a/tests/integration_tests/user_group_tests/test_group_preferences.py
+++ b/tests/integration_tests/user_group_tests/test_group_preferences.py
@@ -31,7 +31,7 @@ def test_preferences_in_group(api_client: TestClient, unique_user: TestUser) -> 
 
 
 def test_update_preferences(api_client: TestClient, unique_user: TestUser) -> None:
-    new_data = UpdateGroupPreferences(recipe_public=True, recipe_show_nutrition=True)
+    new_data = UpdateGroupPreferences(recipe_public=False, recipe_show_nutrition=True)
 
     response = api_client.put(api_routes.groups_preferences, json=new_data.model_dump(), headers=unique_user.token)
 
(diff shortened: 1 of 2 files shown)


tests/integration_tests/user_group_tests/test_group_preferences.py
```diff
@@ -31,7 +31,7 @@ def test_preferences_in_group(api_client: TestClient, unique_user: TestUser) ->
 
 
 def test_update_preferences(api_client: TestClient, unique_user: TestUser) -> None:
-    new_data = UpdateGroupPreferences(recipe_public=False, recipe_show_nutrition=True)
```
michael-genson: Is there a reason for this change? `recipe_public` is `True` by default, so shouldn't we be testing that setting it to `False` works? This is a different field than the one you changed (`private_group`)

michael-genson (changes requested):
Change looks good, this also has the added advantage that new instances kick you to the login screen immediately instead of going to a blank homepage first. Just a comment on the modified test

I also went through the setup wizard and made sure both the "make group private" and "make group public" paths still work as expected

tests/integration_tests/user_group_tests/test_group_preferences.py
p0lycarpio: Indeed, it is not related. I thought I saw a test on group structure but no. I will correct the test

michael-genson (approved):

PR #3496, opened 2024-04-19
Add new OIDC TLS CA Certfile option

### LINKED ISSUE #3489: [BUG] - OIDC is not usable with private CAs
### First Check

I've run Mealie as part of my setup for a while now, since the `omni-nightly` images were available. I use Authelia and LLDAP for authentication and SSO across my self-hosted services, including Mealie. Previously, I had it set up directly against the LDAP backend, which worked fine but was higher-friction than a true SSO experience. I was excited by the new OIDC support, so I tried to switch over to that, but after fixing a few issues I'm now stuck on an error in Mealie's logs when I try to log in, which results in an infinite redirect loop for the user.

The issue seems to be that Mealie's backend request to get the OAuth information from the server is rejecting the certificate, which is signed by a private certificate authority that backs all of my services. I've tried a number of things to get the root CA in a place where the Mealie container would see it, to no avail. Eventually I dug in and found that the requests library uses `certifi`, which explicitly does **not** use the system CA bundle at all. Oops.

Perhaps this could be solved by adding an `OIDC_TLS_CACERTFILE` config, similar to the one that exists for LDAP?

### Steps to Reproduce

1) Deploy Authelia
2) Front Authelia with a reverse proxy using a TLS certificate from a private CA
3) Deploy Mealie, configured to use OIDC
4) Try to log in

> urllib3.exceptions.MaxRetryError: HTTPSConnectionPool(host='auth.example.com', port=443): Max retries exceeded with url: /.well-known/openid-configuration (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1007)')))

### Mealie Version

v1.5.1

### Deployment

Docker (Linux)

_No response_

cmintey:
## What type of PR is this?

- feature

Some users have their IdPs secured with a private certificate authority. Currently, TLS verification fails for those users. Similarly to LDAP, this PR adds an option to specify a path to a CA cert bundle to verify requests against.

Closes https://github.com/mealie-recipes/mealie/issues/3489

## Testing

Created a new image and the OP of the issue verified the changes worked


boc-the-git (approved):
LGTM 👍

PR #3540, opened 2024-05-01
fix: Shopping list labels reordering dialog

### LINKED ISSUE #3426: [BUG] - UX: Re-ordering labels on shopping list
### First Check

The UX for re-ordering shopping lists is bad.
There is only a "cancel" button, though its effect is "close". Updates to the order appear to be saved in real time when you drag and drop; though I didn't look into it extensively.

Hitting "cancel" certainly does not "undo" the changes you'd made.

There's probably 2 solutions here.. either reword "cancel" to "close" (easier), or implement actual "cancel" and "save" functionality. The former is fine.

image

### Steps to Reproduce

On a shopping list, toggle label sorting, then re-order the labels.

N/A

### Mealie Version

Assume this is applicable to all versions.

Version
nightly
Build
[de142c47df8486ee8dad17ad9947bcc4d5b8a6b2](https://github.com/mealie-recipes/mealie/commit/de142c47df8486ee8dad17ad9947bcc4d5b8a6b2)

### Deployment

Docker (Linux)

_No response_

p0lycarpio:
## What type of PR is this?

_(REQUIRED)_

- bug

_(REQUIRED)_

This PR fixes the modal allowing you to reorganize the labels of a shopping list. As mentioned in #3426, a "Save" button has been added to call the API with the new labels order. The cancel button closes the modal and reactivates the API polling.

_(REQUIRED)_

Fixes #3426

If we change the order of the labels and cancel, the local state is still modified (this can be seen if we re-edit the order quickly). The next API poll (every 5 seconds) returns the original order. It was chosen not to call `refresh()` when canceling or saving to limit requests.

## Testing

Manually tested on the frontend


frontend/pages/shopping-lists/_id.vue
```diff
@@ -58,9 +58,15 @@
       </div>
 
       <!-- Reorder Labels -->
-      <BaseDialog v-model="reorderLabelsDialog" :icon="$globals.icons.tagArrowUp" :title="$t('shopping-list.reorder-labels')">
+      <BaseDialog v-model="reorderLabelsDialog"
```
michael-genson: Add a linebreak before v-model (see GitHub Actions Warning)

frontend/pages/shopping-lists/_id.vue
```diff
@@ -103,7 +109,8 @@
         />
       </div>
       <div v-else class="mt-4 d-flex justify-end">
-        <BaseButton v-if="preferences.viewByLabel" edit class="mr-2" @click="reorderLabelsDialog = true">
+        <BaseButton v-if="preferences.viewByLabel" edit class="mr-2"
```
michael-genson: Add a linebreak before v-if (see GitHub Actions Warning)

michael-genson (changes requested):
Generally looks good! GH Actions threw two lint warnings which I commented on. Otherwise, the only issue I have is regarding the cancel behavior you mentioned - I would expect cancel to instantly revert my changes in the dialog (what if I make some changes, change my mind, then want to reset?).

What if instead of modifying the list in-place, we make a copy of the list which is displayed in the dialog, overwrite it when we open the dialog (so it resets), then save the "copied" list to the database on save? Check out the "Edit List Item" flow in the `ShoppingListItem` component, this is more-or-less what we do to achieve the same kind of thing.

p0lycarpio:
Thanks for your feedback! Indeed, it looks better this way. What do you think now?

frontend/pages/shopping-lists/_id.vue
```diff
@@ -525,16 +539,31 @@ export default defineComponent({
...
 
       loadingCounter.value += 1;
-      const { data } = await userApi.shopping.lists.updateLabelSettings(shoppingList.value.id, labelSettings);
+      const { data } = await userApi.shopping.lists.updateLabelSettings(shoppingList.value.id, localLabels.value);
       loadingCounter.value -= 1;
 
       if (data) {
-        refresh();
+        // update shoppingList labels using the API response
+        shoppingList.value.labelSettings = (data as ShoppingListOut).labelSettings
```
michael-genson: Missing a semicolon here

```suggestion
        shoppingList.value.labelSettings = (data as ShoppingListOut).labelSettings;
```

michael-genson (changes requested):
Yeah I like this much better, just one tiny thing and then I think this is good to go. I appreciate the work!

michael-genson (approved):

PR #3332, opened 2024-03-17
fix: Recipe Search URL State

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- bug
- feature

## What this PR does / why we need it:

_(REQUIRED)_

This fixes a bunch of issues with the homepage's recipe search and URL state. Namely:
- The URL always reflects the current search params
- Navigating directly to a URL with search params will correctly apply those search params
- If already on the homepage, clicking on the Mealie logo will reset your search params (and correctly show all recipes)
- Clicking on a tag will correctly update your search params
- Clicking on a tag as a public user sets your tag filter (this was previously only a feature for logged-in users)
- The "created at" param is no longer always shown in the URL (this issue was due to a misused camelCase param, whereas all search params use snake_case)

This also adds a user state for the last set of search params, i.e. next time you open Mealie your last search will be restored. There are two exceptions to this:
1. If you're _already_ on the home page, and you click the Mealie logo, your last search is ignored, and your filters are reset
2. If you navigate to the home page and provide query params (e.g. `https://MyMealieSite.com/g/home?tags=abc-123`) then the query params are respected, and your last search is ignored

## Which issue(s) this PR fixes:

_(REQUIRED)_

Fixes https://github.com/mealie-recipes/mealie/issues/2447
Fixes https://github.com/mealie-recipes/mealie/issues/2352
Fixes https://github.com/mealie-recipes/mealie/issues/2259
Fixes https://github.com/mealie-recipes/mealie/issues/931

## Testing

_(fill-in or delete this section)_

Lots of clicking

PR #3541, opened 2024-05-01
fix: Recipe Search Quirks and Session Storage

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- bug
- feature

_(REQUIRED)_

This fixes a few issues with the recipe search state from https://github.com/mealie-recipes/mealie/pull/3332. https://github.com/mealie-recipes/mealie/pull/3332 fixed most issues, but introduced a few smaller ones. This mostly had to do with the `search` param not updating properly when cleared.

@Kuchenpirat also mentioned in https://github.com/mealie-recipes/mealie/pull/3332 that persisting state _forever_ seems weird (e.g. if I search for a recipe, then come back to Mealie the next day, I don't really want to see that search again). This moves the search params to session storage so it's cleared when you leave (note: if you just refresh the page it may not look cleared since your query params are there. This behavior is only when navigating back to the home page _without_ query params).

I also refactored a bit and moved default values to an object, since it made some of these changes easier. As a bonus, if we ever implement user-preferred default sorts, this will be very easy to merge in (just load the defaults).

_(REQUIRED)_

N/A, I'm a little surprised no one's opened an issue for the search being wonky.

## Testing

Manually poked around the search a lot.


Kuchenpirat (approved):
LGTM 👍
Never really realised some search quirks were annoying but this feels quite a bit more polished 🧹

PR #3579, opened 2024-05-10
fix: Fix file not found error with individual recipe export/download.  

### LINKED ISSUE #3095: [BUG] - Cannot Download Recipe(s)
### First Check

When I click on 'Download' on a recipe, it does nothing. I wanted to download a Mealie recipe, and give the ZIP to a friend, so that they could import the ZIP into their Mealie, but it just won't download.

### Steps to Reproduce

1) Open recipe
2) Click ... > Download
3) Does nothing

INFO:     192.168.29.230:59323 - "POST /api/recipes/shrimp-fettuccine-alfredo/exports HTTP/1.1" 200 OK
INFO:     192.168.29.230:59323 - "GET /api/recipes/shrimp-fettuccine-alfredo/exports/zip?token=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzbHVnIjoic2hyaW1wLWZldHR1Y2NpbmUtYWxmcmVkbyIsImV4cCI6MTcwNjgwOTkxNX0.sLcmPW4e1l60K2LSy1gEqEUCvdnFx4I-x7OuhsKTDK0 HTTP/1.1" 500 Internal Server Error
ERROR:    Exception in ASGI application
  File "/opt/pysetup/.venv/lib/python3.10/site-packages/starlette/responses.py", line 323, in __call__
    stat_result = await anyio.to_thread.run_sync(os.stat, self.path)
  File "/opt/pysetup/.venv/lib/python3.10/site-packages/anyio/to_thread.py", line 33, in run_sync
    return await get_asynclib().run_sync_in_worker_thread(
  File "/opt/pysetup/.venv/lib/python3.10/site-packages/anyio/_backends/_asyncio.py", line 877, in run_sync_in_worker_thread
    return await future
  File "/opt/pysetup/.venv/lib/python3.10/site-packages/anyio/_backends/_asyncio.py", line 807, in run
    result = context.run(func, *args)
FileNotFoundError: [Errno 2] No such file or directory: '/app/data/.temp/my_zip_archive.zip'
During handling of the above exception, another exception occurred:
  File "/opt/pysetup/.venv/lib/python3.10/site-packages/uvicorn/protocols/http/httptools_impl.py", line 419, in run_asgi
    result = await app(  # type: ignore[func-returns-value]
  File "/opt/pysetup/.venv/lib/python3.10/site-packages/uvicorn/middleware/proxy_headers.py", line 84, in __call__
    return await self.app(scope, receive, send)
  File "/opt/pysetup/.venv/lib/python3.10/site-packages/fastapi/applications.py", line 1054, in __call__
    await super().__call__(scope, receive, send)
  File "/opt/pysetup/.venv/lib/python3.10/site-packages/starlette/applications.py", line 123, in __call__
    await self.middleware_stack(scope, receive, send)
  File "/opt/pysetup/.venv/lib/python3.10/site-packages/starlette/middleware/errors.py", line 186, in __call__
    raise exc
  File "/opt/pysetup/.venv/lib/python3.10/site-packages/starlette/middleware/errors.py", line 164, in __call__
    await self.app(scope, receive, _send)
  File "/opt/pysetup/.venv/lib/python3.10/site-packages/starlette/middleware/gzip.py", line 24, in __call__
    await responder(scope, receive, send)
  File "/opt/pysetup/.venv/lib/python3.10/site-packages/starlette/middleware/gzip.py", line 44, in __call__
    await self.app(scope, receive, self.send_with_gzip)
  File "/opt/pysetup/.venv/lib/python3.10/site-packages/starlette/middleware/exceptions.py", line 62, in __call__
    await wrap_app_handling_exceptions(self.app, conn)(scope, receive, send)
  File "/opt/pysetup/.venv/lib/python3.10/site-packages/starlette/_exception_handler.py", line 64, in wrapped_app
    raise exc
  File "/opt/pysetup/.venv/lib/python3.10/site-packages/starlette/_exception_handler.py", line 53, in wrapped_app
    await app(scope, receive, sender)
  File "/opt/pysetup/.venv/lib/python3.10/site-packages/starlette/routing.py", line 762, in __call__
    await self.middleware_stack(scope, receive, send)
  File "/opt/pysetup/.venv/lib/python3.10/site-packages/starlette/routing.py", line 782, in app
    await route.handle(scope, receive, send)
  File "/opt/pysetup/.venv/lib/python3.10/site-packages/starlette/routing.py", line 297, in handle
    await self.app(scope, receive, send)
  File "/opt/pysetup/.venv/lib/python3.10/site-packages/starlette/routing.py", line 77, in app
    await wrap_app_handling_exceptions(app, request)(scope, receive, send)
  File "/opt/pysetup/.venv/lib/python3.10/site-packages/starlette/_exception_handler.py", line 64, in wrapped_app
    raise exc
  File "/opt/pysetup/.venv/lib/python3.10/site-packages/starlette/_exception_handler.py", line 53, in wrapped_app
    await app(scope, receive, sender)
  File "/opt/pysetup/.venv/lib/python3.10/site-packages/starlette/routing.py", line 75, in app
    await response(scope, receive, send)
  File "/opt/pysetup/.venv/lib/python3.10/site-packages/starlette/responses.py", line 326, in __call__
    raise RuntimeError(f"File at path {self.path} does not exist.")
RuntimeError: File at path /app/data/.temp/my_zip_archive.zip does not exist.

### Mealie Version

**Details**
Version: v1.1.0
Build: 
Application Mode: Production
Demo Status: Not Demo
API Port: 9000
API Docs: Enabled
Database Type: sqlite
Recipe Scraper Version: 14.53.0

**Checks**
Secure Site: Yes
Server Side Base URL: Yes
LDAP Ready: No
Email Configured: No
Docker Volumes: error

### Deployment

Docker (Linux)

_No response_

nephlm:
* Fix file not found error with individual recipe export.  (#3095 )
* Add a suitable test.  
* Split `temporary_zip_path` into `temporary_zip_path` and `unlinking_temporary_zip_path`.

## What type of PR is this?

_(REQUIRED)_
- bug

_(REQUIRED)_

This PR fixes the issue that was preventing the recipe download and the `/api/recipes/{slug}/exports/zip` endpoint more generally from working.  As described in #3095, it would fail with a file not found.  This was caused by the file being deleted before the Response read the file from the filesystem.  

_(REQUIRED)_

Fixes #3095 

I created a version of `temporary_zip_path` that doesn't unlink when the view completes and added background task to the Response to clean up the file that is run after the Response is delivered to the client.  

Renamed the original `temporary_zip_path` to `unlinking_temporary_zip_file` and updated references to point to the new name.  This seemed like clearer naming given the changes. 

## Testing

* Manually tested and ran the test suite.
* Added `test_get_recipe_as_zip` that failed with the error described in #3095 before the fix, and succeeded afterwards.
* Verified the file was deleted after download. 
* Manually checked the endpoints potentially impacted by the dependency name change.


mealie/routes/recipe/recipe_crud_routes.py
```diff
@@ -131,7 +132,9 @@
             if image_asset.is_file():
                 myzip.write(image_asset, arcname=image_asset.name)
 
-        return FileResponse(temp_path, filename=f"{slug}.zip")
+        return FileResponse(
+            temp_path, filename=f"{slug}.zip", background=BackgroundTask(temp_path.unlink, missing_ok=True)
```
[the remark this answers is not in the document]
nephlm: * It depends on the `slug` to the same degree the previous code did.  
* Slug here is not the passed in string, but the value in the encrypted jwt token.  However that encrypted value is created from user input so that's no reason to trust it.
* We are not using the slug to read or write from the mealie volume, but only to supply an advisory filename to the client.  
* We are using slug to lookup a value in the db, so the slug must match a slug in the db or we will not reach this code.
    * I have not verified, but I'm assuming sql injection from slug lookups has been mitigated. 
* Slugify should prevent the creation of any recipe title -> slug that would allow directory traversal or similar dangerous derived filenames from being sent to the user.

The change does not seem to have changed the risk profile of this function, and it appears safe to me, though if someone sees a risk here please let me know. 

There is an odd bit of code that is `if slug != slug:` which I suspect is supposed to verify that the slug in the token matches the slug passed in the path, but at present that if clause can never trigger and we just use the the slug from the token.  I can fix that comparison if people agree with my assessment of what it is supposed to be doing.

mealie/routes/recipe/recipe_crud_routes.py
```diff
@@ -118,20 +119,22 @@
...
         image_asset = recipe.image_dir.joinpath(RecipeImageTypes.original.value)
         with ZipFile(temp_path, "w") as myzip:
             myzip.writestr(f"{slug}.json", recipe.model_dump_json())
 
             if image_asset.is_file():
                 myzip.write(image_asset, arcname=image_asset.name)
 
-        return FileResponse(temp_path, filename=f"{slug}.zip")
+        return FileResponse(
+            temp_path, filename=f"{recipe.slug}.zip", background=BackgroundTask(temp_path.unlink, missing_ok=True)
```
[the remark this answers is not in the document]
nephlm: If anyone has an idea how to get CodeQL to pass, let me know.  `recipe.slug` is a cleaned/slugified value from the db.  I wouldn't really characterize that as user provided value.  The path isn't used to access any resource in the mealie container, but to provide an advisory filename to the client.  By definition (since it was slugified), it can't contain any characters that would allow accessing different directories.  

At this point the only option is to change the interface to return a uuid name instead of a slug name, and I don't feel comfortable doing that without input from someone more involved in the project.

mealie/routes/recipe/recipe_crud_routes.py
```diff
@@ -118,20 +119,22 @@ def get_recipe_as_format(self, slug: str, template_name: str, temp_dir=Depends(t
     @router_exports.get("/{slug}/exports/zip")
     def get_recipe_as_zip(self, slug: str, token: str, temp_path=Depends(temporary_zip_path)):
```
michael-genson: The issue is here: technically a user can provide the `temp_path` var explicitly which can be used maliciously. You should be able to fix this super easily by just removing the param and calling `temporary_zip_path` in the route:

```python
@router_exports.get("/{slug}/exports/zip")
def get_recipe_as_zip(self, slug: str, token: str):
  """Get a Recipe and It's Original Image as a Zip File"""
  temp_path = temporary_zip_path()
  ...
```

mealie/routes/recipe/recipe_crud_routes.py
```diff
@@ -118,20 +119,22 @@ def get_recipe_as_format(self, slug: str, template_name: str, temp_dir=Depends(t
...
         """Get a Recipe and It's Original Image as a Zip File"""
```
michael-genson:
```suggestion
        """Get a Recipe and Its Original Image as a Zip File"""
```

mealie/core/dependencies/dependencies.py
```diff
@@ -205,9 +205,13 @@ def validate_recipe_token(token: str | None = None) -> str:
     return slug
 
 
-async def temporary_zip_path() -> AsyncGenerator[Path, None]:
+def temporary_zip_path() -> Path:
```
michael-genson: Can we change this to `get_temporary_zip_path` and not use it as a dependency? More info on the other comment

michael-genson (changes requested):
This should resolve the security issue. The code QL comment is placed at the end of the method because technically that's where the "exploit" is, but the source of the issue is in the method definition

mealie/core/dependencies/dependencies.py
nephlm: Yep.  Though I should note that `temporary_dir` is used as a route parameter for `POST /users/{id}/image` and `GET /recipe/{slug}/exports`.  I'll have to figure out exactly what `Depends` does to figure out if I can remove it from the parameter list and still have it clean up after itself.

mealie/routes/recipe/recipe_crud_routes.py
nephlm: Completely missed that `temp_path` could be passed in by a user.  Should be no problem to fix.

mealie/core/dependencies/dependencies.py
michael-genson: It looks like only the async + yield does the cleanup, so currently `temporary_dir` doesn't do any cleanup with the new behavior.

nephlm: That's what I'm seeing.  I think I'm going to have to push the cleanup code into those two routes.

mealie/core/dependencies/dependencies.py
```diff
@@ -205,24 +204,15 @@ def validate_recipe_token(token: str | None = None) -> str:
...
-
-    try:
-        yield temp_path
-    finally:
-        temp_path.unlink(missing_ok=True)
+    return app_dirs.TEMP_DIR.joinpath("my_zip_archive.zip")
 
 
-async def temporary_dir(background_tasks: BackgroundTasks) -> AsyncGenerator[Path, None]:
+def get_temporary_dir() -> Path:
```
michael-genson: Can we make this a context manager so we don't have to unlink the file every time this is called? Something like:

```python
from contextlib import contextmanager
...
@contextmanager
def get_temporary_path(auto_unlink=True) -> Path:  # also if we could rename this, since it's a temp path not a temp dir
    temp_path = app_dirs.TEMP_DIR.joinpath(uuid4().hex)
    temp_path.mkdir(exist_ok=True, parents=True)

    try:
        yield temp_path
    finally:
        if auto_unlink:
            temp_path.unlink(missing_ok=True)
```

Then in the code use it like this:
```python
with get_temporary_dir() as temp_dir:
    ...
```

mealie/routes/recipe/recipe_crud_routes.py
```diff
@@ -103,35 +103,39 @@ def get_recipe_zip_token(self, slug: str):
...
         `template_name`: The name of the template to use to use in the exports listed. Template type will automatically
         be set on the backend. Because of this, it's important that your templates have unique names. See available
         names and formats in the /api/recipes/exports endpoint.
 
         """
+        temp_dir = get_temporary_dir()
         recipe = self.mixins.get_one(slug)
         file = self.service.render_template(recipe, temp_dir, template_name)
-        return FileResponse(file)
+        return FileResponse(file, background=BackgroundTask(rmtree, temp_dir))
```
michael-genson: Since this needs to unlink after the file is served you can use `get_temporary_path(auto_unlink=False)` here (see above)

mealie/routes/recipe/recipe_crud_routes.py
```diff
@@ -103,35 +103,39 @@ def get_recipe_zip_token(self, slug: str):
...
         image_asset = recipe.image_dir.joinpath(RecipeImageTypes.original.value)
         with ZipFile(temp_path, "w") as myzip:
             myzip.writestr(f"{slug}.json", recipe.model_dump_json())
 
             if image_asset.is_file():
                 myzip.write(image_asset, arcname=image_asset.name)
 
-        return FileResponse(temp_path, filename=f"{slug}.zip")
+        return FileResponse(
+            temp_path, filename=f"{recipe.slug}.zip", background=BackgroundTask(temp_path.unlink, missing_ok=True)
```
michael-genson: Same here

michael-genson (changes requested):
A few comments that I think will make the code more readable using a context manager. This is what FastAPI was _supposed_ to be doing with the dependency injection, but it's broken (hence the bug)

mealie/core/dependencies/dependencies.py
nephlm: Yeah, context manager should be much cleaner. I should have a new commit with that solution soon.

michael-genson (approved):
Looks great, thanks for sticking with it!

PR #3592, opened 2024-05-13
chore: Remove Server Tasks

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- cleanup

_(REQUIRED)_

Removes all mentions of the old server tasks API, except for the database model. I left the database model because otherwise Alembic migrations are going to whine about it, though I added a comment.

_(REQUIRED)_

Mentioned in https://github.com/mealie-recipes/mealie/discussions/3442


mealie/db/models/group/group.py
```diff
@@ -67,7 +73,7 @@ class Group(SqlAlchemyBase, BaseMixins):
     webhooks: Mapped[list[GroupWebhooksModel]] = orm.relationship(GroupWebhooksModel, **common_args)
     recipe_actions: Mapped[list["GroupRecipeAction"]] = orm.relationship("GroupRecipeAction", **common_args)
     cookbooks: Mapped[list[CookBook]] = orm.relationship(CookBook, **common_args)
-    server_tasks: Mapped[list[ServerTaskModel]] = orm.relationship(ServerTaskModel, **common_args)
+    server_tasks: Mapped[list["ServerTaskModel"]] = orm.relationship("ServerTaskModel", **common_args)
```
boc-the-git, line 76: Why did this need to change?
(I don't have a problem with it, just not sure why the change is being made)

boc-the-git (approved):
Have one comment seeking clarification, but ultimately this LGTM

mealie/db/models/group/group.py
michael-genson: Ah yeah, that's because I removed the line by accident then put it back (and accidentally did it slightly differently). Specifically it's different because of how imports work with SQLAlchemy: I could've put it back how it was, but I also moved the import to typing only (which means it's not actually imported at runtime):

```python
# earlier in the file
if TYPE_CHECKING:
    ...
    from ..server.task import ServerTaskModel
```

This is a better way of handling imports in this scenario since it's slightly more performant and avoids circular imports (though that's not actually an issue here). In theory we should replace all of our SQLAlchemy imports to do this, but the performance benefit is negligible so it's not worth the effort.

---

TL;DR it's a totally unnecessary change I did by accident, but I'm inclined to leave it because it's technically better anyway

michael-genson: Or not at all and prod breaks. Cool
https://github.com/mealie-recipes/mealie/pull/3630

boc-the-git: At least we had fun 😅

PR #3603, opened 2024-05-15
feat: Data Management from Shopping List

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- bug
- feature

_(REQUIRED)_

This PR enables managing data from a shopping list. Namely:
1. Foods and units can be created from the shopping list (similar to how they can be created from recipes)
2. Food labels can be changed if the shopping list item uses a different label than the food

This also fixes a visual bug when creating new foods/units on recipes by force-closing the dropdown.

Creating foods from the shopping list (with fixed visual bug):
2024-05-15_11h27_50

Assigning label to food (note that other instances of "apple" are not updated because those items are already set to produce):
2024-05-15_11h28_28

(not pictured: the food "apple" is now re-labeled as "Juice")

_(REQUIRED)_

Closes https://github.com/mealie-recipes/mealie/discussions/3602

## Testing

Manually


boc-the-git (changes requested):
Quite a few linter issues on this one @michael-genson

boc-the-git (commented):
I'm effectively happy to approve this, bar the linting issues previously mentioned and ..

The "Copy to Food" functionality. I like it, but I wonder it it's going to not be clear to a user what it's actually doing? Particularly for say the 2nd and 3rd user of a system, not the "power user".. Thoughts? Wondering if there's an alternate wording, maybe something like "Permanently add Label to Food" (..but not exactly that)

michael-genson:
Yeah I couldn't come up with anything concise enough other than "Copy to Food". Some others:
- Assign to Food
- Apply Label to Food
- Update Food Data
- Change Food Label
- Change Food Data Label
- Move Food to Label

I was avoiding "permanently" because it seems like a bulky word here, but that could work too. Also maybe adding a tooltip can work?

Kuchenpirat:
I would suggest going for something shorter like `Label Food` or `Add Label`

I think you kinda don't need to reference Food since this is the toggled option. But a tooltip might be a nice middleground, but are not that practical on mobile.

michael-genson:
Maybe "Set Label" or "Apply Label"? Or "Save Label"? I agree with dropping Food. It's more than "Add" because a Food might already have a Label, just a different one.

michael-genson:
Updated to "Save Label":
image

boc-the-git (approved):

PR #3638, opened 2024-05-23
docs: add possible log levels

DennisGaida:
## What type of PR is this?

- documentation

Found the documentation here: https://www.uvicorn.org/settings/#logging with the possible logging values (tried `WARN` only to figure out it needs to be `warning` or `WARNING`).

none


docs/docs/documentation/getting-started/installation/backend-config.md
```diff
@@ -16,7 +16,7 @@
 | TZ                            |          UTC          | Must be set to get correct date/time on the server                                  |
 | ALLOW_SIGNUP<super>\*</super> |         false         | Allow user sign-up without token                                                    |
 | LOG_CONFIG_OVERRIDE           |                       | Override the config for logging with a custom path                                  |
-| LOG_LEVEL                     |         info          | logging level configured                                                            |
+| LOG_LEVEL                     |         info          | logging level (e.g. critical, error, warning, info, debug, trace)                   |
```
michael-genson: I don't think Python actually supports "trace" out of the box: https://docs.python.org/3/library/logging.html#levels

```suggestion
| LOG_LEVEL                     |         info          | logging level (e.g. critical, error, warning, info, debug)                          |
```

michael-genson (changes requested):

docs/docs/documentation/getting-started/installation/backend-config.md
DennisGaida: That may be so, but as stated in the original post including URL the Uvicorn library is used for logging which includes those levels.

michael-genson: Wow I'm blind, thanks!

michael-genson (dismissed):

boc-the-git (approved):

PR #3650, opened 2024-05-26
feat: Set default number of days on meal planner

boc-the-git:
## What type of PR is this?

- feature

Allows a user to set their default number of days for the meal planner - which previously was always a week.
The implementation was workshopped on Discord.
Importantly I made the default something used on page load, and independent of what someone might choose in the calendar on that instance of the page. (see 2nd gif)

meal_planner_default_days

meal_planner_default_days2

None.

## Testing

Manual.


boc-the-git:
.. I'm not opposed to UI improvements.. I just wasn't seeing a lot of scope for it with what Vuetify had available to me.

michael-genson:
Should it maybe go into the settings menu instead? Otherwise LGTM, though I'd like to get eyes from someone else since I don't use the meal planner

boc-the-git:
> Should it maybe go into the settings menu instead? 

On Discord the consensus was to put it here (in development channel). 

I agree in theory settings would make sense but the current settings is actually very specific to "meal plan rules" and not really a general settings page.

michael-genson:
Ah okay, I vaguely remember that. Thanks!

michael-genson (approved):

PR #3656, opened 2024-05-27
Setting backend passwords with docker compose secrets

### LINKED ISSUE #3606: Setting backend passwords with docker compose secrets
### Discussed in https://github.com/mealie-recipes/mealie/discussions/3291

<div type='discussions-op-text'>

<sup>Originally posted by **Weizenritter** March 11, 2024</sup>
Hey there. Since I like checking in all my configs I like them to be free of secrets in clear text. I wonder if it's possible to provide the value of the variable SMTP_PASSWORD with an docker compose secret. Here an example of what I would imagine:
```yaml
---
services:
  mealie:
    image: ghcr.io/mealie-recipes/mealie:v1.3.2
    container_name: mealie
    ports:
      - "9925:9000"
    volumes:
      - ./mealie_data:/etc/mealie/data
    environment:
      - TZ=Europe/Berlin
      - BASE_URL=https://mealie.example.com
      - SMTP_AUTH_STRATEGY=TLS
      - SMTP_FROM_EMAIL=user@example.com
      - SMTP_HOST=smtp.example.com
      - SMTP_PASSWORD_FILE=/run/secrets/smtp_password  # <-- Look here
      - SMTP_PORT=587
      - SMTP_USER=user@example.com
    secrets:
      - source: smtp_password
secrets:
  smtp_password:
    name: my_smtp_password
    file: /path/to/smtp_password
```

Note the `_FILE` suffix, it's a convention that some official images implement, as stated in the [documentation of docker compose](https://docs.docker.com/compose/use-secrets) (scroll to the bottom for the note).

If it's possible in some other way I'd like to here how, thanks in advance!</div>

aljora:
## What type of PR is this?

- feature

This change would allow administrators to share their secrets with the mealie container without needing to enter them through the environment variables, which could possibly expose them to other processes.

- mealie/core/settings/settings.py has the addition of a secrets directory per the [pydantic documentation](https://docs.pydantic.dev/latest/concepts/pydantic_settings/#use-case-docker-secrets)
- docs/docs/documentation/getting-started/installation/backend-config.md documents the use of the new feature

It was chosen to hard-code the path to avoid needed to process environment variables before the pydantic object was created. By default Docker places all secrets in files according to their names into a single directory. This supports the default use case but would need further effort to support other paths.

Fixes #3606

Automated testing was considered but I did not want to inconvenience all other developers on this project by making them declare secrets in their dev containers or machines. Is there a need to create a CI workflow that uses a custom container to run tests?

## Testing

The change was verified by running the production container with and without the secrets directory present. With the secrets directory, the content of the secret was evident in the logs showing that the information reached the backend successfully. Note that this PR does not address whether or not secrets are compromised by appearing in the logs. With the directory absent, the logs include a warning per the pydantic docs.


boc-the-git:
@aljora this PR is in draft state, is that intentional?

mealie/core/settings/settings.py
```diff
@@ -236,7 +236,7 @@ def OPENAI_ENABLED(self) -> bool:
     # Testing Config
 
     TESTING: bool = False
-    model_config = SettingsConfigDict(arbitrary_types_allowed=True, extra="allow")
+    model_config = SettingsConfigDict(arbitrary_types_allowed=True, extra="allow", secrets_dir="/run/secrets")
```
boc-the-git, line 245: I haven't given thought to the full implications of this (or even looked at what SettingsConfigDict does), but I think we'd want this information reflected in the doco in some way.

aljora:
> @aljora this PR is in draft state, is that intentional?

Yes, I wanted to put this up there while I was confirming usage and writing documentation. I'll move it to the ready state.

mealie/core/settings/settings.py
aljora: Done, thank you.

Kuchenpirat (approved):

Kuchenpirat:
Hey, thanks for your work on this :)

Tested it and works for me 👍 
Will leave it open for @boc-the-git to merge :)

PR #3657, opened 2024-05-28
docs: Update documentation link on main readme

DataDalton:
## What type of PR is this?

_(REQUIRED)_

- documentation

_(REQUIRED)_

- Updates the link in the main readme to go to the actual documentation rather than the homepage of mealie nightly

_(REQUIRED)_

N/A

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@
 Mealie is a self hosted recipe manager and meal planner with a RestAPI backend and a reactive frontend application built in Vue for a pleasant user experience for the whole family. Easily add recipes into your database by providing the URL and Mealie will automatically import the relevant data, or add a family recipe with the UI editor. Mealie also provides an API for interactions from 3rd party applications.
 
 - [Remember to join the Discord](https://discord.gg/QuStdQGSGK)!
-- [Documentation](https://nightly.mealie.io)
+- [Documentation](https://docs.mealie.io/documentation/getting-started/introduction/)
 
 
 <!-- CONTRIBUTING -->
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@
   <p align="center">
     A Place for All Your Recipes
     <br />
-    <a href="https://nightly.mealie.io"><strong>Explore the docs »</strong></a>
+    <a href="https://docs.mealie.io/documentation/getting-started/introduction/"><strong>Explore the docs »</strong></a>
   <a href="https://github.com/mealie-recipes/mealie">
   </a>
     <br />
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -20,7 +20,7 @@
   <p align="center">
     A Place for All Your Recipes
     <br />
-    <a href="https://nightly.mealie.io"><strong>Explore the docs »</strong></a>
+    <a href="https://docs.mealie.io/documentation/getting-started/introduction/"><strong>Explore the docs »</strong></a>
```
boc-the-git: I'm fine with changing `nightly` -> `docs`, but I'm not sure it's correct to be adding all the other stuff there. Do you have a rationale behind it?

boc-the-git:
@DataDalton this PR is in draft state, is that intentional?

DataDalton:
> @DataDalton this PR is in draft state, is that intentional?

Apologies! The contributor guide says to open as a draft until checks are completed, will take out of draft now

README.md
DataDalton: My reasoning for directly linking to that is because that's the first page of the actual docs, going to just docs.mealie.io takes you to the base page of docs which has no content of the docs themselves from what I can tell and you have to press getting started to go to the docs, so it just saves an extra click

boc-the-git: So my concern with that is we're stating that we're linking to the docs, and that's what we're doing. Changing it to link to the getting started guide is well, not _the docs_ but _the getting started guide_.
I don't think we need to be/should be adding `/documentation/getting-started/introduction/`

boc-the-git (changes requested):

README.md
DataDalton: I was under the impression that the requested change is the docs while just being worded as "getting started" otherwise I cannot find any other documentation. The existing link (nightly.mealie.io) simply redirects to docs.mealie.io so why use this even in the first place? The new link also directly takes you to the 'docs' rather than the menu to get to the docs. Personally, when I click on a link to go somewhere like docs I expect it to directly go there not a page that redirects me and then allows me to find it and then go there.

michael-genson: I agree with @boc-the-git, I don't see a reason for this change. It makes sense to me to change `nightly` to `docs` but I don't think it makes sense to take them to a subpage rather than the homepage

DataDalton: I understand, perhaps it's better to have the explore the docs as docs.mealie.io and the documentation link to go to the docs.mealie.io/documentation/getting-started/introduction/ ?

michael-genson: I don't see a reason to change either link, besides switching from `nightly` to `docs`

DataDalton: Sounds good, will change it to docs

boc-the-git (approved):

Issue #1500, opened 2022-07-22
[v1.0.0b] - Postgresql backup restore fails when not user postgres superuser


wipash:
### First Check

- [X] This is not a feature request
- [X] I added a very descriptive title to this issue.
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Mealie documentation, with the integrated search.
- [X] I already read the docs and didn't find an answer.

### What is the issue you are experiencing?

Postgresql version: 14.4
Mealie version: 1.0.0beta-3

When restoring a backup using the builtin backup restore functionality, the backup fails if Mealie is configured to use a user other than the postgres superuser.

The cause seems to be in the `drop_all()` function of `alchemy_exporter.py`. Postgres doesn't allow `SET session_replication_role = 'replica'` if you are not a superuser.

Log from postgres server:
```
2022-07-22 04:42:24.951 GMT [424] ERROR:  permission denied to set parameter "session_replication_role"
2022-07-22 04:42:24.951 GMT [424] STATEMENT:  SET session_replication_role = 'replica'
2022-07-22 04:42:24.952 GMT [424] ERROR:  current transaction is aborted, commands ignored until end of transaction block
2022-07-22 04:42:24.952 GMT [424] STATEMENT:  SET session_replication_role = 'origin'
```

Mealie is deployed with these environment variables:
```
        TZ: Pacific/Auckland
        TOKEN_TIME: 43830
        DB_ENGINE: postgres
        POSTGRES_USER: mealie
        POSTGRES_PASSWORD: longandcomplexpassword
        POSTGRES_SERVER: my.postgres.server
        POSTGRES_PORT: 5432
        POSTGRES_DB: mealie
        BASE_URL: https://mealie.myhouse.com
```

If I set mealie to use the postgres superuser account, the restore works as expected.

### Deployment

Docker (Linux)

wipash:
Easy enough to work around

```sql
ALTER USER mealie WITH SUPERUSER;

# Run restore from Mealie

ALTER USER mealie WITH NOSUPERUSER;
```

Taylor-Tyson:
Newbie question, how do I connect to the postgres database to alter the user?

yottabit42:
I am logged into the mealie database with pgadmin, but the mealie user doesn't have the permissions to grant superuser to the mealie user. 😄 There is also a postgres user, but I don't know what the password would be...

yottabit42:
[...]

PR #3674, opened 2024-05-31
feat: Handle Safari-mangled backup ZIPs and improve backup UI

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- feature

_(REQUIRED)_

We've had several instances [like this](https://github.com/mealie-recipes/mealie/issues/1500#issuecomment-2142960420) where Safari mangles the ZIP file structure, breaking our backups. This PR adds some logic to detect when this happens and work around it.

The UI is also very unhelpful when backups fail, so this PR fixes some state issues and flashes an alert to check server logs when a backup fails.

_(REQUIRED)_

No open issues, several closed ones where the solution was just to fix the ZIP structure

## Testing

I added a mangled zip file test before implementing the workaround, confirmed it failed, and now it passes


tests/data/__init__.py
```diff
@@ -22,6 +22,9 @@
 backup_version_bcfdad6b7355_1 = CWD / "backups/backup_version_bcfdad6b7355_1.zip"
 """bcfdad6b7355: remove tool name and slug unique contraints"""
 
+backup_version_09aba125b57a_1 = CWD / "backups/backup_version_09aba125b57a_1.zip"
+"""09aba125b57a: add OIDC auth method (Safari-mangled ZIP structure)"""
```
boc-the-git, line 26: This might be a silly question, but what's the relevance of `add OIDC auth method`?? 😕
Appears in a couple files.

boc-the-git (approved):
This LGTM, though I am a bit curious re my other comment.
Assuming there's no cause for concern there, happy for you to merge @michael-genson 👍

tests/data/__init__.py
michael-genson: The original purpose of having a bunch of example backup files in the tests data was to test for migration issues, so for all of the migration test data I include the alembic version (`09aba125b57a` in this case) as well as the migration message. This is basically saying "this backup is a snapshot of a database that is at the migration where the OIDC auth method was added".

For this specific test it's kind of useless, though in the future if this migration fails it's easier to debug

boc-the-git: Ah, makes sense. I tried to check if it was a commit hash and obviously it wasn't! Didn't think of alembic

PR #3721, opened 2024-06-10
docs: Add doco for db changes

boc-the-git:
## What type of PR is this?

- documentation

Adds some documentation for how to go about DB changes.
I wrote this ages ago as part of https://github.com/mealie-recipes/mealie/pull/2930 which is on hold, so splitting it out separately to unblock the docs.

None.


Kuchenpirat (changes requested):
#3437 added `task py:migrate`

Might be good to use that in the dev guide as it sets the appropriate flags automatically and we can extend that if needed in the future. 

https://github.com/mealie-recipes/mealie/blob/21f1d46b6de06f954895e8dc79451a0676101aa1/Taskfile.yml#L150-L154

boc-the-git:
> #3437 added `task py:migrate`
> 
> Might be good to use that in the dev guide as it sets the appropriate flags and we can extend that if needed in the future.
> 
> https://github.com/mealie-recipes/mealie/blob/21f1d46b6de06f954895e8dc79451a0676101aa1/Taskfile.yml#L150

Good spot. Updated.

Kuchenpirat (approved):
Looks Good 👍

PR #3725, opened 2024-06-11
fix: Make fractions in the ingredient list look a tiny bit nicer

bartkummel:
Make fractions in the ingredient list look a tiny bit nicer by making the spacing a bit more tight.

## What type of PR is this?

_(REQUIRED)_

- feature

_(REQUIRED)_

The goal of this PR is to make fractions in the ingredient list look a bit nicer. Only one file is changed:
- `frontend/components/Domain/Recipe/RecipeIngredientListItem.vue`: Added a CSS class to the quantity part of the ingredient item, and added some CSS to make the spacing in and around the fraction a bit tighter.

_(REQUIRED)_

Fixes #3724

### Before
Scherm­afbeelding 2024-06-12 om 17 08 35

### After
Scherm­afbeelding 2024-06-12 om 17 07 48


michael-genson:
Would you be able to post some screenshots of before and after? I see a couple in the discussion, but a few more here would be nice. Also, can you show how this affects decimals/whole numbers? (if at all)

> Awaiting approval of feature request https://github.com/mealie-recipes/mealie/discussions/3724

You can just mark this as ready (assuming it's ready)

bartkummel:
##Screenshots
###Before
Scherm­afbeelding 2024-06-12 om 17 08 35
###After
Scherm­afbeelding 2024-06-12 om 17 07 48

frontend/components/Domain/Recipe/RecipeIngredientListItem.vue
```diff
@@ -45,6 +45,23 @@ export default defineComponent({
   .d-inline {
     & > p {
       display: inline;
+      &:has(>sub)>sup {
+        letter-spacing: -0.05rem;
+      }
+    }
+    &:has(sub) {
+      &:after {
+        content: "\00a0";
```
michael-genson: Can we just replace this in the actual method, rather than doing this with CSS?

See here: https://github.com/mealie-recipes/mealie/blob/b3241d3e8b7e708e0442d8d09d2c851fff765705/frontend/composables/recipes/use-recipe-ingredients.ts#L66C33-L66C37

michael-genson (changes requested):
LGTM. Small change to make the change consistent across pages

frontend/components/Domain/Recipe/RecipeIngredientListItem.vue
bartkummel: That makes a lot of sense indeed. After some testing, it turns out the extra space was not really needed at all, so I just removed it.

michael-genson (approved):
Looks great, thanks!

PR #3732, opened 2024-06-13
feat: Insert instruction step above or below. #3731

derpferd:
## What type of PR is this?

- feature

- This PR fixes the feature requested in #3731.
- It adds two options to the step menu: "Insert Above" and "Insert Below"
- Each option adds a new step either above or below the current step.

Screenshot 2024-06-12 at 9 22 01 PM

Fixes #3731

## Testing

- Tested in UI
- Added step above first step, below last step and above and below a step in the middle.


michael-genson (commented):
Overall looks good. Would you be able to add the same functionality to ingredients? I feel like we should have this for both

derpferd:
Yes, will do.

derpferd:
Added same functionality to ingredients in latest commit

michael-genson (approved):
LGTM, thanks!

PR #3740, opened 2024-06-16
feat: Show nutrition on recipe print

Asdoos:
## What type of PR is this?

- feature

I wan't to have the nutritions on the print for a recipe - so I add a option to do it:

image

image

No issue

I needed to extraxt the labels from RecipeNutrition Component. May I rewrite the component also to use the extraxted method ?

## Testing

Tested manually.


Kuchenpirat:
Hey thanks for your work on this.
Two little things.

1. I would prefer if the table borders are not shown in the print out. This matches the pretty clean look of the existing print better IMHO.
2. I would suggest that we hide the non populated items. There is no point in having them I would say.

Asdoos:
One Nutrition:
image

Some Nutritions:
image


image

frontend/components/Domain/Recipe/RecipePrintView.vue
```diff
@@ -83,19 +83,43 @@
         </div>
       </section>
     </div>
+
+    <!-- Nutrition -->
+    <div v-if="preferences.showNutrition">
+      <v-card-title class="headline pl-0"> {{ $t("recipe.nutrition") }} </v-card-title>
+
+      <v-divider v-if="hasNutrition" class="grey my-4"></v-divider>
```
Kuchenpirat: this will never get shown because `hasNutrition` will always be negative as props.recipe.nutrition.length is undefined. 

I also think the divider is not necessairy 
```suggestion
```

frontend/components/Domain/Recipe/RecipePrintView.vue
```diff
@@ -216,12 +244,18 @@ export default defineComponent({
       return props.recipe.notes && props.recipe.notes.length > 0;
     });
 
+    const hasNutrition = computed(() => {
+      return props.recipe.nutrition && props.recipe.nutrition.length > 0;
+    });
+
```
Kuchenpirat: will always compute to false. Is also only used in the divider that is no needed.
```suggestion
```

frontend/components/Domain/Recipe/RecipePrintView.vue
```diff
@@ -216,12 +244,18 @@ export default defineComponent({
...
     function parseText(ingredient: RecipeIngredient) {
       return parseIngredientText(ingredient, props.recipe.settings?.disableAmount || false, props.scale);
     }
 
     return {
+      labels,
       hasNotes,
+      hasNutrition,
```
Kuchenpirat:
```suggestion
```

frontend/composables/recipes/use-recipe-nutrition.ts
```diff
@@ -0,0 +1,47 @@
+import { useContext } from "@nuxtjs/composition-api";
+
+
+type NutritionLabelType = {
```
Kuchenpirat: can we use an interface here, i think this matches better what we used in the rest of the use-x files. 

```suggestion
export interface NutritionLabelType {
```

frontend/composables/recipes/use-recipe-nutrition.ts
Kuchenpirat, line 1: This file uses an indentation of 4 spaces instead of 2.

You can fix this in VS Code by clicking on the bottom right on `Spaces: 4` and then selecting `Indent Using Spaces` and then `2`. After that you can format this file by right-clicking and `format document`

Kuchenpirat (changes requested):
This looks pretty good, just a few changes before we can merge this. 

1. Can you add showNutrition to the UserPrintPreferences interface (https://github.com/asdoos/mealie/blob/eba84432bdd6a574f575be22154657f55255f6ae/frontend/composables/use-users/preferences.ts#L5-L9)
2. Can we use the `useNutritionLabels()` also in the RecipeNutrition file so we have less duplication in the code base:  https://github.com/asdoos/mealie/blob/eba84432bdd6a574f575be22154657f55255f6ae/frontend/components/Domain/Recipe/RecipeNutrition.vue#L61-L90

There are also just a few minor things down below, nothing major.

frontend/components/Domain/Recipe/RecipePrintView.vue
Kuchenpirat: BTW if you would want to compute if there are existing nutrition values there is an implementation of that here: 

https://github.com/asdoos/mealie/blob/eba84432bdd6a574f575be22154657f55255f6ae/frontend/components/Domain/Recipe/RecipeNutrition.vue#L91-L99

frontend/composables/recipes/index.ts
```diff
@@ -2,4 +2,5 @@ export { useFraction } from "./use-fraction";
 export { useRecipe } from "./use-recipe";
 export { useRecipes, recentRecipes, allRecipes, useLazyRecipes } from "./use-recipes";
 export { parseIngredientText, useParsedIngredientText } from "./use-recipe-ingredients";
+export { useNutritionLabels, NutritionLabelType } from "./use-recipe-nutrition";
```
Kuchenpirat:
```suggestion
export { useNutritionLabels } from "./use-recipe-nutrition";
```

frontend/components/Domain/Recipe/RecipeNutrition.vue
```diff
@@ -34,17 +28,10 @@
 </template>
 
 <script lang="ts">
-import { computed, defineComponent, useContext } from "@nuxtjs/composition-api";
+import { computed, defineComponent } from "@nuxtjs/composition-api";
+import { useNutritionLabels, NutritionLabelType } from "~/composables/recipes";
 import { Nutrition } from "~/lib/api/types/recipe";
```
Kuchenpirat:
```suggestion
import { useNutritionLabels } from "~/composables/recipes";
import { Nutrition } from "~/lib/api/types/recipe";
import { NutritionLabelType } from "~/composables/recipes/use-recipe-nutrition";
```

Kuchenpirat (changes requested):
Hey, thanks for the quick turnaround of the review. 
Looks all pretty great besides the Nutrition Label Type in the RecipeNutrition file. 

The export in the index.ts file does fail. 
You can just import this directly in RecipeNutrition and it should solve this problem.

Kuchenpirat (approved):
Again, thanks alot for your work on this.

Looks all good now 🚀👍

Asdoos:
Thanks for the input!

PR #3750, opened 2024-06-18
feat: bulk assign label to foods

Kuchenpirat:
## What type of PR is this?

- feature

This adds the possibility to assign multiple foods the same label at once. 
I come accross this quite frequently when we try some new fancy recipe and have to create around 5 new foods for it and many of them belong to the same category. So this allows the user to select multiple foods and assign a single label to all of them. 

None

Reactivity for the Manage Data Tabels seems to be somewhat broken again. So i just reloaded the page afterwards. If there is a better way to do this please let me know. 

## Testing

Manual 

## Screenshots
image


michael-genson (approved):
LGTM!

PR #3761, opened 2024-06-21
chore: remove gunicorn in favor of uvicorn workers

hay-kot:
## What type of PR is this?

- chore

Removes dependency of gunicorn and resolves a bug where the gunicorn logs wouldn't be formatted correctly based on the log config.

This may be considered breaking given that the behavior _does_ change, but I don't think in a way that really matters. Those that have configured environments for multiple workers will still get that behavior, but it will be different.  

N/A

## Testing

- [x] local build still works


michael-genson (approved):
LGTM. "Breaking" changes seem reasonable

PR #3764, opened 2024-06-21
feat: If only 1 shopping list, navigate straight to it (Shopping List QoL)

boc-the-git:
## What type of PR is this?

- feature

## What this PR does / why we need it:

If you only have one shopping list available to you at `/shopping-lists`, this will navigate you directly into it. There is also a button to go back to the list that disables that behaviour (via a query param) should you wish.

shopping_list_redirect

## Which issue(s) this PR fixes:

None.

## Special notes for your reviewer:

Credit @Kuchenpirat for the v-container stuff on _id.vue <3

## Testing

Manual

PR #3768, opened 2024-06-21
feat: Prevent Shopping Lists From Rendering If Redirecting

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- feature

_(REQUIRED)_

With https://github.com/mealie-recipes/mealie/pull/3764, if the user is redirected to their sole list, you'll see the shopping list page for a brief moment before redirecting. This PR removes that so it looks more seamless.

_(REQUIRED)_

N/A

## Testing

Tested all scenarios I can think of


boc-the-git:
@michael-genson can you please explain the specific intention with `overrideDisableRedirect`? I feel I might be missing something obvious.
Was there an issue without it, in terms of struggling to load that a user wanted to "view all lists", before determining there was only one to view?

boc-the-git (approved):
Happy for you to merge btw, if you're confident in your decision/testing re my prior question. Still keen to hear an answer though :)

michael-genson:
Moving the redirect to a watcher sped up the redirect a bit, but as a consequence it's checking every time the choices list changes. There was an edgecase where:
- You're currently viewing all lists (of which there is more than one)
- You only have one list
- If you toggle to show only your lists (which is one list) you'd get redirected immediately

The override fixes this

PR #3777, opened 2024-06-23
feat: Add the ability to flag a food as "on hand", to exclude from shopping list

boc-the-git:
## What type of PR is this?

- feature

Adds the ability to mark a food as "on hand" (terminology taken from tools). Then when adding a recipe to a shopping list, on hand foods will be unticked by default. Current behaviour is that every ingredient is ticked by default. The intent is to use this for your staple ingredients that you rarely need to buy, sugar, salt, spices, rice, ...

Don't ask me how this recipe would turn out!! 🙈 

foods_marked_as_staples2

None.

I think the change is fairly straight forward. There are a couple files that had a lot of formatting updates applied, so I've put some comments to indicate the small change actually made in there.

## Testing

Locally, with SQLite.


frontend/lang/messages/en-US.json
```diff
@@ -985,7 +985,9 @@
       "food-data": "Food Data",
       "example-food-singular": "ex: Onion",
       "example-food-plural": "ex: Onions",
-      "label-overwrite-warning": "This will assign the chosen label to all selected foods and potentially overwrite your existing labels."
+      "label-overwrite-warning": "This will assign the chosen label to all selected foods and potentially overwrite your existing labels.",
+      "on-hand": "On Hand",
```
boc-the-git: 'on-hand' exists elsewhere for the tool.. I decided to add it again in this section as it's specific to the foods, and allows for it to potentially change in future.

mealie/db/models/recipe/ingredient.py
```diff
@@ -144,12 +146,15 @@ class IngredientFoodModel(SqlAlchemyBase, BaseMixins):
     name: Mapped[str | None] = mapped_column(String)
     plural_name: Mapped[str | None] = mapped_column(String)
     description: Mapped[str | None] = mapped_column(String)
+    on_hand: Mapped[bool] = mapped_column(Boolean)
```
boc-the-git, line 149: This is the only actual functional change in this file

mealie/schema/recipe/recipe_ingredient.py
```diff
@@ -36,6 +36,7 @@ class UnitFoodBase(MealieModel):
     plural_name: str | None = None
     description: str = ""
     extras: dict | None = {}
+    on_hand: bool = False
```
boc-the-git, line 39: This is the only actual functional change in this file

frontend/lang/messages/en-US.json
michael-genson: I don't think this is necessary, I'd rather just change it in the future if we go that route. We have a _ton_ of mixed translations where it's in one category but used for something else

michael-genson:
LGTM, besides the one comment I had. Once this is ready I'll pull it locally and test it out, but reviewing the dif looks good

frontend/lang/messages/en-US.json
boc-the-git: Updated

frontend/pages/group/data/foods.vue
```diff
@@ -134,6 +139,11 @@
...
           >
           </v-autocomplete>
+          <v-checkbox
+            v-model="editTarget.onHand"
+            :label="$t('data-pages.foods.on-hand-checkbox-label')"
+          >
+          </v-checkbox>
         </v-form>
       </v-card-text>
       <template #custom-card-action>
```
michael-genson, line 155: Can we add a custom display option for the boolean in the `CrudTable` (line 230) similar to how tools does it? It's below this comment, but GH won't let me add a comment there since you didn't modify it.

```html
<template #item.onHand="{ item }">
  <v-icon :color="item.onHand ? 'success' : undefined">
    {{ item.onHand ? $globals.icons.check : $globals.icons.close }}
  </v-icon>
</template>
```

You should be able to insert this right under the label template (line 256)

frontend/pages/group/data/foods.vue
```diff
@@ -87,6 +87,11 @@
             :label="$t('data-pages.foods.food-label')"
           >
           </v-autocomplete>
+          <v-checkbox
+            v-model="createTarget.onHand"
+            :label="$t('data-pages.foods.on-hand-checkbox-label')"
```
michael-genson: Can we give this a simple "On Hand" label and then add a caption? Something like:

```html
<v-checkbox
  v-model="createTarget.onHand"
  hide-details
  :label="$tc('tool.on-hand')"
/>
<p class="text-caption mt-1">
  {{ $tc("data-pages.foods.on-hand-checkbox-label") }}
</p>
```

frontend/pages/group/data/foods.vue
```diff
@@ -134,6 +139,11 @@
             :label="$t('data-pages.foods.food-label')"
           >
           </v-autocomplete>
+          <v-checkbox
+            v-model="editTarget.onHand"
+            :label="$t('data-pages.foods.on-hand-checkbox-label')"
```
michael-genson: same as above

```html
<v-checkbox
  v-model="editTarget.onHand"
  hide-details
  :label="$tc('tool.on-hand')"
/>
<p class="text-caption mt-1">
  {{ $tc("data-pages.foods.on-hand-checkbox-label") }}
</p>
```

frontend/lang/messages/en-US.json
```diff
@@ -985,7 +985,8 @@
       "food-data": "Food Data",
       "example-food-singular": "ex: Onion",
       "example-food-plural": "ex: Onions",
-      "label-overwrite-warning": "This will assign the chosen label to all selected foods and potentially overwrite your existing labels."
+      "label-overwrite-warning": "This will assign the chosen label to all selected foods and potentially overwrite your existing labels.",
+      "on-hand-checkbox-label": "Flag this as a food you have 'on hand', to set it as unchecked by default when adding a recipe to a shopping list."
```
michael-genson: Maybe something like:

"Setting this flag will make this food unchecked by default when adding a recipe to a shopping list."

(See other comments which include the text "On Hand" already)

michael-genson (changes requested):
Functionality looks great. A few UI things which I think will make it look more in line with other similar data features

boc-the-git:
Changes made @michael-genson; good suggestions!

image

image

michael-genson (approved):
👍

PR #3786, opened 2024-06-24
feat: check all in shopping list view

ollywelch:
## What type of PR is this?

- feature

This PR adds a new 'check all' button to the button group in the shopping list view. This allows checking all unchecked items in a shopping list, so they can then be deleted using the existing 'delete checked items' button for instance.

image

N/A

## Testing

Created a shopping list, added some items, then pressed the check all button. This moved all items from the checked to unchecked state.


boc-the-git:
Just reading what the idea is, I think I'd want this to have a confirmation button. It serves an edge case for general usage (IMO), and is fairly destructive if clicked by accident

ollywelch:
Isn't it equally destructive as unchecking all checked items? If you check everything by mistake, you can always uncheck everything again.

boc-the-git:
I think there's a good argument for them both to have a confirmation, yes!
Marking as checked is marginally more destructive though, as combined with delete the items are gone. Unchecked doesn't have that.

michael-genson:
+1 to putting both of those options behind a confirmation button

ollywelch:
I've added dialogue boxes for:

- Checking all items
- Unchecking all items
- Deleting checked items

Not really sure what I'm doing with the text in the dialogue boxes + translations, please lmk if I've missed something!

boc-the-git:
Thanks for taking on that revision @ollywelch, this looks good!

There's a couple scenarios where we probably want a tiny bit more logic though. Right now I can get the popups even though their actions would do nothing, e.g.
- I can get the "delete all" dialog even though nothing is checked
- I can get the "uncheck all" dialog even though nothing is check
- I can get the "check all" dialog even though there's nothing unchecked

Other than that though, functionally this works quite nicely 👍

boc-the-git (changes requested):
.. it's forcing me to add a comment as part of changing status 😞

ollywelch:
@boc-the-git thanks for reviewing! I've just added some checks at the start of each open dialogue method, so should only display the prompt when necessary now

boc-the-git (approved):
Thanks for this! 🚀

PR #3799, opened 2024-06-26
fix: Lint Python code with ruff

cclauss:
## What type of PR is this?

_(REQUIRED)_

- cleanup

_(REQUIRED)_

Lint Python code with `ruff` to make it faster, easier to understand, and less error-prone.

_(REQUIRED)_

## Testing

% `ruff check`


cclauss:
Rebased to fix git conflicts.

PatrickDeTomasi:
I'm a professional that uses Ruff in the workplace, I HIGHLY recommend implementing this, it has significantly improved my own quality and speed

michael-genson:
I'm not sure this PR is necessary; we already do this for every pull request:

https://github.com/mealie-recipes/mealie/blob/mealie-next/Taskfile.yml#L106-L114

The difference between this PR and the existing workflow is that the existing workflow only looks at changes, whereas this PR blanket applies rules to the entire codebase, which makes some unnecessary changes.

Can you go into more detail what issue this PR is trying to address?

cclauss:
Given that ruff is superfast (1/3 second for the CPython codebase), why limit its visibility?  40 files are not yet ruff compliant which seems a shame.  This PR is attempting to get all files to be ruff-compliant and ensure they stay that way even when ruff rules are upgraded.

michael-genson (approved):
Fair enough. This might cause some merge conflicts with https://github.com/mealie-recipes/mealie/pull/3970 but that's fine

PR #3800, opened 2024-06-26
feat: Auto-label new shopping list items

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- feature

_(REQUIRED)_

This PR automatically adds a label to a shopping list item if it isn't provided. [We already did similar work for this a while ago](https://github.com/mealie-recipes/mealie/pull/2535) for the ingredient parser, so I extracted that into its own service and used it here.

We don't want to autocorrect the user's input (since we could be wrong), so we only use the result to choose the label, rather than assign the food. See below, where I spell "oranges" wrong; the proper label is assigned, but the word is still as-written:

2024-06-26_10h54_11

This only works on item creation, so if a user decides "nah I don't want a label on this" and removes it, it stays removed. In general I don't know why you _wouldn't_ want to try to apply a label to an item, hence the PR. This is similar to other shopping lists which auto-categorize your items (e.g. the Alexa shopping list does this).

_(REQUIRED)_

N/A

I've actually been using this logic for a year or so now as a part of my [Unified Shopping List](https://github.com/michael-genson/Unified-Shopping-List/blob/45e1deebe09e8290e6c4463b8feec7aff41fa027/AppLambda/src/services/mealie.py#L89-L107). I've been meaning to get it into Mealie. Now that [Amazon is killing all shopping list integrations for Alexa](https://developer.amazon.com/en-US/docs/alexa/custom-skills/access-the-alexa-shopping-and-to-do-lists.html), I felt now was a good time to get this in here.

## Testing

Added tests for the new feature and manually tested the parser to make sure I didn't break anything with the refactor.

### DIFF
--- a/mealie/services/group_services/shopping_lists.py
+++ b/mealie/services/group_services/shopping_lists.py
@@ -112,7 +112,27 @@ class ShoppingListService:
         if list_refs_to_delete:
             self.list_refs.delete_many(list_refs_to_delete)
 
-    def bulk_create_items(self, create_items: list[ShoppingListItemCreate]) -> ShoppingListItemsCollectionOut:
+    def find_matching_label(self, item: ShoppingListItemBase) -> UUID4 | None:
+        if item.label_id:
+            return item.label_id
+        if item.food:
+            return item.food.label_id
+
+        query = PaginationQuery(page=1, per_page=1)
+        food_search = self.repos.ingredient_foods.page_all(query, search=item.display)
+        if not food_search.items:
+            return None
+
+        return food_search.items[0].label_id
+
+    def bulk_create_items(
+        self, create_items: list[ShoppingListItemCreate], auto_find_labels=True
+    ) -> ShoppingListItemsCollectionOut:
+        """
+        Create a list of items, merging into existing ones where possible.
+        Optionally try to find a label for each item if one isn't provided using the item's food data or display name.
+        """
+
         # consolidate items to be created
         consolidated_create_items: list[ShoppingListItemCreate] = []
         for create_item in create_items:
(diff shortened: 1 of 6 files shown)


mealie/services/group_services/shopping_lists.py
```diff
@@ -108,7 +118,23 @@ def remove_unused_recipe_references(self, shopping_list_id: UUID4) -> None:
         if list_refs_to_delete:
             self.list_refs.delete_many(list_refs_to_delete)
 
-    def bulk_create_items(self, create_items: list[ShoppingListItemCreate]) -> ShoppingListItemsCollectionOut:
+    def find_matching_label(self, item: ShoppingListItemBase) -> UUID4 | None:
+        if item.label_id:
+            return item.label_id
+        if item.food:
+            return item.food.label_id
```
boc-the-git, line 125: @michael-genson do I understand correctly that this will only be labelling "notes" and not "foods that don't already have labels"?

boc-the-git: I wonder if it might make sense to also do foods if their label is null? (helps those of us who haven't got all our data sorted yet!)

michael-genson: The way this works is by fuzzy-matching whatever the user types in to the food store, and if it finds a match, assign it the label of that food. Because of this, we can't really do anything if it already has a food.

This PR basically helps label shopping list items that aren't parsed, or items that are entered manually (without a food). Since we still depend on the food store, we can't do anything for items that have a food already without some sort of external data.

boc-the-git (approved):
LGTM! 🚀

PR #3816, opened 2024-06-29
feat: improve italian units

simgunz:
## What type of PR is this?

- feature

- Add plurals to Italian units
- Translate unit "splash" to Italian

Not sure if "pluralName" is allowed in the JSON seed files.

### DIFF
--- a/mealie/repos/seed/resources/units/locales/it-IT.json
+++ b/mealie/repos/seed/resources/units/locales/it-IT.json
@@ -84,10 +84,10 @@
         "pluralName": "milligrammi"
     },
     "splash": {
-        "name": "splash",
+        "name": "spruzzo",
         "description": "",
         "abbreviation": "",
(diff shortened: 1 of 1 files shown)


Kuchenpirat:
Hey thanks a lot for your work on this! 

We are using crowdin for our localization so for your changes to get into the code you would need to make your changes over there.

If anything is updated over there the new translations will roll into mealie automatically. 

https://crowdin.com/project/mealie

simgunz:
I have now fixed the splash translation in crowdin. 

What about plurals? Given that the do not exist in English they are not proposed for translation in Italian withing crowdin.

Kuchenpirat:
Perfect 👍

Plurals are not included due to our old seed file structure that was designed when plurals where not yet a thing (in mealie I mean, not in general 😅) 

So we would first need to alter that

PR #3837, opened 2024-07-03
fix: Offline Shopping List Fixes V2 - Electric Boogaloo

### LINKED ISSUE #3859: [BUG] - Offline detection of shopping list is broken
### First Check

When accessing the Shopping list on a Android device (in my case Xiaomi 13t) and then switch to offline mode (turn off Wifi and mobile data), it doesn't recognize it properly. The offline message is only displayed for an short time after a request was initiated. It resets directly to the previous value before the change. I can reproduce it on Firefox and Chrome for Android.

Screen recording: 

https://github.com/mealie-recipes/mealie/assets/23395732/5b0c2def-3bd9-46e8-b2ba-a8efef3828c4

I can not reproduce it when using the desktop version (Linux). If I set the throttling to "Offline", then it all works fine.

### Steps to Reproduce

1. Open the Shopping List
2. Turn off Wifi and mobile data or turn on flight mode
3. It's not possible to modify list items as described above.

Console log: [demo.mealie.io-1720262329786.log](https://github.com/user-attachments/files/16115701/demo.mealie.io-1720262329786.log)

Looks like the service worker caches the requests. In offline mode the requests still return 200 OK (from service worker). Screenshot from Network tab:

image

Zipped HAR file (you can extract it and then import the HAR file in the network tab of Chrome): 
[demo.mealie.io.zip](https://github.com/user-attachments/files/16115716/demo.mealie.io.zip)

Might be related to https://github.com/mealie-recipes/mealie/issues/3842.

### Mealie Version

v1.10.2, can also be reproduced with the demo version. 

### Deployment

Docker (Linux)

_No response_

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- bug

_(REQUIRED)_

Fixed the bug that broke 1.10, then went and re-wrote the state management to be a lot simpler (and avoid that kind of bug in the future).

_(REQUIRED)_

Fixes https://github.com/mealie-recipes/mealie/issues/3859

"But it worked on my machine!"

## Testing

Tested from a browser with pre-existing state (i.e. has been using the offline shopping list) as well as from a browser with no state (i.e. a new incognito tab).


michael-genson:
In draft because this will overwrite https://github.com/mealie-recipes/mealie/pull/3850, which we don't want to do until https://github.com/mealie-recipes/mealie/pull/3847 is merged

michael-genson:
Merged and made sure the temp fix isn't overwritten. Will add that to https://github.com/mealie-recipes/mealie/pull/3847

michael-genson:
This also fixes offline state not being detected on the PWA properly

hay-kot (approved):

PR #3840, opened 2024-07-03
fix: Bump other version numbers in GH workflow

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- bug

_(REQUIRED)_

This PR fixes two missing version bumps in:
- pyproject.toml
- package.json

These both expect a string without the "v", so I added an extra step to strip the v away (and add it back in for the docs, since they contain the v). Please double check my work here, if you can!

I also added a weekly scheduled run to update pre-commit hooks. Our ruff version falls behind on every update, so this should fix that. Ideally we update the ruff version at the same time as this, but I'm not sure there's an easy way to do that (and in the rare case the mismatch messes something up, it will be caught in the lint workflow run anyway).

_(REQUIRED)_

N/A


boc-the-git (approved):
I wasn't able to easily test the package.json one, but this pyproject.toml looks good :+1: 

image

PR #3847, opened 2024-07-04
fix: Make Mealie Timezone-Aware

### LINKED ISSUE #3842: [BUG] - Changes to shopping list items immediately overwritten
### First Check

When editing a shopping list, any changes to the items get immediately overwritten. Having checked the server logs no requests are being made other than a periodic GET of the shopping list by ID, i.e. no POST or PUT to update the items. Weirdly I was able to add the first item, but nothing worked after that. Not sure if this is always the case.

https://github.com/mealie-recipes/mealie/assets/56311963/abc2d142-9025-4e24-8234-4359ad7f1a97

### Steps to Reproduce

Create a shopping list, and try and add or update some items.

See screen recording :)

### Mealie Version

v1.10.1

### Deployment

Docker (Linux)

Docker compose behind nginx proxy manager for reverse proxy

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- bug
- cleanup
- feature

_(REQUIRED)_

Much of our frontend <--> backend logic assumes Mealie stores datetimes in UTC. Well it does in the dev server... because that's just the default. As soon as you set the timezone (e.g. TZ=America/Chicago) we start running into issues. Most notably it breaks the new offline shopping list feature (see https://github.com/mealie-recipes/mealie/issues/3842, shoutout to @ollywelch for figuring this one out).

Annoyingly, this isn't a super easy fix. This PR makes several changes to how Mealie handles timezones:
- Added linter rules to enforce all Python dates/datetimes to use timezone info (https://docs.astral.sh/ruff/rules/#flake8-datetimez-dtz)
- Replaced all instances of naive datetimes (e.g. `datetime.now()` -> `datetime.now(timezone.utc)`)
- Added a model validator (to `MealieModel`) to automatically convert all naive datetimes to datetimes with tzinfo=utc

One breaking change this causes is that the API returns datetime strings with a "Z" now (indicating that it is, in fact, in UTC). Some of our frontend logic intentionally adds/strips the Z, so I updated everywhere I could find that logic (which makes the logic simpler).

e.g. currently timestamps look like this from the API: `2024-07-02T16:06:53.431631` and now they'll look like this: `2024-07-02T16:06:53.431631Z`

_(REQUIRED)_

Fixes https://github.com/mealie-recipes/mealie/issues/3842

This is definitely a breaking change because of the API returning a Z, but I don't think there's a cleaner way to fix this problem. In a new application, we should set the database to store TZ info, and we should make Pydantic require TZ-aware datetimes (Pydantic 2.0 supports this) but there's no way to do that now without drastically changing/breaking things.

Also there were a ton of import issues that I found and I fixed those. Adding a model util for getting the current time in UTC exposed them.

## Testing

Tested everything I could; frontend seems to be handling everything fine. I'm sure some bugs will pop up but all major functionality works fine, at least.


michael-genson:
This also reverts https://github.com/mealie-recipes/mealie/pull/3850, which is a temp fix while this gets reviewed

frontend/pages/shopping-lists/_id.vue
```diff
@@ -868,7 +868,6 @@ export default defineComponent({
 
         // set a temporary updatedAt timestamp prior to refresh so it appears at the top of the checked items
         item.updateAt = new Date().toISOString();
-        item.updateAt = item.updateAt.substring(0, item.updateAt.length-1);
       }
```
boc-the-git, line 871: Does this comment become no longer relevant? I'm not 100% sure what's meant to be happening here tbh.

boc-the-git (approved):
I don't see any issues here. I'd love to get a second opinion before we merge though!
Think this should spend at least a few days on nightly before we do a release.

@michael-genson can you please add a simple example of a before and after api response, into the description?
I've also got a tangential question with you on Discord.

frontend/pages/shopping-lists/_id.vue
michael-genson: Yeah that comment still matters. The checked-off items show their `updateAt` time, so if we don't do this the item looks weird until we get a backend response

PR #3862, opened 2024-07-06
feat: Push On Hand Items to Bottom Of Add To Shopping List Dialog

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- feature

_(REQUIRED)_

Pushes all "on hand" ingredients to the bottom of their section when adding a recipe to a shopping list:
image

Works with meal plans as well.

_(REQUIRED)_

N/A

## Testing

Tested the following scenarios manually:
- recipe with sections and on hand items
- recipe with no sections and on hand items
- recipe with no on hand items
- recipe with only on hand items


boc-the-git:
Instinctively, I don't love the idea. I feel this is a personal preference thing.
Maybe I need to give it more thought.

michael-genson:
Fair, happy to close unless there's a way to make the configuration not clunky.

My biggest issue with the current implementation is that I still have to manually parse every ingredient and determine if I need it or not (since they're arbitrarily mixed in), hence the PR

I meant to include that in the body but I was running late to dinner 😅

boc-the-git:
Haha, all good. 
Maybe there's a happy medium which is putting them to the bottom of the section they're in? 🤔 
Worthy of a group or user setting?

michael-genson:
> putting them to the bottom of the section they're in

I like this idea a lot. Just updated the PR and the screenshot. I tried doing a "top half and bottom half", but it looked super awkward when there were an odd number of non-on-hand items (since the top right corner would be empty):

image

frontend/components/Domain/Recipe/RecipeDialogAddToShoppingList.vue
```diff
@@ -237,30 +237,40 @@ export default defineComponent({
...
           }
 
-          // append new section if first
-          if (sections.length === 0) {
+          // If this is the first item in the section, create a new section
+          if (sections.length === 0 || currentTitle !== sections[sections.length - 1].sectionName) {
+            if (sections.length) {
+              // Add the on-hand ingredients to the last section
+              sections[sections.length - 1].ingredients.push(...onHandIngs);
+              onHandIngs.length = 0;
```
boc-the-git, line 252: I'd never seen this syntax before for clearing an array.. did some digging. TIL.

frontend/components/Domain/Recipe/RecipeDialogAddToShoppingList.vue
```diff
@@ -237,30 +237,40 @@ export default defineComponent({
...
+          }
+
+          // Store the on-hand ingredients for later
+          if (ing.ingredient.food?.onHand) {
+            onHandIngs.push(ing);
             return sections;
           }
 
-          // otherwise add ingredient to last section in the array
+          // Add the ingredient to last section
```
boc-the-git:
```suggestion
          // Add the ingredient to previous section
```

Using `previous` instead of `last` is more intuitive to me; `last` gives an implication of "the final one"

frontend/components/Domain/Recipe/RecipeDialogAddToShoppingList.vue
```diff
@@ -237,30 +237,40 @@ export default defineComponent({
...
           sections[sections.length - 1].ingredients.push(ing);
           return sections;
         }, [] as ShoppingListIngredientSection[]);
 
+        // Add remaining on-hand ingredients to the last section
```
boc-the-git:
```suggestion
        // Add remaining on-hand ingredients to the previous section
```

Using `previous` instead of `last` is more intuitive to me; `last` gives an implication of "the final one"
Admittedly, less important in this scenario where the difference between "last" and "previous" doesn't matter!

frontend/components/Domain/Recipe/RecipeDialogAddToShoppingList.vue
```diff
@@ -237,30 +237,40 @@ export default defineComponent({
...
-            return sections;
+            currentTitle = ing.ingredient.title;
           }
 
-          // append new section if first
-          if (sections.length === 0) {
+          // If this is the first item in the section, create a new section
+          if (sections.length === 0 || currentTitle !== sections[sections.length - 1].sectionName) {
+            if (sections.length) {
+              // Add the on-hand ingredients to the last section
```
boc-the-git:
```suggestion
              // Add the on-hand ingredients to the previous section
```

Using `previous` instead of `last` is more intuitive to me; `last` gives an implication of "the final one"

boc-the-git (dismissed):
LGTM. Have some suggested changes to the comments, but happy for you to ignore if you feel strongly.

frontend/components/Domain/Recipe/RecipeDialogAddToShoppingList.vue
michael-genson: Yeah it's weird, I was surprised when I saw it for the first time

boc-the-git (approved):

PR #3875, opened 2024-07-10
fix: Follow redirects during scraping

zeskeertwee:
## What type of PR is this?

- bug

URL's from albert heijn such as `https://www.ah.nl/r/1195504` are shared from the albert heijn app, and fail to scrape because they return a 301 moved permanently which eventually leads to `https://www.ah.nl/allerhande/recept/R-R1195504/romige-linguine-met-brie-artisjok-en-parmaham`, which scrapes perfectly fine. This is fixed by telling `httpx` to follow redirects.
This allows users to directly import recepies from the app instead of having to first follow the redirects in a browser.

URL's like `https://www.ah.nl/r/1195504` from the albert heijn app fail to scrape

## Testing

The aformentioned URL now scrapes fine as the redirects are followed, as seen in the logs:
```
INFO     2024-07-10T14:23:26 - HTTP Request: GET https://www.ah.nl/r/1195504 "HTTP/1.1 301 Moved Permanently"
INFO     2024-07-10T14:23:27 - HTTP Request: GET https://www.ah.nl/allerhande/recept/R-R1195504 "HTTP/1.1 301 Moved Permanently"
INFO     2024-07-10T14:23:27 - HTTP Request: GET https://www.ah.nl/allerhande/recept/R-R1195504/romige-linguine-met-brie-artisjok-en-parmaham "HTTP/1.1 200 OK"
```


michael-genson:
LGTM, other than the lint issue. Once that's fixed we can merge

zeskeertwee:
Thanks, formatting has been fixed in 1cf28a659dc148ab81bb39120564b5e8351574cc

michael-genson (approved):
Looks good, thanks!

More info about the dismissed security advisory, copy-pasted for visibility:

This is part of the recipe scraper code so it's expected that we fetch user-supplied URLs. We don't authenticate these requests, so there is no elevated permissions, and we validate the HTML before importing a recipe

niekniek89:
i have this issue with the latest version also with dishes from ah.nl.
what can i do?

PR #3879, opened 2024-07-11
feat: Add environment variables to enable ingredients/steps placeholders

choikangjae:
## What type of PR is this?

_(REQUIRED)_

- documentation
- feature

_(REQUIRED)_

This PR allows seasoned users to disable placeholders when creating a recipe manually. When a user set `False` to environment variables, `ENABLE_INGREDIENTS_PLACEHOLDER` and `ENABLE_STEPS_PLACEHOLDER`, placeholders won't be placed. 

Screenshot_20240710_181746

Screenshot_20240710_192835

Default value is `True`, so not to change the app's previous behavior. And only texts will be affected, not cards even if they are disabled, so users don't need to click `+Add` to create the first ingredient or step.

N/A

## Testing

Ran `task py:check`

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -9,8 +9,8 @@ env:
   API_PORT: 9000
   API_DOCS: True
   TOKEN_TIME: 256 # hours
-  ENABLE_INGREDIENTS_HOLDER: True
-  ENABLE_STEPS_HOLDER: True
+  ENABLE_INGREDIENTS_PLACEHOLDER: True
+  ENABLE_STEPS_PLACEHOLDER: True
   # mailplit SMTP config
   # start dev:services to use mailpit
   SMTP_HOST: localhost
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -9,6 +9,8 @@ env:
   API_PORT: 9000
   API_DOCS: True
   TOKEN_TIME: 256 # hours
+  ENABLE_INGREDIENTS_HOLDER: True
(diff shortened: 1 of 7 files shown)


boc-the-git:
My first thought is whether an environment variable is really the best option here, or whether it should be a group setting :thinking: 
I also don't think we should necessarily have the step/ingredient created at all, if they're blank :thinking:

boc-the-git:
Merged mealie-next, which hopefully re-evaluates the PR title

choikangjae:
At first, I was trying to add these options as a preference from a user or an admin which can be set from a web UI, but I couldn't find a suitable place to put these options. Existed environment variables like `ALLOW_SIGNUP`, `DEFAULT_GROUP`, or `DAILY_SCHEDULE_TIME` made me think this is where the preferences go to. If it is not the place for them, could you instruct me where should I put them?

Secondly, I thought pre-made blank step/ingredient cards are convenient (to me), I have to add those anyway if I want to make a proper recipe. So it would be better not adding cards at all? It was convenient to add cards for my workflow, but since I don't know how the other users use it, I will follow your guidance.

Kuchenpirat:
i second bocs opinion that this is something we should implement via a group setting.

They can be found at `/group` or by clicking on your profile picture in the top left and then *Group Settings*

choikangjae:
It seems that group settings are the best suitable place for these options. Thanks for letting me know. I will open new PR after I changed it.

PR #3882, opened 2024-07-11
fix: Use env variable to get alembic config file in exporter

litchipi:
## What type of PR is this?

- bug

On any case where the alembic config file is not on expected `PROJECT_DIR + /alembic.ini` path, the restore code wouldn't work.
For this purpose, an `ALEMBIC_CONFIG_FPATH` env var was introduced in the code of `init_db`, but has not been used for the restore code, which is the purpose of this PR

Possible explanation of https://github.com/NixOS/nixpkgs/issues/325391

In the continuation of https://github.com/mealie-recipes/mealie/pull/3102


boc-the-git (changes requested):
Lacking an `import os` by looks of the CI.
Otherwise, this is probably fine to merge; it's all but identical to https://github.com/mealie-recipes/mealie/pull/3102

litchipi:
Yep my bad, sorry for that :sweat_smile:

boc-the-git (approved):

PR #3895, opened 2024-07-15
fix: prevent postgres credentials leak

hay-kot:
## What type of PR is this?

_(REQUIRED)_

- bug

I received a security report that in some cases the API could return the password of the Postgres database. This only showed up when the user was an administrator. I modified the logic of the public URL to not use replace logic and just construct a dummer representation of the db url. Additionally, if they are deploying with a custom Postgres URL we just return a place holder value. I think this is the best approach instead of trying to parse out the username and password. 

Internal security report

## Testing

- [x] Manual testing


michael-genson (approved):
Is there a reason we even need to include this in the API? Can we just return an empty string?

Approved since I think this is fine, but it may be simpler to just return no info unless we rely on it somewhere

PR #3914, opened 2024-07-20
fix: Convert Daily Schedule Time to UTC

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- bug
- cleanup
- documentation

_(REQUIRED)_

https://github.com/mealie-recipes/mealie/pull/3847 changed Mealie to use explicitly UTC on the server. However, previously we assumed the `DAILY_SCHEDULE_TIME` ENV var was local time, not UTC. This PR maintains that assumption and provides a parsed "UTC" version (stored in `DAILY_SCHEDULE_TIME_UTC`).

Docs are also updated to reflect this and make the assumption explicit.

_(REQUIRED)_

Fixes issue reported on Discord.

At first I thought we should just re-define daily schedule time to inherently mean UTC, but since this is being defined in the ENV vars (where we already need to define the TZ) I think it's clearer to define the daily time in local time, rather than UTC. That's also how we think about it naturally anyway (e.g. "run tasks at 11pm").

Another approach that could have worked was to do the scheduler calculation in local time (replacing `datetime.now(timezone.utc)` with the local tz instead), but IMO converting it to UTC upfront is a lot clearer (since everything else in the app uses UTC now).

## Testing

Added/updated logs to verify conversions are happening properly.

### DIFF
--- a/docs/docs/documentation/getting-started/installation/backend-config.md
+++ b/docs/docs/documentation/getting-started/installation/backend-config.md
@@ -4,20 +4,20 @@
 
 ### General
 
-| Variables                     |        Default        | Description                                                                         |
-| ----------------------------- | :-------------------: | ----------------------------------------------------------------------------------- |
-| PUID                          |          911          | UserID permissions between host OS and container                                    |
-| PGID                          |          911          | GroupID permissions between host OS and container                                   |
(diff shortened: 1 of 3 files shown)


boc-the-git (approved):
LGTM! Thanks @michael-genson

PR #3932, opened 2024-07-23
fix: task py:migrate description

Kuchenpirat:
## What type of PR is this?

- bug
- cleanup

The description of the `task py:migrate` is missing something. Cli arguments in Task files work only after `--` (see also [here](https://github.com/Kuchenpirat/mealie/blob/07b65a02ecee82cd2ef0c978cb3e96f278ae3295/Taskfile.yml#L102) where this is described correctly). This fixes the description and corrects the example command in the docs. 

None


michael-genson (approved):
Side note - the quotes around the message are actually optional (e.g. `task py:migrate -- your migration message` is valid) but having them doesn't change anything

PR #3933, opened 2024-07-23
feat: Seed support for plural units

Choromanski:
## What type of PR is this?

- feature
- l10n

This PR allows the seeder to import plural names and plural abbreviation for units.
This allows for new user have cleaner ingredient lists without having to do any configuration. 

N/A

In commit [21216a6](https://github.com/mealie-recipes/mealie/pull/3933/commits/21216a6954df6b27766101712c80185c766a3d8d) I made some changes I'd be happy to revert:
- changed the abbreviation from cup to c
- Added the units "bunch", "pack", and "pinch"

I modified one of the crowdin files, not sure if there is anything additional I need to do.

I would like to also do the same with food seeder but that would require restructuring the local files and I'm unsure if that would cause problems with crowdin.

## Testing

Ran a fresh install and all of the unit data was seeded properly:

image


michael-genson (approved):
Looks great, thanks!

Choromanski:
@michael-genson I would like to also do the same with food seeder but that would require restructuring the local files and I'm unsure if that would cause problems with crowdin.

michael-genson:
@Choromanski Hey sorry, meant to get back to you on this. In theory as long as they keys are the same, the translations should copy over. We should be able to fix it manually if not

PR #3950, opened 2024-07-28
fix: Reduce search tolerance on organizers page

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- bug

_(REQUIRED)_

Currently it's pretty much impossible to search for anything on the organizers page due to false positives. Here's an example:

2024-07-27_20h15_07

Here's what it looks like with the reduced tolerance:

2024-07-27_20h15_30

It's still reasonably forgiving on minor typos.

_(REQUIRED)_

N/A

## Testing

Manually


boc-the-git (approved):
Don't see a reason not to give this a go!

PR #3958, opened 2024-07-29
feat: If there's only one shopping list, navigate directly to it

boc-the-git:
## What type of PR is this?

- feature

Similar to https://github.com/mealie-recipes/mealie/pull/3764, and using some of the enhancements Michael added in https://github.com/mealie-recipes/mealie/pull/3768, this will now take you directly to the only list when you are adding a recipe to a shopping list.
Works by any method of adding to list, as the implementation is on the "add to list" pop up itself.

I've not recorded a gif.. I don't have software set up currently to do so! :see_no_evil: Need to rectify that.

None.

## Testing

Manual


michael-genson:
Will this redirect you if you uncheck the "show all" box? e.g. you have "show all" toggled, then you toggle it (so only one list shows). We probably only want to check the list length once

boc-the-git:
@michael-genson good call. This latest commit should address that.

michael-genson (approved):

PR #3986, opened 2024-08-04
feat: Added fr-BE 

TheSuperBeaver:
## What type of PR is this?
- feature

There is a small difference between french from France, and french from Belgium
In France : 
breakfast : "Petit-déjeuner",
lunch: "Déjeuner",
dinner: "Dîner"

In Belgium : 
breakfast: "Petit-déjeuner",
lunch: "Dîner",
dinner: "Souper"

The lang should be created in crowdin also

## Testing

Switched to language, and verified that dinner is now "Souper"


michael-genson (changes requested):
Hey there!

Can you remove the added `fr-BE.json` file from this PR? It was added in: https://github.com/mealie-recipes/mealie/pull/3995

Otherwise looks good!

PR #4000, opened 2024-08-07
feat: fr-BE addition

TheSuperBeaver:
## What type of PR is this?
- feature

There is a small difference between french from France, and french from Belgium
In France : 
breakfast : "Petit-déjeuner",
lunch: "Déjeuner",
dinner: "Dîner"

In Belgium : 
breakfast: "Petit-déjeuner",
lunch: "Dîner",
dinner: "Souper"

The lang should be created in crowdin also

## Testing

Switched to language, and verified that dinner is now "Souper"

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -1594,13 +1594,13 @@ signedtoken = ["cryptography (>=3.0.0)", "pyjwt (>=2.0.0,<3)"]
 
 [[package]]
 name = "openai"
-version = "1.40.0"
+version = "1.40.1"
 description = "The official Python library for the openai API"
 optional = false
 python-versions = ">=3.7.1"
 files = [
-    {file = "openai-1.40.0-py3-none-any.whl", hash = "sha256:eb6909abaacd62ef28c275a5c175af29f607b40645b0a49d2856bbed62edb2e7"},
-    {file = "openai-1.40.0.tar.gz", hash = "sha256:1b7b316e27b2333b063ee62b6539b74267c7282498d9a02fc4ccb38a9c14336c"},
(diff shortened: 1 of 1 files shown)


michael-genson (changes requested):
Changes to the language have to go through Crowdin; you can't directly modify them here. We do need to update the other files to make sure it works properly. Can you remove the changes to fr-BE.json?

dev/code-generation/gen_ts_locales.py
```diff
@@ -35,6 +35,7 @@ class LocaleData:
     "es-ES": LocaleData(name="Español (Spanish)"),
     "fi-FI": LocaleData(name="Suomi (Finnish)"),
     "fr-FR": LocaleData(name="Français (French)"),
+    "fr-BE": LocaleData(name="Français (Belgique)"),
```
michael-genson: The convention we're using here is "Localized Name (English Name)". Something like "Belgique Français (Belgian French"). I don't speak French so that translation is probably wrong, but you get the idea.

frontend/composables/use-locales/available-locales.ts
```diff
@@ -162,6 +162,12 @@ export const LOCALES = [
     progress: 100,
     dir: "ltr",
   },
+  {
+    name: "Français (Belgium)",
```
michael-genson: Same here, should match `gen_ts_locales.py`

michael-genson (changes requested):
Two minor changes, then we can get this merged. Appreciate you sticking with it!

dev/code-generation/gen_ts_locales.py
TheSuperBeaver: I'm gonna go with "Belge (Belgian)" then :)

PR #4057, opened 2024-08-20
docs: fix import bookmarklet trailing slashes

vahtos:
## What type of PR is this?

documentation

Fixes the case where the recipe URL has a trailing slash. Previously, that would make the keywords and edit settings be ignored, because the resulting URL was malformed.

## Testing

Example url with trailing slash: https://www.allrecipes.com/recipe/8849/baked-chicken-nuggets/


michael-genson (approved):
Looks good, thanks!

PR #4072, opened 2024-08-21
fix: removed version line from docker compose yml files

volumedata21:
## What type of PR is this?

- documentation

Removed line 1 "_version: "3.4"_" in docker compose files. Version is now obsolete in Docker Compose and will throw a warning when launching with docker compose up.


Kuchenpirat (approved):

Kuchenpirat:
Thanks for the easy PR 😊

PR #4088, opened 2024-08-23
fix: Image height in home screen mobile format

p0lycarpio:
## What type of PR is this?

_(REQUIRED)_

- bug

_(REQUIRED)_

Adding recipe tags to recipe cards #3864 can make the recipe card bigger and create a blank space below the image.
This PR fixes this bug by filling the height of the card with the image.

**Before:**
Capture d’écran du 2024-08-23 18-37-37

**After:**
Capture d’écran du 2024-08-23 18-37-27

Also the button for toggling the mobile card view was fixed using another breakpoint.

_(REQUIRED)_

N/A


michael-genson:
LGTM. The other places this gets used are the normal (non-condensed) view, and the meal planner edit view. Would you be able to share screenshots of the differences there? (if any)

p0lycarpio:
Other locations like meal planner, recipes timeline and search results have not changed because the default value (125px) has been set for the image size. Normally only the home screen uses the 'fill-height' property.

michael-genson (approved):
Ah yep you're right, thanks!

Issue #4081, opened 2024-08-22
[Task] - Enable Cross-Household Recipes


michael-genson:
### What is the problem this task addresses?

Currently logged-in users can only see their household's recipes (similar to how it was with groups). One of the main reasons for adding households was specifically to have a common pool of recipes.

### Proposed/Possible Solution(s)?

Features to address:
- [x] View all recipes should include all households' recipes
- [x] Household should be filterable when viewing all recipes (probably a multi-select)
- [x] Shopping lists should support recipes from other households
- [x] Meal Plans should support recipes from other households
- [x] Meal Plan Rules should include a household filter (probably a multi-select)
- [x] User actions should support other households' recipes, including:
  - [x] Favorites
  - [x] Comments
  - [x] Timeline Events

PR #4089, opened 2024-08-23
feat: Cross-Household Recipes

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- feature

_(REQUIRED)_

Partial implementation of https://github.com/mealie-recipes/mealie/issues/4081. This PR allows users to see recipes from other households. More specifically:
- When searching, filtering, etc. _all_ recipes are shown, rather than just your households' recipes
- You can "make" the recipe (i.e. add a timeline event and update the last made date)
- You can add the recipe to your shopping list
- You can add the recipe to your meal plan (only when added directly, see below)
- You can rate the recipe (and your rating is factored into the global "average" rating)
- You can favorite the recipe (and it will appear on your favorites list)
- You can comment on the recipe

There are a few restrictions:
- You cannot edit recipes if they don't belong to your household. There is no way to override this (not even admins can do this, and they are not on the data management page)
- There's no way to _not_ see all recipes (I plan to address this separately, the API supports this)
- These recipes are not considered when choosing a random meal plan recipe (I plan to address this separately)
- These recipes are not included in cookbooks (I don't have plans to change this unless there's community interest)

_(REQUIRED)_

Partial implementation of https://github.com/mealie-recipes/mealie/issues/4081

This implementation assumes we're never going to allow users to hide recipes from other household members. Users _can_ hide the entire household from public users. Supporting hiding from other household users would add an immense amount of complexity (e.g. what happens if a recipe on my shopping list suddenly becomes hidden?), so personally I never see that being an option.

## Testing

Pytest for backend. Checked around the frontend a bunch and everything works as expected.


frontend/components/Domain/Recipe/RecipePage/RecipePageParts/RecipePageHeader.vue
```diff
@@ -100,6 +100,31 @@ export default defineComponent({
...
+        return false;
+      }
+
+      // Check recipe
+      if (props.recipe.settings.locked) {
+        return false;
+      }
+
+      return true;
+    })
```
hay-kot: would it be useful to move this into either a function or composable? I'd assume we'd need this in other paces, like dropdown menu's and what not? Moving it into a function would also make it easier to test the logic here.

mealie/services/recipe/recipe_service.py
```diff
@@ -46,6 +47,9 @@ def __init__(self, repos: AllRepositories, user: PrivateUser, household: Househo
         if repos.household_id != user.household_id != household.id:
             raise Exception("household ids do not match")
 
+        self.group_recipes = get_repositories(repos.session, group_id=repos.group_id, household_id=None).recipes
+        """Recipes repo without a Household filter"""
```
hay-kot, line 51: 🔥

hay-kot (dismissed):
LGTM, one small comment on the frontend stuff, but not a blocker. Leave it to you.

hay-kot (approved):

PR #4130, opened 2024-08-31
docs: boolean value in docker-compose examples not deserializable in env variables

ursuscamp:
## What type of PR is this?

- documentation

Booleans values in docker-compose environment variables are not deserializable and results in an error like this

```
ERROR: The Compose file './docker-compose.yaml' is invalid because:
services.mealie.environment.ALLOW_SIGNUP contains false, which is an invalid type, it should be a string, number, or a null
```

N/A


michael-genson:
Are you using `docker compose` or `docker-compose`? The latter is deprecated.

Otherwise, make sure docker is up to date.

michael-genson (changes requested):
Thanks! I noticed it works fine in mine without quotes, but I think it's because I use list syntax instead of map syntax? Maybe it's undefined behavior? Who knows, I can barely find docs about it.

Would we be able to use double quotes? Only because we already have the port in double quotes.

docs/docs/documentation/getting-started/installation/postgres.md
```diff
@@ -20,7 +20,7 @@ services:
       - mealie-data:/app/data/
     environment:
       # Set Backend ENV Variables Here
-      ALLOW_SIGNUP: false
+      ALLOW_SIGNUP: 'false'
```
hay-kot:
```suggestion
      ALLOW_SIGNUP: "false"
```

docs/docs/documentation/getting-started/installation/sqlite.md
```diff
@@ -24,7 +24,7 @@ services:
       - mealie-data:/app/data/
     environment:
       # Set Backend ENV Variables Here
-      ALLOW_SIGNUP: false
+      ALLOW_SIGNUP: 'false'
```
hay-kot:
```suggestion
      ALLOW_SIGNUP: "false"
```

hay-kot (changes requested):

hay-kot (approved):

michael-genson (approved):

PR #4147, opened 2024-09-03
feat: Automate release drafting

boc-the-git:
## What type of PR is this?

- feature

This PR will create draft releases for us based on the merged pull requests. It writes the release notes in a configurable format.
This does not automatically perform the release! We still maintain full, manual control over that.

I based a bunch of this on Frenck's Spook repo as I loved his release notes: https://github.com/frenck/spook/tree/main
I expect in future we might also want to add some form of validation on PRs having appropriate labels (like [this](https://github.com/frenck/spook/blob/main/.github/workflows/pr-labels.yaml)), but I will work on that separately.

None.

My suggestion is we merge this in with the defaults I've put in place and then we can tweak when we're seeing it in practice. I am less wedded to the labels suggested though.

## Testing

In my fork. 
image


.github/release-drafter.yml
```diff
@@ -0,0 +1,55 @@
...
+      - "breaking-change"
+      - "major"
+  - title: "✨ New features"
+    labels:
+      - "feature"
+      - "minor"
+  - title: "🐛 Bug fixes"
+    labels:
+      - "bugfix"
+  - title: "🧰 Maintenance"
```
michael-genson, line 19: Can we add `collapse-after: 3` here? Since Crowdin usually floods the release notes

```
- title: "🧰 Maintenance"
  collapse-after: 3
  labels:
    - ...
```

.github/release-drafter.yml
```diff
@@ -0,0 +1,55 @@
...
+    labels:
+      - "ci"
+      - "chore"
+      - "l10n"
+  - title: "📚 Documentation"
+    labels:
+      - "documentation"
+  - title: "⬆️ Dependency updates"
```
michael-genson, line 28: Similarly here, I imagine most folks don't care about dependency bumps

.github/release-drafter.yml
```diff
@@ -0,0 +1,55 @@
...
+      - "minor"
+  - title: "🐛 Bug fixes"
+    labels:
+      - "bugfix"
+  - title: "🧰 Maintenance"
+    labels:
+      - "ci"
+      - "chore"
+      - "l10n"
+  - title: "📚 Documentation"
```
michael-genson, line 25: Maybe here too, although we don't usually have tons of docs updates, so maybe not

michael-genson (commented):
Looks sweet. I take it this means we need to label PRs before merging them, yeah?

boc-the-git:
> Looks sweet. I take it this means we need to label PRs before merging them, yeah?

Technically they can be labeled after too, but the release drafter won't run again until the next push to mealie-next.
That's why I think in future we probably want a PR action validating that we have a valid label.

I've added the `collapse-after` tag in a couple places, that's one of the things I knew I'd want to add in future but wanted to see how a v0.1 of this looked in practice.
We might want to just separate crowdin PRs into their own category or possibly hide them altogether.. 🤔

.github/release-drafter.yml
boc-the-git: Chose not to do this

michael-genson (approved):
> We might want to just separate crowdin PRs into their own category or possibly hide them altogether

Yeah I can see hiding them making sense, but we can always do that later. Approved, left for you to merge in case you had any last minute changes!

PR #4154, opened 2024-09-04
chore: Set 'chore' label on PR

boc-the-git:
## What type of PR is this?

- cleanup

Sets the 'chore' label on the pull request created for pre commit hooks.

None.

Ties in to the release-drafter PR, https://github.com/mealie-recipes/mealie/pull/4147. And in the same vein as https://github.com/mealie-recipes/mealie/pull/4153


boc-the-git:
FYI @michael-genson , as the original work here was yours. Chore makes a bit more sense to me. Any concerns?

michael-genson (approved):
Yeah this makes sense to me. I did "fix" because renovate bot uses fix to bump dependencies, but I think chore makes more sense here.

PR #4158, opened 2024-09-04
feat: Additional Household Permissions

### LINKED ISSUE #4082: [Task] - Household Permissions

Our current permissions model doesn't account for households at all. You're pretty much either a super admin (cross-group), an elevated user, or a regular user.

Separate permissions into a few different buckets:
- Group Admin; today's admin, basically a super admin, can manage groups, households for all groups, users for all groups/households, and group data (foods, units, etc.).
- Household Admin; can manage their own household, as well as all users in the household
- User; maintain existing permissions (e.g. manage, organize, invite)

We have two different ways we can approach this:
1. To keep things simple, we can probably just add a user-level permission that lets them do household management tasks (similar to existing user permissions)
2. Otherwise, we probably want to develop more isolated roles, which would be a bigger task, but set us up for easier permissions management down the line.

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- bug
- cleanup
- feature

_(REQUIRED)_

This PR adds two household-related permissions:
- Household Preference: Lock Recipe Edits From Other Households
- User Permission: Can Manage Household

### Lock Recipe Edits From Other Households
This flag lets households decide if other households can only view their recipes, or if they can edit them. With this unflagged, any user from any household in the group can edit their recipes.

When set, it behaves like it was implemented in https://github.com/mealie-recipes/mealie/pull/4089 (users from other households can't edit). When unset, edits (including deletes) are allowed (this is a new feature from this PR).

### Can Manage Household
This flag allows users to manage household settings (such as the new lock recipe edits flag). Previously this was tied to the generic `Can Manage` permission, but it makes sense to want users to be able to manage their own households, but not necessarily manage the group.

---

Both flags are added to their respective settings pages:
image
image
image
image

I also merged the two household settings pages; one was using a component, and one was just baked into the page, so I replaced the page one with the component. The component was missing all of the help text, so I re-wrote it to support the help text.

For reference, it currently looks like this:
image

And now it looks like this:
image

_(REQUIRED)_

Closes https://github.com/mealie-recipes/mealie/issues/4082

I added a new route to fetch a household by id, since we need to check permissions of the recipe's household (which isn't always your household anymore). I also fixed the inconsistency where getting a recipe by id works, but updating/deleting only works by slug. All recipe routes should accept either slugs or ids now.

## Testing

frontend and backend tests


frontend/lang/messages/en-US.json
```diff
@@ -241,13 +241,14 @@
...
     "manage": "Manage",
+    "manage-household": "Manage Household",
     "invite": "Invite",
     "looking-to-update-your-profile": "Looking to Update Your Profile?",
     "default-recipe-preferences-description": "These are the default settings when a new recipe is created in your group. These can be changed for individual recipes in the recipe settings menu.",
     "default-recipe-preferences": "Default Recipe Preferences",
     "group-preferences": "Group Preferences",
     "private-group": "Private Group",
-    "private-group-description": "Setting your group to private will default all public view options to default. This overrides any individual households or recipes public view settings.",
+    "private-group-description": "Setting your group to private will default all public view options to default. This overrides any individual households or recipes public view settings",
```
Kuchenpirat:
```suggestion
    "private-group-description": "Setting your group to private will default all public view options to private. This overrides any individual households or recipes public view settings",
```

Aside from the error that it defaults to default this is a bit confusing in general. Is it defaulting to private and gets overwritten by individual recipe settings. Or does it set all recipes to private and it cannot be overwritten by individual settings.

frontend/components/Domain/Recipe/RecipePage/RecipePageParts/RecipePageHeader.vue
```diff
@@ -64,12 +64,14 @@
 <script lang="ts">
 import { defineComponent, useContext, computed, ref, watch } from "@nuxtjs/composition-api";
 import { useLoggedInState } from "~/composables/use-logged-in-state";
+import { useUserApi } from "~/composables/api";
 import { useRecipePermissions } from "~/composables/recipes";
 import RecipeRating from "~/components/Domain/Recipe/RecipeRating.vue";
 import RecipeLastMade from "~/components/Domain/Recipe/RecipeLastMade.vue";
 import RecipeActionMenu from "~/components/Domain/Recipe/RecipeActionMenu.vue";
 import RecipeTimeCard from "~/components/Domain/Recipe/RecipeTimeCard.vue";
 import { useStaticRoutes } from "~/composables/api";
```
Kuchenpirat:
```suggestion
import { useRecipePermissions } from "~/composables/recipes";
import RecipeRating from "~/components/Domain/Recipe/RecipeRating.vue";
import RecipeLastMade from "~/components/Domain/Recipe/RecipeLastMade.vue";
import RecipeActionMenu from "~/components/Domain/Recipe/RecipeActionMenu.vue";
import RecipeTimeCard from "~/components/Domain/Recipe/RecipeTimeCard.vue";
import { useStaticRoutes, useUserApi  } from "~/composables/api";
```

Dupicate Import from "~/composables/api"

Kuchenpirat (commented):
This was alot, but luckily most of it was pretty self-explanatory. Just a few mini-comments. Did not find anything major. Will do a bit more real-world testing now but i think we are looking good 👍

frontend/lang/messages/en-US.json
michael-genson: I changed this to:
> Setting your group to private will disable all public view options. This overrides any individual public view settings

Which I think works better. There's also the household variant:
> Setting your household to private will disable all public view options. This overrides any individual public view settings

Kuchenpirat: Works for me 👍

Kuchenpirat (approved):
Looks good 🚀

PR #4163, opened 2024-09-05
feat: Use Backend for Recipe Post Actions

### LINKED ISSUE #3936: [BUG] - Recipe Actions don't work with Home Assistant due to incorrect content type header
### First Check

Due to how [recipe actions are implemented](https://github.com/mealie-recipes/mealie/pull/3448/files#diff-496ea00629c62a10748ecc0e0d481b7d4fcd0c32996d11e33e99fcd47aa54d96R68-R71), we currently send `POST` actions with the `text/plain` content type header. This is to get around a CORS preflight issue (since we might not get a response back).

Home Assistant requires using `application/json` to receive the webhook data. It may be as simple as having the backend do the `POST` request (instead of the frontend).

### Steps to Reproduce

1) Create a recipe `POST` action
2) Send the request to HA
3) Observe that the content can't be read (https://www.home-assistant.io/docs/automation/trigger/#webhook-trigger)

N/A

### Mealie Version

nightly

### Deployment

Docker (Synology)

_No response_

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- bug
- feature

_(REQUIRED)_

This leverages the backend to make `POST` requests for recipe actions. The previous implementation had issues requiring us to send the content type `text/plain`. Using the backend we get around this. Consequently, it's also a lot faster for the user since we don't have to wait for the response from the external server (the backend uses background tasks from FastAPI so it can return a `202` response immediately).

Since we now send the correct content header (`application/json`) it should work nicely with Home Assistant. Since this feature was already slightly broken because of the content header, I took this time to improve the actual payload to include the recipe action. This results in a breaking change, though a minor one (since it was impossible to extract the body previously anyway in a lot of applications). We also don't use merge fields in the POST type, since we're already sending all the data in the payload anyway.

_(REQUIRED)_

Fixes https://github.com/mealie-recipes/mealie/issues/3936

## Testing

Pytest backend, manual frontend


frontend/components/Domain/Recipe/RecipeContextMenu.vue
```diff
@@ -376,7 +376,7 @@ export default defineComponent({
       const response = await groupRecipeActionsStore.execute(action, props.recipe);
 
       if (action.actionType === "post") {
-        if (!response || (response.status >= 200  && response.status < 300)) {
+        if (!response?.error) {
```
boc-the-git, line 379: Is there ever a scenario where an `error` output could come back despite say a 200 code? I'd hope not, but just asking..

boc-the-git (approved):
LGTM! :rocket:

frontend/components/Domain/Recipe/RecipeContextMenu.vue
michael-genson: Thankfully not, the error is always empty for 2xx responses

PR #4183, opened 2024-09-09
fix: Broken Social Preview Links

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- bug

_(REQUIRED)_

Households broke the social preview links due to changes to how we access repos. This PR fixes them and adds a few tests.

_(REQUIRED)_

N/A


tests/integration_tests/test_spa.py
```diff
@@ -63,8 +101,91 @@ def test_spa_recipe_json_injection():
...
+
+    set_group_is_private(unique_user, is_private=True)
+    set_recipe_is_public(unique_user, recipe, is_public=True)
+    response = await spa.serve_recipe_with_meta(
+        user.group_slug, recipe.slug, user=None if use_public else user, session=unique_user.repos.session
+    )
+    if use_public:
+        assert response.status_code == 404
+    else:
+        assert response.status_code == 200
```
boc-the-git: Is this implying that a private group's public recipe, is not public? Is that the intention?

boc-the-git (dismissed):
I'm not super familiar with this area of the code base, but LGTM!

One question.

tests/integration_tests/test_spa.py
michael-genson: Yeah that's right, since private group trumps public recipe.

I updated the variable names, "use_public" is confusing

boc-the-git (approved):

PR #3970, opened 2024-07-31
feat: Add Households to Mealie

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- feature

## What this PR does / why we need it:

_(REQUIRED)_

Base implementation for the households feature: https://github.com/mealie-recipes/mealie/discussions/2134. This feature is massive and touches pretty much everything. I will explain the high-level changes, as well as break down into some more specifics.

### What is a household?

A household is a subdivision of groups with a collection of users. While groups can be considered as completely separate instances, households mix their data with each other. Another way to think of a household is a family: your group may contain different families, each with their own shopping lists, meal plans, and recipes.

These previously group-level features now belong to a household:
- Notifiers
- Webhooks
- Cookbooks
- Recipe Actions
- Users
- Shopping Lists (tied to a user)
- Meal Plans (tied to a user) and Meal Plan Rules (not tied to a user)
- Recipes (tied to a user)

Of note, these features are not tied to a household, and are shared amongst households:
- Organizers (Tags, Categories, Tools)
- Ingredient Stores (Foods, Units, Labels)

### What does it mean for a feature to belong to a household?

Features that belong to households (e.g. Notifiers) are only relevant in a household context. This means:
- When querying for documents, only those that belong to _your_ household are returned
- When performing actions (e.g. editing a recipe) only documents in your household are affected (e.g. if a recipe is edited in someone else's household, your notifiers don't fire)
- If the feature is tied to a user (e.g. Shopping Lists), if that user switches households, so does that feature. For instance, if I own a shopping list, only my household can see it. If I switch households, my _old_ household can't see the shopping list anymore

### What are some common use-cases this solves?

Generally speaking, families typically don't prepare meals for other families. This means they shop for their own family, plan meals for their own family, have their own rules on how meal plans work, etc. This allows each family to have their own shopping lists and meal plans, as well as their own automations (i.e. webhooks, notifiers, and actions).

However, in terms of data, there's no reason to have separate stores for foods, units, tags, etc. So those are all shared amongst families. Families also typically want to share recipes with each other, so everyone has their pool of recipes and cookbooks that they draw from to create their meal plans, shopping lists, etc.

### What use-cases aren't solved by this PR? What are the limitations?

Despite the ridiculous diff, this PR _attempts_ to limit its scope and leave some things for future PRs (see tasks in the special notes section below). There are known/intentional limitations to this implementation (some to be addressed in future PRs):
[...]

PR #4186, opened 2024-09-09
feat: Allow Cookbooks To Share Names

### LINKED ISSUE #4083: [Task] - Allow Cookbooks to Share the Same Name

Currently cookbooks will throw an error if there's another cookbook with the same name. Similar to recipes, we should allow this as long as the slug is different.

We want this since two different households might have the same common cookbook name (e.g. Drinks). Since cookbooks are tied to households, they may not know a cookbook name is taken already.

- Make sure the API handles duplicate names correctly (e.g. incrementing the slug, like it does for recipes)
- Make sure the frontend handles this fine

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- feature

_(REQUIRED)_

This PR allows cookbooks to share the same name. We want this since cookbooks are per-household, and multiple households may want the same cookbook name (e.g. "Sides").

When multiple cookbooks share the same name, they are assigned different slugs (so slugs are still unique, which we enforce at the DB layer [in a previous PR](https://github.com/mealie-recipes/mealie/pull/3970)). The logic is similar to how duplicate recipes are handled (although recipes de-dupe name, too). This PR also makes it so when a cookbook's name is updated, so is its slug (so it matches the name).

_(REQUIRED)_

Closes https://github.com/mealie-recipes/mealie/issues/4083

Technically we're not stopping users from creating the same cookbook name in one household, so it is possible to do this:
image

I don't really see this as a problem, just figured I'd point it out.

## Testing

Pytest for the new cookbooks repo with custom dedupe logic. I also confirmed the frontend handles this fine with cookbook CRUD operations (both in the same household and on two different households).


mealie/repos/repository_cookbooks.py
```diff
@@ -0,0 +1,65 @@
...
+    def create_many(self, data: Iterable[ReadCookBook | dict]) -> list[ReadCookBook]:
+        return [self.create(entry) for entry in data]
+
+    def update(self, match_value: str | int | UUID4, data: SaveCookBook | dict) -> ReadCookBook:
+        if isinstance(data, dict):
+            data = SaveCookBook(**data)
+
+        new_slug = slugify(data.name)
+        if not (data.slug and data.slug.startswith(new_slug)):
+            data.slug = slugify(data.name)
```
boc-the-git: [question] Because this is `startswith` rather than an exact match, could that lead to problems?

boc-the-git (dismissed):
LGTM :rocket:

One question.

mealie/repos/repository_cookbooks.py
michael-genson: I was going to answer why this logic makes sense, but then I realized using regex makes way more sense here, so I updated it.

We can still run into weird behavior if the cookbook name actually contains a number in the title (e.g. if you rename a cookbook from "My Cookbook" to "My Cookbook 1" the slug won't update) but this is a tiny edge case, and it doesn't really cause any problems anyway besides the slug being slightly different than the name

boc-the-git (approved):

PR #4192, opened 2024-09-10
chore: Improve Alembic Migration Generation

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- cleanup

_(REQUIRED)_

This PR modifies Alembic to make generating migrations easier. It makes the following changes:
- Adds some type hints to make the linter quit whining (even though we don't actually enforce these in migration scripts)
- Modifies the migration template to output linter-friendly migrations
- Adds exclusions to skip known false migrations (e.g. removing unique constraints that we don't want to remove)
- Defaults to batch mode (so migrations _should_ work as-is for SQLite)

I also updated the `database-changes` docs to remove the blurb about batch operations, since that's now done automatically.

_(REQUIRED)_

N/A

## Testing

I created a migration using these changes and confirmed:
1. The migration is empty (no undesired changes)
2. The migration is fully formatted/linted


alembic/versions/2024-03-18-02.28.15_d7c6efd2de42_migrate_favorites_and_ratings_to_user_.py
```diff
@@ -202,8 +202,6 @@ def downgrade():
     )
     op.drop_index(op.f("ix_recipes_rating"), table_name="recipes")
     op.alter_column("recipes", "rating", existing_type=sa.Float(), type_=sa.INTEGER(), existing_nullable=True)
-    op.create_unique_constraint("ingredient_units_name_group_id_key", "ingredient_units", ["name", "group_id"])
-    op.create_unique_constraint("ingredient_foods_name_group_id_key", "ingredient_foods", ["name", "group_id"])
```
michael-genson, line 206: Other than type hints, this is the only change made to actual migration code. This is in the downgrade method though, which we don't use. These lines shouldn't be there.

boc-the-git:
Is there any regression testing that can be/was done on this?
I haven't fully reviewed it as yet (from a skim, the changes look mostly trivial), but I note your testing notes mostly relate to forward looking issues.

alembic/script.py.mako
boc-the-git, line 1: If it doesn't cause a problem, I don't care about it!
Just want to highlight that most of the actual "versions" files lack a type against `down_revision`, whereas they specify it for `branch_labels` and `depends_on`

boc-the-git (approved):
LGTM - will be great to have Alembic be a smoother process! :rocket: 

Couple comments.

alembic/script.py.mako
michael-genson: Yup, it's fine because they're all strings and have values. The reason I added types to the other ones is because they can be null (and usually are), and that's when the linter gets mad. It doesn't care if it's a string

michael-genson:
Regarding regression testing - the closest thing we have is some migration tests. We have test data from various migrations which are all run against the existing migration scripts, and we check some known previous data issues.

So since test pass we know older versions can still migrate to the latest version, and that the data isn't messed up (in specific instances)

PR #4200, opened 2024-09-12
feat: Reorder ShoppingListItemEditor

Kuchenpirat:
## What type of PR is this?

- cleanup

The ShoppingListItemEditor field order does not make much sense from an input sense (Especially for the Food input)

The current order to add a food is: Food ➡️ Unit ➡️ Note ➡️ Quantity

I reordered it to use a more logical input order of: Quantity ➡️ Unit ➡️ Food ➡️ Note, while also keeping Quantity, Unit and Food in one row. 
This has the added benefit that it is way easier to read the edit field is now matching the displayed order of the shopping list making edits of existing field more intuitive. 

The current order to add a Note is: Note ➡️ Quantity which i changed arround to have it match the shoping list item. 

Screenshots 

| Type | Old | New |
| --- | --- | --- |
| Food |  image| image |
| Note | image  | image |

None

## Testing

Manual testing, i pretty much just reordered things so it **should** be fine.


michael-genson (approved):
Me looks good to!

PR #4207, opened 2024-09-14
feat: Filter Recipes By Household (and a ton of bug fixes)

michael-genson:
Malcom in the Middle Changing a Lightbulb

## What type of PR is this?

_(REQUIRED)_

- bug
- cleanup
- feature

_(REQUIRED)_

This PR adds a household filter to the recipe search. It's a radio select, so you can only select one household or no households. This lets users find recipes based on a specific household (e.g. show me only my household's recipes):
image

Note there's a new "Clear Selection" button there too. You can click on the radio button a second time to deselect it, but I thought an explicit clear button would be easier to understand. This households filter only appears if there's two or more households to choose from (otherwise the filter would be useless). As expected, on public instances, only non-private households are shown here.

You wouldn't think this would be a pain to implement... but boy howdy did I uncover so many layers of jank. In addition to implementing the filter on the recipes page, here are all the things I changed on the frontend:
- Stores (food store, unit store, etc.) are now generic, so we can easily add more of them with little code reuse (consequently this fixed a reactivity bug on the food data management page)
- Data management tables use explicit initial sorting (because the generic stores made changes look jank)
- Fixed household actions mixing admin and non-admin routes
- Fixed filter reactivity
  - filters were getting overwritten if you tried changing them while recipes were loading from a previous filter change
  - recipes seem to jump around after selecting new filters, before they're all fully loaded
- Fixed infinite scroll breaking when you change filters (try adding filters such that there are only a few recipes, then change back to no filters; you can't get past the first page of recipes anymore)
- Converted the recipe favorites page to use lazy recipes like the rest of the app does

And on the backend:
- re-organized some household routes to make more sense (i.e. match the store/actions pattern)
- added a households filter to recipe get_all routes
- fixed cookbooks returning recipes from multiple households (they should be limited to only their own household)

_(REQUIRED)_

Partially implements https://github.com/mealie-recipes/mealie/issues/4081

I swear this PR wasn't supposed to be very big

## Testing

Frontend manual testing; added a bunch to pytest


frontend/components/Domain/Recipe/RecipeCardSection.vue
```diff
@@ -69,50 +69,52 @@
...
-              :rating="recipe.rating"
-              :image="recipe.image"
-              :tags="recipe.tags"
-              :recipe-id="recipe.id"
-            />
-          </v-lazy>
-        </v-col>
-      </v-row>
+    <div v-if="recipes && ready">
+      <div class="mt-2">
```
Kuchenpirat, line 73: Any reason to add this to the new div, doesn't really matter, just interested

frontend/components/Domain/Recipe/RecipeCardSection.vue
```diff
@@ -261,8 +269,12 @@ export default defineComponent({
 
     async function initRecipes() {
       page.value = 1;
+      hasMore.value = true;
+
+      // we double-up the first call to avoid a bug with large screens that render
+      // the entire first page without scrolling, preventing additional loading
       const newRecipes = await fetchRecipes(2);
```
Kuchenpirat: Not part of the commit but since you added an explanation i thought you had your problems determining why. I think this makes it a bit more clear what is happening here by not having a "random" 2 as value. 

```suggestion
      const newRecipes = await fetchRecipes(page.value + 1);
```

frontend/components/Domain/Recipe/RecipeDataTable.vue
```diff
@@ -3,6 +3,8 @@
     v-model="selected"
     item-key="id"
     show-select
+    sort-by="dateAdded"
+    sort-desc
```
Kuchenpirat, line 7: Makes sense to sort this way, but as far as i can see this was already the case.
Was this just because the api defaulted to this?

Anyways, propably not bad to have this setup in case something changes in the backend.

frontend/pages/group/data/foods.vue
```diff
@@ -241,6 +241,7 @@
         {icon: $globals.icons.delete, text: $tc('general.delete'), event: 'delete-selected'},
         {icon: $globals.icons.tags, text: $tc('data-pages.labels.assign-label'), event: 'assign-selected'}
       ]"
+      initial-sort="name"
```
Kuchenpirat: for my usecase i would prefer it to be sorted by dateAdded desc (like recipes) because i often create ingredients during recipe creation (like most people i guess) and want to come later to the food screen to add multiples, labels etc. To have them all at the top, without searching is really practical for this. 

This might also be a case to add dateAdded in the v-data-table but unselect it by default. Especially with the foods.

frontend/components/Domain/SearchFilter.vue
```diff
@@ -11,28 +11,42 @@
...
+            <v-switch
+              v-if="requireAll != undefined"
+              v-model="requireAllValue"
+              dense
+              small
+              hide-details
+              class="ml-2 my-auto"
+              :label="`${requireAll ? $tc('search.has-all') : $tc('search.has-any')}`"
+            />
+          </div>
```
Kuchenpirat: Love the clear selection button 😍 
i think having it on the right side would separate it better from the any/all button

```suggestion
          <div class="d-flex py-4">
            <v-switch
              v-if="requireAll != undefined"
              v-model="requireAllValue"
              dense
              small
              hide-details
              class="my-auto"
              :label="`${requireAll ? $tc('search.has-all') : $tc('search.has-any')}`"
            />
            <v-spacer />
            <v-btn
              small
              color="accent"
              class="mr-2 my-auto"
              @click="clearSelection"
            >
              {{ $tc("search.clear-selection") }}
            </v-btn>
          </div>
```

frontend/composables/partials/use-store-factory.ts
Kuchenpirat, line 1: Well that just eliminated a hell of a lot of redundant code 😍

Kuchenpirat (commented):
I think i managed to have my review match the randomness of the gif. 

Frontend is all i manage today. Will try to finish the rest  tomorrow. 

Its all just little comments. No obvious bugs yet. But will do some real world testing as well and see what i find.

frontend/components/Domain/Recipe/RecipeCardSection.vue
michael-genson: I think I originally had added a new element in there but removed it before committing

frontend/components/Domain/Recipe/RecipeCardSection.vue
michael-genson: I think I had moved the comment which was just in the wrong place, but I agree this is better anyway

frontend/components/Domain/Recipe/RecipeDataTable.vue
michael-genson: Yeah it was screwy for one of the tables (I think it was the foods table) so I went and added an explicit sort to all of them, to prevent any issues in the future

frontend/pages/group/data/foods.vue
michael-genson: Yeah this makes sense. We'll have to add it to the table but that's fine

frontend/components/Domain/Recipe/RecipeDataTable.vue
Kuchenpirat: Yeah, I wrote this comment before I saw the rest of them 👍

frontend/pages/group/data/foods.vue
```diff
@@ -326,8 +331,21 @@ export default defineComponent({
...
+      {
+        text: i18n.tc("general.date-added"),
+        value: "createdAt",
+        show: false,
+      }
     ];
 
+    function formatDate(date: string) {
+      try {
+        return i18n.d(Date.parse(date), "medium");
```
Kuchenpirat, line 343: I am not sure the weekday is actually needed here. 
But it should also not matter much as the actual date created value is not really of interest as long as you can sort this way.

Kuchenpirat (approved):
took me a bit longer than a day but i have no blocking comments. 

Will leave it up to you to merge 🚀

frontend/pages/group/data/foods.vue
michael-genson: I'm going to leave this only because it uses i18n so it's translated. We could modify it to only display yyyy-mm-dd but as you said I'm not sure how much it matters since it's really just for sorting anyway

Kuchenpirat: Yeah, that sounds more important than removing the weekday

jpriebe:
@michael-genson - this looks like an excellent PR.  It was 9 months ago, labeled "Merged", and a quick spot check shows some of these code changes in the current codebase.  

But I just spun up v2.8.0 of mealie, and I don't see the "Household" button on my recipe search.

Am I doing something wrong?  Is there a feature flag I need to enable?  Being able to limit my search results to my household would be a huge help now that I have adult kids outside the house putting in their own recipes.

michael-genson:
No it should definitely be there, can you send a screenshot of where it's missing? Can you also share your logs?

michael-genson:
Also, can you confirm that you have more than one household in your group?

jpriebe:
Hi, @michael-genson  - thank you for such a rapid response. 

I'm missing the "household" button on the main recipe search page.  (please bear with me -- I'm upgrading from 0.5.6 to 2.8.0, so I have a long way to go to catch up to the new data model and UI changes).

image

I do have a second household, and for a minute, I thought that the problem was that the second household has no recipes (I imported all of the 0.5.6 recipes into a single household, although I'm thinking about trying to partition the export into two households before importing).

At any rate, I created myself a test user under the second household, and I created a dummy recipe just to see if maybe the "household" button only shows up when there are recipes belonging to multiple households.  That did not seem to make the button available.

I took a look at my logs, and now I'm not confident the app is logging correctly.  I ran into SQLite performance issues, so I switched over to postgresql, and I don't see much logged after that, despite the fact that I've been using the UI actively for about an hour.

jpriebe:
BTW, if it's helpful at all, I don't see a "Households" filter button on the households demo site, either (https://demo.mealie.io/admin/manage/households)

jpriebe:
OK, I just closed out mealie and returned to it, and now I see the households filter button.  Is it possible that there is a client-side cache of recipes, and when I was originally working with the UI, it didn't see any recipes from households other than mine?  And once I added recipes to the other household, I had to clear/reload that cache to get the households button to show up?

michael-genson:
Yeah it sounds like something was cached weird which got resolved. If you see it again let me know, and keep an eye on your backend logs, but seems like it's working!

jpriebe:
Thanks for your work on this great product!  I'm glad to see such momentum behind Mealie -- not sure what I was doing running 0.5.6 for so long!  :-)

PR #4227, opened 2024-09-17
feat: OpenAI Custom Headers/Params and Debug Page

### LINKED ISSUE #4119: Mealie Azure OpenAI bad connection url
I would like to use this feature, but I want to use Azure openAI resource since I have credits with work. I tried to use it, however, for it to work in azure, we need to specify the api version as a query params:
https://domainname.openai.azure.com/openai/deployments/gpt-4o/chat/completions?api-version=2023-03-15-preview

However, when I look at your code, you take the full given url and append /chat/completions at the end. This ends up with a call to:
https://domainname.openai.azure.com/openai/deployments/gpt-4o/chat/completions?api-version=2023-03-15-preview/chat/completions which for obvious reasons, doesn't work.

Would there be a way to specify, in environment variables, the api-version used? so that you can append it at the end without us having to go play in py files (we're not all software developers ;) )

_Originally posted by @AntoineBriseboisRoy in https://github.com/mealie-recipes/mealie/discussions/1588#discussioncomment-10474879_

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- documentation
- feature

_(REQUIRED)_

This PR adds a few customization options to the OpenAI config to allow for more custom integrations (e.g. with Azure OpenAI). It also adds an admin OpenAI debug page (I also moved the parser debug page there):

image

image

Optionally you can provide an image:
image

And here's an error (I added the word "test" to the end of my API key):
image

_(REQUIRED)_

Fixes https://github.com/mealie-recipes/mealie/issues/4119

## Testing

Manually


AntoineBriseboisRoy (approved):

Kuchenpirat (approved):
Looks all good to me. 
OpenAI debugging does work as expected. I have no way of real-world testing custom headers and params but the implementation makes sense to me and i hope we have some people testing that out in nightly for us.

AntoineBriseboisRoy:
I tried testing custom params this morning. I can't figure out the required format to put in my docker-compose since it is not specified in the doc. I think it would be important to include an exemple in the doc. 

Can you give me a hint here? ;)

Kuchenpirat:
> I tried testing custom params this morning. I can't figure out the required format to put in my docker-compose since it is not specified in the doc. I think it would be important to include an exemple in the doc.
> 
> Can you give me a hint here? ;)

There is an example in the docs in the backend config page on the openai section.
https://docs.mealie.io/documentation/getting-started/installation/backend-config/#openai

OPENAI_CUSTOM_PARAMS='{"k1": "v1", "k2": "v2"}'

PR #4231, opened 2024-09-18
feat: Add Household Filter to Meal Plan Rules

### LINKED ISSUE #4081: [Task] - Enable Cross-Household Recipes

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- feature

_(REQUIRED)_

Adds a household filter to meal plan rules. Similar to tags and categories, if no household is selected, then all households are considered. I had to re-write the part of the recipes repo that does the filtering by tag/category (since we need to support household, too) and ended out just removing it and using the normal `page_all` route, since that already supports tags, categories, households, and random ordering.

New rule form:
image
image
image

I also tweaked how meal plan rules render on the frontend to call out when there are no filters (e.g. "All Tags" when no tags are selected):
image
image

_(REQUIRED)_

Closes https://github.com/mealie-recipes/mealie/issues/4081

## Testing

Pytest coverage. I was surprised to find we didn't have any existing coverage for meal plan rules other than basic CRUD, so I added tests for other filters too (type, tag, etc.).


frontend/lib/api/types/meal-plan.ts
```diff
@@ -31,31 +31,44 @@ export interface ListItem {
   quantity?: number;
   checked?: boolean;
 }
+export interface PlanCategory {
+  id: string;
+  name: string;
+  slug: string;
+}
```
boc-the-git, line 38: Does this make our future maintenance more complicated? :thinking: 
Is it creating duplication in the code base?

boc-the-git (commented):
This is an incomplete review.. it'll be at least a few days before I can fully/properly review it.

frontend/lib/api/types/meal-plan.ts
michael-genson: We were already doing this, the class names were just identical so the auto-generation wasn't handling it properly. We already had tag/category subclasses for meal plans.

This part of the code is autogenerated though so maintenance isn't an issue

mealie/routes/households/controller_mealplan.py
```diff
@@ -40,57 +41,77 @@ def mixins(self):
...
         )
 
+    def _get_random_recipes_from_mealplan(
+        self, plan_date: date, entry_type: PlanEntryType, limit: int = 1
+    ) -> list[Recipe]:
+        """
+        Gets rules for a mealplan and returns a list of random recipes based on the rules.
+        May return zero recipes if no recipes match the filter criteria.
+
+        Recipes from all households are included unless the rules specify a a household filter.
```
michael-genson:
```suggestion
        Recipes from all households are included unless the rules specify a household filter.
```

mealie/routes/households/controller_mealplan.py
```diff
@@ -40,57 +41,77 @@ def mixins(self):
...
         return self.repo.get_today()
 
     @router.post("/random", response_model=ReadPlanEntry)
     def create_random_meal(self, data: CreateRandomEntry):
         """
-        create_random_meal is a route that provides the randomized functionality for mealplaners.
-        It operates by following the rules setout in the Groups mealplan settings. If not settings
-        are set, it will default return any random meal.
+        `create_random_meal` is a route that provides the randomized functionality for mealplaners.
+        It operates by following the rules set out in the household's mealplan settings. If no settings
```
boc-the-git, line 93: This implies to me that the meal plan settings live at the household level, but that's not true is it? (mental blank as I write this)

tests/integration_tests/user_household_tests/test_group_mealplan.py
```diff
@@ -167,3 +206,128 @@ def test_get_mealplan_today(api_client: TestClient, unique_user: TestUser):
...
+        unique_user.repos.tags.create(TagSave(name=random_string(), group_id=unique_user.group_id)) for _ in range(4)
+    ]
+    categories = [
+        unique_user.repos.categories.create(CategorySave(name=random_string(), group_id=unique_user.group_id))
+        for _ in range(4)
+    ]
+    [
+        create_recipe(unique_user, tags=[tag], categories=[category])
+        for tag, category in zip(tags, categories, strict=True)
+    ]
```
boc-the-git, line 222: I'm sure the answer is "some python crap that I'm not too familiar with".. can you explain what this is doing?

boc-the-git (approved):
LGTM :rocket:

mealie/routes/households/controller_mealplan.py
michael-genson: Yes, meal plan rules are at the household level. All meal plan related things are, since each household plans for meals differently. We moved them from group to household in the original households PR

tests/integration_tests/user_household_tests/test_group_mealplan.py
michael-genson: `zip(tags, categories, strict=True)` means "take these two lists which are the same length and iterate over them returning a tuple of one of each item". So in this case if you have:

`[tag1, tag2, tag3]` and `[cat1, cat2, cat3]` then `zip(tags, categories, strict=True)` returns `[(tag1, cat1), (tag2, cat2), (tag3, cat3)]`. `strict=True` here just means if the lists are different sizes, throw an error (in this test this never happens, but our linter wants it).

The rest of the function is just looping over the zipped tags and categories and creating recipes from it.

PR #4077, opened 2024-08-22
feat:  Shopping list UI overhaul 

Wetzel402:
## What type of PR is this?

- feature

## What this PR does / why we need it:

The shopping list UI seemed a bit complex.  This PR's goal is to...
1. Simplify the UI by moving buttons to an overflow menu in the upper corner
2. ~~List items are in v-card now to help with readability~~
3. Each label now has it's own add button that should add to that label by default to speed up/simplify adding under that label. (in progress) 
4. Default to the label sorted view rather than the unsorted view
5. Add keep screen awake 
<details><summary>Screenshots</summary>

image
image
image

</details>

## Which issue(s) this PR fixes:

None

## Testing

Tested in dev container

PR #4238, opened 2024-09-19
feat: Shopping list UI overhaul - label sorted by default 

Wetzel402:
## What type of PR is this?

- feature

- Default to the label sorted view rather than the unsorted view

https://github.com/mealie-recipes/mealie/pull/4077
## Testing

Tested in dev container


Kuchenpirat (approved):
We are ignoring the case that many people don't label their foods, so for them everything will always be grouped under "No Label"
But since preferences are persistent it should be fine. 
If not it is super easily reversible

PR #4252, opened 2024-09-22
feat: Groups/households custom invitations

p0lycarpio:
## What type of PR is this?

- feature

This PR adds the ability for admins to invite users to join a specific group/household on Mealie. Previously, invite links were only valid for the user who generated it.

N/A

The invitation endpoint has been changed to take into account group and household directly in the body. This should not be a breaking change.

## Testing

Pytests and manual testing.


michael-genson (changes requested):
Makes sense to me, should make inviting other users easier, especially for new instances. A couple comments:
- Can we add this to the normal "get invite link" button on the user profile? We can do a check for whether or not the user is an admin to conditionally render the dropdowns (`$auth.user?.admin`)
- On the user management page you can't get an invite link, you can only send an email. Can we add a link to that dialog as well? (not all users have SMTP set up)

I'll give it a more thorough review later but those are the two things that jumped out at me

p0lycarpio:
Thanks for your feedback! Indeed, I think it would be interesting to add the invitation link in all cases. However, I find that adding the two drop-down lists for administrators on the "user" page can be confusing because this page is not linked to administration actions and should be the same for everyone...
I will try to make the changes when I have time ;)

michael-genson:
Yeah I see your point, but I think admins will have a hard time finding the feature otherwise (or just not know about it)

p0lycarpio:
I implemented the requested changes, and I took the liberty of modifying the API to return all tokens if an admin calls it.
I'm busy at the moment but I'll try to read your feedback.

mealie/routes/households/controller_invitations.py
```diff
@@ -21,18 +21,31 @@
...
     @router.post("", response_model=ReadInviteToken, status_code=status.HTTP_201_CREATED)
-    def create_invite_token(self, uses: CreateInviteToken):
+    def create_invite_token(self, body: CreateInviteToken):
         if not self.user.can_invite:
             raise HTTPException(
                 status.HTTP_403_FORBIDDEN,
                 detail="User is not allowed to create invite tokens",
             )
 
+        if body.group_id and body.household_id and not self.user.admin:
```
michael-genson: Can we modify this to compare against `self.group_id` and `self.household_id`, rather than just checking if they're provided? That way if a user provides their own group/household it doesn't error out. Something like:

```python
body.group_id = body.group_id or self.group_id
body.household_id = body.household_id or self.household_id

if not self.user.admin and (body.group_id != self.group_id or body.household_id != self.household_id):
    ...
```

mealie/routes/households/controller_invitations.py
```diff
@@ -21,18 +21,31 @@
 class GroupInvitationsController(BaseUserController):
     @router.get("", response_model=list[ReadInviteToken])
     def get_invite_tokens(self):
-        return self.repos.group_invite_tokens.page_all(PaginationQuery(page=1, per_page=-1)).items
+        if self.user.admin:
```
michael-genson: Can we move this to an admin route? I'm worried it might get confusing if we respond with all tokens sometimes, and tokens filtered by group other times. Maybe make a new admin controller for invitations?

michael-genson (changes requested):
I have a few notes on the backend; I still need to fully review the frontend, but preliminarily it looks good. Apologies for the delay in reviewing, I've been pretty busy the past week or so.

mealie/routes/households/controller_invitations.py
p0lycarpio: Thanks for the feedback! Indeed it seems more appropriate to me in a separate route (see `mealie/routes/admin/admin_invitations.py`). I also added a `get_one()` operation that allows to specify the *group_slug* or the *id*.

mealie/routes/admin/admin_invitations.py
```diff
@@ -0,0 +1,25 @@
...
+router = APIRouter(prefix="/invitations")
+
+
+@controller(router)
+class AdminInvitationsController(BaseAdminController):
+    @router.get("", response_model=list[ReadInviteToken])
+    def get_all(self):
+        """Get all invite tokens"""
+        self.repos.group_id = None
+        self.repos.household_id = None
```
michael-genson: By default admin controllers don't filter by group or household, so you can remove these lines

mealie/routes/admin/admin_invitations.py
```diff
@@ -0,0 +1,25 @@
...
+        return self.repos.group_invite_tokens.page_all(PaginationQuery(page=1, per_page=-1)).items
+
+    @router.get("/{group_slug}", response_model=list[ReadInviteToken])
+    def get_one(self, group_slug: str):
+        """Get all invite tokens for a specific group"""
+        self.repos.group_id = self.repos.groups.get_by_slug_or_id(group_slug).id
+        return self.repos.group_invite_tokens.page_all(PaginationQuery(page=1, per_page=-1)).items
```
michael-genson: We can get rid of this route; `get_one` implies you're getting a single invitation (rather than all invitations for a particular group). At the admin-level I don't think this is necessary. If we do need this functionality I'd rather build it into the `get_all` route using pagination, rather than having a separate route for it.

frontend/components/Domain/User/UserInviteDialog.vue
```diff
@@ -0,0 +1,195 @@
+<template>
+  <BaseDialog
+    v-model="inviteDialog"
+    :title="$tc('profile.get-invite-link')"
+    :icon="$globals.icons.accountPlusOutline" color="primary">
+    <v-container>
+      <v-form class="mt-5">
+        <v-select
+v-if="groups && groups.length" v-model="selectedGroup" :items="groups" item-text="name"
```
michael-genson: Can we break these long lines up with line breaks? To stay consistent with other components. e.g.
```js
<v-select
  v-if="groups && groups.length
  v-model="selectedGroup"
  :items="groups"
  item-text="name"
  item-value="id"
  :return-object="false"
  filled
  :label="$tc('group.user-group')"
  :rules="[validators.required]"
/>
```

frontend/components/Domain/User/UserInviteDialog.vue
```diff
@@ -0,0 +1,195 @@
...
+  },
+  props: {
+    value: {
+      type: Boolean,
+      default: false,
+    },
+    isAdmin: {
+      type: Boolean,
+      default: false,
+    },
```
michael-genson: You should be able to detect this instead of passing a prop. In setup:

```js
    const { $auth, i18n } = useContext();
    const isAdmin = computed(() => $auth.user?.admin);
```

michael-genson (changes requested):
A few more comments, but it's close! I appreciate all the work you've put into this!

Are you able to add tests for the new route(s)? Also a test for trying to set the group/household as a non-admin on the old route. If not I'm happy to get those in there myself.

mealie/routes/admin/admin_invitations.py
p0lycarpio: I removed this route and added `PaginationQuery` in `get_all`

p0lycarpio:
Thanks for your review!
I've taken the requested changes into account and written some tests. You can edit them if you have better suggestions or other cases that aren't covered.

mealie/routes/admin/admin_invitations.py
```diff
@@ -0,0 +1,21 @@
...
+
+@controller(router)
+class AdminInvitationsController(BaseAdminController):
+    @router.get("", response_model=list[ReadInviteToken])
+    def get_all(
+        self,
+        q: PaginationQuery = Depends(make_dependable(PaginationQuery)),
+    ):
+        """Get all invite tokens"""
+        return self.repos.group_invite_tokens.page_all(pagination=q).items
```
michael-genson: We should return the pagination object as-is instead of the items. It should be:

```python
response = self.repos.group_invite_tokens.page_all(pagination=q)

response.set_pagination_guides(router.url_path_for("get_all"), q.model_dump())
return response
```

This will also need to be updated on the frontend to accept the pagination response (rather than the list of invitations)

mealie/routes/admin/admin_invitations.py
```diff
@@ -0,0 +1,21 @@
...
+    ReadInviteToken,
+)
+from mealie.schema.make_dependable import make_dependable
+from mealie.schema.response.pagination import PaginationQuery
+
+router = APIRouter(prefix="/invitations")
+
+
+@controller(router)
+class AdminInvitationsController(BaseAdminController):
```
michael-genson: Maybe I'm just missing it, but where in the frontend is this endpoint being called? I only see us using the old (non-admin) route

michael-genson (changes requested):

mealie/routes/admin/admin_invitations.py
p0lycarpio: This admin route is not called via the frontend just like the non-admin one to get all the created invitations.
There is no page associated with these operations and they are only available via API. The admin route was added to give access to all groups/households, not the current user's.
I don't think an admin page containing all current or past invitation links would be useful on the frontend side.
Also, I don't think it is necessary to duplicate the endpoint to create an invitation on the admin controller?

michael-genson: Yeah I agree that would be unnecessary. Let's go ahead and remove the new backend route/controller then if it's not being used in the app

michael-genson (changes requested):
Looks great! Last thing is the test for non-admins attempting to create invites for other groups/households. I think it got removed in your revert. Once we have that this is good to go. I appreciate all the work you've put into this one!

p0lycarpio (commented):
You're right! Here is the test I added. I used random uuids, I don't know if there was another way to use existing groups/households.

michael-genson (approved):
Thank you! Excited to get this one in there

PR #4253, opened 2024-09-22
feat: sort labels by name,asc in shopping lists

seamuslowry:
## What type of PR is this?

- feature

This is an initial (partial) implementation of the request in this discussion: https://github.com/mealie-recipes/mealie/discussions/3297.

~In `controller_labels.py` it extends the `PaginationQuery` class to override the default values for `order_by` and `order_direction` so that call to this endpoint will use those as the default, rather than using nothing.~

Update `frontend/pages/shopping-lists/_id.vue` to pass `orderBy: "name"` and `orderDirection: "asc"` on label refreshes

No issues reference this specifically. This is intended as an initial step towards the feature requested in This is an initial implementation of the request in this discussion: https://github.com/mealie-recipes/mealie/discussions/3297.

~I am not experienced with FastAPI or Mealie conventions. I am happy to apply the same pattern to the paginated GET requests of the other listed models, but I wanted to get some input on the pattern and implementation before copying the pattern around.~

## Testing

- ~in Swagger UI, I tested passing other values for `order_by` and `order_direction` to verify they were used over the new defaults. Similarly, tested with other pre-existing values like `per_page` to ensure they were still respected.~
- in the UI, I validated that the request to get labels when adding a new item to a shopping list (which wasn't specifying `order_by` and `order_direction` is now specifying `name` and `asc`


michael-genson:
I don't think this is something we want to do on the backend; is there somewhere in particular on the frontend you labels aren't sorted alphabetically? If so, we should address it there

seamuslowry:
I've only noted seeing it when trying to add an item to a shopping list

image

I thought it would make sense to change the defaults for the endpoint, get it for free everywhere that doesn't specify. But I'm happy to make it in that one spot on the frontend instead

michael-genson:
Ah okay, that makes sense. Is it possible to use the label store here instead of using `userApi`? The store defaults to name ascending, and it would save us an API call (we're already using the label store in that same file). Something like:

```javascript
const { store: allLabels, actions: labelActions } = useLabelStore();  // this already partially exists on line 605
```

And then we can just call `actions.refresh()` and get rid of `refreshLabels` (and `allLabels` will automatically be updated).

seamuslowry:
Updated. But I used `labelActions.getAll` rather than `labelActions.refresh`

The `refresh` call seems to call the api directly without specifying any defaults: https://github.com/mealie-recipes/mealie/blob/mealie-next/frontend/composables/partials/use-actions-factory.ts#L106. So it does not resolve the issue.

frontend/pages/shopping-lists/_id.vue
```diff
@@ -792,15 +792,7 @@ export default defineComponent({
...
-    async function refreshLabels() {
-      const { data } = await userApi.multiPurposeLabels.getAll();
-
-      if (data) {
-        allLabels.value = data.items ?? [];
-      }
-    }
-
-    refreshLabels();
+    labelActions.getAll()
```
michael-genson: Actually, do we even need this here at all? Can we just use the store without calling refresh?

Originally I had thought that we used the refresh function elsewhere, but upon further review it looks like we only use it here anyway

seamuslowry: you are correct. removing the call here has the list sorted alphabetically as expected.

it makes sense that this was superfluous now. i saw two requests in the network tab prior to this change, one with and without the ordering query params. i could never track down where the two calling locations were. but it makes sense that one was the original store request and the other was this one.

TL;DR removed

michael-genson: Sweet. Yeah that file is a bit of a mess, largely my fault

michael-genson (approved):
Thanks for keeping with this one!

PR #4274, opened 2024-09-26
feat: Create Recipe From HTML or JSON

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- feature

_(REQUIRED)_

This PR adds the ability to create a recipe from raw HTML or JSON. We leverage the existing URL scrapers as-is, but bypass calling the target URL and instead provide the HTML/JSON directly (JSON must be in schema.org format, although if OpenAI is enabled it might figure it out regardless).

This is suggested on the create by URL page:
image

Which brings you here:
image

You can put HTML or JSON into that text field, however you may optionally use the JSON editor mode for JSON:
image

The main advantages of this are:
1. If you have raw HTML or JSON from a local recipe (or if you fed something to OpenAI) you can import it here without spinning up a webserver or similar
2. If our scrapers fail for some reason (e.g. the target URL has anti-bot protection, or recipes are locked behind a login) you can use this as a workaround

_(REQUIRED)_

Closes https://github.com/mealie-recipes/mealie/discussions/3224
Closes https://github.com/mealie-recipes/mealie/discussions/2391

I changed some of the "create" route names because they're getting a bit unorganized. All recipe create routes are now under `/recipe/create/...`. This is a breaking change, so I added it to the migration docs.

## Testing

Pytest


frontend/pages/g/_groupSlug/r/create.vue
```diff
@@ -52,6 +52,11 @@ export default defineComponent({
         text: i18n.tc("recipe.bulk-url-import"),
         value: "bulk",
       },
+      {
```
Kuchenpirat, line 55: at some point we should try to simplify that list somehow. It kepps getting longer and  longer...

Kuchenpirat (approved):
This was way more straigt-foreward than expected 😊
All  good on this one 🚀

frontend/pages/g/_groupSlug/r/create.vue
michael-genson: Yeah I agree, we're gonna have to do something about it soon

PR #4278, opened 2024-09-28
fix: Hide rating on notes

boc-the-git:
## What type of PR is this?

- bug

I saw that note only "recipes" only the meal planner were allowing me to give a rating.. which resulted in a 400 or 404 (can't remember).
This stops the rating from appearing (making it similar to favourite which was already hidden).

Before:
Screenshot from 2024-09-28 15-27-56

After:
Screenshot from 2024-09-28 15-28-30

None.

The favourites badge a couple lines above this code has a limitation of `isOwnGroup`, whereas rating doesn't. I believe that was intentional from Michael when initially implementing per-user ratings.

## Testing

The screenshots basically. Confirmed no impact against a real recipe.


boc-the-git:
I'll have more time to deal with the merge conflict later tonight.

michael-genson:
IIRC the component handles whether or not you can set a rating, which is why it's not hidden from users from other groups (so they can still see the rating, just not set it)

michael-genson (approved):

PR #4282, opened 2024-09-28
feat: Support additional compression algorithms

HowRuck:
## What type of PR is this?
- feature

This PR introduces support for newer compression algorithms (Brotli and Zstd) in addition to the currently supported Gzip. By dynamically selecting the best compression algorithm based on the Accept-Encodings header from the request, this change aims to improve overall performance, particularly in low bandwidth scenarios

Not really any issue per se, it just improves performance a little.

I've done a quick scan over the `starlette-compress` library used here, as it is still pretty new and very small. I haven't found any issues with it. It literally just chooses the best compression and compresses the content using standard libraries. Feel free to have another look around, in case I missed anything :-)

## Testing

I've been using this change in my private mealie instance for the last week without any issues.


hay-kot:
Do you have any benchmarks for the performance difference here? If we’re going to add a dependency id be interested to know if the tradeoff is worth it. 

In my mind setting up your reverse proxy to handle compression might be a better bet.

HowRuck:
> Do you have any benchmarks for the performance difference here? If we’re going to add a dependency id be interested to know if the tradeoff is worth it.

I've done some *very* basic testing, and so far I am coming to the conclusion that adding a new dependency is probably not worth it (benefits are there but relatively minor). The solution below, however, might be an interesting alternative :thinking: 
I'll do some more testing in the coming days if needed. 

> In my mind setting up your reverse proxy to handle compression might be a better bet.

That would most definitely be the best bet :+1: 

The only problem with this is that traefik, or most other reverse proxies, don't compress data that is already compressed. A good solution for this would be to add an environment variable (defaulting to `true`) that controls whether gzip compression is active or not. If not, compression can then be handled by the reverse proxy.

This would enable the use of modern compression algorithms without adding additional dependencies, and would also allow a nice separation of concerns while increasing flexibility. 

What would you think of this approach? :thinking:

hay-kot:
> A good solution for this would be to add an environment variable (defaulting to true) that controls whether gzip compression is active or not. If not, compression can then be handled by the reverse proxy.

This sounds good to me, Sorry for the delay!

michael-genson:
Converted to draft since it looks like the changes were reverted. Once it's ready feel free to switch it back!

hay-kot:
Closing for now, feel free to reopen or create a new PR with the changes discussed. Thanks!

PR #4292, opened 2024-09-30
dev: add setting to fix mypy pydantic warning on vs code dev container setup

tjb1982:
## What type of PR is this?

- bug

Adds a setting for the mypy extension to fix an import error in dev container setup.

N/A

## Testing

Manual via vs code under dev container.


Kuchenpirat:
@tjb1982 can you prefix your title with `fix:` to have the title linter not fail?

tjb1982:
> @tjb1982 can you prefix your title with `fix:` to have the title linter not fail?

I had to add an empty commit to retrigger it

Kuchenpirat (approved):
Thanks for this fix. 
The warning has been around since forever and its pretty nice to have it gone 😊

PR #4298, opened 2024-10-01
feat: refactor recipe scaling

### LINKED ISSUE #1390: [v1.0.0b] [Task] - Recipe scaling UX

This is an extension ticket for #1345.

In this task we need to find a good way of showing the scaled servings and making clear what the user can configure.

At the moment the only thing the user can change is the **scale**.
This scale is then multiplied with the “base-servings”, configured in the recipe editor to calculate the final servings amount.

I would propose having two ways to configure the servings amount:
1. Configure a scale
2. Configure the amount of servings

Option 2 is only possible if the recipe has a “base-servings” configured, Option 1 is used when no “base-servings” are given.

This could then look like this:
image

When clicking on the "Scale.." button, it will show the modal with the configuration of a scale.
The scale is then used for the ingredient scaling.

When clicking on the "Servings.." button, it will show the modal with the configuration of target servings.
The scale is here calculated automatically.

In addition to these features I would hide the scale button and hide/disable the "+/-" + tooltip functionality from the servings button if no scaling can be applied (if the recipe does not have its ingredients configured with amount).

Kuchenpirat:
## What type of PR is this?

- feature

This PR refactors the recipe scaler for a smoother user experience.

### Previous Behavior:
The old scaler was based on a base scale factor of 1, which would autmatically double/half the recipe using the `+ | -` Buttons. However, to scale a recipe from 4 servings to 6, for example, users had to manually calculate the scale (e.g., entering 1.5). This process could be cumbersome, especially when the math wasn't straightforward.

### New Behavior:
With this update, the scale is inferred directly from the current serving size, allowing users to adjust servings more intuitively by using the + | - buttons which will now increase/decrease the serving size by one. Additionally, the manual input field now allows users to directly enter the desired number of servings, eliminating the need to calculate a scale value.

This update also removes the restriction that required a serving count for the component to be rendered. Now, the component will default to a serving size of 1 if no serving count is provided.

If the serving size cannot be infered from the servings count string (eg. six servings) the scaler will default to a value of one. If the serving size is increased it will prepend the Yield string with `<number>x` resulting in e.g. "2x six servings" which is ugly but is the best we can do i think. 

An alternative implementation was discussed in #3755, where scaling below the original serving size would be handled by halving. However, this current implementation supports fractional servings (e.g., entering 0.5 manually), which should cover most use cases. For large numbers one can also directly set the number of serving sizes so clicking `+` should not be required. 

- closes #3755 (superseeded by this)
- closes #1390 

Discussions: 
- closes #2861 
- closes #3754 

- Removed the note added in #4287 as it is no longer applicable.
- The findMatch function from the use-extract-recipe-yield composable was used. We could alternatively return this number automatically with with the useExtractRecipeYield function.
- This change may affect recipes with fractional serving sizes (e.g., 1/2 baking tray). Previously, the scaler increased in increments like 0.5 -> 1 -> 1.5, but now it increases as 0.5 -> 1.5 -> 2.5. I feel like this should not concern the majority of recipes. 

## Testing

Manual testing, including fractional serving sizes.

### DIFF
--- a/frontend/components/Domain/Recipe/RecipeScaleEditButton.vue
+++ b/frontend/components/Domain/Recipe/RecipeScaleEditButton.vue
@@ -8,7 +8,7 @@
               <span v-if="!recipeYield"> x {{ scale }} </span>
               <div v-else-if="!numberParsed && recipeYield">
                 <span v-if="numerator === 1"> {{ recipeYield }} </span>
-                <span v-else> {{ numerator }}x {{ recipeYield }} </span>
+                <span v-else> {{ numerator }}x {{ scaledYield }} </span>
               </div>
               <span v-else> {{ scaledYield }} </span>
 
@@ -97,13 +97,12 @@ export default defineComponent({
       },
     });
 
-    const numerator = ref<number>(props.basicYieldNum ?? 1);
-    const denominator = props.basicYieldNum ?? 1;
+    const numerator = ref<number>(parseFloat(props.basicYieldNum.toFixed(3)) ?? 1);
+    const denominator = parseFloat(props.basicYieldNum.toFixed(32)) ?? 1;
     const numberParsed = !!props.basicYieldNum;
 
-    watch(() => numerator.value, (newValue) => {
-      console.log("newValue", newValue);
-      scale.value = numerator.value / denominator;
+    watch(() => numerator.value, () => {
+      scale.value = parseFloat((numerator.value / denominator).toFixed(3));
     });
     const disableDecrement = computed(() => {
       return numerator.value <= 1;
--- a/frontend/components/Domain/Recipe/RecipeScaleEditButton.vue
+++ b/frontend/components/Domain/Recipe/RecipeScaleEditButton.vue
@@ -16,7 +16,7 @@
           </template>
           <v-card min-width="300px">
             <v-card-title class="mb-0">
-              {{ $t("recipe.edit-scale") }}
+              {{ $t("recipe.servings") }}
             </v-card-title>
             <v-card-text class="mt-n5">
               <div class="mt-4 d-flex align-center">
@@ -29,7 +29,7 @@
                       </v-icon>
                     </v-btn>
                   </template>
-                  <span> {{ $t("recipe.reset-scale") }} </span>
+                  <span> {{ $t("recipe.reset-servings-count") }} </span>
                 </v-tooltip>
               </div>
             </v-card-text>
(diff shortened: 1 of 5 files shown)


Kuchenpirat:
@bartkummel if you have any thought on this feel free to share. 
Always valuable to have somebody with a different view contribute 👍

bartkummel:
Generally speaking, I think this is a better solution. However, I’m not sure if 1 is the best default value. In my experience most recipes serve 4.

Kuchenpirat:
That might make sense, my thought process for going with 1 was that if we scale up from one and i want to comunicate that back to the user the scaler would show something like "3/4x four muffins" which looks even more complicated than just "2x four muffins"

Happy to change this tho, if it is clear that this would be prefered.

bartkummel:
Or we should add a global setting for default serving size

Kuchenpirat:
The default serving size is just the fallback so i don't think that makes too much sense since recipes will naturally have different serving sizes and my guess is that the people who do not fill out the serving sizes will not bother to recalculate the recipe to a common basis. 
This will make this unreliable. 

AND some people will 100% think that this will recalculate their recipes with serving sizes to the default serving size, which might actually be an interesting idea.

frontend/components/Domain/Recipe/RecipeScaleEditButton.vue
```diff
@@ -94,9 +97,25 @@ export default defineComponent({
       },
     });
 
+    const numerator = ref<number>(props.basicYieldNum ?? 1);
+    const denominator = props.basicYieldNum ?? 1;
+    const numberParsed = !!props.basicYieldNum;
+
+    watch(() => numerator.value, (newValue) => {
+      console.log("newValue", newValue);
+      scale.value = numerator.value / denominator;
```
michael-genson: Should we round this? On the backend we round decimals to 3 points and denominators to a max of 32.

michael-genson (dismissed):
This looks good to me. I see no issue with the default serving size being 1, it's the closest we're going to get to something that makes sense (if no serving is specified, it's "1x the current recipe").

In the future we can look into improving the `findMatch` method to account for words, etc, or maybe even attempt to parse it server-side.

---

Approved, though I left one comment. May also want to keep this one open for discussion.

frontend/components/Domain/Recipe/RecipeScaleEditButton.vue
michael-genson: Also we may want to ditch the log

Kuchenpirat: rounded numerator and scale to three, denominator to 32 👍
An removed the log

Kuchenpirat:
> In the future we can look into improving the `findMatch` method to account for words, etc, or maybe even attempt to parse it server-side.

i think having this eventually in a backend db might make sense. We don't do much with that info currently, but trying to parse it on creation and then have separate entries for amount and text makes sense to me. 
When the parsing failed the user can input a number (followed by text) which we know to be a number and don't have to find in some string and save that to the backend

michael-genson (approved):

PR #4301, opened 2024-10-01
feat: adding the rest ofthe nutrition properties from schema.org

tjb1982:
## What type of PR is this?

- feature

Some of the most important schema.org nutrition properties were missing. (See https://schema.org/NutritionInformation)

This adds the remaining ones from schema.org, except for `servingSize` which would cause problems with the current design wrt translating the unit of measure, which is variable. It's also easy to confuse the `servingSize` (i.e., "500 g", "4 cups", etc.) concept with the existing and unrelated "servings" (i.e., "serves 4") scaling concept, so I left it out (for now?).

In the future it might be nice to allow an EAV type of design to allow users to create arbitrary nutrition entries, but that's more difficult and definitely an edge-case. Out of scope for this regardless.

**update**: added tests and updated cleaner, migration services, etc.

N/A

## Testing

Manually tested with the following URL import: https://www.twopeasandtheirpod.com/vegetarian-taco-skillet/

This recipe included saturated fat and cholesterol.

Updated autoated unit/integration tests to account for these new entries.


tjb1982:
Another thing I forgot to mention is that I don't know what to do for crowdin translations. I updated the en-US for the affected front-end messages with the assumption that en-US is the source for the rest of the translations (I think I saw that configured somewhere). But that will need to be run at some point if this PR is accepted.

tjb1982:
image

Kuchenpirat:
> Another thing I forgot to mention is that I don't know what to do for crowdin translations. I updated the en-US for the affected front-end messages with the assumption that en-US is the source for the rest of the translations (I think I saw that configured somewhere). But that will need to be run at some point if this PR is accepted.

You just need to update the en-US file. Everything else will be handled by Crowdin automatically 👍

Kuchenpirat (approved):
Hey, thanks for your contribution to mealie. 
Looks all good 👍

Sorry it took me a bit longer to get to it.

PR #4302, opened 2024-10-01
fix: Shopping List Label Text Color

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- bug

_(REQUIRED)_

Fixes the shopping list label text color to be black or white depending on the background color. We were already doing this on the `MultiPurposeLabel` component, so I just extracted that logic.

Before:
image

After (notice the "Alcohol" label is still white):
image

I tweaked a few other style things as well.

_(REQUIRED)_

N/A

### DIFF
--- a/frontend/pages/shopping-lists/_id.vue
+++ b/frontend/pages/shopping-lists/_id.vue
@@ -58,7 +58,15 @@
       <div v-else>
         <div v-for="(value, key) in itemsByLabel" :key="key" class="mb-6">
           <div class="text-left">
-            <v-btn :color="getLabelColor(value[0]) ? getLabelColor(value[0]) : '#959595'">{{ key }}</v-btn>
+            <v-btn
+              :color="getLabelColor(value[0]) ? getLabelColor(value[0]) : '#959595'"
+              :style="{
+                'color': getTextColor(getLabelColor(value[0])),
+                'letter-spacing': 'normal',
+              }"
+            >
+              {{ key }}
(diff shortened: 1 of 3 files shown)


frontend/pages/shopping-lists/_id.vue
```diff
@@ -58,7 +58,15 @@
       <div v-else>
         <div v-for="(value, key) in itemsByLabel" :key="key" class="mb-6">
           <div class="text-left">
-            <v-btn :color="getLabelColor(value[0]) ? getLabelColor(value[0]) : '#959595'">{{ key }}</v-btn>
+            <v-btn
+              :color="getLabelColor(value[0]) ? getLabelColor(value[0]) : '#959595'"
+              :style="{
+                'color': getTextColor(getLabelColor(value[0])),
+                'letter-spacing': 'normal',
```
Kuchenpirat, line 65: The only thing I'm a bit unsure about is that it feels slightly cramped, though that might just be because it's such a contrast to how it was before.

Let's merge it as is, and if it ends up feeling too tight in the long run, we can adjust the spacing later.

Kuchenpirat (approved):

frontend/pages/shopping-lists/_id.vue
michael-genson: It's funny, I felt the opposite probably because that's how it appears elsewhere on labels, lol. Yeah we can always change it

PR #4339, opened 2024-10-09
dev: cancel running pull request jobs on new push

Kuchenpirat:
## What type of PR is this?

- internal feature 

With some of our jobs running for quite some time, it makes sense to cancel running jobs on a new push to free up resources. This will cancel all running jobs on a pull request if a new push is made to it.

I also moved CodeQL and end-to-end tests to be called by the pull-request workflow instead of spawning on their own for pull requests.

None


michael-genson (approved):

boc-the-git:
Looks like a good change @Kuchenpirat! 🙌

PR #4341, opened 2024-10-09
docs: Update rest call for home assistant

PurelyNicole:
## What type of PR is this?

_(REQUIRED)_

- documentation

The method shown in the guide is not the best practice. It calls the API twice when one call will do. This change calls the API once, and then parses the JSON to get the sensors required.

## Testing

Used GitHub preview to ensure formatting was correct. As for content, this is how I have set up the sensors in my instance of Home Assistant.


michael-genson (approved):
LGTM, thanks!

PR #4342, opened 2024-10-09
fix: Fix Locked User Reset

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- bug

_(REQUIRED)_

Resetting locked users fails because we don't pass a `group_id`. This fix passes a null `group_id` and `household_id`.

_(REQUIRED)_

N/A


hay-kot (approved):
Shouldn't this show up with MyPy or something?

michael-genson:
No because you don't always have to pass `group_id` and `household_id`. For instance if you need group repos there is no `household_id` to pass. And `household_id=None` is different than not setting it (`None` means "no filter" whereas unset means you forgot to set it).

We throw a special exception for it though so it's very easy to catch (with tests/etc.) we just don't have coverage on this one.

PR #4370, opened 2024-10-16
fix: Prevent login via credentials when Auth Method is Mealie

cmintey:
## What type of PR is this?

- bug

The intended functionality of the AuthMethod with relation to OIDC is that when the auth method is not OIDC, then the user is able to log in with OIDC *and* the other other auth method (LDAP or Mealie) as a fallback. However, if the auth method is OIDC and the user was able to log in previously with credentials, then they should no longer be able to use credentials. This was working as intended with the LDAP auth method, but not for Mealie.

Fixes https://github.com/mealie-recipes/mealie/discussions/4368

## Testing

Added unit tests for the credentials provider


aclerici38:
Thank you for working on this! 1 thing I noticed is when the admin user is created on install, if done so through OIDC and not given a mealie password the container throws some errors. Would this be the place to resolve that as well or should I open a seperate issue?

Edit since I worded that oddly: I think the issue is moreso allowing an admin with no default mealie pass at the initial 1st account creation, that seems like a reasonable thing to require and check for.

michael-genson (approved):

cmintey:
> Thank you for working on this! 1 thing I noticed is when the admin user is created on install, if done so through OIDC and not given a mealie password the container throws some errors. Would this be the place to resolve that as well or should I open a seperate issue?
> 
> Edit since I worded that oddly: I think the issue is moreso allowing an admin with no default mealie pass at the initial 1st account creation, that seems like a reasonable thing to require and check for.

@ant385525 No, this PR does not address this. If the only admin's auth method is OIDC and they no longer can access their account for whatever reason, they can use the `change_password.py` script, which will allow them to set a password and change their auth method back to `Mealie` as well.

PR #4393, opened 2024-10-19
fix: Bulk Update Owner Removes Some Recipe Data

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- bug

_(REQUIRED)_

For more insight into why this happened, see: https://github.com/mealie-recipes/mealie/pull/2168/files

_(REQUIRED)_

N/A, reported on Discord

## Testing

Manually verified data is not lost

### DIFF
--- a/tests/integration_tests/user_recipe_tests/test_recipe_crud.py
+++ b/tests/integration_tests/user_recipe_tests/test_recipe_crud.py
@@ -483,7 +483,8 @@ def test_read_update(
         assert cats[0]["name"] in test_name
 
 
-def test_update_many(api_client: TestClient, unique_user: TestUser):
+@pytest.mark.parametrize("use_patch", [True, False])
+def test_update_many(api_client: TestClient, unique_user: TestUser, use_patch: bool):
     recipe_slugs = [random_string() for _ in range(3)]
     for slug in recipe_slugs:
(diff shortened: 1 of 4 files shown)


tests/integration_tests/user_recipe_tests/test_recipe_crud.py
boc-the-git, line 1: Seems like we expect a different end result from patch, compared to put. Could be good to add validation of that difference in the test?

boc-the-git (approved):
Can you summarise the problem in a couple sentences?

Why is patch working for you in a way out didn't? Patch updates only the fields changed, whereas put is replacing the whole object?

I added a note suggesting adding logic into the test confirming those differences could be good.

michael-genson:
Basically on the frontend we only have a `RecipeSummary` which doesn't have instructions, assets, etc. When calling update the backend assumes this is a `Recipe`, and defaults to empty instructions, assets, etc.

The `patch` method merges the old data with the new data before committing to the database

tests/integration_tests/user_recipe_tests/test_recipe_crud.py
michael-genson: Unfortunately this is a frontend issue, not backend, so I can't really write a test for it. The frontend is sending a valid payload to the backend, which is accepted. It's just missing data

PR #4402, opened 2024-10-20
dev: Improve Docs Generation

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- documentation

_(REQUIRED)_

Modifies our docs generation to be consistent. Prior to this PR, if you run `task docs:gen` twice in a row, you will get two different outputs. This is due to two things:
1. We use `datetime.now` as a default, so the current time is always inserted into the docs
2. There were some tag conflicts on the routes which are added in an inconsistent order

_(REQUIRED)_

N/A

This should enable us to add docs generation to a workflow, since now it won't needlessly add to a PR.

## Testing

I ran `docs:gen` multiple times locally and only the first time did it actually update the docs.


Kuchenpirat (approved):
looks all good and tested locally 👍
Note: I renamed the PR to dev:... to have it in the internal dev section of the release docs

PR #4415, opened 2024-10-22
feat: Shopping list UI overhaul - three dot menu

Wetzel402:
## What type of PR is this?

- feature

Moved shopping list action bar into a three dot menu to clean up the UI/UX.

None

## Testing

Tested in docker container


Kuchenpirat:
This has been marked as draft for quite some time without progress, anything you need from us to get this over the finish line?

Wetzel402:
> This has been marked as draft for quite some time without progress, anything you need from us to get this over the finish line?

I know there was some discussion among the core team regarding what should be in the menu vs what should be exposed as a separate button.  I posted in the Discord asking for feedback and then honestly forgot about it.

Kuchenpirat (commented):
Ok, so I finally made some free time and looked at your proposal, I think it is a good idea to tidy up the interface, but I think putting absolutely everything into the three dot menu is a bit too much.
So, I tried to move stuff around a bit and add it to its relevant places.

## Delete & Uncheck All Icon
The delete icon (to delete all checked items) as well as the box to uncheck all items should be where the checked items are. Especially with the delete items people will assume that it deletes the shopping list instead of the checked items.
I’d suggest something like this:
image

 
## Owner Menu
I would move the “owner” menu to the shopping lists overview page. I’d say it makes much more sense over there. 
Something like this should be pretty intuitive

image

 
## Copy & Check All
I'd think the copy and check all buttons are use pretty frequently so i would like them to not be hidden. I think a slimmed down version of the old button group is fine. 
image


## Remaing Buttons
The remaining buttons should be moved into the three dot menu since those will not be used on a super regular basis and don't need to take up prime real estate. 
The result should look something like this which should be pretty clean.
The three dot menu contains `Toggle Label Sort`, `Reorder Labels` and `Manage Labels`

## Miscellaneous
- I would also rename the button to the shopping lists overview to “All Lists” because it is not a back button since it does not go back where the user came from. 
- The image on the top should be only shown on desktop view. On mobile it takes up to much of the most important space and makes the top section look crammed. 
- I think we can remove the manage labels button from the shopping list overview page. There is nothing there happening with labels, so it is a bit confusing to have it over there. 

---

I hope you like where i am going with this. As always happy for feedback from you or other members of the community. Always happy to have better ideas suggested.

Kuchenpirat:
hey @Wetzel402 any progress on this one?
Happy to lend a helping hand.

Wetzel402:
> hey @Wetzel402 any progress on this one? Happy to lend a helping hand.

Sorry, between the holidays and work I haven't had any time to work on it.  I will try to find some time to work on this.

Wetzel402:
This is ready for review and I love how it all came together.  So much cleaner.  Thanks!

frontend/pages/shopping-lists/_id.vue
```diff
@@ -17,11 +17,74 @@
...
+                        <v-list-item-title>{{ $t("shopping-list.copy-as-markdown") }}</v-list-item-title>
+                      </v-list-item>
+                    </v-list>
+                  </v-menu>
+                  <v-btn text @click="openCheckAll()">
+                    <v-icon>
+                      {{ $globals.icons.checkboxOutline }}
+                    </v-icon>
+                  </v-btn>
+                  <v-menu offset-y>
```
Kuchenpirat: Its pretty confusing that any item inside the three dot menu has the same icon, i would suggest either not using icons or we need to switch it up a little, otherwise there is no advantage in using icons.

frontend/pages/shopping-lists/_id.vue
```diff
@@ -17,11 +17,74 @@
...
           <v-row>
             <v-col cols="3" class="text-left">
-              <ButtonLink :to="`/shopping-lists?disableRedirect=true`" :text="$tc('general.back')" :icon="$globals.icons.backArrow" />
+              <ButtonLink :to="`/shopping-lists?disableRedirect=true`" :text="$tc('shopping-list.all-lists')" :icon="$globals.icons.backArrow" />
             </v-col>
-            <v-col cols="6" class="d-flex justify-center">
+            <v-col cols="6" class="d-none d-lg-flex justify-center">
               <v-img max-height="100" max-width="100" :src="require('~/static/svgs/shopping-cart.svg')"></v-img>
             </v-col>
+            <v-col class="d-flex justify-end">
```
Kuchenpirat, line 25: any reason to not use a BaseButtonGroup here?

Kuchenpirat (changes requested):
Thanks for the quick work. 

There is one major problem with updating the shopping list owner. 
The list items are deleted if an api call is made (every time you click confirm on that dialog). You probably have to get the full list(including its items) and send that to the api instead of the summary that is used by the shoping list overview page. 
You probably can also skip the api call if the owner has not been updated. 

Besides that some minor comments for the first feedback round.

Wetzel402:
Hopefully this one is ready to pull now.

frontend/pages/shopping-lists/index.vue
```diff
@@ -6,6 +6,27 @@
       </v-card-text>
     </BaseDialog>
 
+    <!-- Settings -->
+    <BaseDialog
+        v-model="settingsDialog"
+        :icon="$globals.icons.cog"
+        :title="$t('general.settings')"
```
Kuchenpirat: As this is no longer a general settings menu i would update the title.

frontend/pages/shopping-lists/index.vue
```diff
@@ -71,6 +100,7 @@ export default defineComponent({
     const overrideDisableRedirect = ref(false);
     const disableRedirect = computed(() => route.value.query.disableRedirect === "true" || overrideDisableRedirect.value);
     const preferences = useShoppingListPreferences();
+    const settingsDialog = ref(false);
```
Kuchenpirat: this should be renamed and pushed to the state const

frontend/pages/shopping-lists/index.vue
```diff
@@ -79,6 +109,7 @@ export default defineComponent({
       deleteTarget: "",
     });
 
+    const shoppingList = ref<ShoppingListOut | null>(null);
```
Kuchenpirat: this should be renamed to "ownerTarget" and pushed to the state const

frontend/pages/shopping-lists/index.vue
```diff
@@ -136,6 +167,53 @@ export default defineComponent({
       }
     }
 
+    async function toggleSettingsDialog(list: ShoppingListOut) {
```
Kuchenpirat:
```suggestion
    async function toggleOwnerDialog(list: ShoppingListOut) {
```

Kuchenpirat (changes requested):
99% there, just a few nitpicky things that are mostly about naming stuff correctly.

frontend/pages/shopping-lists/index.vue
Wetzel402: Changed to...
```
:icon="$globals.icons.admin"
:title="$t('user.edit-user')"
```

Kuchenpirat (approved):
Looking all good now :)
This really is a big UI improvement, thanks for your work on this 😊

Wetzel402:
@Kuchenpirat, Thank you as well.  Excited to see Mealie improve.

boc-the-git:
Good stuff @Wetzel402 ! Looking good!

PR #4444, opened 2024-10-24
fix: Disable Foreign Key Checks During Restore

### LINKED ISSUE #4433: [BUG] - Restoring mealie 2.0 does not work. Users are imported before households
### First Check

I wanted to migrate my Mealie instance from sqlite to postgres. However, I cannot restore the created backup. It seems the new households feature is not yet properly implemented in the restore. 

It tries to import users before households, leading to a FK error as the households do not yet exist.

### Steps to Reproduce

1) Create a backup
2) Delete DB
3) Restore

sqlalchemy.exc.IntegrityError: (psycopg2.errors.ForeignKeyViolation) insert or update on table "users" violates foreign key constraint "fk_users_household_id"
DETAIL:  Key (household_id)=(ec8b8b72-1cd1-4081-80f0-fec7adae6f52) is not present in table "households".

### Mealie Version

ghcr.io/mealie-recipes/mealie:latest, version 2.0.0, Production
Build [a7c8b33cca37882f687a0528f93631545de74f1d](https://github.com/mealie-recipes/mealie/commit/a7c8b33cca37882f687a0528f93631545de74f1d)

### Deployment

Unraid

_No response_

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- bug

_(REQUIRED)_

In certain instances, when restoring from backup, users may be restored before households. Since users have a fk constraint with households, this will fail, because the household doesn't exist yet.

This PR disables fk constraints during restore, then re-enables them. It uses a context manager to do this, so even if there's an error it will always be restored to its initial value.

_(REQUIRED)_

Fixes https://github.com/mealie-recipes/mealie/issues/4433

Since it's possible to disable fk constraints manually (and they are disabled on sqlite by default) I made sure to preserve the existing setting rather than disable/re-enable. This approach ensures if fk constraints are disabled before Mealie touches them, then they stay disabled.

## Testing

I used the backup related to https://github.com/mealie-recipes/mealie/issues/4433 to reproduce the issue locally on postgres, then observed that it's fixed now. I also added an anonymized backup to our test suite.


Kuchenpirat (approved):
Looks all good 👍

PR #4452, opened 2024-10-25
feat: Show Cookbooks from Other Households

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- feature

_(REQUIRED)_

This PR allows users to see all cookbooks from all households. They're organized into menus as not to bog down the UI too much:
image

Note that in the screenshot above, "Bento", "Cocktails", and "Sides" are all cookbooks belonging to the currently logged-in user's household, so they're not in a menu (just like before this PR). If a user is not logged in, they'll see all cookbooks in a menu (since they don't belong to a household).

Cookbook order is preserved within each household, and each household is sorted by household name (other than your household, which is always on top).

_(REQUIRED)_

N/A, discussed in Discord.

I removed the edit button from cookbooks that aren't yours. In the cookbook edit page, only your households' cookbooks appear, since you can't edit other households' cookbooks. To make this cleaner I added a new composable `useMyCookbooks` (instead of `useCookbooks`) and created a factory function (since the only difference between the two is having a filter or not).

## Testing

Updated tests since the API returns all cookbooks in the group now (rather than filtered by household).


Kuchenpirat (commented):
Looks pretty good overall, just a few minor things. 

1. I think it makes sense to add a toggle to the cookbook page to hide/show cookbooks from other households. I have five households and i don't see myself using their cookbooks frequently. 
2. When creating a new cookbook, one has to refresh the page to have it shown in the navigation drawer. Having this pop up directly would make feel this a bit more streamlined.

michael-genson:
A toggle makes sense to me. For now do you think a frontend toggle is sufficient? I can see a world where we add this as a configuration option, a la hiding other features users don't use

Kuchenpirat:
Yeah, frontend only sounds fine to me 👍

michael-genson:
Added the checkbox here:
image

I also fixed the reactivity for adding/editing/removing cookbooks. My new store was causing issues so I just reverted the entire store change and made some changes on how cookbooks are accessed on the cookbooks page.

Kuchenpirat (dismissed):
Looks all good 👍🚀
Let's get this in, finally! 😊

michael-genson:
I just realized I broke re-ordering cookbooks; should be fixed now though

Kuchenpirat (approved):
Next try 😊👍

PR #4453, opened 2024-10-25
fix: Drag Delay Only Apply On Touch

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- feature

_(REQUIRED)_

Apparently there's a bug with Vue Draggable's `delay-on-touch-only` boolean; you have to set it to `true` to get it to work (usually just putting `my-bool-val` works, but instead we have to do `:my-bool-val="true"`).

_(REQUIRED)_

N/A


Kuchenpirat (approved):
Next try then 👍

Kuchenpirat:
Works like a charm now 👍

PR #4456, opened 2024-10-26
feat: implement the possibility to add tls

wim-de-groot:
## What type of PR is this?

- feature

Mealie needs to have the option to use a TLS certificate, for the following cases:
- applications that expose their mealie not via an ingress controller but via tools like TwinGate or TailScale
- if you run your ingress controller in transparant mode
- if you do not use Kubernetes and you just want to run mealie

Fixes: #2294

I am not sure if the naming of the properties is in line with the project, I would like feedback on that. 

## Testing

I've tested this change by deploying the app to my homelab. Since the features of uvicorn are mostly used here I am not sure if there are more unit tests needed.


wim-de-groot:
@michael-genson, you  seem to contribute a lot to this project. I'am not able to find anything about a process on asking for a review. Do you know how this works?

michael-genson:
Hey there!

All of the maintainers contribute to Mealie in their spare time, so we don't really have a review schedule. We will get to it eventually, promise!

michael-genson (changes requested):
Looks good, thanks for this! Could you add the new variables to the docs?

`docs/docs/documentation/getting-started/installation/backend-config.md`.

You can put it between Webworker and LDAP (unless you think it's better somewhere else). And maybe add a small blurb that this is only for serving Mealie without a webserver/reverse proxy.

michael-genson (approved):

PR #4467, opened 2024-10-29
feat: Automatic instruction timers

codetakki:
## What type of PR is this?

_(REQUIRED)_

- feature

_(REQUIRED)_
This pr is my suggestions for timers that are added directly to steps, making the timer feature more automatic.
I suggest we attempt to parse the instructions text to find out timers. The regex is currently very basic, and uses locales to support single and plural of minute in different languages. I hope this catches most cases. Timers with hours could easily be added to. For seconds i dont think its  necessary. 
I have placed the timer below the step, as i don't like the suggestion of adding it inline in the text, as it can be hard to click,  does not look good and easy to miss. 
The timer is simple, but does allow for reducing or adding some seconds if needed. 
It might also be a option to only show it during cooking mode, but as cooking mode only is available if i link products to steps i don't like this restriction. 

#### images

15 min timer detected
image
Timer is running
image
Timer is paused
image
Timer is done, Clock shakes to call attention
image
Multiple timers are supported:
image

_(REQUIRED)_

This is just a suggestion, but having a timer in directly in the instructions would be a great improvement. I'm happy to adjust after feedback.

## Testing

Checkout pr.
Go to a recipe
include "5 min" in a instruction step 
You should see a timer appear under the instruction. 
It should be able to be started, paused, and reset. 


Jaykurb:
This looks great!! Nice work!

boc-the-git:
Haven't looked at the code yet, but like the look of this!

Your description mentions clock shaking. Do they make noise? I think the current timer does.

Kuchenpirat:
You only have to provide the translation in the enUs file.
The rest is handled by crowdin

frontend/components/Domain/Recipe/RecipePage/RecipePageParts/RecipePageInstructionsTimer.vue
```diff
@@ -0,0 +1,191 @@
...
+      // Makes regex work with multiple languages via locales
+      const minuetStrings = computed<string[]>(() => {
+        console.log(i18n.availableLocales);
+        const result: TranslateResult[] = [];
+        i18n.availableLocales.forEach((locale) => {
+          result.push(i18n.t("recipe.timer.minute", locale ))
+          result.push(i18n.t("recipe.timer.minutes", locale ))
+        })
+        return result
+      })
```
michael-genson: Love this. Might be worth considering hours and seconds, too

boc-the-git (dismissed):
Please remove all translation files other than en-US. 
Translation are done via Crowdin.

frontend/components/Domain/Recipe/RecipePage/RecipePageParts/RecipePageInstructionsTimer.vue
```diff
@@ -0,0 +1,191 @@
...
+        for (const match of matchesMin) {
+          timers.push(createTimer(Number(match[1]) * 60));
+        }
+        return timers;
+      };
+      const timers = computed(() => {
+        return extractMinutes(props.text as string);
+      });
+
+      const createTimer = (totalSeconds: number) => {
```
michael-genson: A lot of this code is re-used from `RecipeTimerMenu`. Can we extract this logic into a composable so we don't have duplicate code?

michael-genson (changes requested):

codetakki:
> Haven't looked at the code yet, but like the look of this!
> 
> Your description mentions clock shaking. Do they make noise? I think the current timer does. 

Yes it makes the same sound as the existing one

frontend/components/Domain/Recipe/RecipePage/RecipePageParts/RecipePageInstructionsTimer.vue
boc-the-git: Yep, if we're inferring time rather than a user configuring it, we need to handle all cases.

codetakki: I have a feeling that hours and second s will add alot of complexity relative to little payoff.
Recipes that need timers over a hour are alot fewer than minute timers. I don't really see what recipe need seconds in precision, and if they do, users will mostly like use dedicated apps as having a web thats unreliable unless it's in focus would impact the precision.

We would also need to adress how to parse out what hour, minutes and seconds are part of the same time in a step.

If a step is: 
"Boil for 3 hours and 30 minutes, rest for 30 minutes"

Parsing this with just some regex seams hard to me. An natural language model might be able to help with this, but now where using complex  tools for little improvement, for a pretty simple feature.

(I could go one, like how some langues format time different, etc) 

Making sure that minutes and mins in multiple languages work seams like a more realistic goal.
If they want hours the can add it with "60 mins" no headache with complex time formats.

If we let the user configure it them self, it will be a lite used feature, as the payoff is so little compared to the setup needed.
These are my thoughts on this.

michael-genson: Would it maybe make sense to parse instructions upfront on the backend and store them in the database? That way we could leverage a few different parsing methods, while also enabling users to manually add timers directly.

For instance, on import, for users with OpenAI integration, we can add an additional processing step which lets an LLM parse instructions for timers.

There's also [this package](https://pypi.org/project/timelength/) which enables time duration parsing (and has support for custom localizations)

codetakki: That would absolutely be a option. Sadly i don't think I have the skills to implement a solution like this.
Possible to start with simpler solution and add better parsing later? This way we can get a feel if more complex parsing Is needed. 
The python package seams like a good middle ground.

michael-genson: Yeah if we go the backend route we can keep it simple for now:
1. Add a database entry to store timers on recipe instructions (we can keep it simple and store them as strings in ISO 8061 duration format)
2. When a recipe is created, parse instructions for time strings (any multi-lingual method works, such as that Python package, or any other package)
3. Allow users to add timers manually

And then on database migration we can parse all existing recipes and add missing time duration strings

boc-the-git: It's clearly incomplete, but this is my WIP branch that would store a user configurable timer in the DB:
https://github.com/boc-the-git/mealie/compare/mealie-next...boc-the-git:mealie:feat/timers-per-recipe-instruction

(just for reference)

codetakki: How could we handle locales? The python package supports including a json file, maybe we can read a key from the i18n locales?

codetakki:
I removed the locales changes except for en. I'm fluent in Swedish, so could  that back in, or should that to be done in Crowdin later?

Kuchenpirat:
Thanks 👍
Everything besides the base en_US file should be done via crowdin. So don't add any of them back in. You are more than welcome to provide translations via crowidn as soon as this PR is merged.

codetakki:
As I want locales to work during development, how would you suggest I try them out, should i just ignore my translation files?

Kuchenpirat:
i am not sure if that works for files that are already tracked via github. 
If not you can just add the translations for swedish and other languages you need and we will ask you to remove them once this is ready to merge.

boc-the-git:
You can have the Swedish language file in your Dev environment and just not commit it. Nothing complicated to do there.

codetakki:
Fair enough 😄

frontend/components/Domain/Recipe/RecipePage/RecipePageParts/RecipePageInstructionsTimer.vue
codetakki: Probably, are you thinking something that just generates the functions and state values and returns them as a object?

michael-genson: Yeah basically keep all of the state management in the composable and only expose a bare-minimum API for the components to use. Something like:

```js
export const useTimer = function () => {
  ...  // your time code
  const controls = {
    updateTimer,
    pauseTimer,
    resetTimer,
    ...
  }
  return { controls }  // or anything else the consumer might want
}
```

Additionally, you could potentially combine `RecipePageInstructionsTimer` and `RecipeTimerMenu` into one component (or two components with a shared sub component). But either way we should probably move the logic/state into a composable.

michael-genson: You can take a look at our store actions factory to see kind of what I'm getting at: https://github.com/mealie-recipes/mealie/blob/mealie-next/frontend/composables/partials/use-actions-factory.ts

codetakki: I have extracted the timer logic from the existing timer, adjusted it a little and reused it in both timers.

boc-the-git:
I'm a bit removed from the detail at present, so don't have anything too constructive to say right now @codetakki, but just wanted to offer some positive words.. this change, once it's merged, will be a huge improvement to Mealie! So thanks for taking on the effort, and being responsive to the feedback offered :heart: :rocket:

michael-genson:
Composable looks good 👍 I'm still interested in parsing upfront using the backend and storing it in a new field on the recipe instruction, though, and letting users add/edit timers explicitly like in the branch @boc-the-git shared. Sort of similar to what I did with my [recipe yield PR](https://github.com/mealie-recipes/mealie/pull/4489).

codetakki:
Hey i spend some time trying to set the backend parsing up, and setup a new column for timers.
I however dont think i will have the motivation to spend much time on the backend side currently. If some want to contribute the backend part i will be glad to make the ui changes for it.

michael-genson:
I implemented the backend using [dateparser](https://github.com/scrapinghub/dateparser). It's made by the same team that maintains extruct, which is used in recipe scrapers. It also supports [a ton of translations](https://dateparser.readthedocs.io/en/latest/supported_locales.html) out of the box. It's not perfect, but it works pretty well. As long as we support editing timers manually I think it's plenty effective.

One caveat to this approach is that parsing takes waaaayyyyy too long to run on existing recipes. On my humble 200 recipe instance it takes several minutes to parse everything (it has to parse every instruction on every recipe), and I guarantee many users will assume something's wrong and kill the container during migration which will cause migration problems. So I decided not to parse existing recipes and instead only parse recipes as they're created.

michael-genson:
The frontend still needs to be updated to read/write the new backend field `timers`. It's a list of integers which represent the timer duration in seconds.

Ignore the new `timers_json` field in the database; that's an implementation detail since SQLite doesn't support JSON objects like postgres does. The backend will automatically translate the list of integers to/from the JSON field.

codetakki:
Thats super! Do you think we can add an option to trigger the parsing for all recipes manually? That way we can display a warning that it will take a while and even a progress indicator, and we can get the timers on existing recipes. We could also add a button under specific recipes to parse its timers?

Does the parser use all the locales for parsing? We might be able to speed it up by allowing the user to limit what languages are used?

michael-genson:
>Do you think we can add an option to trigger the parsing for all recipes manually

Yeah this isn't a bad idea. We could add some sort of startup task which prompts the user about the new feature. Alternatively we could add a button in the recipe to trigger it on a per-recipe basis.

> Does the parser use all the locales for parsing?

Great point, hadn't thought of that. It definitely does

michael-genson:
Oh you literally suggested the button on each recipe in the next sentence. Missed that, haha

codetakki:
I have added ui to change timers:
image
Is there a existing endpoint for parsing the timers for all/singe recipes?
Do we want to parse the timers when a step is edited?

michael-genson:
I'm open for feedback on that one. I think either we need to parse everything on migration (which means we need to figure out how to make it faster, _or_ telegraph to the user after the app loads up to trigger it manually one-time) or have an endpoint to parse the entire recipe.

I'm leaning towards "figure out how to make it go faster" I just haven't looked into it too much. You can try playing around if you have the time: https://github.com/scrapinghub/dateparser

codetakki:
i ran some simple tests and tried to run the dateparse 100 times  the same string that around the length of a instruction.
It seems that limiting the number of locales improves the times ALOT. 
Parsing  "this step will take around 5 mins, and then it will be done. Add 10 hours to your timer if not done and then add the eggs" * 100 Without defining locales took around 3.9 s 
With languages set to ["en", "sv"] it took 0.15s to parse it 100 times
So i would say if we start there, it might be fast enough to work at migration. Is int there a global set language that we can use? That plus English should cover most cases

michael-genson:
Awesome, I should have time today or tomorrow to give that a whirl.

Off the top of my head there's no language setting the backend can reach (without a request from the frontend) but I'm thinking we can do something like pre-scan recipes for language and then use those

codetakki:
Hmm the date parser did already have function to do that, so it might be that it's guessing the language each time. So the improvement there would be that we do it once per recipe.

michael-genson:
I went and added this to the migration. We take 20 random instructions (each from a unique recipe) and try to detect languages from them. We then parse all ingredients using those detected languages.

In my testing it was pretty fast (10-15 seconds).

codetakki:
Not bad! When a new recipe is added it uses all of them like before?

michael-genson:
Yup, I only added it to the migration, otherwise it's the same as before.

codetakki:
Alright, so now we have a ui, functional backend and UI for editing timers. 

What would be left to change/add? I feel like additional features could be added later if wanted, as this seams pretty good for a first version.

codetakki:
i was just reminded of the mass import feature, there we might still see long loading times. i have personally imported over 100 recipes once. And it already felt like that might have frozen the system.  But then again, i dont know how common that is.

michael-genson:
For the sake of keeping this PR in a reasonable scope I think we can leave it for now. That's something we can easily address later.
In terms of getting this merged I think it's good to go, pending a review

michael-genson:
It looks like you'll have to fix the merge conflict in `RecipePageInstructions.vue`. It's related to the new cook mode changes. I'd also verify the timers look okay in cook mode.

For the poetry conflict, just run `poetry lock --no-update` and commit whatever that spits out

codetakki:
Is there any documentation that we want to update?

hay-kot (commented):
Sorry I'm late to the party on this one but I have some general concerns about functionality here and the overall UX. I think my main questions are

- If I start a timer and close my phone or it goes to sleep what happens to that timer? Does it keep ticking, will it notify me when it goes off?
- If I refresh the page, what happens to the timer?

I've tried implementing this kind of thing before and there was no way to keep the timer running when the phone went to sleep or the browser app was left. 

Overall, I think that this restriction makes this kind of change confusing for users and unpredictable. What I would like to avoid is having users filing issues about timers "not working" when implementing a fully functional timer isn't possible.

I know a ton of work has gone into this and I don't want to be the one to say we should throw this all away this late in the PR process, but I'd at least like to have a conversation about how this would get used and what the unexpected limitations are.

michael-genson:
> If I start a timer and close my phone or it goes to sleep what happens to that timer?

Yeah after thinking about it more this is definitely a big concern. As a user I would assume the timer would continue to work, but as a developer I'm pretty sure there's no way to get that to work without an actual app. The issue with refreshing the page is less important to me, but I lock my phone instinctively, and not all devices support wakelock

codetakki:
I have had the same concern and even looked into a solution but turning up empty-handed, even though pwa have come a long way, having it track time in the background seams impossible/impractical. Unless we implement web notifications. 
I want to mention that this issue was also present in the current timer implementation, which presents the same issues. 
Mealie does have the feature where it keeps the screen on while looking at a recipe, but more than that will be hard.

I have some things in mind to improve it somewhat:
Looking into: https://www.npmjs.com/package/worker-timers, it might be able to improve it somewhat, but needs support for web workers. 
Warn the user first time they use the timer that its not perfect, and should keep the screen on.
Add a "timeStarted" variable that we can double-check time progression, so if we close the app and open it again, as long as that variable is kept (local storage, session etc) we can at least notify the user that the timer has progressed. 

Or: Offload the timer to a separate app, using app scheme URL or something.

codetakki:
I was looking for some kind of feedback to be able to get this pr merged. So let me know what should be done.

Kuchenpirat:
Sorry for our delayed response.
All maintainers are currently pretty swamped with life, work, and other responsibilities outside of Mealie, which is why progress has been slow.

While I like most of the ideas on how to minimize the problem, I am somewhat hesitant to approve a timer with restrictions on functionality.
This could lead to situations where the timer fails to ring, causing user frustration. On the other hand, I feel that using the system timers on phones, tablets, or voice assistants isn’t much of a hurdle, and they are far more reliable.

We could require wake-lock, but this wouldn’t help in cases where users lock their phones or switch apps. Additionally, we must consider the current state of iOS wake-lock, which is far from ideal.

I’ll push for input from the other maintainers so this doesn’t stay in limbo much longer.

Again, sorry for the delay, and thank you for your efforts to add new features to Mealie, making it better than it was before! 😊🚀


(Considering my reasoning above, we should also remove the existing timer to remain consistent.)

Kuchenpirat:
Hey @codetakki, 
Thank you so much for your work on this feature. We truly appreciate the effort and thought you put into it. 🙏
We had some internal discussion and came to the conclusion that we are not able to include this feature into mealie with the limitations that come by being a pwa / webpage only. 

Having a timer that works only under certain situations is not something we can support and the understandable user frustration due to missed timers we can support or effectively manage at this time. 
If there is a way in the future to have timers ring from mealie on the lock screen we are more than happy to revisit this. 

Again thanks a LOT for your work on this feature, it really could have made a big positive impact. 
Hope to see you around some more.

CloCkWeRX:
https://issues.chromium.org/issues/41417116 tracks the (non)availability of 'scheduled tasks' as an API in chromium based browsers.

Alternatively, if mealie's UI was packaged or used with a chrome/firefox/etc extension; https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/alarms could be used possibly to set up alarms for a future point in time.

Potentially publishing JSON LD/hevent/etc data for when an event would happen could allow for opt in to this.

Alternatively (not needing an extension), add ical reminder buttons to rely on the user's calendar app?
https://stackoverflow.com/questions/55109527/ical-add-event-to-calendar-via-webcal
(Better suited for "I am cooking something in the oven for X hours" perhaps, rather than "check in 3 minutes")

codetakki:
i have already been thinking about possibility to use a system like "[capacitorjs](https://capacitorjs.com)" to bundle the app into a native package. This allows us to access native apis while keeping the frontend running as a website, but its loaded localy.

atiris:
Hi everyone,  
I realize I'm joining the conversation a bit late, but I wanted to share a few ideas regarding the timer functionality.

**Balancing Robustness and Practicality**  
I understand the desire for a truly robust solution. However, striving for perfection can sometimes become counterproductive - if we aim to cover every edge case and eliminate even the smallest imperfection, we risk delaying implementation indefinitely. A straightforward approach, even if it isn’t “perfect” in a technical sense, may be more practical and serve the vast majority of use cases effectively.

**Embracing the Nature of a Web-Based Timer**  
The concept of a timer tied to an open page aligns well with user expectations for many web applications. Users intuitively understand that a timer operates only while the page is active. In this case, a simple yet transparent solution would be to:  
- Clearly **indicate that the timer runs** solely as long as the **page is open**.
- Display a **notification or warning** if the user attempts to leave the page while the timer is still active, thus informing them that the timer will stop.

Just like battery-operated devices don’t suddenly alert you when you remove batteries with an alarm set up, users should be aware of and responsible for managing a web-based timer’s limitations.

**Alternative Mechanisms**  
Have we considered other possible options aside from leveraging a web app framework or using something like Capacitor to run a native timer? For example **Web Notifications**: By integrating web push notifications, we could alert users even after the page has been closed. This wouldn’t necessarily require the timer to run persistently on the client side—rather, we could schedule a server-based process that triggers a notification once the time is up.

**Fundamental Solution**  
Personally, I prefer a web page where I can simply click on a timer, and it automatically counts down the set time. This is far more intuitive than a time, you have to manually retype to the another clock app, and then figure out which timer to refer to when it starts beeping.

How about as a solution having the option to enable/disable in the settings: **Enable web timer when cooking** _(this timer will only remain active on the open page, we recommend using your system's native timer)_

Have you tried letting users who use Mealie when cooking vote on whether they prefer to be able to run such a simple timer, or be forced to always use guaranteed timers?

atiris:
Since I have a strong opinion on this (and it may be distorted by actual usage of the app), I wrote a [simple post with a vote](https://github.com/mealie-recipes/mealie/discussions/5356) in discussion. I'm interested in what others think. Please participate (the more diverse the audience, the more representative the result). Thank you.

bugs181:
> Hi everyone, I realize I'm joining the conversation a bit late, but I wanted to share a few ideas regarding the timer functionality.
> 
> **Balancing Robustness and Practicality** I understand the desire for a truly robust solution. However, striving for perfection can sometimes become counterproductive - if we aim to cover every edge case and eliminate even the smallest imperfection, we risk delaying implementation indefinitely. A straightforward approach, even if it isn’t “perfect” in a technical sense, may be more practical and serve the vast majority of use cases effectively.
> 
> **Embracing the Nature of a Web-Based Timer** The concept of a timer tied to an open page aligns well with user expectations for many web applications. Users intuitively understand that a timer operates only while the page is active. In this case, a simple yet transparent solution would be to:
> 
>     * Clearly **indicate that the timer runs** solely as long as the **page is open**.
> 
>     * Display a **notification or warning** if the user attempts to leave the page while the timer is still active, thus informing them that the timer will stop.
> 
> 
> Just like battery-operated devices don’t suddenly alert you when you remove batteries with an alarm set up, users should be aware of and responsible for managing a web-based timer’s limitations.
> 
> **Alternative Mechanisms** Have we considered other possible options aside from leveraging a web app framework or using something like Capacitor to run a native timer? For example **Web Notifications**: By integrating web push notifications, we could alert users even after the page has been closed. This wouldn’t necessarily require the timer to run persistently on the client side—rather, we could schedule a server-based process that triggers a notification once the time is up.
> 
> **Fundamental Solution** Personally, I prefer a web page where I can simply click on a timer, and it automatically counts down the set time. This is far more intuitive than a time, you have to manually retype to the another clock app, and then figure out which timer to refer to when it starts beeping.
> 
> How about as a solution having the option to enable/disable in the settings: **Enable web timer when cooking** _(this timer will only remain active on the open page, we recommend using your system's native timer)_
> 
> Have you tried letting users who use Mealie when cooking vote on whether they prefer to be able to run such a simple timer, or be forced to always use guaranteed timers?

Chiming in my thoughts on the matter too. Although Mealie aims to be freely accessible, and is by some other users - I'd argue the vast majority of them are running this as self-hosted. This goes to say, most people that are using Mealie are somewhat technical. I'm sure that the most of them would even understand the timer feature limitations and expect them. I for one, really wish to have this feature. I run Mealie on an iPad, running in a nested Home Assistant screen. The screen stays active ALL the time. I wish this feature was added, even under experimental section of settings which goes on to further explain the limitations. It truly is unfortunate that this isn't added because it doesn't cover every edge case.

maciel310:
Just curious, has anyone explored using the Media Session and Web Audio APIs as a route to having a background-compatible alarm system? Based on [this page](https://progressier.com/pwa-capabilities/audio-player-pwa) it sounds like it should work on both Android and iOS, and would have the added benefit of showing the timer status on the lock screen as well.

Basically the idea would be to use the Web Audio API to dynamically generate an audio file with X seconds of silence (based on timer length), followed by the alarm tone. Then pass that into the Media Session API to start playing that audio when the timer is started.

There would still potentially be issues where the user could pause the player, and possibly some issues if they started audio in another app (ex: music, podcast, etc). Some of this could be mitigated by automatically updating the playback position using the Web Audio API if these are detected (should be able to compute when the timer should end and compare that with how much time is left in the audio track).

Any thoughts if this would sufficiently address your concerns, @Kuchenpirat?

JuMi46:
Hello everyone, chiming in with my possible solution to make timers fully reliable in Mealie; with webhooks and a newly created timer app solely for this purpose. Check it out and let me know. [discussions/7540](https://github.com/mealie-recipes/mealie/discussions/7540)

PR #4489, opened 2024-11-01
feat: Structured Yields

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- feature

_(REQUIRED)_

This PR adds a new column on the recipe `recipe_yield_quantity`. This splits the recipe's yield into two fields: qty and description/unit (where the old column of `recipe_yield` serves as the latter). We also attempt to parse existing yields upon migration.

For parsing, we assume the yield string starts with the quantity, and the rest of it is the unit. We also only consider numerical quantities (so we aren't parsing "One Serving" into "1" and "Serving"). Users can manually update the quantity/unit themselves, so it doesn't have to be perfect. Backwards compatibility is maintained since both the quantity and the unit are displayed, although there are a few edge-cases where it's not perfect (e.g. "3-4 servings" will display as "3 -4 servings", because we parse "3" as the qty, and "-4 servings" as the unit). I think it's preferred to parse these edge-cases and expect the extra space, rather than not parse them, since it populated the yield qty (which we'll leverage in future PRs).

For scaling, the user experience hasn't changed much, but I had to re-work how we do it since we don't extract yields anymore. Instead scaling requires that the recipe has a yield quantity. The only user-facing change is that previously we allowed scaling even if we couldn't parse the servings string (by assuming the quantity is 1). Since it's user-editable now I don't see why we need to support this anymore (if you want it to work, set the quantity).

For editing it's a simple UI change:
image

When displayed it looks the same as before:
image

While this PR doesn't add a ton of value upfront, it opens the door to a lot of enhancements (e.g. filtering by yield).

_(REQUIRED)_

N/A

## Testing

Frontend tests and pytest.


michael-genson:
I don't know how I feel about "Servings" and "Servings Text" on the UI. Suggestions for that are welcome!

FuzzyWereBeast:
When I inquired about this on discord, the reason I was asking is for some recipes you will have makes 4 servings and yields 8 cups.  This implementation gives the ability to list "4" "2 cup servings" and have the 4 scale with the recipe.  The other option would be leave the "servings" as is, but provide a 2nd field for total yield.  I understand both directions.  I'm just better explaining the question I had earlier.  Either can work.

In the current implementation, "servings" and "servings text" would work.  "servings type" would also work.

hay-kot:
Some context from working in [Recipinned](https://app.recipinned.com) I decided to split "servings" from "yields" as they serve different purposes (at least that's what I've talked myself into thinking!)

Here's an example from our editor, it gives a pretty decent description of how we see each type

> Recipe Yields indicate the total amount of food a recipe makes (e.g., 12 muffins, 24 cookies). Servings specify how many people the recipe will serve. We use the Servings field to scale recipes, ensuring you know the exact quantity to prepare.

## Examples

### Editor

CleanShot 2024-11-07 at 11 38 11@2x

### View 

CleanShot 2024-11-07 at 11 41 51@2x


### Scale

CleanShot 2024-11-07 at 11 41 00@2x

michael-genson:
Ahh interesting. If we go this route with Mealie, we'll probably have to assume the yield and serving are the same when migrating (because we don't have different values). But this would allow users to specify.

Curious, when importing a recipe, how do you decide between the two? Or do you just leave "yield" blank?

FuzzyWereBeast:
This type of changes shown in the photos would be welcome. Especially the ability to just indicate how many servings instead of a straight doubling + when scaling. It will streamline the process of making sure you have the correct amounts instead of needing to check how many servings are in the original recipe.

hay-kot:
> Ahh interesting. If we go this route with Mealie, we'll probably have to assume the yield and serving are the same when migrating (because we don't have different values). But this would allow users to specify.
> 
> Curious, when importing a recipe, how do you decide between the two? Or do you just leave "yield" blank?

Nothing super sophisticated, here's the relevant go code

```go
	servingsMatch := []string{
		"servings", "serves", "serving", "yield", "makes",
	}

	// If text is empty we assume that it's talking about how many people it serves
	// generally, if that's not the case it will specify the amount of servings in the
	if servingsText == "" {
		create.Serves = create.Servings
	} else {
		lowerText := strings.ToLower(servingsText)
		for _, match := range servingsMatch {
			if strings.Contains(lowerText, match) {
				create.Serves = create.Servings
				break
			}
		}
	}
```

michael-genson:
I made several changes to incorporate this:
1. I added a new `recipe_servings` field
2. I updated the parser (using the example from @hay-kot above as a reference)
3. I refactored how we render servings vs yield, see below

image

Adjusting the scale also adjusts the yield. You still need a servings count to enable scaling

Kuchenpirat (commented):
Finally had time to look at this (just a little bit). 

I think the upgrade path adds quite a bit of friction to actually using this (mostly to do with separating it into yield and servings). 

<hr>

All recipes that were previously scalable will now no longer be scalable and one has to manually add the serving to re-enable that functionality. 
I think a fallback to the Yield instead of disabling this completely would make sense to have have it match the old behaviour. 

Same thing for recipes that have no serving count or yield at all. Defaulting to 1x Recipe and allow user to scale would feel more intuitive then not allow scaling at all. I don't see a reason why we should disable scaling in this situation if it is technically possible (Have not looked at the code at all but can't imagine setting a default to 1 would be to hard)


<hr>

Another Observation after filling out the yields on some recipes. Most recipes yields are the Serving count. After adding the servings there are two elements on top of each other displaying the same information. 

image

Maybe displaying only one of the two if the field matches would make sense, or moving the yield towards the Recipe Info Panel with cooktimes etc would loosen that up a bit. 

<hr>

Just to be not all negative, i think this has quite some potential and its a logical step to have the yield extracted into its own field.

michael-genson:
Thanks @Kuchenpirat! Love the feedback. I've gone ahead and implemented these suggestions. Refactoring the recipe info at the top was long overdue, and I really needed to do it to get the yield in there, so that's part of this PR now too. Happy to take feedback on the new layout.

This is the new recipe info layout, including the yield (if set):
image

The yield will no longer scale, since it's in the info card. I don't think changing this value with scaling makes sense, because it's no longer visually anywhere near the scale controls.

With no yield:
image

The time cards on the right are always stacked, and always in their own part of the info card. If there's not enough room it wraps:
image

---

With the refactor, we don't have a bunch of duplicated elements when in/out of landscape. I also separated the info card display from the editor. This led to some... weirdness in the editor with the layout, so I added some text to the top of the editor:

image

This also resulted in some tweaks to landscape mode:
image

---

Additionally, scale now works as you suggested: we look for servings qty first, then yield qty, then a default of "1". The text always reads "Serves X" regardless of which value it uses (I think this makes sense since you're scaling how much you're making, however regardless changing this would be a nightmare).

frontend/components/Domain/Recipe/RecipePage/RecipePageParts/RecipePageHeader.vue
```diff
@@ -1,46 +1,12 @@
...
-      >
-      </v-img>
-    </div>
-    <v-divider></v-divider>
+    <RecipePageInfoCard v-if="!isEditMode" :recipe="recipe" :landscape="landscape" />
+    <v-container v-else width="100%">
+      <v-card-title class="headline d-flex justify-center align-center pt-0 pb-12 mb-4">
+        {{ $t("recipe.editing-recipe", [recipe.name]) }}
+      </v-card-title>
+    </v-container>
```
Kuchenpirat: I'd like to still have access to the image while editing to be able to gauge whether all relevant details are still present in the image crop

frontend/components/Domain/Recipe/RecipePage/RecipePageParts/RecipePageInfoCard.vue
Kuchenpirat, line 1: I feel like the refactor this is not there yet all the way and needs some future work. Especially if the two columns are no longer besides each other it looks a bit off to have one column centered to the right and the other centered. 

I try to think about my ideas a bit and do a mockup of what i have in mind.

frontend/components/Domain/Recipe/RecipePage/RecipePageParts/RecipePageInfoCard.vue
```diff
@@ -0,0 +1,95 @@
...
+                </v-col>
+              </v-row>
+              <v-row no-gutters>
+                <v-col cols="12" class="d-flex flex-wrap justify-center">
+                  <RecipeLastMade
+                    v-if="isOwnGroup"
+                    :class="true ? undefined : 'force-bottom'"
+                    :value="recipe.lastMade"
+                  />
+                </v-col>
```
Kuchenpirat: Somewhere along the way this got broken. I am no longer able to add a recipe to the timeline via the Dialog on the recipe page.

frontend/components/Domain/Recipe/RecipeScaleEditButton.vue
```diff
@@ -1,16 +1,12 @@
...
+        <v-menu v-model="menu" :disabled="!canEditScale" offset-y top nudge-top="6" :close-on-content-click="false">
           <template #activator="{ on, attrs }">
             <v-card class="pa-1 px-2" dark color="secondary darken-1" small v-bind="attrs" v-on="on">
-              <span v-if="!recipeYield"> x {{ scale }} </span>
-              <div v-else-if="!numberParsed && recipeYield">
-                <span v-if="numerator === 1"> {{ recipeYield }} </span>
-                <span v-else> {{ numerator }}x {{ scaledYield }} </span>
-              </div>
-              <span v-else> {{ scaledYield }} </span>
+              <!-- eslint-disable-next-line vue/no-v-html -->
```
Kuchenpirat: Can we add an edit icon to this? 
Had user feedback that it is not intuitive that this can be clicked to open the scaling menu. 

This would help people figure this out easier.

frontend/components/Domain/Recipe/RecipePage/RecipePageParts/RecipePageInfoCard.vue
```diff
@@ -0,0 +1,95 @@
...
+          <SafeMarkdown :source="recipe.description" />
+          <v-divider />
+          <v-container class="d-flex flex-row flex-wrap justify-center align-center">
+            <div>
+              <v-row no-gutters class="mb-1">
+                <v-col v-if="recipe.recipeYieldQuantity || recipe.recipeYield" cols="12" class="d-flex flex-wrap justify-center">
+                  <RecipeYield
+                    :yield-quantity="recipe.recipeYieldQuantity"
+                    :yield="recipe.recipeYield"
+                  />
```
Kuchenpirat: This does currently not scale when increasing/decreasing the servings. 
It seems like the component already supports it, so we should probably make use of it.

Kuchenpirat (commented):
Happy to hear my feedback was well recieved and i think we are heading towards the right direction, but i feel a few things have not yet clicked into their right place. 

So onto my next round of Feedback, i have skimmed over the backend and it seems pretty solid (as usual on a michael PR 👍) but i have a few more notes on the Frontend UI/UX

frontend/components/Domain/Recipe/RecipePage/RecipePageParts/RecipePageInfoCard.vue
michael-genson: I can change those back to centered, I like it in that the times are aligned, but you're right it looks weird, especially when it wraps on smaller screens

In general I would love someone to just re-arrange the whole thing and make it better, but I don't have enough confidence in my UI/UX to do it

frontend/components/Domain/Recipe/RecipePage/RecipePageParts/RecipePageInfoCard.vue
michael-genson: I can add that back in, but it was a conscious choice on my part to omit it since it's no longer near the scale controls. To me it reads "this recipe normally yields X Amount" and then the scale over by the ingredients says "this is what it's set to right now"

But maybe I'm overthinking it and it should just scale everywhere

Kuchenpirat: I think it makes sense to have it scale. 
Just to avoid any confusion and have it display data that is coherent with the ingredient amounts.

michael-genson:
I restored the full info card to edit mode; I don't really like how information is duplicated, but it's like that now, so it's not a regression. I'll probably tackle this in a later PR:
image

(The recipe times are now centered as well, as seen above)

Scale with edit icon:
image

frontend/components/Domain/Recipe/RecipeScaleEditButton.vue
```diff
@@ -1,16 +1,13 @@
...
+        <v-menu v-model="menu" :disabled="!canEditScale" offset-y top nudge-top="6" :close-on-content-click="false">
           <template #activator="{ on, attrs }">
             <v-card class="pa-1 px-2" dark color="secondary darken-1" small v-bind="attrs" v-on="on">
-              <span v-if="!recipeYield"> x {{ scale }} </span>
-              <div v-else-if="!numberParsed && recipeYield">
-                <span v-if="numerator === 1"> {{ recipeYield }} </span>
-                <span v-else> {{ numerator }}x {{ scaledYield }} </span>
-              </div>
-              <span v-else> {{ scaledYield }} </span>
+              <v-icon small class="mr-1">{{ $globals.icons.edit }}</v-icon>
```
Kuchenpirat: Might be nice to gie this a bit more space. 

```suggestion
              <v-icon small class="mr-2">{{ $globals.icons.edit }}</v-icon>
```

Kuchenpirat (dismissed):
Yeah, i agree lets push the RecipeInfo redesign into a separate PR. 
There are probably a lot of things that can be done there (e.g. the two separate buttons for last made and "I made this", all time elements having the same icon ...)

For the main part of the PR this is all working as expected and i couldn't find a way to break it 😊🚀

There is one super small suggestion on the edit button. Feel free to merge if you think this is not necessairy. 

The next logical step for me to make this even more powerfull out of the box would be to parse the ingredient quantities on recipe creation. This would enable full scaling capabilities even for non parsed recipes 🤔

Kuchenpirat (approved):

michael-genson:
> parse the ingredient quantities on recipe creation

That's not a bad idea. While the parser isn't the most consistent, it's much better at the quantity portion

Kuchenpirat:
Yeah, i just thought, what we are doing here could work for ingredient quantities, which is really the only thing needed for scaling. 

The actual ingredient can stay a note. 
Plurals, aliases and all the other smart ingredient handling wont work, but scaling should work pretty nicely which is a function that is pretty essential to a recipe manager imho.

PR #4504, opened 2024-11-03
feat: adds descriptions to feature checks and add them to logs

cmintey:
## What type of PR is this?

- feature

Adds some additional logging around our application features on startup. This is similar to the app checks in the admin UI, but I think its nice to log them as well and the descriptions can be helpful in debugging. We could also show these descriptions in the app checks on the settings page, but we'd need to consider translations.

Also, instead of excluding secrets from the dump, this PR will instead mask the secrets with `"*****"` or if the secret is None, will show it as `null`. This again will help with debugging to show when values are set without logging secrets.

image

Not an issue, but a discussion: https://github.com/mealie-recipes/mealie/discussions/3334#discussioncomment-11116751
This doesn't directly fix it, but hopefully will help in debugging

### DIFF
--- a/mealie/app.py
+++ b/mealie/app.py
@@ -64,17 +64,10 @@ async def lifespan_fn(_: FastAPI) -> AsyncGenerator[None, None]:
         settings.model_dump_json(
             indent=4,
             exclude={
-                "LDAP_QUERY_PASSWORD",
-                "OPENAI_API_KEY",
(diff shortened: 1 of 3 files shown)


cmintey:
~~Hold on this, thinking of adding something else here~~
Implemented the changes with tests

michael-genson (approved):

PR #4517, opened 2024-11-06
feat: Improve Recipe Imports with Cleaner

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- feature
- bug

_(REQUIRED)_

We have a pretty robust recipe data cleaner for migrations, but we aren't actually using it for our scraping. This refactors the scraping/cleaning process a bit to make sure the cleaner actually runs on all imported recipes.

This also fixes a bug with OpenAI scraping which would fail on the `ld+json` import. We now include `ld+json` data along with the HTML text.

_(REQUIRED)_

Fixes issue raised in https://github.com/mealie-recipes/mealie/discussions/4490

I feel like we used to run all recipes through the cleaner, I'm not sure if/when/how that changed. But it works now!

## Testing

Manually with a bunch of local data


mealie/services/scraper/recipe_scraper.py
```diff
@@ -41,9 +44,23 @@ async def scrape(self, url: str, html: str | None = None) -> tuple[Recipe, Scrap
         raw_html = html or await safe_scrape_html(url)
         for scraper_type in self.scrapers:
             scraper = scraper_type(url, self.translator, raw_html=raw_html)
-            result = await scraper.parse()
 
-            if result is not None:
-                return result
+            try:
+                result = await scraper.parse()
+            except Exception:
```
hay-kot, line 50: Is it worth catching the specific exception? If not maybe throw that exception in the log so we know what the Exception is.

mealie/services/scraper/recipe_scraper.py
```diff
@@ -41,9 +44,23 @@ async def scrape(self, url: str, html: str | None = None) -> tuple[Recipe, Scrap
...
+                self.logger.exception(f"Failed to scrape HTML with {scraper.__class__.__name__}")
+                result = None
+
+            if result is None or result[0] is None:
+                continue
+
+            recipe_result, extras = result
+            try:
+                recipe = cleaner.clean(recipe_result, self.translator)
+            except Exception:
```
hay-kot, line 60: Is it worth catching the specific exception? If not maybe throw that exception in the log so we know what the Exception is.

hay-kot (changes requested):

mealie/services/scraper/recipe_scraper.py
michael-genson: `logger.exception` logs the exception in addition to the log message, so the logs should show what it is

hay-kot: You don't have to pass it in or anything, it just magically knows???

hay-kot (approved):

mealie/services/scraper/recipe_scraper.py
michael-genson: Python is magic

```python
from logging import getLogger

logger = getLogger()

try:
    assert False
except Exception:
    logger.exception("bad exception")
```

```
bad exception
  File "...", line 6, in <module>
    assert False
AssertionError
```

PR #4539, opened 2024-11-11
fix: Cookmode hide additional ingredients if all ingredients are linked

Kuchenpirat:
## What type of PR is this?

- bug

The headline for additional ingredients in the new cookmode was shown regardless of whether there were any ingredients or not. This fixes that

None

Also removed the remaining console logs that slipped under my radar. 

## Testing

Manual


codetakki:
Damn you where faster than me

michael-genson (approved):

PR #4544, opened 2024-11-12
fix: Enable OIDC with Synology SSO Server

### LINKED ISSUE #4543: [BUG] - OIDC with Synology not working
### First Check

I Configured the SSO Server on my Synology, but the Loging won't function.

Mealie requires the `name` claim from an OIDC Provider, but the Synology SSO Server doesn't Provide this claim. And to my knowlegde this isn't configurable.

Would it be possible to not require `name` as a claim from the OIDC Provider? On a first glance it doesn't seam required.

Maybe it would be possible to set Full Name equal to username if no `name` claim is present. For the `username` claim there is something similar configured: https://github.com/mealie-recipes/mealie/blob/622c1b11f5f84f73529e0c5567d34f11f9685c79/mealie/core/security/providers/openid_provider.py#L64C1-L77C38

### Steps to Reproduce

1) Configure Synology SSO Server
2) Try to Loging via SSO

[OIDC] Required claims not present. Expected: {'email', 'name'} Actual: dict_keys(['aud', 'auth_time', 'email', 'exp', 'iat', 'iss', 'nonce', 'sub', 'username'])

### Mealie Version

2.2.0

### Deployment

Docker (Synology)

_No response_

JoTec2002:
## What type of PR is this?

_(REQUIRED)_

- feature
- bug

Mealie required the `name` claim from an OIDC Provider, but the Synology SSO Server doesn't Provide this claim.

After this Pull Request Mealie will no longer required the `name` claim to be present.

To not cause any issues or lose any OIDC user information the following logic is used:
If the `name` claim is present it will be used, otherwise the username will be used as name.

Fixes #4543


MuadDibVV:
+1

cmintey:
I think it would be better and more extensible to introduce a new env var `OIDC_NAME_CLAIM`, which would default to `name`, but you could override it to whatever you want.

michael-genson:
@JoTec2002 if you could take a look at the feedback from @cmintey we should be able to get this merged!

JoTec2002:
@cmintey I completed the changes you suggested.

mealie/core/security/providers/openid_provider.py
```diff
@@ -96,7 +98,7 @@ def authenticate(self) -> tuple[str, timedelta] | None:
     def required_claims(self):
         settings = get_app_settings()
 
-        claims = {"name", "email", settings.OIDC_USER_CLAIM}
+        claims = {"email", settings.OIDC_USER_CLAIM}
```
cmintey: Can you add the new env var to this set, since it's required?

cmintey (changes requested):
Looking good! One more small thing to address

mealie/core/security/providers/openid_provider.py
JoTec2002: Done

cmintey (approved):
Thanks for working on this!

michael-genson (approved):

PR #4554, opened 2024-11-14
fix: Wakelock Toggle Broken

### LINKED ISSUE #4553: Keep Screen Awake Toggle is Unresponsive

**Information** 

The Keep Screen Awake  Toggle is designed to prevent the device screen from going to sleep. However while attempting to switch the toggle it does not move from its current position, resulting in an inability to enable or disable the feature as needed.

**Steps to reproduce:**

1. Navigate to any recipe you have saved and clicked on it.
2. Locate the button at the bottom of the recipe.
3. Attempt to switch the toggle from right position to left position.
4. Observe the behavior.

**Expected Behavior**

- The toggle should change positions when tapped. 
- When toggled to the left, the feature should be disabled, allowing the screen to sleep
- When toggled to the right, the feature should be enabled, preventing the screen to sleep after the designated timeout.

**Actual Behavior**

- The toggle does not change position when tapped i.e. it stays on the right position.
- As a result, the keep screen is enabled all the time.

**Mealie Version: ** 2.2.0

first_issue

**Possible Issues**
This may be due to the toggle control not updating its state in the UI or an issue in the binding with the underlying logic.

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- bug

_(REQUIRED)_

Our computed wakelock state doesn't use the `isActive` correctly. This fixes that.

_(REQUIRED)_

Fixes https://github.com/mealie-recipes/mealie/issues/4553

### DIFF
--- a/frontend/components/global/WakelockSwitch.vue
+++ b/frontend/components/global/WakelockSwitch.vue
@@ -16,7 +16,7 @@ export default defineComponent({
     setup() {
         const { isSupported: wakeIsSupported, isActive, request, release } = useWakeLock();
         const wakeLock = computed({
-            get: () => isActive,
+            get: () => isActive.value,
             set: () => {
                 if (isActive.value) {
                     unlockScreen();
--- a/frontend/components/global/WakelockSwitch.vue
+++ b/frontend/components/global/WakelockSwitch.vue
@@ -27,13 +27,13 @@ export default defineComponent({
         });
         async function lockScreen() {
             if (wakeIsSupported) {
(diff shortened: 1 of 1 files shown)


amishabagri (commented):
Hi,
Thanks for fixing the issue this fast. I tested it out on my setup and the toggle button is now working as expected.

michael-genson:
Thanks! I'm honestly surprised it went this long without being noticed and/or reported. I'm not sure when it broke

Kuchenpirat (approved):

PR #4616, opened 2024-11-27
feat: Move "on hand" and "last made" to household

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- feature
- bug

_(REQUIRED)_

This feature migrates food on-hand and tool on-hand to be configured per-household, as well as the recipe's "last made" property. The user experience is mostly the same, 

~~(this plays nicely with the [cocktail builder PR](https://github.com/mealie-recipes/mealie/pull/4542), although it may generate some merge conflicts)~~ fixed now that it's merged.

While working on this PR I also uncovered/fixed a few existing issues:
- Currently the `orderBy` param must be in snake case, while filters can be in either snake case or camel case. This PR fixes this by allowing camel case in the order by
- When sorting by rating, we already correctly sort by the _user's_ rating (rather than the global rating), however we didn't account for this in filters (i.e. if you included rating in a query filter, it would only use the global rating). This is fixed by this PR (since the same issue occurred using last made in a filter).

_(REQUIRED)_

N/A

~~If this is merged before the cocktail builder, I will update that PR to use the new on hand properties. If the cocktail is merged before this, I will fix it in this PR~~ fixed.

## Testing

Pulled up two browsers with users from different households, as well as a bunch of backend tests.


michael-genson:
Merged this with main now that the cocktail builder is in and updated the logic as needed

frontend/pages/household/mealplan/planner/edit.vue
```diff
@@ -22,7 +22,6 @@
         }
         resetDialog();
       "
-      :submit-disabled="isCreateDisabled"
```
michael-genson: This was duplicated above somehow

Kuchenpirat (dismissed):
I think this is good to go, was more straight foreward than i expected an 2k line PR to be 😊

Kuchenpirat:
welp should have merged the ruff stuff afterwards. 
There are a few conflicts now.

michael-genson:
All good, it probably would've messed the ruff PR up anyway

Kuchenpirat (approved):

PR #4703, opened 2024-12-10
fix: Add Formatting for Markdown Tables

michael-genson:
## What type of PR is this?

_(REQUIRED)_

- bug

_(REQUIRED)_

We currently support markdown in most places (recipe ingredients, steps, description, notes, etc.). Technically this includes markdown tables, but they're basically useless in their current form due to formatting:

```markdown

| First Header  | Second Header |
| ------------- | ------------- |
| Content Cell  | Content Cell  |
| Content Cell  | Content Cell  |
```
image

This PR adds some formatting to make them look like how you'd expect:
image
image

And in other places:
image
image

An example from GH:
image

This is only applied to the `SafeMarkdown` component, so if we use tables elsewhere (e.g. nutrition on the print view) the styling won't be affected.

_(REQUIRED)_

Closes https://github.com/mealie-recipes/mealie/discussions/4702

## Testing

Tested using the examples above

### DIFF
--- a/frontend/components/global/SafeMarkdown.vue
+++ b/frontend/components/global/SafeMarkdown.vue
@@ -48,3 +48,20 @@ export default defineComponent({
   },
 });
 </script>
+
+<style scoped>
+::v-deep table {
+  border-collapse: collapse;
+  width: 100%;
+}
+
+::v-deep th, ::v-deep td {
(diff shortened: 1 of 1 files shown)


Kuchenpirat (approved):
Looks Good 👍

PR #4800, opened 2024-12-30
feat: Add recipe as ingredient

parumpum:
## What type of PR is this?

- feature

Adds the ability to reference another recipe in a recipe's ingredients. A reference to the recipe is added to a new column for table recipes_ingredients. This can be utilized to determine if the ingredient is a food, a recipe, or just a note.

Adding recipe to shopping list will loop through the ingredients list, cascading down into recipes if it finds any to add individual ingredients to the shopping list.

image

image

Clicking the referenced recipe will open a new tab to the recipe

Linked recipes will expand into ingredients to add to shopping lists.

https://github.com/mealie-recipes/mealie/discussions/1761

## Testing

Tested by linking recipes together, adding ingredients to shopping lists
Unit test that adds a recipe to another recipe as an ingredient
Modified referenced recipe name and ingredients, verified changes were picked up on main recipe page


parumpum:
Addressed feedback in Discord:

- Added select all / none to Shopping List sections
- Added selectable list of child recipes on "I Made This" dialog
- Images shown when searching for child recipes to help pick the correct item

felixschndr:
Thanks for the great work! Looking forwards to this being merged

felixschndr:
@parumpum would you mind fixing the merge conflicts so the MR can be reviewed (@hay-kot)?

Kuchenpirat:
we won't merge anything with frontend code before the nuxt 3 upgrade has been done (which will bring new conflicts)

felixschndr:
I see, thanks for the clarification (FYI for anyone looking for that PR: https://github.com/mealie-recipes/mealie/pull/5184)

jawilson:
Please don't close this PR! I really want this ability.

Kuchenpirat:
No worries we are not closing this.
Maintainers are currently just pretty busy with stuff outside of mealie as well as the nuxt 3 upgrade.

tecosaur:
As one of the people eagerly awaiting this, now that v3 is out, how's this PR looking?

parumpum:
Once I find some free time to update it I’ll get it going!

michael-genson:

tecosaur:
Friendly poke :point_right: @parumpum :slightly_smiling_face:

parumpum:
Just got my dev environment set back up to work on this. Definitely some broken pieces to work on with all the Nuxt 3 changes that went in!

1112luke:
I would love this!! Any information on the timeline for this to be merged?

Coryrichter94:
This is a great feature! Can not wait

mealie/schema/recipe/recipe_ingredient.py
```diff
@@ -155,15 +156,49 @@ class RecipeIngredientBase(MealieModel):
...
+            self.is_recipe = bool(self.referenced_recipe)
+
+        return self
+
+    @model_validator(mode="after")
+    def change_note_templates(self):
+        if self.referenced_recipe:
+            self.note = self.referenced_recipe.name
+
+        return self
```
michael-genson: Is this desired behavior? What if we want to add a note to the recipe and have it render? e.g. "Make this thicker since we're using it as a base" or "slice extra thin".

michael-genson (changes requested):
Looks good so far! I made a few cleanup changes, including removing the `isRecipe` flag. We used to have a bunch of flags on the backend for things like this and it got super messy so we removed them. Similarly to how we handled those, I adjusted the frontend logic to accommodate.

For adding a recipe as an ingredient, instead of adding a dropdown to the "Add" button, would it be better to add it to the existing context menu? This would have the advantage of using an existing ingredient as a recipe instead of adding a new one. Something like "toggle recipe"
image

Another question: when we set a recipe to qty 0, that's signaling to the frontend that we should just not render the quantity (good for things like "teriyaki sauce" which might not have a defined amount). If we add that to the shopping list with 0 qty, will that work, or will nothing for that recipe be added? Might want an edge case for when the qty is 0 to use a shopping list qty of 1.

One more note below as well. I haven't fully finished reviewing this but these were my first impressions :) Love how it looks so far!

parumpum:
> Looks good so far! I made a few cleanup changes, including removing the `isRecipe` flag. We used to have a bunch of flags on the backend for things like this and it got super messy so we removed them. Similarly to how we handled those, I adjusted the frontend logic to accommodate.
> 
> For adding a recipe as an ingredient, instead of adding a dropdown to the "Add" button, would it be better to add it to the existing context menu? This would have the advantage of using an existing ingredient as a recipe instead of adding a new one. Something like "toggle recipe" image
> 
> Another question: when we set a recipe to qty 0, that's signaling to the frontend that we should just not render the quantity (good for things like "teriyaki sauce" which might not have a defined amount). If we add that to the shopping list with 0 qty, will that work, or will nothing for that recipe be added? Might want an edge case for when the qty is 0 to use a shopping list qty of 1.
> 
> One more note below as well. I haven't fully finished reviewing this but these were my first impressions :) Love how it looks so far!

Setting qty to 0 seems to work as expected. It doesn't show the quantity on the ingredient list, and still allows adding to shopping list as expected.

For the Add button, I feel like expanding the Add Food button is more intuitive than moving this function entirely to the Toggle Section/Insert menu. It does make sense to add a Toggle Recipe option there, so I'll work on that soon.

frontend/components/Domain/Recipe/RecipePrintView.vue
```diff
@@ -262,32 +262,60 @@ const ingredientSections = computed<IngredientSection[]>(() => {
...
+              ingredients: [ingredient],
+            });
+          }
+          else {
+            sections[sections.length - 1].ingredients.push(ingredient);
+          }
+        }
+      }
+    });
+  };
```
michael-genson, line 309: When printing the linked recipes, if the linked recipe appears in the middle of a section, the section titles become mangled. For instance:
```markdown
Recipe
 - [Section A]
    - ing 1
    - ing 2
    - linked recipe
    - ing 3
```

Becomes:
```markdown
[Section A]
- ing 1
- ing 2
[Linked Recipe]
- linked recipe ing 1
- ...
- ing 3 (from the recipe, not from the linked recipe)
```

Maybe we just keep it simple and remove this functionality from printing?

michael-genson (changes requested):
Pushed a few more changes/fixes. Just two more things I ran into that I didn't get time to address:
- One regarding printing, see below
- When saving a recursive ingredient, the backend appropriately handles this case and returns an error, but the frontend still looks like it updated successfully. Can we catch these errors on the frontend so the user doesn't think it worked? (we probably don't want to leave edit mode)

frontend/components/Domain/Recipe/RecipePrintView.vue
parumpum: I will remove this functionality. It can be returned with a rework of Sections that better defines which section each ingredient is part of instead of depending on a single ingredient with the "title" property set to define a Section.

Appreciate your assistance getting this through to the finish line.

michael-genson (approved):
LGTM, thanks for sticking with this one! I know it took a while but we got it in there 😅

ambroisie:
It's unclear to me how to handle linked recipes amounts. Or rather, I think I understand the intended way to deal with it but find it awkward to read.

Here are a couple of examples:

* "Salted caramel macarons": depends on "Macaron shells". The (unscaled) recipe yield 50 top/bottoms. The salted caramel recipes is scaled for 50 top/bottoms. If I link "1 Macaron shells" that is the correct amount of that recipe to make, but reads weirdly on the page. If I link "50 macaron shells" it links correctly but doesn't reflect the fact that I should 50x the shell recipe.
* "Pain aux raisins": depends on both "Croissant dough" and "Pastry cream". The scales should be 0.5x dough (500 grams from a 1kg yield recipe) and 0.25x cream (200 grams from an 800 gram yield recipe). Once again it reads weirdly to only have "0.5 Croissant dough" on the recipe page, instead of "500 gram croissant dough".

parumpum:
> It's unclear to me how to handle linked recipes amounts. Or rather, I think I understand the intended way to deal with it but find it awkward to read.
> 
> Here are a couple of examples:
> 
>     * "Salted caramel macarons": depends on "Macaron shells". The (unscaled) recipe yield 50 top/bottoms. The salted caramel recipes is scaled for 50 top/bottoms. If I link "1 Macaron shells" that is the correct amount of that recipe to make, but reads weirdly on the page. If I link "50 macaron shells" it links correctly but doesn't reflect the fact that I should 50x the shell recipe.
> 
>     * "Pain aux raisins": depends on both "Croissant dough" and "Pastry cream". The scales should be 0.5x dough (500 grams from a 1kg yield recipe) and 0.25x cream (200 grams from an 800 gram yield recipe). Once again it reads weirdly to only have "0.5 Croissant dough" on the recipe page, instead of "500 gram croissant dough".

I think this is more of a servings vs yield issue. Showing "0.5 Croissant dough" on your Pain Aux Raisins recipe indicates that you would scale your croissant dough recipe to 0.5 and make that. If you utilize the shopping list, the ingredients will be scaled to this value.

Likewise for the macarons, you could create or change the recipe to have amounts for 1 macaron shell so that it makes sense to enter 50 macaron shells on the salted caramel macaron recipe. You would be able to set the scale of the macaron shell recipe to the number of shells you ultimately want.

Linking a recipe is simply the quantity of that recipe to make. The current idea behind the feature is that you would be able to go into the main recipe you are ultimately going to make and click on the linked recipe to open a new tab to make that item in the required quantities. One enhancement I could see here would be to pull up the recipe scaled as indicated in the main recipe to save a manual adjustment there.

ambroisie:
Yeah I understand, I just find it awkward.

I wish the linked recipe would surface the yield amount and be able to scale that instead.

xF4m3:
Would it be possible to show the ingredients of the linked recipie aswell? Sortes under the linked recipie title or something? I think this would make working with it easier as you do not necessarly have to open the linked recipie.

Also just wanted to say thank you guys for getting this feature implemented in the First place, have been following and waiting for this to happen all this time. Great work, really appreciate this.

ambroisie:
> Sorted under the linked recipie title or something

In a foldable menu?

xF4m3:
Doesn't have to be foldabale but that could be one way to do that. In my mind i was just thinking about an expanded static list that is maybe writtin in a smaller or grayer text so it doesn't draw the main attention, but i am probably fine with whatever idea like being collapsible like you mentioned.

MatzeG1t:
> Would it be possible to show the ingredients of the linked recipie aswell? Sortes under the linked recipie title or something? I think this would make working with it easier as you do not necessarly have to open the linked recipie.
> 
> Also just wanted to say thank you guys for getting this feature implemented in the First place, have been following and waiting for this to happen all this time. Great work, really appreciate this.

I would suggest to think about the usecase of this. Seeing all ingredients might sound helpful at first, but whats the goal for the user? If its "adding also the ingredient/recipe to the shopping list" a confirmation would make sense at another spot.
Just seeing the ingredients might be only helpful if its a simple ingredient-recipe. If I think about "turkey" or something, the sauce and side dishes are complex as well - only seeing the ingredients in the "turkey recipe" wont help I'd say.

Did the dev team ever thing about adding a feedback function in the tool? This way maybe more user feedback could be sourced. Not just by the mealie-hosts (who check release logs) but rather also the cooks using it - if they are not the same person :)

PR #4809, opened 2025-01-01
fix: Move sort direction to new button

### LINKED ISSUE #4806: [BUG] - Sort Order (Ascending vs Descending) for Categories (Last Made, Alphabetical) Feels Counterintuitive
### First Check

When viewing recipes, retaining the default sort of 'Ascending' or 'Descending' is counterintuitive when switching between some of the options.  Though not necessarily a 'bug', it feels odd to have it done this way. 

For example, the 'Last Made' being in Descending Order shows the latest recipes I have made recently. (Is there any way to show the DATE here in this screen? This is more a Feature Request)

image

This works fine for Last Made, switching to 'Created' to show me the latest one I have made as well. Though it is properly sorted in Descending, it also feels off that the dropdown says "Sort Ascending" making me feel its actually sorted by Ascending, but its not, thats the action I can take. Only the Header Icon is indicating the up or down arrow to visualize the sort order. 

However, when then switching to something like 'Alphabetical' it shows from Z-A.

image

When in these views, I would typically want A-Z as default. If you then change the sort order, it affects the 'Last Made' where you would have to change it back (obviously depending on preferences of how you want to see the data).

Some ways I can think of to resolve this: 

Could 1. Each filter option (Last Made, Alphabetical, etc) retain its own Sort (Ascending vs Descending) 

AND/OR:

Breakout buttons for this so you have two filters instead of one (One for the Category, then one for sort order). This could also help to visualize by simply being an UP arrow or DOWN arrow as when you have that as its own separate button, you could enable it to be a click to change sort order instead of having to click the button, then press 'Sort Ascending' / 'Sort Descending' 

### Steps to Reproduce

1. Go to recipes
2. Sort by 'Last Made'
3. This will default to show 'Descending' which is typically what I would expect to see and want
4. Switch to Alphabetical
5. It is sorted by Z-A instead of A-Z

n/a

### Mealie Version

[23a85f51d20b12ee93f7b25d312d70d18be0e92d](https://github.com/mealie-recipes/mealie/commit/23a85f51d20b12ee93f7b25d312d70d18be0e92d)

### Deployment

Docker (Linux)

_No response_

parumpum:
## What type of PR is this?

_(REQUIRED)_

- bug

On Recipe Explore page, moved the Sort Ascending/Descending to its own button outside of the sort options menu. This allows you to clearly see at a glance how items are sorted.

Before:
image

After:
image

Fixes #4806


Kuchenpirat:
Hey, thanks for your work on this, but i actually don't think this is needed. 
Adding another button to this already quite complex menu that users are interacting with a lot is actually counter productive to keeping the usability high and having everything sorting related in one button group standard on menus that allow for additional filters.

PR #4842, opened 2025-01-06
feat: add delete option to three dots menu

johnpc:
## What type of PR is this?

- feature

This PR adds the "Delete" option to the three dots menu. Excluding delete from the three dots menu has caused confusion as seen in https://github.com/mealie-recipes/mealie/discussions/3984#discussioncomment-11748286

Before: 

Screenshot 2025-01-06 at 12 40 27 PM

After: 

Screenshot 2025-01-06 at 12 33 26 PM

No issue, but feature request is here: https://github.com/mealie-recipes/mealie/discussions/3984#discussioncomment-11748286

This may have been omitted intentionally from the three dots menu, considering all the plumbing was already in place and I just had to flip the value to `true`. Was this omitted for a good reason to where we actually don't want to allow this?

## Testing

I tested manually, but I'm not sure where test coverage should be added. Please advise.


simgunz:
Great to see this feature implemented. A minor suggestion: I believe that the best practice in terms of UX is to place the 'Delete' action as the last item in the menu.

Kuchenpirat (commented):
While i agree with  the general sentiment that the delete option might be a bit too hidden i don't think we need to include it on the RecipeExplore page. 
The implementation on this page also comes with the problem that the state of the recipe page does not reflect that the recipe has been deleted. 

Where i can see the delete button making sense is in the RecipeActionMenu, no longer hiding it behind the edit button. 
As @simgunz suggested i would move it into last position (above the divider) and i would keep it grey instead of red (so that it isnt the first thing your eyes get drawn towards when opening the menu).
Here we also need add functionality to navigate away from the page and reload the recipes as the current delete button does as well.

johnpc:
@Kuchenpirat I've made the update as you've suggested. 

* Only added the delete action to the three dots in the detail view (not list view)
* Added handling so that that the delete action navigates back to the list view once deleted
* Changed the delete color from red to gray

Screenshot 2025-01-14 at 7 22 24 PM

ChipSkylark37:
Does this still prompt the normal "Are you sure" confirmation on this? 

With it being close to the 'Share' button, I know I would accidentally do this _at least_ once.

michael-genson:
Might also be worth putting a divider in-between share and delete to visually separate it (similar to data management)

image

johnpc:
@ChipSkylark37  Yes, the confirmation prompt is still there. Here is a screen recording


https://github.com/user-attachments/assets/850d747e-ed82-42c6-9088-d7ddceba95a4

johnpc:
@michael-genson I'm not a designer but in my mind, the sections in the example you shared work well because those dividers actually divide logical sections. In this example it would just be to create space away from the delete action, which doesn't feel like the same thing. I think the confirmation dialog is enough to prevent disaster in the case of accidental clicks.

michael-genson:
Yup, that makes sense! Adding a divider probably doesn't do much

I was thinking of some apps where "dangerous" operations are visually separate from everything else, but other than delete we don't really have anything like that

simgunz:
> While i agree with the general sentiment that the delete option might be a bit too hidden i don't think we need to include it on the RecipeExplore page.

Is the RecipeExplore page the one that lists all the recipes (the one in the picture below)?

Screenshot_20250115_082027

I believe that the delete button should _also_ be present in that three-dots menu. From my user experience, it is a common UX pattern that of being able to delete an item from a list view (see the Gmail UI for example where there are quick actions to delete an email).

Screenshot_20250115_082409

Kuchenpirat:
I don't think this is an equal situation as in my experience I am way more likely to delete a mail than a recipe I or someone in my household added with some intend.
But I'll think about it 👍

simgunz:
You convinced me. Deleting a recipe is indeed a seldom action. Adding the
action in the detail view menu is already good enough.

johnpc:
I think it's fine to only have it on the detail page personally.

When I was first trying to delete a recipe, the recipe detail page three dots menu (the one in this PR in current form) was the first place I checked. When it wasn't there I also checked the list view page three dots menu. 

When it wasn't there I got very confused, poked around a while longer to no avail, and eventually googled it and found https://github.com/mealie-recipes/mealie/discussions/3984#discussioncomment-11748286

Then a few weeks later, I ran into the _same situation again_! I'd forgotten how to delete a recipe, and had to google it again to re-find that discussion thread explaining how! That's when I was like okay I'll just cut a PR.

ChipSkylark37:
> I think it's fine to only have it on the detail page personally.
> 
> When I was first trying to delete a recipe, the recipe detail page three dots menu (the one in this PR in current form) was the first place I checked. When it wasn't there I also checked the list view page three dots menu.
> 
> When it wasn't there I got very confused, poked around a while longer to no avail, and eventually googled it and found [#3984 (comment)](https://github.com/mealie-recipes/mealie/discussions/3984#discussioncomment-11748286)
> 
> Then a few weeks later, I ran into the _same situation again_! I'd forgotten how to delete a recipe, and had to google it again to re-find that discussion thread explaining how! That's when I was like okay I'll just cut a PR.

I like the option into the three dots menu (without going into 'Edit').

I also agree it should only be on the Individual Recipe Page itself, not from the page that shows all Recipes.

Kuchenpirat (approved):
Thanks for helping out with the project 😊👍

LGTM 🚀

PR #4853, opened 2025-01-07
feat: add make_admin script

Kuchenpirat:
## What type of PR is this?

- feature

In the discord we had multiple people loose access to the last admin account. This adds a small script to promote a user to admin. This prevents users from needing to interact directly with their db which has greater potential of breaking stuff. 

None, but the problem came up in Discord quite frequently. 

## Testing

Manual


michael-genson (approved):
Should probably add this to the Discord bot too

PR #4891, opened 2025-01-13
docs: move faq into details / spoilers

Kuchenpirat:

This changes the layout of the faq section in the docs. 
We are adding more and more faqs which is good but the usability of the page is decreasing because of its length. This moves everything into expandable spoilers that contain just the question when not expanded. 

None, i am happy to discuss if it makes sense to do this. 

- Makes the page look quite repetitive, but i think it is easier to find the relevant questions. 
- we could also remove the toc to reduce redundant data. I kept it there to be able to get a link to a specific question (which is valuable at least to me). Happy to be overruled here as well. This is also the reason i kept the additional titles within the spoiler. 

## Testing

Manual.


ChipSkylark37:
I like this idea and was thinking it while working on some of the documentation that it was getting too long to scroll effectively. Even by use of spoiler tags, I think the organization could be tidied up with sections or sub-pages, or merging into the Features section.

What if we had different sub-pages for Features / FAQ? So those exist at a top level, and links to sub-pages that relate to the item in question (that could also have the spoiler tag).

Instead of just 'Here's the 2 steps to enable and turn on this feature' under the Features section is "Here are all the advanced features / usage of a Cookbook and examples" That blurs the line a bit more between documenting it under Features vs 'FAQs'. 

If something is so frequently asked, then why isn't it just in the normal 'Features' section? That doesn't apply to _everything_ but definitely some of whats being asked commonly should, I think, be in the normal features we want people to know about. 

I would love to document and detail some example scenarios (with pictures, etc) but that would make it unruly, which I think the spoiler tag is good for. But I think before that, rethinking where the data should go and if it should still be on a single page (Features / FAQ) makes sense.

ChipSkylark37:
A general "how to" section may also make sense as a breakout?

So we have:

Features: a list and short description of each

How-To: more in-depth about how to do certain things

FAQs: Questions people ask that don't necessarily go over features but could be things like "hey I'm having this weird issue with my setup that I installed onto Windows 10" and there's information that only the Docker method is supported, etc.

Kuchenpirat:
Yeah it's a first step of making that page more palatable.

michael-genson (dismissed):
Yeah I'm not big on how repetitive it looks but either way it's substantially better than how it is now so I don't really care

> I kept [the toc] there to be able to get a link to a specific question

Absolutely, I use that all the time. Plus clicking on it auto expands the collapsed section

Approved, will let you merge in case you want to change anything

ChipSkylark37:
At least from an organization standpoint it makes sense to me to break things out to something like this:

Pre-Install (Section / Header)
> General Install Info

> Install Methods
>> SQLlite
>> Postgres

Post-Install
> First Time Setup
>> Restore from Backup
>> Guided Walkthrough / Tutorial for common things to do

> Features

> How To
>> How to perform specific functions as it relates to feature 1
>> More In-Depth - Exploring Features, etc

> FAQS
>> Do I need an email Server?
>> How can I access Mealie outside of my house?

> Community / Outside Guides 
>> Things like third party website / videos for other NON-SUPPORTED / MAINTAINED articles?

Kuchenpirat:
Divided the docs up into to following general topics. I used h2 for the titles and moved the question title to h3. This will add bit of clutter to the toc. We could also just use large text if it is prefered. 

- Features and Functionality
  - How do I enable 'smart' ingredient handling?
  - How do I enable Nutritional Values?
  - What is fuzzy search and how do I use it?
  - How can I attach an image or video to a Recipe?
- Customization and Configuration
  - How can I change the theme?
  - How can I change the login session timeout?
  - Can I serve Mealie on a subpath?
  - How can I install Mealie without Docker?
- Account Management
  - How can I unlock my account?
  - How can I reset admin privileges for my account?
  - How can I change my password?
  - I can't log in with external auth. How can I change my authentication method?
- Collaboration and Privacy
  - How do private groups, households, and recipes work?
- Security and Maintenance
  - Can I use fail2ban with Mealie?
  - Is it safe to upgrade Mealie?
- Technical Considerations
  - Why an API?
  - Why a database?
- General Questions
  - Why is there no 'Keep Screen Alive' button when I access a recipe?

michael-genson (approved):
I like this _much_ better!

ChipSkylark37:
I see its now live - Awesome. I'll be reviewing to add in some things (like the most recently asked question of linking ingredients to steps). 

Nice work @Kuchenpirat !

PR #4907, opened 2025-01-17
feat: Show the current week instead of today + 7 days in meal planner

simon-zumbrunnen:

The meal planner currently shows the next 7 days. With this pull request I propose to change this behaviour so the default is a weekly overview similar to how calendars usually work. The default setting for the time range is changed so it starts at the most recent "start of the week" according to the settings.

If this pull request is accepted I plan to implement buttons to get to the previous / next week.

This does not directly fix any open issue but it is a step towards weekly shopping lists discussed in #1899.


boc-the-git (changes requested):
The concept sounds like it would be a significant regression on my own UX. So _if_ this is to become the default, I think you need to implement a lot of configurability.
It's uncommon to need to look at the history in the meal planner, and current functionality allows a user to set how many future days to include (I have it set to 12, FWIW).

DenuxPlays:
Just a small Note:

Wouldnt it be possible to let the user define their own schedule?

So in my case I plan from thursday to thursday.

Kuchenpirat:
The most flexible solution would probably to have a checkbox that can be enabled disabled that toggles a start day. 
This should enable a dropdown menu where the user can choose the day the week should start with. This would offer flexibility for half of the world where the start of the week is not monday.

simon-zumbrunnen:
@boc-the-git It is also uncommon to need to look at what meetings you had yesterday. But still every calendar app shows you the whole week. But yes, usually you have the option to see a single day, a week or a month. But yes, an option to choose whether to start at today or at the start of the week could make sense? Would this option make sense as a global setting or in this calendar dropdown?

@DenuxPlays This would already be possible by setting the start of the week to thursday. I don't know what other consequences this would bring though.

simon-zumbrunnen:
@Kuchenpirat Somehow your comment was not yet there when I wrote my text. There is already the option to configure what day is the start of the week. In my code I already take this into account (household.value?.preferences?.firstDayOfWeek).

Kuchenpirat:
Ah, yeah, I forgot about that option. 
Then it would be just a toggle to enable start with the first day of the week.

simon-zumbrunnen:
@boc-the-git What would you think about somthing like this:

Screenshot 2025-03-20 at 20 46 55

The default behaviour would stay the same, but instead of choosing "Today" you could also choose "Start of Week" or any other day of the week:

Screenshot 2025-03-20 at 20 47 11

michael-genson:

mecman92:
I don't personally have the expertise to contribute to this (but I hope to learn and contribute one day 😊). Also thank you so much to all that do contribute, am newish to mealie from Tandoor.

I want to share that this expanded meal plan UI would be enormously helpful!

I have to think that more than a minority of folks don't stick to the scheduled meal for the day. I like certainly many others make a meal plan for the week, shop for said meal plan, but then cook whatever meal during that week that sounds good on the day. Of course I could shuffle meals around from day to day but this is more effort to maintain. I am constantly adjusting the date in meal plan backwards to see recipes that are planned for the week but skipped because another recipe sounded better at the time. Not the end of the world obviously but naturally more flexibility for various ways to use meal plan would be a huge upgrade!

michael-genson:

Issue #4789, opened 2024-12-29
[BUG] - Cookbook using a Tag filter doesn't find all recipes with that tag


jwismar:
### First Check

- [X] This is not a feature request.
- [X] I added a very descriptive title to this issue (title field is above this).
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Mealie documentation, with the integrated search.
- [X] I already read the docs and didn't find an answer.
- [ ] This issue can be replicated on the demo site (https://demo.mealie.io/).

### What is the issue you are experiencing?

I have a couple hundred recipes in Mealie, including many cocktails. 34 of the the cocktails have the tag "Tiki" amongst other tags. I created a cookbook called "Tiki Cocktails" which should pull recipes that have "contains all of" the single tag "Tiki". Only 22 recipes get pulled into the cookbook.

I created a second cookbook called Tiki 2 using the same filter criteria, and it ends up containing the same 22 recipes.  I also attempted this using an "is one of" filter, and get the same results.

If I look at the main recipes page, and filter using tag "Tiki", all 34 recipes get displayed.

I was not able to replicate this behavior on the demo site, but my test recipes there are obviously much simpler and don't have multiple Tags amongst other confounding factors.

### Steps to Reproduce

Create a new cookbook. Select filter criteria for Tag called "Tiki". Save the cookbook. View the cookbook. Observe that only a subset of expected recipes are included.

### Please provide relevant logs

Docker log contains no error messages, just access logs. Relevant request looks like:
INFO     2024-12-29T12:42:24 - [10.111.1.5:54334] 200 OK "GET /api/recipes?page=1&perPage=64&orderBy=name&orderDirection=asc&cookbook=tiki-cocktails HTTP/1.1"

### Mealie Version

Docker image: ghcr.io/mealie-recipes/mealie:nightly
Build: [23a85f51d20b12ee93f7b25d312d70d18be0e92d](https://github.com/mealie-recipes/mealie/commit/23a85f51d20b12ee93f7b25d312d70d18be0e92d)

### Deployment

Unraid

fakethinkpad85:
Not sure but same bug might apply to Categories as well, having the same issue using categories. Guessing they use the same filter logic.

![image](https://github.com/user-attachments/assets/0f28484c-2a7f-4671-9bed-01a0252da970)

PancakeZik:
+1

ChipSkylark37:
I wonder how this could be replicated.

[...]

PR #4908, opened 2025-01-17
fix: Show All Recipes in Cookbook Regardless of Sort

michael-genson:

_(REQUIRED)_

Removes old code that filtered out nulls when using the old recipe sort. Also force a refresh if a cookbook is edited (since recipe criteria will have changed).

_(REQUIRED)_

Fixes part of the issue raised in https://github.com/mealie-recipes/mealie/issues/4789#issuecomment-2599012513

Not sure if all issues related to https://github.com/mealie-recipes/mealie/issues/4789#issuecomment-2599012513 are resolved, but at least half of them are.

## Testing

Manually

### DIFF
--- a/frontend/components/Domain/Cookbook/CookbookPage.vue
+++ b/frontend/components/Domain/Cookbook/CookbookPage.vue
@@ -104,9 +104,12 @@
         }
         const response = await actions.updateOne(editTarget.value);
 
-        // if name changed, redirect to new slug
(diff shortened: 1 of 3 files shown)


michael-genson:
I decided to leave that debug log in since it could be helpful for diagnosing other recipe query issues

Kuchenpirat (approved):

PR #4676, opened 2024-12-03
feat: Option to automatically create share links for recipes in recipe actions

ianp1:
## What type of PR is this?

- feature
- bug

## What this PR does / why we need it:

Currently, when using a recipe action on a recipe that is not publicly available (due to recipe, group or household being private), the receiving service will not be able to read the recipe data. 

As discussed in #4619, this PR implements an additional template option for recipe actions. The option `${share-link-expires-seconds-<x>}` is automatically replaced by a newly created share link that expires after `x` seconds. This makes share links usable even on private recipes, groups and households. 

## Which issue(s) this PR fixes:

Fixes  #4619

## Special notes for your reviewer:
I am not completely certain, if the method to obtain the group slug is correct. Please check this part thoroughly.
Also, i have not yet created documentation for the new option - is the website documentation part of this repository?

## Testing
I created a private recipe, a public recipe and multipe recipe actions

- One containing a single share link template
- One containing multiple share link templates with different expiry times
- And one containing multiple share link templates, some of which having the same expiry times

I also tested with different privacy settings in group and household. The receiver of the given action was always able to read the recipes, without being logged in using the share link.

PR #4920, opened 2025-01-20
docs: Add community guide for integrating the Mealie-Bring-API

felixschndr:

See https://github.com/mealie-recipes/mealie/discussions/3627#discussioncomment-11698342

I was not sure whether there is some crucial information missing. I din't want to add to many details as all the information is described in the repository the docs link to and I didn't want to add any unnecessary duplication. Do you think anything shall be added?

### DIFF
--- a/docs/docs/documentation/community-guide/bring-api.md
+++ b/docs/docs/documentation/community-guide/bring-api.md
@@ -1,0 +1,8 @@
+!!! info
+This guide was submitted by a community member. Find something wrong? Submit a PR to get it fixed!
+
+Mealie supports adding the ingredients of a recipe to your [Bring](https://www.getbring.com/) shopping list, as you can
+see [here](https://docs.mealie.io/documentation/getting-started/features/#recipe-actions).
+However, for this to work, your Mealie instance needs to be exposed to the open Internet so that the Bring servers can access its information. If you don't want your server to be publicly accessible for security reasons, you can use the [Mealie-Bring-API](https://github.com/felixschndr/mealie-bring-api) written by a community member.
+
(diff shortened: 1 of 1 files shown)


michael-genson:
Placement looks good, can you elaborate more on the specific use-case isn't resolved by https://github.com/mealie-recipes/mealie/pull/4676? That PR will make it so Mealie doesn't have to be publicly accessible to use the Bring integration, but it _does_ need to be exposed to the internet (whereas, if I understand correctly, your integration allows Mealie to stay locked behind local network access only)

Otherwise I agree, no need to duplicate docs when you link to the project. Maybe a quick "check out the project for installation instructions" or something similar, just to clarify that you won't find them in the Mealie docs.

felixschndr:
> That PR will make it so Mealie doesn't have to be publicly accessible to use the Bring integration, but it does need to be exposed to the internet

Can you please explain what you mean by that? Does that mean Mealie has to be exposed to the internet and the recipe can be shared even though it normally would require a user login?

> whereas, if I understand correctly, your integration allows Mealie to stay locked behind local network access only

exactly, no service has to be exposed to the internet at all. I will add a comment to make this clearer in the docs (edit: done in https://github.com/mealie-recipes/mealie/pull/4920/commits/d369cab436a065248450961dc09ae8c0bf7eb75e)

michael-genson (approved):
Thank you!!

> Does that mean Mealie has to be exposed to the internet and the recipe can be shared even though it normally would require a user login?

Yup, that's right. Once that PR is merged Mealie will still need to be exposed to the internet, however it won't require a login for the Bring integration (and similar usecases). Your web server adds an extra layer of protection though since it doesn't require Mealie to have external access

Issue #4593, opened 2024-11-22
[SECURITY] 4 Broken Access Control Vulnerabilities


m10x:
A few minutes ago I've emailed you the details to one high risk broken access control vulnerability and will now write you the details to three more (but low to lower medium risk) broken access control vulnerabilities.

m10x:
I've sent an email for each of the 4 vulns :)

hay-kot:
Thanks for the report! I've reviewed the info and passed it along to the other maintainers, we'll follow up once we've fixed that issue.

m10x:
Hi, any update? :)

michael-genson:
Hey there! These are still being worked on, thank you for following up!

Madj42:
I have to say, I'm a little more than concerned that these vulnerabilities haven't been fixed yet.  Is there any plan to have this in a release soon?

michael-genson:
These are actively being worked on. All four reported vulnerabilities require authenticated access, and there are no known vulnerabilities exploitable by unauthenticated users.

While we intend to resolve the reported issues, [please see our security documentation](https://docs.mealie.io/documentation/getting-started/installation/security/) which addresses best practices which cover the reported issues.

m10x:
I just tested the 3 fixes and can confirm that 3 of the 4 reported vulnerabilities are fixed :)

PR #4928, opened 2025-01-22
fix: prevent users from updating their own household privileges

Kuchenpirat:

Fixes 2 security vulnerabilites reported in #4593 
Those have both to do with an existing user escalating their household priviledges (canManage, advanced etc...)

adresses 2/4 vulnerabilities reported in #4593 

Not the most experienced in our backend, so there is probably a more elegant solution to this but this should address the problem 🤷‍♂️

## Testing

Added tests that cover the reported vulnerabilites.


michael-genson:
Looks good! I refactored a little and also removed the check from the delete endpoint because I don't think it does anything (you have to be an admin to reach that endpoint anyway).

Also, I don't think "advanced" is a permission, pretty sure users are supposed to be able to toggle that themselves, so I removed it. Let me know if I'm mistaken on that

tests/integration_tests/user_tests/test_user_crud.py
```diff
@@ -106,7 +106,7 @@ def test_admin_updates(api_client: TestClient, admin_user: TestUser, unique_user
     assert response.status_code == 403
 
     # admin updating their own permissions
-    permissions = ["canInvite", "canManage", "canManageHousehold", "canOrganize", "advanced", "admin"]
+    permissions = ["canInvite", "canManage", "canManageHousehold", "canOrganize", "admin"]
```
michael-genson, line 109: Oops, thanks!

Kuchenpirat:
thanks for the helping hand. Removed another advanced that was hidden in one of the tests

michael-genson (approved):

PR #4946, opened 2025-01-26
feat: Add recipe count to header of recipe card section

jannon:

This just adds the recipe count to the header of the recipe card section.  It provides a useful piece of information right at the top of the section so users can see how many total recipes or how many match the filters/search.

So it goes from 'Recipes' to 'Recipes (n)' where n is the number of recipes in the result.

This is a small change looking to enhance the usability/convenience of the UI.  It's a common UI paradigm to display the number of items in a collection somewhere near the top.

I just starting using mealie and found myself wanting this.  I thought this was a nice simple change to start to familiarize myself with the project.  I don't think it warranted any test or doc changes, but let me know if there's anything else I should do.

## Testing

Visual testing with the development UI:
- Upon start, displays 'Recipes (0)'
- After adding a recipe, displays 'Recipes (1)'
- After adding another, displays 'Recipes (2)'
- After searching for something specific to one of the recipes, displays 'Recipes (1)'


frontend/components/Domain/Recipe/RecipeCardSection.vue
```diff
@@ -5,7 +5,7 @@
         <v-icon v-if="title" large left>
           {{ displayTitleIcon }}
         </v-icon>
-        <v-toolbar-title class="headline"> {{ title }} </v-toolbar-title>
+        <v-toolbar-title class="headline"> {{ title }} ({{ recipes.length }})</v-toolbar-title>
```
michael-genson, line 8: Won't this only show the currently fetched recipes (i.e. _after_ pagination)? We return the true "total" with the pagination API, should we use that value instead?

michael-genson:
I like the idea, see my comment though, I'm not sure this will work as-is. Also, would you be able to include a screenshot of what it looks like?

frontend/components/Domain/Recipe/RecipeCardSection.vue
michael-genson: This is what the API returns: https://github.com/mealie-recipes/mealie/blob/93c2df41c3ea44288d21618cd81c656e9a8e1ffd/mealie/schema/response/pagination.py#L53-L60

jannon: Ah, I see. I hadn't encountered any pagination yet in my usage.  I'll have a look and take another pass

michael-genson: IIRC the frontend pulls two pages at once (with a page size of 32 that's 64 recipes) so you might want to add 100 recipes or so to your dev instance

michael-genson:

PR #4961, opened 2025-01-27
feat: Ingredients parser UX - Slightly improve controls

CloCkWeRX:

Rearranges the save/parse/etc controls on ingredient editing so that:
- Save/cancel is always on screen for long lists
- Parse all is shifted to its own "form"

After:
image

image

image

Part of https://github.com/mealie-recipes/mealie/discussions/4954

If there's a preferred style or background for floating action buttons/navs; happy to implement. Current approach overlays content in a weird way.

## Testing

Regression tested locally with chrome dev tools (desktop,, mobile).


frontend/pages/g/_groupSlug/r/_slug/ingredient-parser.vue
```diff
@@ -21,15 +21,16 @@
...
+          <BaseButton color="info" class="mb-4" :disabled="parserLoading" @click="fetchParsed">
+            <template #icon> {{ $globals.icons.foods }}</template>
+            {{ $tc("recipe.parser.parse-all") }}
+          </BaseButton>
         </div>
       </BaseCardSectionTitle>
 
-      <div class="d-flex mt-n3 mb-4 justify-end" style="gap: 5px">
+      <div class="d-flex mt-n3 mb-4 justify-end" style="position: sticky; top: 65px; z-index: 2;">
         <BaseButton cancel class="mr-auto" @click="$router.go(-1)"></BaseButton>
```
CloCkWeRX:
```suggestion
        <BaseButton cancel @click="$router.go(-1)"></BaseButton>
```

haileyrs:
Would you consider leaving the Save and Cancel buttons at the top and including a copy of those buttons at the end of the ingredients list? In my experience using the parser, I scroll to the end of the list every time to make sure that every ingredient is correct, and I think it's likely that a lot of people do the same. In that situation, floating the buttons isn't really necessary, and doing it this way would alleviate the issue of having a strange overlay.

michael-genson:
Hey there, it looks like this PR is no longer needed with the new ingredient parser. If you're running into the same issues with the new parser feel free to create a new PR or re-open this one with the changes for the new parser.

PR #4962, opened 2025-01-27
fix: Address Ingredients Parser UX overflow issues on mobile

CloCkWeRX:
### Before

Desktop was ok

image

But mobile cropped error messages if there was more than one or they were long
image

After:
image

image

https://github.com/mealie-recipes/mealie/discussions/4954

This has a slight tradeoff in that the existing call to action ("Create") now relies on the user understanding what the + icon means, unless they hover over and read the title attribute.

The use of `flex-wrap` also looks a bit stupid, but the alternative was a list group; which looked bad on desktop.

If the CSS framework has a btn-group-vertical on small screens class; I couldnt find it.

## Testing

Tested locally, w/Chrome dev tools for mobile UX.


CloCkWeRX:
Let me know if this should have a general "errors" or "Add missing" prefix when errors.any?

frontend/pages/g/_groupSlug/r/_slug/ingredient-parser.vue
```diff
@@ -97,7 +99,12 @@
     </v-container>
   </v-container>
 </template>
-
+<style lang="css">
+.v-card__actions {
+  flex-wrap: wrap;
+  overflow: truncate;
+}
+</style>
```
Kuchenpirat: Styles should be the last item below <script>

Kuchenpirat (changes requested):
I like the concept of this quite a bit but there are still a few things that need improving. 

- I feel like it should be visually possible to determin if the parser wants to create a new food or unit (either swaping the plus icon in the button for the food | icon button or appending an icon that matches the type)
- We need to add vertical margin so the cards are no longer right against each other. 
- The cards are not all aligned to the right side if they stack which looks quite ugly (screenshot)

--- 

image

frontend/pages/g/_groupSlug/r/_slug/ingredient-parser.vue
CloCkWeRX: Moved, though I could swear vscode rearranged things on me at some point - perhaps a different branch.

Kuchenpirat: It should automatically move it to the bottom 🧐
Thats at least what it has done when i opened your PR.

Kuchenpirat:
Not stale, just dependent on the nuxt 3 pr

michael-genson:

Same with your other PR for the ingredient parser: https://github.com/mealie-recipes/mealie/pull/4961

We're considering re-doing the parsing flow which may eliminate this page entirely, but I'm not sure how far out that is, so I think it's still worth getting these changes in if you have the time.

michael-genson:
Hey there, it looks like this PR is no longer needed with the new ingredient parser. If you're running into the same issues with the new parser feel free to create a new PR or re-open this one with the changes for the new parser.

PR #4980, opened 2025-01-29
fix: Fixed LastMade recipes sorting order

### LINKED ISSUE #4937: [BUG] - Sorting by "Last Made" Places "Never Made" Recipes Incorrectly
### First Check

When sorting recipes by the "Last Made" date, recipes that have never been made are currently positioned at the end of the list. This behaviour is counterintuitive because the concept of "Last Made" is intended to highlight recipes based on how recently they were made. Recipes that have "never been made" are the ones I would most likely want to revisit first, as they are essentially the ones that haven't been given a chance yet.

Currently, the sorting order looks like this:

Recipe made in 2022
Recipe made in 2023
Recipe never made
Instead, the expected behaviour should position "never made" recipes before the earliest made recipe, treating them as if they were made the longest time ago. The expected sorting order would look like this:

Recipe never made
Recipe made in 2022
Recipe made in 2023
This approach makes sense because the purpose of sorting by "Last Made" is to allow users to quickly find recipes they haven’t tried yet, as well as revisit older ones. By placing "never made" recipes at the start of the list, users can more easily identify recipes they might wish to try.

### Steps to Reproduce

Add recipes to Mealie with "Last Made" dates, including some with no "Last Made" date at all.
Sort the recipes by "Last Made."
Observe that recipes with no "Last Made" date are placed at the end of the list.

-

### Mealie Version

Nightly

### Deployment

Docker (Synology)

Expected Behaviour:
Recipes with no "Last Made" date should appear before the earliest made recipe when sorting by "Last Made," as they represent the recipes that have never been made and are likely to be of higher interest in this context.

PancakeZik:

Fixes postgres incorrectly sorting null as the most recent.

-->

Fixes #4937

The original problem was only reproduced in postgres - in sqlite it was fine.

## Testing

Ran checks


michael-genson (changes requested):
I don't think we want to bake this into the repository on the backend. [The backend already supports specifying null ordering](https://docs.mealie.io/documentation/getting-started/api-usage/#order-by). Can we make this change to the frontend query instead?

From the API docs:
image

PancakeZik:
With my limited knowledge I couldnt find a way to make it work on the front end, but I will experiment with what you showed in the screenshot.

michael-genson:
You can probably add some logic here that handles null sorting:

https://github.com/mealie-recipes/mealie/blob/4229061377ddee3dded2198cd71c50664635b7c0/frontend/components/Domain/Recipe/RecipeCardSection.vue#L238-L248

You can calculate the order by null, something like:

```js
const orderDir = props.query?.orderDirection || preferences.value.orderDirection;
const orderByNull = orderDir === "asc" ? "first" : "last"
```

Then pass `orderDir` and `orderByNull` to `fetchMore`. You will also have to modify `fetchMore` to accept null ordering ([see the API docs](https://docs.mealie.io/documentation/getting-started/api-usage/#order-by))

PancakeZik:
See what you think now?

frontend/components/Domain/Recipe/RecipeCardSection.vue
```diff
@@ -236,12 +236,18 @@ export default defineComponent({
...
         page.value,
         perPage * pageCount,
         props.query?.orderBy || preferences.value.orderBy,
-        props.query?.orderDirection || preferences.value.orderDirection,
-        props.query,
+        orderDir,
+        {
+          ...props.query,
+          // Pass the null-position to the backend
+          orderByNullPosition: orderByNull,
```
michael-genson: Would we be able to clean this up so that the pagination params are together? Something like:

```
page.value,
perPage * pageCount,
props.query?.orderBy || preferences.value.orderBy,
orderDir,
orderByNullPosition,
props.query,
```

Instead of mixing them. You'll have to change the signature of `fetchMore` but it isn't used anywhere else in the app currently so that should be fine

frontend/components/Domain/Recipe/RecipeCardSection.vue
```diff
@@ -236,12 +236,18 @@ export default defineComponent({
     });
 
     async function fetchRecipes(pageCount = 1) {
+      const orderDir = props.query?.orderDirection || preferences.value.orderDirection;
+      const orderByNull = orderDir === "asc" ? "first" : "last";
```
michael-genson:
```suggestion
      const orderByNullPosition = orderDir === "asc" ? "first" : "last";
```

michael-genson (changes requested):
Looks good! Just a couple small code cleanliness things

frontend/components/Domain/Recipe/RecipeCardSection.vue
PancakeZik: sorry, I'm quite new, would you mind doing that?

michael-genson: Sure thing! No worries

michael-genson (dismissed):
Looks great now, thanks for this! I think since I pushed the last bit someone else will have to approve it, but putting my approval here anyway for visibility

Kuchenpirat (approved):

PR #4985, opened 2025-01-29
fix: Remove API Tokens from User APIs

### LINKED ISSUE #4981: [Security] Email to follow
### First Check

Emailed

### Steps to Reproduce

Emailed

Emailed

### Mealie Version

Build: `d2ced50fcceceb35e742765d99fd76378284f228`
Image: `mealie-recipes/mealie@sha256:0a9839f19e74d5c4a818161a5850b16d9608b240381691c317d3c23cd62e0c0f`

### Deployment

Other (please specify below)

Kubernetes (k3s)

michael-genson:

_(REQUIRED)_

Removes tokens from all APIs other than the create API (since you need the token exactly once, upon creation).

_(REQUIRED)_

Fixes https://github.com/mealie-recipes/mealie/issues/4981

The API still returns a list of all user tokens to the admin due to how our models are set up. We should _probably_ stop doing this, but now that the actual token string isn't being returned it's not a security risk (and it's restricted to admins). Removing the tokens from this API will take some more effort/investigation since that endpoint is used in several places in the app.

## Testing

Tested the frontend locally (to make sure you can still copy the token, etc.) and added backend tests to assert that the token is private.


hay-kot (approved):

disconn3ct:
FWIW I would not have reported if it was only names :+1: so this works for me. Thanks!

PR #5065, opened 2025-02-16
fix: Invalidate Expired Shared Links

michael-genson:

_(REQUIRED)_

Currently, while share links have an expiration date, nothing actually happens when they expire (and they continue to work). This PR:
1. verifies that tokens are valid before fetching the recipe (and instead returns a 404 if they're expired)
2. purges expired share links daily

I also ran into a bug where trying to share a recipe that belongs to a different household silently fails. I fixed this.

_(REQUIRED)_

N/A

## Testing

Pytest. I also verified via the frontend that sharing works as expected.


Kuchenpirat (approved):
Looks all good 🚀👍
Crazy that this got not caught earlier.

PR #5125, opened 2025-02-27
feat: Support User-Level Default Activities

miah120:
This feature adds a piece of user configuration to specify a default activity. 
image

On Login, a user will be taken to the activity set here instead of the recipe list. 
The activity defaults to the Recipes list if none is specified. 

None

I'm not the most familiar with how this might interact with the groups/households infrastructure. 
I took my routes from what the sidebar menu used, so I would hope everything just works. 

## Testing

Verified each route took me where I expected to go on login. 
Tested in docker container on desktop.

### DIFF
--- a/frontend/pages/user/profile/edit.vue
+++ b/frontend/pages/user/profile/edit.vue
@@ -171,13 +171,26 @@
         class="mt-10"
         :title="$t('profile.preferences')"
       />
-      <v-checkbox
-        v-model="userCopy.advanced"
-        class="mt-n4"
-        :label="$t('profile.show-advanced-description')"
-        color="primary"
-        @change="updateUser"
-      />
+      <v-card variant="outlined" style="border-color: lightgrey;">
+        <v-card-text>
+          <v-combobox
+            v-model="selectedDefaultActivity"
+            :label="$t('user.default-activity')"
+            :items="activityOptions"
+            :hint="$t('user.default-activity-hint')"
+            density="comfortable"
+            variant="underlined"
+            validate-on="blur"
+            persistent-hint
+          />
+          <v-checkbox
+            v-model="userCopy.advanced"
+            :label="$t('profile.show-advanced-description')"
+            color="primary"
+            @change="updateUser"
+          />
+        </v-card-text>
+      </v-card>
       <nuxt-link
         class="mt-5 d-flex flex-column justify-center text-center"
         :to="`/group`"
@@ -207,6 +220,8 @@ import UserAvatar from "~/components/Domain/User/UserAvatar.vue";
 import UserPasswordStrength from "~/components/Domain/User/UserPasswordStrength.vue";
 import { validators } from "~/composables/use-validators";
 import type { VForm } from "~/types/auto-forms";
+import { useUserActivityPreferences } from "~/composables/use-users/preferences";
+import { getDefaultActivityLabels, getActivityLabel, getActivityKey, ActivityKey } from "~/lib/api/activity/get-default-activity-route";
 
 export default defineNuxtComponent({
   components: {
@@ -220,6 +235,13 @@ export default defineNuxtComponent({
 
     useSeoMeta({
       title: i18n.t("settings.profile"),
+    });
+
+    const activityPreferences = useUserActivityPreferences();
+    const activityOptions = getDefaultActivityLabels(i18n);
+    const selectedDefaultActivity = ref(getActivityLabel(i18n, activityPreferences.value.defaultActivity));
+    watch(selectedDefaultActivity, () => {
+      activityPreferences.value.defaultActivity = getActivityKey(i18n, selectedDefaultActivity.value) ?? ActivityKey.RECIPES;
     });
 
     watch(user, () => {
@@ -273,6 +295,8 @@ export default defineNuxtComponent({
       updateUser,
       updatePassword,
       userCopy,
+      selectedDefaultActivity,
+      activityOptions,
       password,
       domUpdatePassword,
       passwordsMatch,
--- a/frontend/pages/user/profile/edit.vue
+++ b/frontend/pages/user/profile/edit.vue
@@ -221,7 +221,8 @@ import UserPasswordStrength from "~/components/Domain/User/UserPasswordStrength.
 import { validators } from "~/composables/use-validators";
 import type { VForm } from "~/types/auto-forms";
 import { useUserActivityPreferences } from "~/composables/use-users/preferences";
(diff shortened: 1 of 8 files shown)


CloCkWeRX (approved):
This looks more or less sensible. Did you want to rebase it?

miah120:
Rebased :)

CloCkWeRX:
In would suggest closing/forking, as your contribution is not valued to the point anyone (else) can be stuffed to looking at it in more than 45 day period, even if to say thanks, no thanks.

I am not a core contributor/maintainer, I cannot merge your changes, I don't speak for the project in any way. I think you have done everything right and your change is worth at the very least a surface level review, feedback or discussion.

Don't worry: I am told this was discussed internally in #5426 and the benefits of having a robot devalue human conversation while everyone else is too busy were once again reinforced and chosen to be a quite valuable service. More so than any potential contributions, effort, goodwill on either of our part. Comforting, right? At least as a minimum, you will now only be reminded your contribution isn't valued rather than it be discarded automatically.

Kuchenpirat:
@CloCkWeRX We had suggested changes to your PR that removed the stalebot, but since there was no follow-up, we went ahead and created a separate pull request to address the issue that we are currently not able to respond to issues within the 30-day timeframe.

Going forward, we're likely moving to a system where only triaged issues that require more information will be automatically marked as stale and then closed. For clarity, stalebot has never been configured to automatically close pull requests—only issues.

We understand that progress on other PRs and issues has been slow, but we are actively working on the Nuxt 3 upgrade. Since this upgrade will introduce a significant number of conflicts with most frontend PRs, we've decided to pause work on those PRs for now to avoid ongoing merge conflicts while fixing bugs in the nuxt 3 upgrade.

michael-genson:

mealie/alembic/versions/2025-08-10-21.22.50_766c23cf0a3a_support_default_activity.py
```diff
@@ -0,0 +1,32 @@
...
+"""
+
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "766c23cf0a3a"
+down_revision: str | None = "7cf3054cbbcc"
+branch_labels: str | tuple[str, ...] | None = None
+depends_on: str | tuple[str, ...] | None = None
```
michael-genson: We've had a migration since you generated yours, so you'll need to update this which should fix your backend tests. Also if you could rename the file to make it later than the new migration (e.g. `2025-08-08-21.22.50`)

```suggestion
"""'support default activity'

Revision ID: 766c23cf0a3a
Revises: d7b3ce6fa31a
Create Date: 2025-02-27 21:22:50.172425

"""

import sqlalchemy as sa
from alembic import op

# revision identifiers, used by Alembic.
revision = "766c23cf0a3a"
down_revision: str | None = "d7b3ce6fa31a"
branch_labels: str | tuple[str, ...] | None = None
depends_on: str | tuple[str, ...] | None = None
```

frontend/pages/user/profile/edit.vue
```diff
@@ -92,25 +65,28 @@
...
+                      v-model="userCopy.defaultActivity"
+                      :label="'Default Activity'"
+                      validate-on-blur
+                      required
+                    />
+                    <v-combobox
+                      v-model="selectedDefaultActivity"
+                      :label="$t('user.default-activity')"
+                      :items="activityOptions"
+                      validate-on-blur
```
michael-genson: Can we add a short description here? Something like "Select which page you'd like to navigate to upon logging in".

michael-genson (changes requested):
Overall I like this, though I have a few suggestions which I think will make this easier to work with as the app develops:

Instead of storing the entire activity path in the database, can we set up an enum system and figure it out with application-level logic?

Something like:

```python
class UserDefaultActivity(Enum):
    default = "_default"

    recipes = "recipes"
    mealplanner = "mealplanner"
    shopping_list = "shopping_list"

...

class UserOut(UserBase):
    ...
    default_activity: str = Field(exclude=True)

    ...

    @computed_field  # type: ignore
    @property
    def default_activity_route(self) -> str:
        GROUP_SLUG_ROUTE = f"/g/{self.group_slug}"

        if self.default_activity == UserDefaultActivity.recipes.value:
            return f"{GROUP_SLUG_ROUTE}/recipes"
        elif self.default_activity == UserDefaultActivity.mealplanner.value:
            return "/household.mealplan/planner/view"
        elif self.default_activity == UserDefaultActivity.shopping_list.value:
            return "/shopping-lists"
        else:
            return "/"
```

That way if routes change we can handle it in the application code rather than via a migration.

---

In terms of which pages can be set to defaults, let's keep it simple for now and just do:
- recipes
- meal planner
- shopping lists

It's possible there's a usecase for the other defaults but I'd rather keep it to just a few for now and add as needed.

michael-genson:
The more I think about this one, have you considered making this browser only rather than adding an entry to the backend? That way it could be device specific. For instance I'm usually using the shopping list on my phone, but recipes on my PC, etc.

miah120:
I'm not really sure what a backend-less, device-specific implementation of this would look like. If you think it makes sense to build it on top of what I'm doing here, then I'd welcome someone else to implement it. If the implementation would be something else entirely from what I have here I'd rather just drop this  PR than starting over from scratch.

michael-genson:
It would be more or less what you have here but using user preferences on the frontend instead. For instance, this is how we store new recipe preferences: https://github.com/mealie-recipes/mealie/blob/b0c78de2dacc802b50183c7dfd3dd830081f58e4/frontend/composables/use-users/preferences.ts#L210-L224

My thought is that there's more utility in doing it this way since, at least in my mind, it's more likely that a particular device wants to open to a specific page rather than all of your devices, if that makes sense.

frontend/pages/user/profile/edit.vue
miah120: Added!

miah120:
Thanks for the tip on where to look! I re-implemented it to be device specific.

michael-genson (changes requested):
Great! I think this is much more flexible. Two comments:
1. Can we move the `get-default-activity-route` logic to composables? e.g. `composables/use-default-activity-route`. And then maybe we should wrap the functions into a `useDefaultActivityRoute` function similar to other composables, just for consistency.
2. Do we maybe want to move this logic to the base `index` page instead of just the `login` page? That way any time a user opens the main page (e.g. if they open the PWA) they'll be redirected instead of only upon login.

Thanks for your work on this so far!

michael-genson (approved):
LGTM, thanks for all your work on this one!

PR #5340, opened 2025-04-13
feat: add the selected recipe servings and yields in the content of the recipe post action

felixschndr:

This change allows the `post` recipe action to send information about the selected recipe `servings` and `yieldQuantity` in the body.

1st for https://github.com/felixschndr/mealie-bring-api/issues/15

Previous discussion regarding this: https://github.com/mealie-recipes/mealie/discussions/3610#discussioncomment-12524763

Since I am unfamiliar with the codebase I'd like to get some feedback and proposed changes if I (unintentionally) dit not comply with coding styles.

## Testing

Took a debugger to see that the data was set correctly. 
I am unsure whether it is possible to check the type script side of things or how I can build a docker container to run locally.

### DIFF
--- a/mealie/routes/households/controller_group_recipe_actions.py
+++ b/mealie/routes/households/controller_group_recipe_actions.py
@@ -1,7 +1,7 @@
 from functools import cached_property
 
 import requests
-from fastapi import APIRouter, BackgroundTasks, Depends, HTTPException, status
+from fastapi import APIRouter, BackgroundTasks, Body, Depends, HTTPException, status
 from fastapi.encoders import jsonable_encoder
 from pydantic import UUID4
 
@@ -67,8 +67,10 @@ class GroupRecipeActionController(BaseUserController):
     # ==================================================================================================================
     # Actions
 
-    @router.post("/{item_id}/trigger/{recipe_slug}/{scaled_amount}", status_code=202)
-    def trigger_action(self, item_id: UUID4, recipe_slug: str, scaled_amount: float, bg_tasks: BackgroundTasks) -> None:
+    @router.post("/{item_id}/trigger/{recipe_slug}", status_code=202)
+    def trigger_action(
+        self, item_id: UUID4, recipe_slug: str, bg_tasks: BackgroundTasks, scaled_amount: float = Body(..., embed=True)
+    ) -> None:
         recipe_action = self.repos.group_recipe_actions.get_one(item_id)
         if not recipe_action:
             raise HTTPException(
@@ -93,12 +95,10 @@ class GroupRecipeActionController(BaseUserController):
                 detail=ErrorResponse.respond(message="Not found."),
             ) from e
 
-        payload = GroupRecipeActionPayload(
-            action=recipe_action, content=recipe, scaled_amount=scaled_amount
-        ).model_dump()
+        payload = GroupRecipeActionPayload(action=recipe_action, content=recipe, scaled_amount=scaled_amount)
         bg_tasks.add_task(
             task_action,
             url=recipe_action.url,
-            json=jsonable_encoder(payload),
+            json=jsonable_encoder(payload.model_dump()),
             timeout=15,
         )
--- a/mealie/routes/households/controller_group_recipe_actions.py
+++ b/mealie/routes/households/controller_group_recipe_actions.py
@@ -67,8 +67,10 @@ class GroupRecipeActionController(BaseUserController):
     # ==================================================================================================================
     # Actions
 
-    @router.post("/{item_id}/trigger/{recipe_slug}", status_code=202)
-    def trigger_action(self, item_id: UUID4, recipe_slug: str, bg_tasks: BackgroundTasks) -> None:
+    @router.post("/{item_id}/trigger/{recipe_slug}/{servings}/{yield_quantity}", status_code=202)
+    def trigger_action(
+        self, item_id: UUID4, recipe_slug: str, servings: int, yield_quantity: int, bg_tasks: BackgroundTasks
+    ) -> None:
         recipe_action = self.repos.group_recipe_actions.get_one(item_id)
         if not recipe_action:
             raise HTTPException(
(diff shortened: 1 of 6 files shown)


felixschndr:
Can one of the maintainers run the publish PR image job so I can test e2e? Thank you:)

felixschndr:
On second thought: I think `servings` and `yieldQuantity` is not what I want. I need this number:
Screenshot_20250413_182807_Chrome
This is probably the `recipeScale`, right?

Kuchenpirat:
The pr image job currently works only on branches that originate from the mealie repo due to authentication issues when using another origin.
We know that it is not hugely useful that way and are working on a mitigation.

felixschndr:
Okay, good to know.
May I get a response to my last question?

Kuchenpirat:
I believe this would be yieldQuantity * recipeScale

But it will fall back onto recipeServings if I remember right 

---

Just on phone so can't check right now

felixschndr:
Hey @Kuchenpirat
I took a look at the source code and would say (as you guessed) that it is indeed servings*scale (https://github.com/mealie-recipes/mealie/blob/mealie-next/frontend/components/Domain/Recipe/RecipeScaleEditButton.vue#L112 and https://github.com/mealie-recipes/mealie/blob/mealie-next/frontend/composables/recipes/use-scaled-amount.ts#L25). Can you confirm this for me as I am not that deep in the codebase?

Kuchenpirat:
yeah, that should be it 👍

felixschndr:
I implemented the scaled amount as a field to be sent. There is currently the following problem: a `Recipe` does not have a `scaled_amount` (thus the one test fails). Shall I add a field to the recipe or shall I reuse one of the existing fields (e.g. `servings` or `yield_quantity`)? I think the first option would be cleaner, but happy to hear your thoughts

felixschndr:
@Kuchenpirat what's your opinion on that?

Kuchenpirat:
Yeah, reusing one of the fields does not make too much sense.
Do I understand you correctly that you want to add a field to the database?

felixschndr:
> Do I understand you correctly that you want to add a field to the database?

I was thinking about adding the field to the `Recipe` class. However, I think this would require a change to the database. Moreover, I don't think this field should be in either the database or the class.

@Kuchenpirat what do you think about the implementation I just committed?

felixschndr:
This is ready from my side, may I get a review? @Kuchenpirat

mealie/routes/households/controller_group_recipe_actions.py
```diff
@@ -67,8 +67,8 @@ def delete_one(self, item_id: UUID4):
     # ==================================================================================================================
     # Actions
 
-    @router.post("/{item_id}/trigger/{recipe_slug}", status_code=202)
-    def trigger_action(self, item_id: UUID4, recipe_slug: str, bg_tasks: BackgroundTasks) -> None:
+    @router.post("/{item_id}/trigger/{recipe_slug}/{scaled_amount}", status_code=202)
```
michael-genson: This would be more appropriate in the request body rather than the URL path. That way if we need to add additional attributes later down the line we don't have to keep changing the path.

mealie/routes/households/controller_group_recipe_actions.py
```diff
@@ -67,8 +67,8 @@ def delete_one(self, item_id: UUID4):
...
+    def trigger_action(self, item_id: UUID4, recipe_slug: str, scaled_amount: float, bg_tasks: BackgroundTasks) -> None:
```
michael-genson: Let's instead make a data class, something like:
```python
class RecipePostActionData(MealieModel):
    scaled_amount: float | None = None
```

michael-genson (changes requested):
Currently changes that involve the frontend are on hold until the [Nuxt 3 upgrade](https://github.com/mealie-recipes/mealie/pull/5184) is completed. However we can still review in the meantime

mealie/routes/households/controller_group_recipe_actions.py
felixschndr: I added the field to `GroupRecipeActionPayload` as this is the only location where it is used and it seems appropriate there. Is that okay for you?
https://github.com/mealie-recipes/mealie/pull/5340/commits/8f8a9c150d8f51730dafffd52cd0a752360d0345

mealie/routes/households/controller_group_recipe_actions.py
felixschndr: Done in https://github.com/mealie-recipes/mealie/pull/5340/commits/23e6be5c97aa1e5a49673b2c1e1c8223f11488de

felixschndr:
@michael-genson Does the new way of implementing it work for you?

I also rebased on main

felixschndr:
Any thoughts?

mealie/routes/households/controller_group_recipe_actions.py
```diff
@@ -68,7 +68,9 @@ def delete_one(self, item_id: UUID4):
     # Actions
 
     @router.post("/{item_id}/trigger/{recipe_slug}", status_code=202)
-    def trigger_action(self, item_id: UUID4, recipe_slug: str, bg_tasks: BackgroundTasks) -> None:
+    def trigger_action(
+        self, item_id: UUID4, recipe_slug: str, bg_tasks: BackgroundTasks, scaled_amount: float = Body(..., embed=True)
```
michael-genson: Should we default this to 1 to maintain old behavior?

```suggestion
        self, item_id: UUID4, recipe_slug: str, bg_tasks: BackgroundTasks, scaled_amount: float = Body(1, embed=True)
```

michael-genson (changes requested):
Looks good, aside from one small change. I'll see if I can get this merged in before the Nuxt 3 upgrade, since the frontend changes aren't too bad

mealie/routes/households/controller_group_recipe_actions.py
felixschndr: Good idea, changed it

michael-genson:
Looks good! Will merge this when I have the time to also update the Nuxt branch (if we're lucky that's today) or after we merge that branch.

Appreciate you sticking with this one!

michael-genson (approved):

felixschndr:
Thanks for merging. Will there be a (minor) release before the `Next` upgrade?

michael-genson:
Most likely not, as the Nuxt upgrade is pretty close to getting merged

Issue #5511, opened 2025-06-08
[BUG] - Changing Shopping list item state in offline mode dosn't work


cm226:
### First Check

- [x] This is not a feature request.
- [x] I added a very descriptive title to this issue (title field is above this).
- [x] I used the GitHub search to find a similar issue and didn't find it.
- [x] I searched the Mealie documentation, with the integrated search.
- [x] I already read the docs and didn't find an answer.
- [x] This issue can be replicated on the demo site (https://demo.mealie.io/).

### What is the issue you are experiencing?

Background information available here: https://github.com/mealie-recipes/mealie/discussions/5291

In summary the issue is that when using the shopping list in offline mode, its not possible to change the checked state on list items from whatever the state was when last online

### Steps to Reproduce

* Install PWA
* Create a shopping list with at least 1 item
* go offline
* check the shopping list item

Expectation is that the item stays checked. 
Actual is that the item is re-checked immediately. 

### Please provide relevant logs

I debugged the issue a-bit, the cause is the interaction between the PWA cache and the local state stored in the update queues in frontend/composables/use-shopping-list-item-actions.ts

Sequence of interaction is

1. Some list state is cached in the PWA cache when online
2. User goes offline
3. User updates list item state (e.g checking an item)
4. Local state in use-shopping-list-item-actions.ts is queued for update
5. component is refreshed
6. Refresh attempts to fetch the latest state from the server 
7. Fetching latest state from server request is handled by the SW cache (because we are offline), which responds with the last state we had when we were online

I think a solution to this might be to merge the local pending state with the state from the server when refreshing, using updatedAt field to show the latest state to the user. 

I have some changes that do the above, which Ill make a PR for. If there is a better way to resolve I'm happy to update. 

### Mealie Version

Confimed this happens here (https://demo.mealie.io/g/home)

Also on mealie next branch

### Deployment

Docker (Linux)

PR #5512, opened 2025-06-08
fix: #5511, list item state doesn't change when offline

cm226:

* Merge pending local state changes with server state before updating UI
* Local state changes will overwrite server state where the updatedAt property is larger

https://github.com/mealie-recipes/mealie/issues/5511

Not sure if the check for updatedAt is required? It might be that we always want local state to overwrite server state, because local state will always be newer ( I think ) 

I don't think I have broken anything by updating the updatedAt property every time, without checking the isChecked, but worth considering, since I'm not very familiar with the code base. The reason I did that was because I found there was a case where if you 
* Checked an item while online
* Went offline 
* Unchecked the same item,
Then the updatedAt times were equal, and so the local state didnt overwrite server state. 
 

## Testing

* Ran steps to reproduce in the issue, and confirmed the this fixes the issue
* Generally tested out shopping list functionality in online and offline modes

### DIFF
--- a/frontend/composables/use-shopping-list-item-actions.ts
+++ b/frontend/composables/use-shopping-list-item-actions.ts
@@ -89,9 +89,31 @@ export function useShoppingListItemActions(shoppingListId: string) {
     return true;
   }
 
+  function mergeListItemsByLatest(
+    list1: ShoppingListItemOut[],
+    list2: ShoppingListItemOut[]
+  ) {
+    const mergedList = [...list1];
+    list2.forEach((list2Item) => {
+      const confictingItem = mergedList.find((item) => item.id === list2Item.id)
+      if (confictingItem &&
+        list2Item.updatedAt && confictingItem.updatedAt &&
+        list2Item.updatedAt > confictingItem.updatedAt) {
+        mergedList.splice(mergedList.indexOf(confictingItem), 1, list2Item)
+      } else if (!confictingItem) {
+        mergedList.push(list2Item)
+      }
+    })
+    return mergedList
+  }
+
   async function getList() {
     const response = await api.shopping.lists.getOne(shoppingListId);
-    return response.data;
+    if (response.data) {
+      const createAndUpdateQueues = mergeListItemsByLatest(queue.update, queue.create);
+      response.data.listItems = mergeListItemsByLatest(response.data.listItems ?? [], createAndUpdateQueues);
+    }
+    return response.data
   }
 
   function createItem(item: ShoppingListItemOut) {
@@ -188,7 +210,7 @@ export function useShoppingListItemActions(shoppingListId: string) {
   }
 
   async function process() {
-    if(queueEmpty.value) {
+    if (queueEmpty.value) {
       queue.lastUpdate = Date.now();
       return;
     }
--- a/frontend/composables/use-shopping-list-item-actions.ts
+++ b/frontend/composables/use-shopping-list-item-actions.ts
@@ -109,7 +109,7 @@ export function useShoppingListItemActions(shoppingListId: string) {
 
   async function getList() {
     const response = await api.shopping.lists.getOne(shoppingListId);
-    if (response.data) {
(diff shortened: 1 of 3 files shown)


frontend/composables/use-shopping-list-item-actions.ts
```diff
@@ -89,9 +89,31 @@ export function useShoppingListItemActions(shoppingListId: string) {
...
+  }
+
   async function getList() {
     const response = await api.shopping.lists.getOne(shoppingListId);
-    return response.data;
+    if (response.data) {
+      const createAndUpdateQueues = mergeListItemsByLatest(queue.update, queue.create);
+      response.data.listItems = mergeListItemsByLatest(response.data.listItems ?? [], createAndUpdateQueues);
+    }
+    return response.data
```
michael-genson: Is there a way to skip this when we're online? i.e. only run this if we went from offline to online

frontend/composables/use-shopping-list-item-actions.ts
```diff
@@ -89,9 +89,31 @@ export function useShoppingListItemActions(shoppingListId: string) {
...
+    list2: ShoppingListItemOut[]
+  ) {
+    const mergedList = [...list1];
+    list2.forEach((list2Item) => {
+      const confictingItem = mergedList.find((item) => item.id === list2Item.id)
+      if (confictingItem &&
+        list2Item.updatedAt && confictingItem.updatedAt &&
+        list2Item.updatedAt > confictingItem.updatedAt) {
+        mergedList.splice(mergedList.indexOf(confictingItem), 1, list2Item)
+      } else if (!confictingItem) {
```
michael-genson:
```suggestion
      const conflictingItem = mergedList.find((item) => item.id === list2Item.id)
      if (conflictingItem &&
        list2Item.updatedAt && conflictingItem.updatedAt &&
        list2Item.updatedAt > conflictingItem.updatedAt) {
        mergedList.splice(mergedList.indexOf(conflictingItem), 1, list2Item)
      } else if (!conflictingItem) {
```

michael-genson (changes requested):
One request to improve performance (and a small typo fix). Otherwise this looks good, thank you!

The reason I implemented comparing update dates is in the scenario where someone updates the list on the server and someone else updates the item locally (i.e. to resolve multiple users making changes). If there's only one user acting on the list I'd agree that it's unnecessary.

cm226:
Thanks for your comments. I updated the PR to only merge local state when offline.

frontend/composables/use-shopping-list-item-actions.ts
```diff
@@ -89,9 +89,31 @@ export function useShoppingListItemActions(shoppingListId: string) {
...
+        mergedList.push(list2Item)
+      }
+    })
+    return mergedList
+  }
+
   async function getList() {
     const response = await api.shopping.lists.getOne(shoppingListId);
-    return response.data;
+    if (window.$nuxt.isOffline && response.data) {
```
michael-genson, line 112: Have you tested to make sure this works? Won't we be online once we're able to receive data?

cm226: hmm I might be missing something, I cant see, right now why this wouldn't work work. 

I tested it offline and online and it seemed to be working. 

My understanding of what this does is:
- will use server state when we are online 
- will merge local state with server state when offline

michael-genson: As long as you're seeing it work it should be fine. My concern was that we won't be able to fetch the server state while offline

cm226: Hi yep I'm confident this is working for offline and online, response.data is populated when offline by the service worker cache.

michael-genson:
LGTM. Will wait to merge until the Nuxt upgrade is completed, unless I can find the time to manually merge it in

michael-genson (approved):

PR #5189, opened 2025-03-08
feat: Update seeding to use new foods list format - round 2

Cameronwyatt:
## What this PR does / why we need it:

[See this as the original PR](https://github.com/mealie-recipes/mealie/pull/4812) with all of the context. After it was merged, Mealie hit the upper-limit of the free-tier of CrowdIn. @hay-kot has since requested the Open Source license and it was granted. 

I rebased on the latest changes, which includes the translation changes since the last PR was submitted

## Which issue(s) this PR fixes:

Fixes https://github.com/mealie-recipes/mealie/issues/1230

## Special notes for your reviewer:

No changes from the last PR other than rebasing

## Testing

Tested seeding en-US.json and a couple other languages from a fresh install.
Tested seeding en-US.json while there were already foods present - no duplicates were created.

`task py:test` - all tests passed.

PR #5514, opened 2025-06-09
feat: Update food seeding logic & increase food seed size

### LINKED ISSUE #1230: [v1.0.0b] [Task] - Link food labels to foods when importing seed data

Cameronwyatt:

See previous 2 PRs...[one here](https://github.com/mealie-recipes/mealie/pull/5189)
Both had issues in the past due to a limit we were running into with CrowdIn. That has supposedly been resolved, [see here](https://community.crowdin.com/t/you-ve-about-to-reach-the-limit/2829/19?u=cameronmwyatt)

I made sure to pull in all of the CrowdIn changes that have happened since the original PRs, then converted those files to use the new format. 

Fixes https://github.com/mealie-recipes/mealie/issues/1230

3rd time's the charm? 

## Testing

Only updated the locale files and the helper script since last PRs


hay-kot (approved):

MeikelLP:
@hay-kot + @Cameronwyatt 
Would it be possible to add a "force" feature that overwrites/merges existing food items? 
I'd love to use the seed and tried to use it but its such a big hassle to fix all recipes (or merge all food items manually) after importing the seed.
I have over 500 recipies already and this is no go for me.

PR #5537, opened 2025-06-17
fix: Fixing the OpenAPI Spec and the Call to delete a shared recipe.

### LINKED ISSUE #5531: [BUG] - OpenAPI Spec for Delete /api/shared/recipes/{item_id} endpoint is defined with a schema that allows for a null value
### First Check

I was trying to use the OpenAPI Spec on the website. I used the json file, but ran into an issue where my compilation is failing. 

Upon further inspection, it looks like this is a required path parameter, that can take a null object -- which I think is a mistake. 

``` 
"parameters": [
    {
        "name": "item_id",
        "in": "path",
        "required": true,
        "schema": {
            "anyOf": [
                {
                    "type": "string",
                    "format": "uuid4"
                },
                {
                    "type": "null"
                }
            ],
            "title": "Item Id"
        }
    },
    ...
]
```

### Steps to Reproduce

I used Swift Open API Generator. 
To reproduce, 
1. Start a brand new iOS project.
2. Add the plugins for Swift OpenAPI Generator and Swift OpenAPI Runtime along with Swift OpenAPI URLSession. Set them up using the build phases, tools.
3. Load the openapi spec json file. 
4. Add the config file ( I just added a basic yaml file )
5. Try to compile to see the error. 

YAML File: 
```
generate:
  - types
  - client
```

This was the error I got

```
Library/Developer/Xcode/DerivedData/mealie-drczqrczyjhgetahibzsrgxikkhu/Build/Intermediates.noindex/BuildToolPluginIntermediates/mealie.output/mealie/OpenAPIGenerator/GeneratedSources/Client.swift:16022:31 Value of type 'Operations.delete_one_api_shared_recipes__item_id__delete.Input' has no member 'path'
```

[openapi.json](https://github.com/user-attachments/files/20751165/openapi.json)

### Mealie Version

_No response_

### Deployment

Docker (Synology)

I don't think these details are relevant for this issue.

sravankaruturi:

This change fixes the function `delete_one(self, item_id)` in `RecipeSharedController`. 

The `item_id` was able to take in Null. A delete item should not be able to take `None` as the id. This is evident by the underlying call to the `self.mixins.delete_one(item_id)` function, which doesn't accept `None`. 

This bug also breaks the OpenAPI spec when generating Swift code ( which is how I found it ).

Fixes #5531

### DIFF
--- a/mealie/routes/shared/__init__.py
+++ b/mealie/routes/shared/__init__.py
@@ -46,5 +46,5 @@ class RecipeSharedController(BaseUserController):
         return self.mixins.get_one(item_id)
 
     @router.delete("/{item_id}")
-    def delete_one(self, item_id: UUID4 | None = None) -> None:
+    def delete_one(self, item_id: UUID4) -> None:
         return self.mixins.delete_one(item_id)


michael-genson (approved):
Thank you!

Issue #5543, opened 2025-06-19
[BUG] - Mealie nightly visual glitches with Nuxt 3 upgrade


cmintey:
I downloaded the latest nightly and found some visual glitches related to Nuxt 3 migration.

1. Admin settings navigation pane doesn't update, so can't actually navigate to any settings
    ![Image](https://github.com/user-attachments/assets/35d94971-649a-4795-a98e-4d1970be8664)

2. Shopping lists "back button" is too small for the text on my iPhone
    ![image](https://github.com/user-attachments/assets/5b97c9ea-2619-48ca-a8d3-06f13fe7f58c)

3. The old font doesn't seem to be working. I see a 404 when trying to fetch the Roboto font

michael-genson:
> Admin settings navigation pane doesn't update

Yeah this is bizarre, happens to me too. It works fine on the dev server so we never caught it

PR #5547, opened 2025-06-20
fix: shopping list button in one row

Kuchenpirat:

fixes the shopping list button alignments

#5543 

## Testing

local


boc-the-git (approved):
:saluting_face:

PR #5570, opened 2025-06-23
fix: Cookbooks not rendering on sidebar

michael-genson:

_(REQUIRED)_

For some reason cookbooks don't always render on the sidebar, or they render funny. Switching to a store seems to fix it. It's really hard to reproduce consistently (some magic combination of opening a new tab and hard refreshing usually does it) but this seems to fix it.

_(REQUIRED)_

N/A

~~This kind of duplicates the cookbook store~~ nevermind we already have something like this, I'm going to implement that here (this is now done).

## Testing

Built the prod image a lot, seems to be consistently fixed now.


Kuchenpirat (approved):
looking at the the lines removed vs lines added i quite like this PR 👍

PR #5590, opened 2025-06-27
feat: create recipe from multiple images

SkepticMystic:
## What this PR does / why we need it:

This PR aims to resolve a much-requested issue of accepting multiple images when creating a recipe from an image.

## Which issue(s) this PR fixes:

Fixes #4858
Relates to discussion #4067

## Special notes for your reviewer:

I'm a JS/TS web dev, mainly using Svelte. So my Vue knowledge is limited. Very happy for feedback of anykind.

https://github.com/mealie-recipes/mealie/issues/4858#issuecomment-2576212809 tells us that the backend is already setup to handle multiple images, so this is purely a frontend, API client change.

I'm aware the i8n copy/content needs to be adjusted for "images" (plural).

`Kuchenpirat`: Added an axios api config option to increase the request timeout. Alternatively we could also just remove the timeout but having a timeout is sensible imho. This means also the current implementation will fail currently after 10s.

## Testing

More details in a comment below. I created 3 simple images, each with a single ingredient. I then uploaded multiple images to create the recipe, and it successfully detected the ingredients from each image.

PR #5595, opened 2025-06-28
feat: new create from image visuals

Kuchenpirat:

New Visuals for the create from image functionality as the old visuals were a bit broken after the nuxt 3 upgrade and did not handle multiple images upload very gracefully. 

None, just didn't like the visuals too much, especially with multiple images adding up one below the other. 

- merge after #5590 

## Testing

Local

## Screenshot
image

### DIFF
--- a/frontend/assets/css/main.css
+++ b/frontend/assets/css/main.css
@@ -64,3 +64,7 @@ a {
 .fill-height {
   min-height: 100vh;
 }
+
+.vue-simple-handler {
+  background-color: rgb(var(--v-theme-primary)) !important;
+}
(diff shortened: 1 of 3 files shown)


michael-genson (changes requested):
Looks SOOO much better! A couple things I noticed:
1. It's possible to delete/edit the photos after submitting them to the backend. Can we disable those buttons after submitting?
2. The save button doesn't seem to be working after cropping/rotating

Kuchenpirat:
fixed the issues you found. 
I also moved the save button out of the menu and made it only visible if the image was changed.

image

michael-genson (approved):
Love it!

Issue #5350, opened 2025-04-15
[BUG] - UX problem: Back from the detail view, scroll position always at the start again


flipzoom:
### First Check

- [x] This is not a feature request.
- [x] I added a very descriptive title to this issue (title field is above this).
- [x] I used the GitHub search to find a similar issue and didn't find it.
- [x] I searched the Mealie documentation, with the integrated search.
- [x] I already read the docs and didn't find an answer.
- [x] This issue can be replicated on the demo site (https://demo.mealie.io/).

### What is the issue you are experiencing?

First of all, thank you for the great software. I'm currently digitising all my recipes. And this is where I come across a very, very unpleasant and annoying UX problem in daily use. If you go back to the overview from a detailed view (in the browser via the back button, mobile via native Android back), you end up at the top of the recipe list again and again.

This is still ‘okay’ up to the point when you only have 10 or 20 recipes, but not with 50 or 100 recipes. It's not uncommon to scroll through the list in search of the right meal for tonight, look at a recipe to see if you have the ingredients, for example, and then want to go back to the list to continue searching. And this is where the frustration factor comes in. Because now you're back at the beginning and have to scroll down 80 recipes again to the position where you were before. That's no fun. Especially not mobile.

The last scroll position is lost somewhere and is not saved.

https://github.com/user-attachments/assets/1121b4e5-52ae-452c-9b3b-1ea573a364bf

### Steps to Reproduce

1. go to the recipe overview
2. scroll to the bottom
3. click on a recipe to see it in detail
4. go back
5. now you are back at the top of the list

### Please provide relevant logs

There are no relevant logs

### Mealie Version

Version v2.8.0
Build [2deb9c276cfff6b1aff473776dd327110f2eb55b](https://github.com/mealie-recipes/mealie/commit/2deb9c276cfff6b1aff473776dd327110f2eb55b)

### Deployment

Docker (Synology)

flipzoom:
Actually, as far as I have read, NUXT has already implemented this option, but another approach may also help: https://levelup.gitconnected.com/nuxt-js-how-to-retain-scroll-position-when-returning-to-page-without-navigation-history-7f0250886d27

death2all110:
Just ran into this myself.

FGOD1983:
I was going to create an issue for this too because it is really annoying if you are going through the recipes looking for what to eat next week and you have to start from the top again every time you opened a recipe. 

Would be great if this can be fixed in a future version

[...]

PR #5601, opened 2025-06-29
feat: Open recipes in new tab for mobile view

razziel89:

This PR addresses issue #5350. There may be older, closed issues about the same UX problem that I didn't find. Please be referred to the aforementioned issue for details. In a nutshell, though: When scrolling through lists of recipes and looking at them in detail, you are always taken back to the top of the list when going back to the list from a detailed view. That is fine for small numbers of recipes, but it is frustrating when browsing large libraries. On a computer, we can right click and then open a recipe in a new tab. That way, one tab can remain open on the recipe list page while the details can be viewed in another tab, thus not losing the position in the list. Unfortunately, the same cannot be done on any of the many mobile devices that I tried. It is not possible to long-press and then open a recipe in a new tab. Thus, this PR causes recipes in the recipe list to open in new tabs on mobile.

- The mobile recipe card's content is wrapped in an anchor tag pointing to the detail page and the `to` attribute is removed. The component still looks the same. Most of the changes pertain to whitespace due to adjusted indentations.

For the reasons mentioned above, I think it makes sense to always open recipe detail views in new tabs on mobile.

- #5350 

I was considering to make this optional and configurable on a per-user basis, but I have to admit to not knowing the frontend's tech stack well enough to do that. I hope it would be simple but I would need advice and help if that were desired.

Any feedback you can give is appreciated! If this change is not desirable, please let me know. Me and my family consider the current behaviour rather frustrating and would very much like to see this change, or a change to a similar effect, introduced in mainline mealie.

Thank you very much for mealie in general and for considering this PR in particular.

I recommend not showing changes due to whitespace when viewing this PR's diff. You can follow [this link](https://github.com/mealie-recipes/mealie/pull/5601/files?diff=unified&w=1) to get to such a view.

## Testing

I built the production docker image locally with my changes and accessed a locally hosted instance both from a computer and from a phone. When clicked, recipes in the mobile view opened in new tabs while ones clicked in the desktop view did not. The mobile recipe cards still look the same as before in my tests.

### DIFF
--- a/frontend/components/Domain/Recipe/RecipeCardMobile.vue
+++ b/frontend/components/Domain/Recipe/RecipeCardMobile.vue
@@ -7,116 +7,127 @@
         :style="{ cursor }"
         hover
         height="100%"
-        :to="$attrs.selected ? undefined : recipeRoute"
         @click="$emit('selected')"
       >
-        <v-img
-          v-if="vertical"
-          class="rounded-sm"
-          cover
+        <a
+          :href="recipeRoute"
+          target="_blank"
+          rel="noopener noreferrer"
+          style="
+           text-decoration: none;
+           color: inherit;
+           display: block;
+           height: 100%;
+           "
         >
-          <RecipeCardImage
-            :icon-size="100"
-            :slug="slug"
-            :recipe-id="recipeId"
-            size="small"
-            :image-version="image"
-            :height="height"
-          />
-        </v-img>
-        <v-list-item
-          lines="two"
-          class="py-0"
-          :class="vertical ? 'px-2' : 'px-0'"
-          item-props
-          height="100%"
-          density="compact"
-        >
-          <template #prepend>
-            <slot
-              v-if="!vertical"
-              name="avatar"
-            >
-              <RecipeCardImage
-                :icon-size="100"
-                :slug="slug"
-                :recipe-id="recipeId"
-                :image-version="image"
-                size="small"
-                width="125"
-                :height="height"
-              />
-            </slot>
-          </template>
-          <div class="pl-4 d-flex flex-column justify-space-between align-stretch pr-2">
-            <v-list-item-title class="mt-3 mb-1 text-top text-truncate w-100">
-              {{ name }}
-            </v-list-item-title>
-            <v-list-item-subtitle class="ma-0 text-top">
-              <SafeMarkdown v-if="description" :source="description" />
-              <p v-else>
-                <br>
-                <br>
-                <br>
-              </p>
-            </v-list-item-subtitle>
-            <div
-              class="d-flex flex-nowrap justify-start ma-0 pt-2 pb-0"
-              style="overflow-x: hidden; overflow-y: hidden; white-space: nowrap;"
-            >
-              <RecipeChips
-                :truncate="true"
-                :items="tags"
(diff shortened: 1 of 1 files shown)


michael-genson:
Since Mealie is also used as a PWA I'm not sure we want this behavior by default, though I'm definitely open to community feedback on this one. A good compromise IMO would be to add this to the context menu (e.g. "Open in new tab") and just only show it on mobile devices. If users feel strongly one way or another I'm happy to change my mind.

The ideal solution to this problem (which I think should definitely be implemented eventually) is making sure the back button brings you back to the correct place, rather than starting you at the top. However, due to how it's implemented it's definitely not a quick fix, so this kind of solution is better in the meantime.

razziel89:
Thanks a lot for the swift reply!

> Since Mealie is also used as a PWA I'm not sure we want this behavior by default

Thanks for raising this good point about the PWA. I agree that users should be able to decide whether they want this behaviour or not. 

> though I'm definitely open to community feedback on this one.

How is such feedback usually being gathered? Do people usually comment on PRs or is there a channel on the discord server where such things are commonly discussed? I’m asking since this is my first time contributing to mealie. Any advice is appreciated!

> A good compromise IMO would be to add this to the context menu (e.g. "Open in new tab") and just only show it on mobile devices. If users feel strongly one way or another I'm happy to change my mind.

Thanks for the suggestion. In my experience, it’s rather easy to fumble the screen on small devices, particularly when on the go. That would mean it’s rather easy to miss hitting the context menu button, opening the recipe in the same tab instead. Consequently, I would like to suggest a new global toggle to select whether to open recipes in a new tab or not. That new toggle could be similar to the one on desktop that lets you select the type of recipe cards to show. If I remember correctly, it’s simply called “toggle view”. If the aforementioned new toggle’s value were persisted via a query parameter like the sorting and the filters, then each user could select the desired behaviour and store their choice in a bookmark via the query parameter, all without changing the default behaviour. What to you think about that alternative compromise?

> The ideal solution to this problem (which I think should definitely be implemented eventually) is making sure the back button brings you back to the correct place, rather than starting you at the top. However, due to how it's implemented it's definitely not a quick fix, so this kind of solution is better in the meantime.

I fully agree. That’s the change I wanted to make initially, but it wasn’t clear to me how and I didn’t have enough time to try and figure it out. 

My alternative suggestion again in a nutshell:
- A new global toggle shown only on mobile. 
- The value of the toggle determine whether recipes clicked are opened in new tabs or not.
- The toggle's state is persisted via a query parameter, which also makes it possible pre-select a value via an appropriate bookmark.

What do you think?

michael-genson:
For community feedback really anywhere is fine, either on the PR, on Discord, or in an issue/discussion.

If we go the route of a toggle state, it would probably be better to just keep it in local storage similar to how we do other ad hoc settings (like showing all shopping lists). That way the implementation is rather simple, and it's per-device so you can have a different setting for desktop vs mobile, etc.

razziel89:
Thanks lot for the reply. I'll try to implement your suggestion soon and update this PR once I have something.

razziel89:
I still intend to implement the feature described in this PR but I do not yet know when I will find the time to do so. For the time being, I will maintain this small diff in a fork in my local namespace. I am thus closing this PR.

PR #5603, opened 2025-06-29
fix: check for OPENAI_MODEL in OPENAI_FEATURE

jknndy:
`mealie/core/settings/settings.py`
- Previously, the method set the description message `"OPENAI_MODEL is not set"` when `OPENAI_MODEL` *was* set due to a missing `not`.
- This updates the condition to correctly trigger the message when `OPENAI_MODEL` is *not* set.


michael-genson (approved):
Thanks!

PR #5609, opened 2025-06-30
fix: recipe page warnings

Kuchenpirat:

After upgrading to Nuxt 3, the Recipe Explorer page started throwing a bunch of warnings, making the browser console difficult to use on that page.

The specific warning was:
`RecipeCard.vue?t=1751270565415:239 [Vue warn]: Component inside <Transition> renders non-element root node that cannot be animated.`

The root issue wasn't with the animation itself. The problem came from the use of the v-hover inside a v-lazy, which doesn't render a proper DOM element that v-hover can use to determine isHovering. To fix this, v-hover needed to be wrapped in an additional div to ensure it had a valid DOM node to work with.

None

This is a lot smaller change than the div it makes out to be. I literally just had to wrap the v-hover with a div. 

## Testing

Local.

### DIFF
--- a/frontend/components/Domain/Recipe/RecipeCard.vue
+++ b/frontend/components/Domain/Recipe/RecipeCard.vue
@@ -1,101 +1,104 @@
 <template>
+  <!-- Wrap v-hover with a div to provide a proper DOM element for the transition -->
   <v-lazy>
-    <v-hover
-      v-slot="{ isHovering, props }"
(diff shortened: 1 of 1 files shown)


michael-genson (approved):
🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉

PR #5611, opened 2025-06-30
fix: Refactor Stores and Fix Missing Public Cookbooks

michael-genson:

_(REQUIRED)_

Refactors the use store functions to accept `i18n` as a param rather than trying in instantiate it. This is useful because `useI18n()` must be done in the `setup` block.

Using this change, I fixed an issue where public cookbooks weren't loading initially because the `groupSlug` is not loaded in the route yet (since the layout runs its setup before the route loads); we now wait until `groupSlug` is a non-empty string, then create the public API (which would fail previously because we would no longer be in `setup`).

_(REQUIRED)_

N/A, noticed this while testing.

## Testing

Was able to reproduce the issue consistently by tearing down both the frontend and the backend, closing the page, then spinning everything back up before opening the page. The first load would consistently fail (subsequent loads usually wouldn't, though occasionally they could). Now the first load consistently works for both logged-in and logged-out users.


frontend/composables/api/api-client.ts
```diff
@@ -52,31 +53,34 @@ function getRequests(axiosInstance: AxiosInstance): ApiRequestInstance {
...
 }
 
-export const useRequests = function (): ApiRequestInstance {
-  const i18n = useI18n();
+export const useRequests = function (i18n?: Composer): ApiRequestInstance {
   const { $axios } = useNuxtApp();
+  if (!i18n) {
+    // Only works in a setup block
+    i18n = useI18n();
+  }
```
michael-genson, line 61: I think ideally we just don't do this, but fully decoupling the i18n instantiation is a lot more work and touches a lot more of the app. Doing it this way preserves old behavior, but lets us use the new behavior wherever we want.

Kuchenpirat (approved):
LGTM 👍

PR #5612, opened 2025-06-30
fix: AppButtonCopy errors in tooltip & console

p0lycarpio:

_(REQUIRED)_

This pull request refines the clipboard functionality in the `AppButtonCopy.vue` component to improve user feedback and simplify the code. The most important changes include adding a new `copiedSuccess` state, enhancing error logging, and modifying the tooltip behavior.

_(REQUIRED)_

N/A

Calling the `.deactivate()` method on the value no longer seems to be correct on Nuxt 3 and was causing an error in the console. Also, the error text was quickly displayed before the copy was actually made. This is why the reference was created.


Kuchenpirat (changes requested):
hey this looks pretty good.  
One small thing. 

V-Tooltip seems to have dropped color support on the migration to v3. Could you remove the color prop from the v-tooltip component and add this to the icon within the popup 

`:color="copiedSuccess ? 'success-lighten-2' : 'error-lighten-2'"`

so we have a bit of glancable color information on success / failure.

p0lycarpio:
The color on the tooltip has been removed. Having tried the color on the icon, it is less visible and unnecessary in my opinion. I increased the display time.

Kuchenpirat (approved):
lgtm

PR #5619, opened 2025-07-02
fix: Remove redundant get_one call in patch_one method

jknndy:
This PR removes an unnecessary call to `get_one()` in `patch_one` to avoid a redundant lookup, as `_pre_update_check` already returns the validated recipe.


michael-genson (approved):
Thank you!

PR #5647, opened 2025-07-06
feat: Button to select recipe cover image when creating recipe from multiple images

SkepticMystic:

Since support was added for creating a recipe from _multiple_ images, it would be convenient to be able to select which image is chosen as the cover image for the created recipe (currently the first image is chosen).

No existing issues

I initially added swapper buttons "<" and ">" to change the order of the list. But this was overkill when the goal is just to select which image is the cover.

## Testing

Tested on local dev. Uploaded multiple images where the cover wasn't the first image. Selected the intended cover, then created the recipe.

## Screenshot

image

- Chosen image has a disabled button with different text and icon
- Non-chosen image can be selected
- If number of uploaded images is < 2, no buttons are shown


Kuchenpirat:
could you add screenshots

SkepticMystic:
Added to OP.
I considered adding it the the Cropper, based on the styling changes you made recently. But felt it wasn't relevant enough to the component

michael-genson (changes requested):
Looks good! Just one suggestion: since we're swapping the cover image to be the first image after submitting, would it be better to do the swap before the user clicks submit? It looks kinda weird to see them re-arrange after submitting.

SkepticMystic:
Makes sense! I'll change that

SkepticMystic:
Lovely! Always happy when a commit is mostly removing code 😋

michael-genson (approved):
Thank you!

PR #5649, opened 2025-07-06
feat: Add plan entry type of snack

cmpickle:

This PR adds a new entry type of snack for snacks that are planned throughout the day that are not a part of any specific meal.

[#5648 ](https://github.com/mealie-recipes/mealie/discussions/5648)

## Testing

I added tests to the test_group_mealplan.py file to test using the new entrytype of snack.


michael-genson:
Would we be better off extending this to allow users to add arbitrary types, rather than adding another preset type? I know this comes up a lot, and something like how we handle other group data (foods, units, etc.) might be a better approach.

michael-genson:
Going to close this in favor of https://github.com/mealie-recipes/mealie/pull/6149 since there's more discussion there and has a few more options. Appreciate your work on this!

PR #5651, opened 2025-07-06
feat: Add delete option to the recipe card context menu for easier deletion of recipies

cmpickle:

Adds delete option to the context menu for recipe cards to make it easier for users to curate their collection of recipes.

[#5650 ](https://github.com/mealie-recipes/mealie/discussions/5650)

## Testing

Ensured that the context menu showed up in both web and mobile views.

image


Kuchenpirat:
We have decided to not include the delete option on the recipe explore page. 
See the discussion on #4842 

Thanks for your PR, but we wont include it at this time.

PR #5687, opened 2025-07-12
feat: Move create-item button in shopping list to the top

lehnerpat:

On the Shopping List page, the "Add" button (to create new shopping list items) is currently located _under_ the list of existing items. This means that every time you want to add a new item, you have to scroll all the way to the bottom.

This PR moves the "Add" button to be _above_ the list, making it easier for users to create new shopping list items because the button is visible right when navigating to this page.

| | Before | After |
|---|---|---|
| Desktop | CleanShot 2025-07-12 at 15 23 12@2x | CleanShot 2025-07-12 at 15 06 00@2x |
| Desktop (editor open) | CleanShot 2025-07-12 at 15 25 11@2x | CleanShot 2025-07-12 at 15 06 11@2x |
| Mobile | CleanShot 2025-07-12 at 15 23 32@2x | CleanShot 2025-07-12 at 15 05 43@2x |

https://github.com/mealie-recipes/mealie/discussions/5528 (discussion)

This is my first code contribution! Please let me know if there's anything I missed, or if there are concerns about this change.

Please let me know if you're okay with categorizing this as a "feat" or if it's something else, like a "fix".

An alternative to _moving_ the button could be to show the "Add" button _both_ above _and_ below the list, which might make the change less jarring if users are really used to the button being below the list.

## Testing

Manually, opening the shopping list page and seeing that the "Add" button is indeed at the top, and that the "new item" editor shows up above the list as well when clicking the "Add" button.


michael-genson (approved):
Yeah I like this much better, thanks!

PR #5699, opened 2025-07-12
fix: user & household creation

### LINKED ISSUE #5681: [BUG] - Household can not be selected on user creation form
### First Check

It is not possible to choose a household in the user creation form. This is reproducible in the demo.
`Failed to load resource: the server responded with a status of 422`

It might be an issue with localstorage / cache or whatever way you're using to locally store the household IDs, as a cache purge fixes the issue.

### Steps to Reproduce

Create a new household.
Try to create a new user.

Failed to load resource: the server responded with a status of 422

### Mealie Version

_No response_

### Deployment

Unraid

_No response_

Kuchenpirat:

Fixes user and household creation. 

- fixes #5681
- fixes broken Household deletion

This was a suuuper weird thing to track down. My best guess is that both used the same asyncKey that was based on time which resulted in both hitting the same cache. I simplified that function to just use a normal api call and refactored some stuff because it took me quite some time to get this :D

## Testing

Local. Could no longer reproduce the error.

### DIFF
--- a/frontend/composables/use-households.ts
+++ b/frontend/composables/use-households.ts
@@ -103,7 +103,7 @@ export const useAdminHouseholds = function () {
   return {
     households,
     useHouseholdsInGroup,
-    getAllGroups: getAllHouseholds,
+    getAllHouseholds,
     refreshAllHouseholds,
     deleteHousehold,
     createHousehold,
--- a/frontend/composables/use-households.ts
+++ b/frontend/composables/use-households.ts
@@ -48,28 +48,11 @@ export const useHouseholdSelf = function () {
 export const useAdminHouseholds = function () {
   const api = useAdminApi();
   const loading = ref(false);
+  const households = ref<HouseholdInDB[] | null>(null);
 
-  function getAllHouseholds() {
+  async function getAllHouseholds() {
     loading.value = true;
-    const asyncKey = String(Date.now());
-    const { data: households } = useAsyncData(asyncKey, async () => {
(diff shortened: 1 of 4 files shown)


frontend/composables/use-households.ts
```diff
@@ -109,10 +96,14 @@ export const useAdminHouseholds = function () {
...
 
+  if (!households.value) {
+    getAllHouseholds();
+  }
+
   return {
     households,
     useHouseholdsInGroup,
-    getAllHouseholds,
+    getAllGroups: getAllHouseholds,
```
michael-genson: Is this right? Or should this still be getAllHouseholds?

michael-genson (changes requested):
Looks good, one question on the households composable.

Looking at this I'm scratching my brain going "Why the heck were we doing it that way? This makes way more sense."

michael-genson:
Also this might be the issue I was facing with cookbooks back in nightly which [I "fixed" by removing an old composable](https://github.com/mealie-recipes/mealie/pull/5570)

Yup looks like I deleted a few `useAsyncKey` references in that PR, I bet that was it

frontend/composables/use-households.ts
Kuchenpirat: oh yeah, this is what you get when you want an AI to apply the changes of one file to another

Kuchenpirat:
> Looking at this I'm scratching my brain going "Why the heck were we doing it that way? This makes way more sense."

yeah i was confused too, after i saw that groups already contained a list of their households. 
Maybe this wasnt the case from the start.

michael-genson (approved):

PR #5709, opened 2025-07-13
fix: dash slug names

### LINKED ISSUE #5532: [BUG] - recipe with name '---' can't be edit
### First Check

hello 
for a test i have created a recipe with name ---.
The recipe is available/visible in the recipes page but i can't access to it. So impossible to delete or update it. I delete it manuelly with sqlite client.

### Steps to Reproduce

1- create a recipe
2- in the name type ---
3- save the recipe
4-a 404 page is displayed
5-in the recipe page you can't do anything with --- recipe

n/a

### Mealie Version

2.6.0

### Deployment

Docker (Linux)

_No response_

fernandom06:

- Check slug in front to show error if is neccessary
- Default slug in backend if generated slug is empty

Before

image

After

image

image

Fixes https://github.com/mealie-recipes/mealie/issues/5532

I have created a function to generate the slug in the frontend to be able to check it before sending it to the backend, I don't know if I should use some library like in the backend

## Testing

Tested locally, w/Chrome dev tools for mobile UX.


michael-genson (changes requested):
Since we already do basically this on the backend, I think it would be better to keep the logic there and throw an error that we catch on the frontend. That way we don't duplicate the slug logic in both the frontend and backend.

We could even do a custom error message like "unable to generate slug" which the frontend could read. We already have some error messages that we directly show the user using i18n on the backend (though not many). 

e.g. from LDAP:
```python
if self.user.password == "LDAP" or self.user.auth_method == AuthMethod.LDAP:
    raise HTTPException(
        status.HTTP_400_BAD_REQUEST, ErrorResponse.respond(self.t("user.ldap-update-password-unavailable"))
    )
```

This will automatically flash an error to the user on the frontend (we have a custom plugin for this).

fernandom06:
> Since we already do basically this on the backend, I think it would be better to keep the logic there and throw an error that we catch on the frontend. That way we don't duplicate the slug logic in both the frontend and backend.
> 
> We could even do a custom error message like "unable to generate slug" which the frontend could read. We already have some error messages that we directly show the user using i18n on the backend (though not many).
> 
> e.g. from LDAP:
> 
> ```python
> if self.user.password == "LDAP" or self.user.auth_method == AuthMethod.LDAP:
>     raise HTTPException(
>         status.HTTP_400_BAD_REQUEST, ErrorResponse.respond(self.t("user.ldap-update-password-unavailable"))
>     )
> ```
> 
> This will automatically flash an error to the user on the frontend (we have a custom plugin for this).

Now when slug is invalid raise a SlugError, that I create in core.exceptions and raise a HTTPException in RecipeController class in recipe_crud_routes.py file

michael-genson (changes requested):
Looks good, except for the language files; translations are handled through Crowdin. You should only add to/update the `en-US.json` file. Everything else shouldn't be modified here

michael-genson (approved):
I went ahead and finished this one up. I ended out removing the custom message and just stuck with the existing flow. The user can always check the logs if they're confused, but at least now it won't create an invalid recipe.

Thanks for your work on this!

fernandom06:
I'm sorry the past few months have been really crazy. Please forgive my absence.

I think your solution is much cleaner and better suited for such a rare case.

Thank you so much for the opportunity to collaborate on this wonderful project.

michael-genson:
No worries, appreciate your work on this!

PR #5715, opened 2025-07-14
fix: Update admin_backups.py to handle API backup file uploads correctly.

duckhead:
The return was missing from the upload_one function, so I added one.

Before this change, files would successfully upload to Mealie via the API, but then would report failure, because the correct object to indicate success was never returned.

This was a bug found in my internal testing and reported on Discord.

I didn't generate any tests for this, because I really don't know python, but I can handle little changes like this for the most part. Given that this bug has likely been present for some time, I think a test would be a great idea. I can work with someone to make that happen, or let someone else take over.

## Testing

I tested this with my application to automatically backup from one Mealie instance to another. This code failed with a 500 error before, but it succeeds now with a 200 status code, as expected. I haven't quite gotten the backup to apply, but I suspect it's an issue in my own code, not with mealie. And if it's not, I'll submit under a new PR, because it's different functionality.


michael-genson (approved):
Good find, thanks!

PR #5716, opened 2025-07-14
fix: Added copy icons to first-login message

### LINKED ISSUE #5551: [BUG] - Nuxt-Frontend: Its allowed to enter "spaces" in username/email and passwort
### First Check

I'm not sure if it's supposed to be like this, but it's “stupid” if you copy out the password above and then copy the space with it.

Another idea would be to equip these fields with a copy flag, like on our nextjs/nuxt website (example: https://community-scripts.github.io/ProxmoxVE/scripts?id=planka) 

Screenshot for example: 
Image

### Steps to Reproduce

Image

Image

-

### Mealie Version

current git clone build

### Deployment

Other (please specify below)

Proxmox with current git clone

RichardVanLeeuwen:
Goal:
Prevent mistakes when copy/pasting the first-login info from the info box to the input fields.
Allows copying the email and password by clicking the clipboard icon.
~~Up for discussion: also puts the info in the inputs. This behaviour is different from what a user expects based on experience with the web. Is this a good idea? If this, is it still neccesary to copy to the clipboard. Perhaps the clipboard can be a button with the text "copy to input"?~~

before:
20250714_22h23m46s_grim
after:
20250714_22h22m20s_grim

Fixes #5551


Kuchenpirat:
Hi, thanks for this PR. 
While i think this makes sense, we already have logic implemented to copy to the clipboard. 

Take a look at the AppCopyButton for the correct copy symbol. You might even be able to just use the button, don't have design of it on hand currently. If not we have a use-copy composable that includes the required logic to copy text to the clipboard.

RichardVanLeeuwen:
Thanks for looking at this. I refactored to use the AppCopyButton, works perfect!
This is how it looks now:
20250722_22h02m24s_grim
I chose to move the buttons over to the right, similar to RecipeIngredients does it. The flex broke the whitespace between the bolded part and non-bolded part, the `&nbsp;` is used as non-collapsible whitespace so it look correct again.
Also modified the height of the icons a bit so they align better with the text.

Kuchenpirat:
This is still marked as draft, anything you still want to update, or is this ready for review?

RichardVanLeeuwen:
Sorry forgot I set that due to the questions.

michael-genson (changes requested):
Generally looks good, but I'm not sure we want the copy button to be all the way to the right:
image

I'd like to see them next to the text (still aligned like they currently are). e.g. (using photoshop)
2025-08-07_10h47_33

Or maybe the alignment would look better with how you used to have it (right next to the text, not aligned with each other). Not sure, open for suggestions there

RichardVanLeeuwen:
Not aligning to the right was easy enough, even code-wise it's very clean
20250810_22h20m12s_grim
Would you agree with this?

michael-genson (approved):
Yup looks great, thank you!

PR #5736, opened 2025-07-16
fix: style of recipe actions to be compliant with design schema

### LINKED ISSUE #5734: [BUG] - Recipe Action menu doesn't follow design scheme
### First Check

When viewing a recipe and then clicking on the three dot menu the recipe actions look _off_:
- The text has no margin to the left
- The text is smaller than the other text
- The curser does not switch to a click style
- There is no `>` that turns when the menu is expanded

Image

This changed with the switch to the new version of Nuxtv3

### Steps to Reproduce

0. Have a recipe action defined 
1. Open any recipe
2. Open the three dot menu on the right

### Mealie Version

Version
v3.0.1

Build
[9ec15994276fb6784c59e83b92ab9a12b8c7ce8d](https://github.com/mealie-recipes/mealie/commit/9ec15994276fb6784c59e83b92ab9a12b8c7ce8d)

### Deployment

Docker (Linux)

_No response_

felixschndr:

The design of the recipe actions in the three dot menu did not look good (not like the other points) and it was unclear that is is possible to click on the text to show the actions. This fixes it:
- The menu for actions is always visible now (if there is at least one recipe action 
- The text is aligned properly
- Removed the header
- The actions have an icon (always the same, the same one as in `/group/data/recipe-actions`), changed that icon as discussed below to `mdiLinkVariantPlus` 

------

Old design:
image
image

New design:
image

image

the menu is always visible now (if there is at least one recipe action and the text is aligned properly

Fixes https://github.com/mealie-recipes/mealie/issues/5734

## Testing

`task ui:test` runs successfully

### DIFF
--- a/frontend/components/Domain/Recipe/RecipeContextMenu.vue
+++ b/frontend/components/Domain/Recipe/RecipeContextMenu.vue
@@ -125,11 +125,6 @@
         </v-list-item>
         <div v-if="useItems.recipeActions && recipeActions && recipeActions.length">
           <v-divider />
-          <v-list-item disabled>
-            <v-list-item-title>
-              {{ $t("recipe.recipe-actions") }}
-            </v-list-item-title>
-          </v-list-item>
           <v-list-item
             v-for="(action, index) in recipeActions"
             :key="index"
@@ -137,7 +132,7 @@
           >
             <template #prepend>
               <v-icon color="undefined">
-                {{ $globals.icons.primary }}
+                {{ $globals.icons.linkVariantPlus }}
               </v-icon>
             </template>
             <v-list-item-title>
--- a/frontend/components/Domain/Recipe/RecipeContextMenu.vue
+++ b/frontend/components/Domain/Recipe/RecipeContextMenu.vue
@@ -125,25 +125,25 @@
         </v-list-item>
         <div v-if="useItems.recipeActions && recipeActions && recipeActions.length">
           <v-divider />
-          <v-list-group @click.stop>
-            <template #activator="{ props }">
-              <v-list-item-title v-bind="props">
-                {{ $t("recipe.recipe-actions") }}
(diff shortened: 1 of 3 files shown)


Kuchenpirat (commented):
Hey, thanks for the catch and the PR. 

I suggest using another icon. I dont think we should use the mealie icon for this. 

I would suggest: [mdiLinkVariantPlus](https://pictogrammers.com/library/mdi/icon/link-variant-plus/)

You can add new icons in the frontend/lib/icons/icons.ts file. 
Would also make sense to update the data management page then.

felixschndr:
Okay, I will change that.
I also suggest removing the header "Recipe actions" (just show the divider and then the actions) since the text is kinda out of place and not that _nice to look at_. What do you think about that?

Kuchenpirat:
yeah, makes sense to me.

Kuchenpirat:
@felixschndr if you have a better icon in mind or come across a better fitting one, im absolutely open for suggestions

felixschndr:
> if you have a better icon in mind or come across a better fitting one, im absolutely open for suggestions

No, I really like the `mdiLinkVariantPlus`, I think it fits in great. 

I changed the icon and updated the design at the top of the MR. Thus, this MR is ready to be merged from my side :)

Kuchenpirat (changes requested):
Looks good so far, 
One minor thing that i catched. 

The recipe actions create and edit dialogs still use the primary icons.

felixschndr:
Good catch, fixed that

Kuchenpirat (changes requested):
please also remove the package-lock.json, package.json and yarn.lock from the pr.

felixschndr:
They slipped in there by accident, removed them

Kuchenpirat (approved):

PR #5737, opened 2025-07-16
fix: send the correct value for recipe scale and ensure the body is sent correctly

felixschndr:

This is a followup of https://github.com/mealie-recipes/mealie/pull/5340

The original MR had two flaws:
1. It sent the wrong value. For the use case of updating the quantity in the Mealie Bring API (see https://github.com/felixschndr/mealie-bring-api/issues/15), the recipe scale is actually needed.
2. The value was not sent correctly to the backend; it never received it. Thus, the scaled amount was always `1`.
    Fixed in https://github.com/mealie-recipes/mealie/pull/5737/files#diff-e695608c33d2ecfa3261e01677242602676ca814f366be97b4628d8ea421a30dR17

The issues mentioned above

## Special notes for your reviewer/Testing

For the original MR I only guessed what could be correct.
This time I took the time to setup the mealie frontend and backend up locally and test the functionality end-to-end, from the frontend via the backend to the mealie bring api.


felixschndr:
This is ready to be merged and and I already implemented support for the feature on the other side (the mealie bring api)

felixschndr:
@Kuchenpirat this is a __friendly__ ping, can we get this merged? I'd really like this functionality and the feature request has been open for more than six months now.
The real changes were reviewed in the original MR and this is only replacing the values that is sent.

Kuchenpirat:
Sure thing, I'll try to get a review round in tomorrow for this (if the internet on Deutsche Bahn is on my side) if not it'll be the day after.

felixschndr:
From my experience: _Thursday it is_ :D

Kuchenpirat (approved):

PR #5739, opened 2025-07-17
fix: make only checkbox reactive

### LINKED ISSUE #5423: [BUG] - UX problem shopping list: Only checkbox to tick off
### First Check

The shopping list is somewhat inconvenient and sometimes a hindrance in real-world use. Especially on a smartphone, the problem is that the entire entry serves as a label for the checkbox. Especially when scrolling around, it is easy to tick an entry and sometimes not even notice it because you have your finger in the middle of the screen. From a UX perspective, only the checkbox itself should be reactive and not the entry itself.

Image

### Steps to Reproduce

1. go to the shopping list
2. click on the entry (not the checkbox)
3. the entry is immediately marked as completed and disappears

None.

### Mealie Version

Version 2.8.0

### Deployment

Docker (Synology)

_No response_

mariodz95:

Make only the checkbox reactive, so clicking on the item name will not check the item.

Fixes #5423 

There is also a problem when quickly clicking on items, some items reappear in the list as unchecked, so they can be checked again. I'm new to Vue, but I will try to fix that as well.


Kuchenpirat (approved):
LGTM 👍
Thanks for the contribution

Issue #5477, opened 2025-05-26
[BUG] - Cloudflare/ Nginx Proxy Manager / Authentik - CSRF Error


m-hesse:
### First Check

- [x] This is not a feature request.
- [x] I added a very descriptive title to this issue (title field is above this).
- [x] I used the GitHub search to find a similar issue and didn't find it.
- [x] I searched the Mealie documentation, with the integrated search.
- [x] I already read the docs and didn't find an answer.
- [ ] This issue can be replicated on the demo site (https://demo.mealie.io/).

### What is the issue you are experiencing?
CSRF-Issue on logging in using Authentik.

My setup is:
Cloudflare <--> Nginx Proxy Manager <--> Authentik <--> Mealie

I've configured Mealie, corresponding to:
https://docs.mealie.io/documentation/getting-started/authentication/oidc-v2/#mealie-setup

#### Authneik config
Authentik is also configured as following:
Authorization flow: default-provider-authorization-implicit-consent (Authorize Application)
Client type: Confidential
Redirect URIs/Origins (RegEx): 
- Strict: https://mealie.domain.com/, https://mealie.domain.com/login, https://mealie.domain.com/login?redirect=1, 
- Regex: https://mealie.domain.com/login*

All of them with https and http.

I've checked a lot of issues here, especially https://github.com/mealie-recipes/mealie/issues/4482
but that's not feasible for me due to the docker image i really like to use.

mealie docker-compose.yaml
```
version: "3.7"
services:
  mealie:
    image: ghcr.io/mealie-recipes/mealie:latest
    ports:
        - "9925:9000"
    depends_on:
      - postgres
    volumes:
      - mealie-data:/app/data/
      - ./backups:/app/data/backups
    networks:
      intern:
      network_self_hosted:
    environment:
    # Set Backend ENV Variables Here
      - ALLOW_SIGNUP=true
      - PUID=1000
      - PGID=1000
      - TZ=europe/Berlin
      - MAX_WORKERS=1
      - WEB_CONCURRENCY=2
    # Database Settings
      - DB_ENGINE=postgres
      - POSTGRES_SERVER=${DB_POSTGRESDB_HOST}
      - POSTGRES_PORT=${DB_POSTGRESDB_PORT}
      - POSTGRES_DB=${DB_POSTGRESDB_DATABASE}
      - POSTGRES_USER=${DB_POSTGRESDB_USER}
[...]

PR #5746, opened 2025-07-18
feat: Allow using OIDC auth cache instead of session

hr-kapanakov:

Add a new option to allow using cache instead of session to store OIDC temporary data.

By default, the authlib OAuth registry will use session to store OAuth temporary credential (request token). However, in this way, there are chances your temporary credential will be exposed.
Enabling this setting, the temporary credentials will be stored in a cache system.

https://github.com/mealie-recipes/mealie/issues/5477
Maybe other OIDC issues as well, which are caused by a lost session.

This change will allow using OIDC with http as well.

## Testing
Built Docker image and ran it with the enabled setting, and verified the OICD login workflow change.


cmintey:
Any reason why we wouldn't just make this the default method and not have this be a configurable option?

michael-genson:
Agreed with cmintey, makes sense to just make this default behavior.

I know this is server-side, but I don't think we should use pickle here, since it can execute arbitrary Python code. While this implementation is _probably_ fine, I think it's better to just avoid it. There are other options that work just as well (keep the original object, use a JSON string, etc.)

hr-kapanakov:
Sure, make sense. Removed the setting and pickle.

michael-genson (dismissed):
Generally looks good, two things:
1. Can you add type hints to the cache class? e.g. `def _normalize_timeout(self, timeout: int):`
2. Is it possible to add tests for this? Just a simple end to end test to make sure it works

If you're not comfortable with either of those let me know, happy to help!

hr-kapanakov:
@michael-genson I just added the type hints, but I am not confident about the end to end test, so I will appreciate your help.

michael-genson:
LGTM, thanks! I added some tests, another maintainer will have to approve since I was the last pusher

cmintey (approved):
My approval won't count towards anything, but it all looks good to me

Kuchenpirat (approved):

Issue #5437, opened 2025-05-11
[BUG] - Admin User Couldn't Delete Other Household Recipes


coastlee:
### First Check

- [x] This is not a feature request.
- [x] I added a very descriptive title to this issue (title field is above this).
- [x] I used the GitHub search to find a similar issue and didn't find it.
- [x] I searched the Mealie documentation, with the integrated search.
- [x] I already read the docs and didn't find an answer.
- [ ] This issue can be replicated on the demo site (https://demo.mealie.io/).

### What is the issue you are experiencing?

I had a user in a different household begin creating a recipe and then immediately abandon effort; it created a few "No Recipe Name Found - {{GUID}}" records in my data. If I tried to delete the record (as an admin user), I was told I didn't have permissions.

Switching my household did enable me to then delete the record.

### Steps to Reproduce

1. Be an admin user
2. Have multiple households in your instance
3. Have a user in a household different from yourself create a recipe
4. Try to delete the recipe as your different-household-admin-user

### Please provide relevant logs

[Discord request](https://discordapp.com/channels/795445547577770014/795489645227671584/1370806424686432286) from Michael. Too lazy to go look this up, but can if it's absolutely necessary.

### Mealie Version

[d2ced50fcceceb35e742765d99fd76378284f228](https://github.com/mealie-recipes/mealie/commit/d2ced50fcceceb35e742765d99fd76378284f228)

### Deployment

Docker (Linux)

mariodz95:
If I understand correctly, an admin can delete a recipe, any recipe?
For example, I have a user in household 1 who creates a recipe, then I log in with the admin user who is in household 2, should he be able to delete the recipe created by user 1?

mariodz95:
After investigating, I found that there's an option when editing a household called "Lock recipe edits from other households." If you uncheck this option, you're able to delete the recipe from other households.

coastlee:
Yes, but then that's not an admin privilege. I don't really want other households editing other households recipes. But I do want to be able to delete recipes as the admin.

klacointe:
> Yes, but then that's not an admin privilege. I don't really want other households editing other households recipes. But I do want to be able to delete recipes as the admin.

I have the same need (edit and delete) 💯

michael-genson:
This was fixed by https://github.com/mealie-recipes/mealie/pull/5767 but there was a typo so it didn't get auto-closed

klacointe:
[...]

PR #5767, opened 2025-07-21
fix: allow admin users to delete other household recipes

mariodz95:

Allow admin users to delete other household recipes.

Fiexes #5437 

<!--
If this PR fixes one of more issues, list them here.
One per line, like so:
Fixes #123
Fixes #39


michael-genson (changes requested):
This allows admins to _edit_ recipes belonging to other households, which I don't think we want. Similarly, just below this, we prevent admins from editing locked recipes:

```python
if recipe.settings.locked:
    return False
```

Should we instead make a separate `can_delete` method which is less restrictive to admins? Something like:
```python
if self.user.admin:
    return True
else:
    return self.can_update(recipe)
```

We should also probably add a warning to the frontend that says something like "You're about to delete a recipe that isn't yours using admin permissions. Are you sure?". I can see confusion stemming from this, especially for smaller instances where users might not even know they're admins.

mariodz95:
Okay, so admins shouldn't be able to delete other household recipes if they are locked.
And if household recipes are unlocked, admin users can delete them, but non-admin users cannot? Only user who owns recipe.
Just to make sure we're on the same page.
Also, is there already some kind of modal I can follow as a pattern?

michael-genson:
That's the current behavior, but as long as we throw in a warning to the admin user I think it would be okay to let them delete a locked recipe (they are an admin, after all).

There's already a delete confirmation dialog (see: `RecipeActionMenu`). You would basically want to add a warning banner only if the user is an admin:

```js
<BaseDialog v-model="deleteDialog" :title="$t('recipe.delete-recipe')" color="error"
  :icon="$globals.icons.alertCircle" can-confirm @confirm="emitDelete()">
  <!-- This is the new part -->
  <BannerWarning v-if="isAdmin">
    {{ $t("recipe.some-new-string-added-to-enUS.json" }}
  </BannerWarning>
  <v-card-text>
    {{ $t("recipe.delete-confirmation") }}
  </v-card-text>
</BaseDialog>
```

Might have to tweak the style a bit.
If you haven't added a new language string before, basically just add it manually to the `en-US.json` file. We have automations that handle all other languages.

mariodz95:
> That's the current behavior, but as long as we throw in a warning to the admin user I think it would be okay to let them delete a locked recipe (they are an admin, after all).
> 
> There's already a delete confirmation dialog (see: `RecipeActionMenu`). You would basically want to add a warning banner only if the user is an admin:
> 
> ```js
> <BaseDialog v-model="deleteDialog" :title="$t('recipe.delete-recipe')" color="error"
>   :icon="$globals.icons.alertCircle" can-confirm @confirm="emitDelete()">
>   
>   <BannerWarning v-if="isAdmin">
>     {{ $t("recipe.some-new-string-added-to-enUS.json" }}
>   </BannerWarning>
>   <v-card-text>
>     {{ $t("recipe.delete-confirmation") }}
>   </v-card-text>
> </BaseDialog>
> ```
> 
> Might have to tweak the style a bit. If you haven't added a new language string before, basically just add it manually to the `en-US.json` file. We have automations that handle all other languages.

Added new warning message if an admin tries to delete another household recipe, and removed delete option if user is not allowed to delete it.

JPorter-02:
Would it also be worth considering that the admin should be able to edit/delete locked recipes across the platform, unless the household has recipes locked? For example if a user account has been deleted but still has recipes assigned to it?

michael-genson (changes requested):
Looks good! Can we add backend tests for this? Just the basic test cases:
- admin can delete a locked recipe that isn't theirs
- admin can delete another household's recipe

And we should already have tests for these, but double check (and add them if they're missing):
- user can delete their own locked recipe
- user cannot delete someone else's locked recipe
- user can delete their own household recipe
- user cannot delete another household's recipe

If you're not sure how to check/add these let me know and I'm happy to help

michael-genson:
> Would it also be worth considering that the admin should be able to edit/delete locked recipes across the platform

@JPorter-02 I think this would be a good change for sure, but not necessarily for this PR

mariodz95:
> Looks good! Can we add backend tests for this? Just the basic test cases:
> 
> * admin can delete a locked recipe that isn't theirs
> * admin can delete another household's recipe
> 
> And we should already have tests for these, but double check (and add them if they're missing):
> 
> * user can delete their own locked recipe
> * user cannot delete someone else's locked recipe
> * user can delete their own household recipe
> * user cannot delete another household's recipe
> 
> If you're not sure how to check/add these let me know and I'm happy to help

Just to clarify, since the test structure doesn't seem to follow the project structure, should these tests be written in test_recipe_crud.py, or is it acceptable to add a new test_recipe_service.py file? From what I can see, most of the tests you've written appear to be missing.

michael-genson:
For the cross-households tests, we already have tests under `test_recipe_cross_household.py`. We can add onto those.
For the non-household tests (e.g. admin can delete a locked recipe) check out `test_recipe_owner.py`. We can add onto those with admin-specific tests.

mariodz95:
> For the cross-households tests, we already have tests under `test_recipe_cross_household.py`. We can add onto those. For the non-household tests (e.g. admin can delete a locked recipe) check out `test_recipe_owner.py`. We can add onto those with admin-specific tests.

Ok will check those files and try to cover all cases. That way looks much more complicated to me than if all the tests were in one file and followed the project structure. Can you give me more informations about groups and households?

michael-genson:
You can check out the docs regarding groups and households: https://docs.mealie.io/documentation/getting-started/features/#groups-and-households

> That way looks much more complicated to me than if all the tests were in one file and followed the project structure

Yeah our test structure is pretty messy. It's a consequence of iterative additions. Households were added pretty recently, as opposed to other core features which have been around since basically forever.

michael-genson (approved):
LGTM - thanks for your work on this!
I moved a couple tests around to better fit the existing test structure and catch a couple edge cases. Another maintainer will have to approve this one.

Kuchenpirat (approved):
Lets get this is then 👍

klacointe:
What about editing of other houhsehols recipes as an admin ? Is this abandoned ?

PR #5768, opened 2025-07-21
fix: cookbook random recipe selector

### LINKED ISSUE #5362: [BUG] - Cookbook Random Recipe Selector doesn't work
### First Check

On desktop browser (unsure on other platforms) random recipe selectors does not randomly select any recipe.

### Steps to Reproduce

1) Create a cookbook containing recipes
2) Select the Random icon in the top right 

N/A

### Mealie Version

V2.8.0

### Deployment

Docker (Linux)

Docker using CasaOS

mariodz95:

Fix random recipe selector in cookbook.

Fixes #5362

### DIFF
--- a/frontend/components/Domain/Recipe/RecipeCardSection.vue
+++ b/frontend/components/Domain/Recipe/RecipeCardSection.vue
@@ -414,8 +414,15 @@ export default defineNuxtComponent({
       loading.value = false;
     }
 
+    const seededQuery = computed(() => {
+      return {
+        ...props.query,
+        _searchSeed: Date.now().toString()
+      };
+    });
+
     async function navigateRandom() {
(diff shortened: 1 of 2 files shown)


michael-genson (changes requested):
Could we instead move this to the `getRandom` function? Since `getRandom` returns a single recipe the seed doesn't actually matter, so it makes sense for the function do to this directly. Saves a bit on performance as well since we're not using a computed prop.

Something like:

```ts
async function getRandom(query: RecipeSearchQuery | null = null, queryFilter: string | null = null) {
  query = query || {};
  query._searchSeed = query._searchSeed  || Date.now().toString();
  ...
}
```

mariodz95:
Refactored as you suggested.

michael-genson (approved):
Thank you!

PR #5783, opened 2025-07-22
fix: Reduce margin in RecipePageInstructions

felixschndr:

Fixes spacing of the ingredient linker. Since Nuxt3 there was a lot of space between ingredients, which becomes a problem for recipes with many ingredients and/or lower resolution screens.

Before:
image
image
image

After:
image
image
image

-

## Testing

`task ui:lint` runs sucessfully


Kuchenpirat (commented):
You can set the density prop on v-checkbox to compact, so you don't have to use negative margins

https://vuetifyjs.com/en/components/checkboxes/#api

felixschndr:
@Kuchenpirat that does not seem to work as I would expect it. This is with `density="compact"` instead of `class="mb-n10 mt-n2"`
image

felixschndr:
Found the solution [here](https://github.com/vuetifyjs/vuetify/issues/15577#issuecomment-1208312444) and implemented it. I updated the images above. Ready to be reviewed @Kuchenpirat

Kuchenpirat (changes requested):
Lets remove the negative margings from the checkbox components. The elements are a bit too close together for touch screen usage.

felixschndr:
Removed the margins and updated the screenshots above.

Kuchenpirat (approved):
LGTM 👍
Thanks for the contribution

PR #5784, opened 2025-07-22
feat: change recipe sorting default to alphabetical order instead of `created_at`

felixschndr:

This PR changes the default sorting from when the newest recipe was created to an alphabetical order. This is much more intuitive (in my opinion) as

- The list of recipes is a list. I expect lists to be sorted by number/name
- Sorting by the newest recipe is not senseful, since the user probably cooked that recipe lately since he added it to mealie afterwards. In this case he wouldn't want to re-cook it now. _I don't know if this sounds senseful, but I think alphabetical is just the logical ordering_

Before:
image

After:
image

-

### DIFF
--- a/frontend/components/Domain/Recipe/RecipeExplorerPage.vue
+++ b/frontend/components/Domain/Recipe/RecipeExplorerPage.vue
@@ -300,8 +300,8 @@ export default defineNuxtComponent({
 
     const queryDefaults = {
       search: "",
-      orderBy: "created_at",
-      orderDirection: "desc" as "asc" | "desc",
+      orderBy: "name",
+      orderDirection: "asc",
       requireAllCategories: false,
       requireAllTags: false,
       requireAllTools: false,


Kuchenpirat:
I am not sure this would improve the UX, at least for me i prefer having the newest recipes up top, instead of alphabetical sorting, which would show me almost always the same recipes. 
Also, i quite frequently add recipes i want to make in the future, which i would want to be at the top of the list.

I think a better way to improve the UX around this is to make the setting persistent, so the user would not have to change it constantly. 

I'll close this for now, but am happy to hear arguments why aphabetical would be better than created at

felixschndr:
Can we make this user configurable somewhere? If so, where would you expect this setting to be?

Kuchenpirat:
I think the best way would be to keep the default the same but persist the setting by using [useLocalStorage](https://vueuse.org/core/useLocalStorage/) to save it to the browser local storage. 
With this we would not need to add new ui elements and just improve what we already have. 

We could also add this database as user setting, but i think this is not necessairy for this setting.

felixschndr:
Replaced by https://github.com/mealie-recipes/mealie/pull/5826

PR #5785, opened 2025-07-22
fix: add confirm button to bulk delete of tags

felixschndr:

Brings back the delete confirm button of bulk deleting tags.

Follow-up of e638df37

Before: 
image

After:
image


Kuchenpirat:
the same error is also on tools, can you add that too this PR too?

felixschndr:
Dont know how I missed that, I thought I checked every component. Added it now

Kuchenpirat (approved):

PR #5816, opened 2025-07-27
fix: Make Sure Test Webhook Always Fires

### LINKED ISSUE #5521: [BUG] - Test Webhook Button No Longer Works (Nuxt Upgrade)
### First Check

I saw in the features that this includes the ability to configure notifications and Webhooks, but I see no where to configure that.

### Steps to Reproduce

1. Login with OIDC as an admin user
2. Check settings (nothing about notifications/webhooks)
3. Click everywhere looking for webhooks or notifications
4. Read the docs (find no docs on configuring the Webhooks)
5. Look at the environment variables (nothing for configuring webhooks)
6. Make sure the active user is an admin (check)
7. Shamefully post a bug ticket when there probably isn't a bug but just a confused foodie

No logs relevant here

### Mealie Version

2.8.0

### Deployment

Docker (Linux)

OIDC via Authentik

michael-genson:

_(REQUIRED)_

Webhooks only fire if there are mealplans within range. This is generally good, but even our test webhooks were constrained by this. This PR does a few things to address this:
1. Adjust the webhook logic to send webhooks if the body `is not None`
2. Add an empty list as the body for test webhooks (so it matches the signature of meal plan webhooks)
3. Use the `test_message` event type for posting a test message, rather than `mealplan`, and change the `post_single_webhook` method to be `post_test_webhook` to reflect this change (it wasn't used for anything aside from this endpoint anyway)

While testing, I also found it annoying that new webhooks are disabled by default. Generally I think it's better UX to enable webhooks by default (I just created it, why would I want it to be disabled?).

_(REQUIRED)_

Fixes https://github.com/mealie-recipes/mealie/issues/5521

## Testing

Added a test for the service


Kuchenpirat (approved):
LGTM

PR #5826, opened 2025-07-28
feat: save default recipe ordering in local storage

felixschndr:

Saves the default recipe orderning (field to sort by and direction) in the local storage of the browser.
Follow-up of https://github.com/mealie-recipes/mealie/pull/5784#issuecomment-3126950950

See the changes on the right
image
image
image

-

## Testing

- Tested manually, see screenshots above
- `task ui:lint` runs successfully


Kuchenpirat (changes requested):
Oh cool, i didn't know we had something already in place. 

This looks mostly good to go, only thing is that this will change the default to name and ascending while the current default is created_at & descending.

Kuchenpirat (changes requested):
Also when clicking the mealie symbol again the current choice is cleared because the state is encoded in the page query which will be removed when clicking the mealie symbol from the main page. 
We should probably remove this from the query and dont reset it when clearing the other search params by clicking on the mealie icon.

felixschndr:
Okidoki, I will implement the requested changes in the upcoming days

felixschndr:
> This looks mostly good to go, only thing is that this will change the default to name and ascending while the current default is created_at & descending.

Is that so? I can't find where I should have changed that. Can you point me to a line in the change of the MR?

Kuchenpirat:
This would come from the composable. 

https://github.com/mealie-recipes/mealie/blob/mealie-next/frontend/composables/use-users/preferences.ts#L99

felixschndr:
> This looks mostly good to go, only thing is that this will change the default to name and ascending while the current default is created_at & descending.
> This would come from the composable.
> https://github.com/mealie-recipes/mealie/blob/mealie-next/frontend/composables/use-users/preferences.ts#L99

Changed that

-----

> We should probably remove this from the query and dont reset it when clearing the other search params by clicking on the mealie icon.

@Kuchenpirat Since I have very few experience with frontend I have some difficulties to understand you^^.
You want that the recipe ordering is not present in the url (e.g. `g/home?orderBy=name`) when selecting an ordering, right? So Mealie should always order by what the local storage demands and not care (since it will not be present any more) about the query parameter?

Kuchenpirat:
> @Kuchenpirat Since I have very few experience with frontend I have some difficulties to understand you^^. You want that the recipe ordering is not present in the url (e.g. `g/home?orderBy=name`) when selecting an ordering, right? So Mealie should always order by what the local storage demands and not care (since it will not be present any more) about the query parameter?

Yeah, i think there are only very little advantages of having this encoded in a query parameter. Only thing that is an advantage is that one can share a link including the sort setting, but i dont think this is a frequent usecase. (if you have more thoughts about this, happy to hear if i am missing something)
I havent looked at all into how much of a lift this would be, but reseting the sort by clicking on the mealie icon is imho against the spirit of this PR.

felixschndr:
> if you have more thoughts about this, happy to hear if i am missing something

I totally agree with you.

----

I implemented the requested change:
- The query parameters `orderBy` and `orderDirection` are gone.
- Instead the ordering is stored in the local storage. This is persistent when navigating through different pages and coming back to the recipe overview, clicking on the Mealie icon, reloading the page and logging out and back in as long it is in the same browser.

Actually the changes are quite small.
Do you approve the implementation?

Kuchenpirat (approved):
Did some testing and LGTM 👍

PR #5829, opened 2025-07-28
fix: Nuxt 3 Ingredient Parsing Issues and Tooltip Positions

### LINKED ISSUE #5761: [BUG] - Several Issues with Parsing
### First Check

When parsing, several things are broken from the Nuxt upgrade:
- "Insert above" and "Insert below" have no effect
- When a food isn't found, the parser still puts the parsed text in the main food field (i.e. the dropdown which is only supposed to contain food objects, not freeform text)

There might be others, but I noticed these the other day,

---

I also manually added a new food in the dropdown, and the food added successfully, but would not show up in the food store. I had to manually edit it in the API with a `PUT` request to get it to show up. I haven't dug into this further to confirm it's consistent behavior. I imagine if this is a problem it's also a problem with units.

### Steps to Reproduce

For the first few issues, simply start parsing a recipe and test each thing.
~~For the bottom (potential) issue:~~ I can't reproduce this anymore.
~~1. Start parsing a recipe~~
~~2. Select any ingredient and type in a food that isn't in your food store manually~~
~~3. Select "create new food" in the dropdown~~
~~4. Observe that it's missing from your food store, but _does_ exist in your database~~

N/A

### Mealie Version

nightly

### Deployment

Docker (Synology)

_No response_

michael-genson:

_(REQUIRED)_

Fixes several issues with the ingredient parser. I've also added a useful icon/tooltip for when manual parsing is needed (so it's easier to see at a glance what needs to be manually addressed).

image
image

I also noticed that the tooltip location wasn't configured correctly from Vue 2 -> 3, so I fixed it in multiple places. The boolean props don't work anymore, they have to be `location="bottom"` or similar.

_(REQUIRED)_

Fixes https://github.com/mealie-recipes/mealie/issues/5761

## Testing

Tested a bunch with missing foods, units, etc.


frontend/components/Domain/ShoppingList/ShoppingListItem.vue
```diff
@@ -59,7 +59,7 @@
                 open-delay="200"
                 transition="slide-x-reverse-transition"
                 density="compact"
-                right
+                location="end"
```
michael-genson, line 62: Switched this from `right` to `end` because the only difference is for RTL languages ("right" is _always_ right, whereas "end" is right for LTR languages, and left for RTL languages).

frontend/components/Domain/User/UserAvatar.vue
```diff
@@ -2,7 +2,7 @@
   <v-tooltip
     v-if="userId"
     :disabled="!user || !tooltip"
-    right
+    location="end"
```
michael-genson, line 5: same here

Kuchenpirat (dismissed):
happy to merge as is, might make sense to set opacity to 100 for the icons so they look less "disabled"
If you dont agree feel free to merge.

michael-genson:
I agree - looks much better
image

Kuchenpirat (approved):

PR #5835, opened 2025-07-29
fix: remove v-lazy from RecipeCard

### LINKED ISSUE #5206: [BUG] - Overviewpage slow to load
### First Check

The overview page of the recipes take a significant amount of cpu time to render (in the browser). 

On the demo page I've added 4 recipes and it is showing already in the browser profiler (firefox, chrome) that the event loop is blocked for like 200ms (it does not sound much here). On my instance with many recipes it is around 2.5s that the page is blocked due to javascript code running. Even more on my mobile phone.

The profiler also shows warnings that the event loop is blocked for a long time 

### Steps to Reproduce

1. add some recipes (for example 20)
2. go to F12 -> performance profiler
3. activate profiling
4. reload overview page
5. check profiling

Performance is not visible in the logs

### Mealie Version

Version
v2.7.1
Build
d24a518bac7c89df27118286c8d9588f2edbbc44 

### Deployment

Docker (Linux)

_No response_

frysztak:
each v-lazy creates a new IntersectionObserver, tanking the performance. there's still some more work to do around fetching the recipes. locally, it takes 35 ms for API to send the response, but somehow that turns into almost 800 ms of JS 'computations' (that's on production build)

_(REQUIRED)_

this change vastly improves loading speed of recipes list

_(REQUIRED)_

(mostly) fixes https://github.com/mealie-recipes/mealie/issues/5206

### DIFF
--- a/frontend/components/Domain/Recipe/RecipeCard.vue
+++ b/frontend/components/Domain/Recipe/RecipeCard.vue
@@ -1,6 +1,5 @@
 <template>
   <!-- Wrap v-hover with a div to provide a proper DOM element for the transition -->
-  <v-lazy>
     <div>
       <v-hover
         v-slot="{ isHovering, props }"
@@ -99,7 +98,6 @@
         </v-card>
       </v-hover>
     </div>
-  </v-lazy>
 </template>
 
 <script lang="ts">


michael-genson (approved):
Wow, it's really that simple, huh? Awesome find. Everywhere we're using this component we're already paginating the API which is more-or-less the same as lazy loading anyway.

While there are still plenty of improvements to be made, this speeds up the main page a _ton_. Thank you!

Issue #5844, opened 2025-07-29
[BUG] - time picker & treeview components not availlable in currently used vuetify version


Kuchenpirat:
### First Check

- [x] This is not a feature request.
- [x] I added a very descriptive title to this issue (title field is above this).
- [x] I used the GitHub search to find a similar issue and didn't find it.
- [x] I searched the Mealie documentation, with the integrated search.
- [x] I already read the docs and didn't find an answer.
- [x] This issue can be replicated on the demo site (https://demo.mealie.io/).

### What is the issue you are experiencing?

The vuetify version we currently use does not include the time picker component yet. It was reintroduced to vuetify v3 on version 3.9 (see [vuetify docs](https://vuetifyjs.com/en/components/time-pickers/#usage))

### Steps to Reproduce

- open webhooks
- edit webhook
- observe missing time picker

### Please provide relevant logs

[Vue warn]: Failed to resolve component: v-time-picker
If this is a native custom element, make sure to exclude it from component resolution via compilerOptions.isCustomElement. 

### Mealie Version

nightly

### Deployment

Docker (Linux)

michael-genson:
Looks like the same issue for `v-treeview` in migrations: https://vuetifyjs.com/en/components/treeview/

PR #5839, opened 2025-07-29
fix: QueryFilter Hydration & script setup

Kuchenpirat:
## What this PR does / why we need it:

- First round of modernizing components to script setup 
  - Cookbook Components, and the components i opened to fix the issue
- fixes hydration issues on the QueryFilterBuilder

## Which issue(s) this PR fixes:

- fixes #5833 

## Special notes for your reviewer:

None

## Testing

Local

PR #5845, opened 2025-07-29
fix: script setup #2 and some fixes

Kuchenpirat:

- Migrates more components to the modern script setup function
- fixes an issue in where the QueryFilterBuilder does not return a QueryFilterString in some circumstances because i removed a event call from the recipe RecipeOrganizerSelector in #5839 
- replaces the timepicker used in the webhook editor with a time input field
- removed the unused GroupHouseholdSelector component

- solves the timepicker issue in #5844 (we might want to leave it open because of the treeview issue)

refactored some thing to make better use of automatic update propagation to the parrent by using defineModels 

## Testing

Local


michael-genson:
Would it be easier/better to upgrade Vuetify and get the time picker back?

Kuchenpirat:
yeah, should be not too hard, but i feel like the time picker is quite a wasteful component. It takes up an incredible amount of space just to input 2 numbers

michael-genson:
Yeah that's fair

michael-genson (approved):
LGTM

PR #5866, opened 2025-08-01
feat: Remove Not-Sort-By-Label and Refactor Shopping List Page

michael-genson:

_(REQUIRED)_

Removes the option to _not_ sort shopping lists by label. There's no real reason for this anymore as we move towards making parsing easier/more automated, and maintaining both views is already a headache.

Along the same lines as maintaining the shopping list view being a headache, this PR also refactors the shopping list page into several composables. This brings the shopping list page down from 1,251 lines to just 425; it now reads from the `useShoppingListPage` composable, which itself is broken into 7 different sub-composables with more focused function.

_(REQUIRED)_

N/A

I used Copilot (Claude Sonnet 4) to do the refactor, then touched it up manually. It did a pretty great job by itself, I really didn't need to modify much. Since this was a refactor (rather than writing new code aside from stitching things together) it makes sense that it could handle it well. I did do a manual review of everything, but worth calling attention to it and double checking since it's largely done by AI.

Breaking down this massive file is already a huge win for code quality/maintainability, but I think there's a lot of room to simplify how each individual composable works.

## Testing

Manually tested existing functionality to make sure everything still works as expected.


Kuchenpirat (changes requested):
This is so much cleaner than before 🚀
Did some testing. 
The uncheck all  as well as the check all button does no longer work for me. 

Also i would suggest moving the drag-handle to the last button position. 
This is also how it was before the nuxt upgrade.

Kuchenpirat (dismissed):
LGTM 👍

Kuchenpirat:
Seems like something messed up the e2e tests

michael-genson:
@Kuchenpirat pretty sure this is related to this: https://github.com/mealie-recipes/mealie/pull/5946
Once that is merged I'll catch this branch up and confirm

michael-genson:
Passing now

Kuchenpirat (approved):

PR #5879, opened 2025-08-02
feat: Add label notifier

cm226:

Adds a notifier for changes to labels.

Allows applications to observe changes to label data. 

Fixes #5822

Found that running 'task dev:generate' before I made any changes was updating some files. So I made 2 commits to separate the changes. 

First commit (0741e07999ac99036958de41247d3cd8a66d0c79) made no changes only ran 'task dev:generate' 
Second (9e04d2b390036807ca14e1666fd4f8b9b208ef57) Adds label notifier

## Testing

Ran the existing UT, Added new notifier to those

Registered a notifier and enabled all the label notifications, checked they all worked. 

Edit: Tests broken :( , will fix.


frontend/lang/messages/he-IL.json
```diff
@@ -79,7 +79,8 @@
     "tag-events": "אירועי תגיות",
     "category-events": "אירועי קטגוריות",
     "when-a-new-user-joins-your-group": "כאשר משתמש חדש מצטרף לקבוצה",
-    "recipe-events": "אירועי מתכון"
+    "recipe-events": "אירועי מתכון",
+    "label-events": "אירועי תווית"
```
DeepSpace2:
```suggestion
    "recipe-events": "אירועי מתכונים",
    "label-events": "אירועי תוויות"
```

Small changes for consistency. I noticed that most other "x-events" used the plural form of the noun.

michael-genson:
Haven't had the chance to review this yet, but you only need to add new strings to the en-US file. The rest is handled by Crowdin. So you can revert the changes to the other language files

cm226:
@michael-genson  Understood thanks

cm226:
@michael-genson any idea whats going on with the postgress failures? 

seems like alembic is complaining about multiple heads. When I check for heads locally there is only 1

```
vscode ➜ /workspaces/mealie/mealie/alembic (add-label-notifier) $ poetry run alembic heads
e6bb583aac2d (head)
```

mealie/alembic/versions/2025-07-30-19.32.37_e6bb583aac2d_add_label_notifier_crud_bools.py
```diff
@@ -0,0 +1,48 @@
...
+Create Date: 2025-07-30 19:32:37.285172
+
+"""
+
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "e6bb583aac2d"
+down_revision: str | None = "7cf3054cbbcc"
```
michael-genson: We added a migration recently, probably merged after you created this branch. Switch this to `d7b3ce6fa31a` and that should fix your Alembic errors.
```suggestion
"""'Add label notifier CRUD bools'

Revision ID: e6bb583aac2d
Revises: d7b3ce6fa31a
Create Date: 2025-07-30 19:32:37.285172

"""

import sqlalchemy as sa
from alembic import op

# revision identifiers, used by Alembic.
revision = "e6bb583aac2d"
down_revision: str | None = "d7b3ce6fa31a"
```

michael-genson:
Haven't had a chance to properly review yet, but the language changes look right now :) Posted a comment on the Alembic error, should be a quick fix

michael-genson (changes requested):
Looks good! One small thing: can you change the name of the migration file so it's later than the current latest migration?

e.g. `2025-08-08-19.32.37_...`
And then if you could change the comment in the file too (on line 5)

Forgot to mention this in my previous comment, we have some logic that assumes migrations are generated in order. This _probably_ wouldn't break anything but it'll save us some trouble in case it does.

cm226:
Great thanks for looking at it. Think that should be done now

michael-genson (approved):
Thanks!

Issue #5803, opened 2025-07-25
[BUG] - Minifying an image to "tiny-original" returns a cropped image


EmilFlach:
### What is the issue you are experiencing?

The current implementation of image minify `tiny` returns a cropped in image. This makes it impossible to use the `tiny` format as a thumbnail in a custom UI. I would like to use this format to reduce image file sizes and improve loading times.

Here is the relevant function doing this cropping:
https://github.com/mealie-recipes/mealie/blob/fb4aa2b7139749b8a80b1c1b64bac0e7f76e45c9/mealie/pkgs/img/minify.py#L100
```
    def crop_center(pil_img: Image, crop_width=300, crop_height=300):
        img_width, img_height = pil_img.size
        return pil_img.crop(
            (
                (img_width - crop_width) // 2,
                (img_height - crop_height) // 2,
                (img_width + crop_width) // 2,
                (img_height + crop_height) // 2,
            )
        )
```

### Steps to Reproduce

1. Go to any image in the demo environment, consider this example example:
https://demo.mealie.io/api/media/recipes/c81988d6-694e-47fb-93ad-0a3891a3630b/images/original.webp?rnd=1&version=YkMV
<img width="624" height="383" alt="Image" src="https://github.com/user-attachments/assets/83d079df-32ac-4799-9d30-f3ec3942ad49" />

2. Change the `original.webp` to `tiny-original.webp`

3. Notice how the returned image is not a tiny image, but one cropped to the center.

<img width="300" height="300" alt="Image" src="https://github.com/user-attachments/assets/5e8693d7-661e-48a8-801a-43acffa01926" />

### Please provide relevant logs

No error with a log occurs, it is just highly unexpected behavior. 

### Deployment

Docker (Linux)

PR #5883, opened 2025-08-03
feat: Improve Image Minification Logic and Efficiency

1-tempest:

This PR simplifies and centralizes the logic used for generating multiple image sizes (`original`, `min`, and `tiny`) during image minification. It improves performance and ensures consistent transformations for all variants.

- Updates logging to consistently include the output path for traceability.
- Reduces disk I/O by only reading the image once to generate variants.
- Adds min-original.webp max dimensions and ratio
- Adds original.webp max dimensions and ratio 
- Fixes tiny-original.webp cropping logic

- #5803
- Dramatically reduces traffic payload on each page with multiple recipes, which increases page responsiveness

## Testing

Tested Locally

## Notes

- It does not replace any existing images, it only applies to new uploads.
- Updating the frontend to use the correct images must be done to take full advantage
- Should decide if keeping original asset would make sense to generate new thumbnails for in the future (currently it is downsized and cropped)


1-tempest:
The previous tiny image just cut out the center pixels of the image instead of resizing it. I rewrote the cropping function to resize the image, rather than just see a tiny piece of a large image. As you can see there is significant savings which would drastically improve page loads over a remote connection.

| Version   | Before       | After        | Savings         |
|-----------|--------------|--------------|------------------|
| Tiny      | 3.9 KB       | 12.3 KB | N/A   |
| Min       | 345 KB       | 23.6 KB  | 93.2% |
| Original  | 345 KB       | 78 KB    | 77.4% |


From what I can tell images should be used in the following locations. I have been unable to confirm that min is used, and it seems that tiny isn't used very much. 

Original
```
 Default
```
Tiny
```
 /g/home/ in mobile view
 /g/home/recipes/finder all views
 /g/home/recipes/timeline Desktop view only
 /g/home/cookbooks/cookbook in mobile / tablet
 /household/mealplan/planner/edit (currently used)
```
Min
```
 /g/home in desktop / tablet view
 /g/home/cookbooks/ desktop view
```

Perhaps it would make sense to add another image file `large` or `header` instead of cropping the original, to preserve the ability to generate future thumbnails without further cropping

michael-genson:
> Updating the frontend to use the correct images must be done to take full advantage

Would you be able to go more in-depth about this? I'm pretty unfamiliar with how we handle images in Mealie

> It does not replace any existing images, it only applies to new uploads

Is it possible to do a one-time batch job, e.g. through a no-op migration? Or are we not able to due to how we're cropping images currently?

1-tempest:
> > Updating the frontend to use the correct images must be done to take full advantage
> 
> Would you be able to go more in-depth about this? I'm pretty unfamiliar with how we handle images in Mealie

I'm unfamiliar with Vue myself. I did sleuth through the pages however, and made the list of image replacements needed above. It looks like images are placed in different divs based on viewports, so just replacing the image type in the template in each location is all that I'm thinking is required.

> > It does not replace any existing images, it only applies to new uploads
> 
> Is it possible to do a one-time batch job, e.g. through a no-op migration? Or are we not able to due to how we're cropping images currently?

I'm not sure if there is a current regenerate thumbs job? If there was then that would surely work. However then we'd also want to add a new file type to allow re-cropping as currently each image type has an aspect ratio, cropping from our largest aspect ratio doesn't necessarily make sense.

michael-genson:
> I'm not sure if there is a current regenerate thumbs job?

No I'm thinking we could write one and add it here. Would we need a new file type if this were a one-type operation? I'm thinking it's not something user-triggerable, it's just something we do once. I'm also open to adding a new file type if needed though

1-tempest:
I ended up cropping the current original to gain further image savings.

I'd personally opt for saving the original file as is without deleting or resizing it, and then just showing smaller images to the client as requested. I believe the original image is currently purged instead of saved. 

Perhaps keeping original assets could be a choice for those conscious about space?

michael-genson:
Yup I think that makes sense. I'm inclined to just keep the original for now like you proposed and if we want to add an option later down the line to opt-in to not including the original we can look at it then.

1-tempest:
To be clear, we would still be saving one additional image - what would you want to name said image before the prefix if we're going this route? Just asking as I think `original` is already used (although I am just running off memory)

michael-genson:
`source` or `uncropped`, maybe? You're correct that `original` is already taken

1-tempest:
Yeah something like that would work.

I wonder if it would be confusing in the future having similar named objects that are actually quite different. 

I believe there is already a difference from the front-end variable naming to the file name, so perhaps keeping a naming convention constant doesn't matter too much?

In any case I'll just follow whatever is decided

michael-genson:
Let's see what you come up with! We can always change it before merging if we need to

1-tempest:
Ok I'm not sure I'll have time to work on this in the next couple weeks, but I can circle back when I have a minute. 
Someone is more than welcome to contribute if they feel inclined in the meantime

1-tempest:
I updated the file to add the uncropped file format. I am currently converting the original image to webp with a quality of 100. This can be changed this if desired.


TODOs related to this PR:
1) Update the the vue templates to reference the correct image size - I haven't checked if this has been done since I started the PR, it was not originally. I did make a list above in the conversation of what I noticed needs changed.
2) Create a migration job to regenerate all thumbnails.

mealie/pkgs/img/minify.py
```diff
@@ -71,82 +67,108 @@ def purge(self, image: Path):
...
+                        self._logger.info(f"{tiny_dest} already exists")
+                    else:
+                        tiny = PillowMinifier.crop_center_square(img.copy(), size=300)
+                        result_path = PillowMinifier.to_webp(dest_path=tiny_dest, quality=80, img=tiny)
+                        self._logger.info(f"{result_path} created")
+                        success = True
+
+        except Exception as e:
+            self._logger.error(f"[ERROR] Failed to minify {image_file.name}. Error: {e}")
+            raise
```
michael-genson, line 203: It looks like this code block never runs (there's a return statement right before it). Is this intentional? Should we remove this code?

michael-genson (changes requested):

mealie/pkgs/img/minify.py
1-tempest: That's a great catch. I don't think it should be there, I'll try to take a look why I added it a little later today

1-tempest: Thanks for catching this. It seems my code partly regressed with one of my commits. This is now untested as I unfortunately don't have the time needed for doing a proper test.
I'll try to test it when I can and mark it as ready for review at that time.

michael-genson (approved):
LGTM, thanks again for your work on this! I made a few changes:
- For simplicity I removed the functionality to keep the original before cropping. We can always add this in later if users want this
- I fixed the frontend references to these images and added a few new ones, to better leverage the work you did
- Fixed a few bugs

I also opted not to add any sort of migration script to keep this PR reasonably sized. It's complex since we need to account for small servers with potentially thousands of recipes and can't assume a simple migration job will complete in a reasonable amount of time.

1-tempest:
Thanks for going through this. Woah I somehow merged a much older draft at the end here that was still ridden with errors. Sorry about all the bugs!

Perhaps it would make sense to have a manually triggered script to regenerate thumbnails? I only kept the original as we now don't have any uncropped original image to run a future thumbnail job from (if that's something that anyone cares about)

michael-genson:
All good! It was only a few minor fixes.

I agree with a manually triggered script, I just didn't want to put it in this PR since I wanted to get it merged sooner rather than later and didn't have too much time to dive into a script for it.

1-tempest:
That's fair. I think it'd be good to note that by removing the uncropped-original.png we have no original image to generate thumbs from after merging the script as it is now, so if you run a manual script in the future you'll get new smaller thumbs with every generation

EmilFlach:
Thanks so much for your efforts on this, very excited to be using this soon!

PR #5900, opened 2025-08-05
fix: CONTAINS ALL doesn't contain all

michael-genson:

_(REQUIRED)_

The `CONTAINS ALL` filter doesn't actually contain all results when exactly one item is passed in and the item belongs to a join. This is most easily observed by creating a cookbook with `CONTAINS ALL` set to exactly one category (or tag, etc.).

This PR updates the `CONTAINS ALL` logic to use the same logic for `IN` when there is exactly one element. I also did a minor refactor on the query filter logic to make it easier to read.

_(REQUIRED)_

N/A, discovered with https://github.com/mealie-recipes/mealie/pull/5899

## Testing

Manually verified on the frontend cookbooks page, and updated our tests to include this edgecase.


mealie/schema/response/query_filter.py
```diff
@@ -351,45 +351,46 @@ def _get_filter_element[Model: SqlAlchemyBase](
...
         elif component.relationship is RelationalKeyword.CONTAINS_ALL:
-            primary_model_attr: InstrumentedAttribute = getattr(model, component.attribute_name.split(".")[0])
-            element = sa.and_()
-            for v in component.validate(model_attr_type):
-                element = sa.and_(element, primary_model_attr.any(model_attr == v))
+            if len(value) == 1:
+                element = model_attr.in_(value)
+            else:
+                primary_model_attr: InstrumentedAttribute = getattr(model, component.attribute_name.split(".")[0])
+                element = sa.and_(*(primary_model_attr.any(model_attr == v) for v in value))
```
michael-genson, line 375: This is the actual fix, the rest is just refactor

Kuchenpirat (approved):

PR #5903, opened 2025-08-06
fix: add confidence calculation for BruteForceParser

### LINKED ISSUE #5402: Brute parser always 0 %
The brute parser percentage is always 0 %, even if 100 % of the data was recognised.

Image

Version
v2.8.0

Skye-Samuels:

- Implements meaningful confidence scoring for the `BruteForceParser`
- Previously, all brute-force parsed ingredients defaulted to 0% confidence, regardless of whether values were successfully extracted.
- This change assigns field-level confidence scores (1.0 if a field is present, 0.0 if missing) and calculates an average confidence across all populated fields.
- Provides a more accurate and interpretable confidence value for brute-force parsing, helping to reduce confusion for end users who might otherwise interpret 0% as a failure.

Fixes [#5402](https://github.com/mealie-recipes/mealie/issues/5402)

This is my first contribution to an open source project! I look forward to hearing your feedback on my simple bug fix, it may be small but I hope it is at least a little helpful! :)

## Testing

- I tested this was non-breaking by running the `task py:check` and it passed 100% of the tests.
- To ensure the bug was fixed I used the web UI to add a new recipe, edit the recipe, parse the ingredients with the Brute parser then observe the ingredients it parsed displaying 100% confidence instead of 0%.


mealie/services/parser_services/ingredient_parser.py
```diff
@@ -31,8 +31,27 @@ class BruteForceParser(ABCIngredientParser):
...
+        qty_conf = 1.0 if bfi.amount else 0.0
+        unit_conf = 1.0 if bfi.unit else 0.0
+        food_conf = 1.0 if bfi.food else 0.0
+        note_conf = 1.0 if bfi.note else 0.0
+
+        for conf in [qty_conf, unit_conf, food_conf, note_conf]:
+            if conf > 0:
+                confidences.append(conf)
+
+        avg_conf = sum(confidences) / len(confidences) if confidences else 1.0
```
michael-genson: Won't this always come out to 1? Since you're only appending to `confidences` if it's 1. We can probably just hardcode `avg_conf = sum(confidences) / 4` and remove the loop.

```python
avg_conf = (qty_conf + unit_conf + food_conf + note_conf) / 4
```
However, this still really isn't accurate (e.g. if I parse `1 carrot` this will only have a 50% accuracy). Can we instead try doing this calculation after we do ingredient matching? Something like:

```python
matched_ingredient = self.find_ingredient_match(parsed_ingredient)

# Always true, since we have no way of validating these
qty_conf = 1
note_conf = 1

# If we parsed a unit and matched it to the unit store, we're confident it's correct
if bfi.unit is None or matched_ingredient.ingredient.unit:
    unit_conf = 1
else:
    unit_conf = 0

# Same for a parsed food
if bfi.food is None or matched_ingredient.ingredient.food:
    food_conf = 1
else:
    unit_conf = 0

avg_conf = (qty_conf + unit_conf + food_conf + note_conf) / 4
matched_ingredient.confidence = ...  # same code you had before
return matched_ingredient
```

michael-genson (changes requested):
Hey there, thanks for this! I have a couple notes on the confidence calculation to maybe make it more meaningful, let me know what you think.

mealie/services/parser_services/ingredient_parser.py
Skye-Samuels: Thanks for the feedback @michael-genson !

I’ve updated the PR to implement your suggested approach:
- The confidence calculation is now performed after ingredient matching.
- Confidence for unit and food is now based on whether they were successfully matched to the store.
- This provides a more meaningful and accurate average confidence score.

Let me know if this aligns with what you had in mind!

michael-genson (changes requested):
LGTM, last thing though - can we add some tests around this? Just basic ones, particularly around whether we do/don't have foods/units. If you're not sure how to do this let me know, happy to help finish it up.

Check out `tests/unit_tests/test_ingredient_parser.py` for some examples of how to test the brute parser (you can just add a couple of tests there)

Skye-Samuels:
I've pushed a new commit that includes:
- Unit tests specifically for the confidence calculation logic in the BruteForceParser
- A small adjustment to the confidence logic, as discussed. The confidence scores are now determined after ingredient matching, based on whether a match was found for unit and food.

Let me know if there's anything else you'd like me to update!

michael-genson (approved):
Looks great, thanks for your work on this!

PR #5911, opened 2025-08-07
fix: import from Paprika not importing some images

brokeh:

This fixes 4 issues that can arise when importing recipes from Paprika:
1.  A silent failure to import the image due to an empty (0 bytes) or incomplete JPEG image. There was nothing in the logs, just no image attached to the imported recipe.
    The problem is that the temp file contents may not have been flushed to disk before copying the temp file to the working directory as `original.jpeg`.
    This was the main problem I faced when importing from Paprika and meant that none of the images would come across in the import.
2.  The below error log when importing a recipe that has no image.
    ```
    Failed to download image for slow-cooker-beef-ragu: slow-cooker-beef-ragu
    ```
    The error log doesn't cause any issues with the import, it's just not particularly helpful and did make me look into it briefly when trying to figure out why the other images failed to import.
    The error message has been replaced by an info message saying
    ```
    Recipe 'Slow-Cooker Beef Ragu' has no image
    ```
3.  Failure to import the image for a recipe when there's another recipe with the same name, either in the same import or pre-existing.
    The failure occurs with the error log
    ```
    Failed to download image for chicken-pot-pie-2: chicken-pot-pie-2
    ```
    The way it was handling images did not take into account that the imported recipe might end up with a slug that is not the same as the slugified name.
4.  Silently failing to import images due to unrecognised file contents when converting to WEBP.
    For my issue, the unrecognised files were all due to issue 1 making them all empty, but this could also just be for other reasons such as a corrupt file or unrecognised format.
    Now when an image being imported can't be processed, the below error will be logged
    ```
    Failed to download image for chicken-and-leek-casserole: cannot identify image file '/app/data/recipes/4598f8cb-fabd-4e1e-bca7-a357d4e09f34/images/original.jpeg'
    ```
    This applies to migrations from all supported formats, not just Paprika.

## Testing
1. Importing my Paprika zip file which failed to import any images previously
    1. Check the imported recipes to ensure all recipes that have an image in Paprika do in Mealie too (test fix 1)
    2. Check the `task py` logs for errors (test fix 2)
2. Importing the same Paprika zip file a 2nd time to import a bunch of duplicate names/slugs
    1. Check the imported recipes to ensure all recipes (both original and duplicates) that have an image in Paprika do in Mealie too (test fix 3)
3. Revert the fix to `temp_file.flush()` and import the same Paprika zip file on a fresh database
    1. Check the `task py` logs for errors (test fix 4)


michael-genson (approved):
LGTM, thank you!

PR #5966, opened 2025-08-16
feat(ui): improve drawer behavior, translations, and cookbook dropdown

sophiafema:

- **Translations:** Updated German language file (`de-DE.json`) with more natural and consistent wording (reviewed by a native speaker).
- **Drawer bottom links:** Links are no longer fixed at the bottom but scroll with the content → improves UX on smaller screens.
- **Main content scrolling:** Disabled scrolling of the main content when the drawer is open on mobile → prevents background scrolling and improves focus.
- **Drawer menu:** Adjusted the order of menu items for better navigation consistency.
- **Cookbook management:** Added the ability to create a new category directly in the dropdown menu → streamlines workflow, no need to leave the page.

These changes improve overall usability and help prevent potential user errors in navigation.

---

- No breaking changes introduced.
- All commits follow **Conventional Commit** syntax.
- Pre-commit hooks pass successfully.

---

## Testing

- Verified in local Docker dev environment.  
- **Manual testing only** (no new automated tests added), since these are small UI adjustments built on existing, tested functionality.  
- Confirmed:
  - Bottom links scroll with content.
  - Main content cannot scroll when drawer is open (mobile).
  - Drawer menu order updated.
  - New category creation in cookbook dropdown works as expected.
- **Translation updates** reviewed and validated by a native German speaker.


michael-genson:
Hey there! Haven't checked out the PR yet, but translations are handled through Crowdin, not through PRs:
https://docs.mealie.io/contributors/translating/

Happy to take a look at the rest of the PR, but the translation changes will have to be removed and done through Crowdin

Kuchenpirat (changes requested):
Without having looked at the rest of the PR, all translations have to be done through crowdin.
So please remove the updates to the translation files from this pr.

Also the translation for 'cookbooks' is not 'Kochbuchverwaltung', so we should either update the wording to 'cookbook settings' for all languages, which is quite long in the sidebar, or leave as is. 

---
Missed on phone that michael already send this message,

Kuchenpirat:
Had a look at the changes and i dont think we would want to include this pr at this point and time. 
- translations: as mentioned above handled by crowdin
- shift & reorder drawer links: while i agree that the scroll behaviour of the sidebar can be confusing if space needed exceeds the availlable vertical space i do not think this is a better UI in general. It is a worse UI on bigger screens and is confusing in that we mix settings and overview pages in the bottom part of the navigation
- Cookbook management: I dont think this makes sense here, since the new category would not have any recipes. So the logical flow would be to create category -> assign recipes -> create cookbook with said category, and we already had an existing add option with and icon that was explicitly disabled on this page.

sophiafema:
Thank you for taking the time to review this PR and for the detailed feedback 🙏.
I completely understand the concerns you raised, but I would like to share a bit more context about the motivation behind these changes and suggest some possible adjustments:

Sticky vs. non-sticky drawer links
I see that on large screens the sticky behavior is preferable, but on mobile the current experience can be confusing when the bottom links (sticky) take up a lot of space. When using mealie first I didn't even notice, that I could scroll the upper part of the drawer. Additionally on mobile the drawer takes up most of the screen and both the background and the drawer scroll at the same time, which can be confusing.
→ A compromise could be: keep the bottom links sticky on desktop, but make them scroll with the content on mobile. This way we get the best of both worlds.

“Add Category” in cookbook management dropdown
My workflow often starts with wanting to create a new cookbook, and only then realizing that I need a new category for it. Having the option to quickly add a category at that point is very convenient, even if the category starts empty and recipes are assigned later.
I believe this small addition could save steps for users who build new cookbooks more flexibly.

Drawer menu order
From my perspective, cookbooks themselves are accessed much more frequently than cookbook management and the timeline. On mobile, this means users currently need to scroll down past these items before reaching the cookbooks they actually want to open. Reordering them would make the most frequently accessed items quicker to reach and reduce friction, especially for mobile-first use.

Of course, I don’t expect all changes to be merged as-is. Even if only some adjustments (like the drawer scroll/background behavior) were accepted, I think they would improve usability without disrupting the overall UI design.
Would you be open to discussing a hybrid approach (sticky on desktop, scrollable on mobile) and possibly reconsidering the drawer/menu adjustments from a mobile-first perspective?

Kuchenpirat:
Hey,
thanks for sharing more of your thought process around this. A quick reply without having thought too much about it. 

**Sticky vs. non-sticky drawer links**
I can absolutely see your point about the sticky drawer. In my opinion it would be more advantageous to have:
1. more flexible cookbooks: make the showing of cookbooks more flexible (add options to hide cookbooks) and make the cookbook page be an actual overview page instead of just a cookbooks settings page so there is one page with all cookbooks and your favorite cookbooks can be shown on the side. 
2. Consolidate the settings into a single settings button with a pop out that lets the user then acces the settings that are currently at the bottom + maybe even link to the normal settings this way since they are currently hidden behind the user name. This has been on my list for quite some time, but havent gotten too it yet. 
3. Add visual hint that the top section can be scrolled or move the consolidated settings button to the top app bar entirely.

I know this would entail a lot more changes than what you are proposing, but i feel the result would be a lot "cleaner".

**Add Category**
I still do not think this is necessairy. If more people differ from my opion here, i am happy to hear about it. Either way, we sould just use the functionality that is already included in the component with the icon instead of using "Add" which will have very different lengths depending on the user language. 

**Drawer Order**
I might be too stuck on the way we have it set up currently, but this too me does not offer too much of an improvement. And grouping the timeline with the settings does not make sense to me.

PR #5974, opened 2025-08-17
fix: OIDC signup fails if OIDC_ADMIN_GROUP set and groups claim empty

LinuxSBC:

When OIDC_ADMIN_GROUP is set to a value (such as "admin") but OIDC_USER_GROUP is unset, this should allow any user to be created when attempting to log in with OIDC. However, if the OIDC_GROUPS_CLAIM is empty, user creation fails even though it should succeed.

For an example, see the test_no_group_new_user test that I added, which fails before this change but succeeds after it.

This PR fixes the issue by skipping the check that required claims not be empty for the OIDC_GROUPS_CLAIM. For the list of groups, it doesn't actually matter if the list is empty or not as long as the group requirements are met, so it's fine to bypass that check.

Other implementations I considered:

1. I considered marking OIDC_GROUPS_CLAIM as required only if both OIDC_ADMIN_GROUP and OIDC_USER_GROUP are set, rather than just if one is set, but that's likely to cause further consequences by allowing an expected field to be missing. It also caused some tests to fail.
2. I considered allowing all arrays to skip the empty check, but again, that's likely to cause other unforeseen consequences, so I wanted to limit the scope of this.

I could not find any issues about this, but I can create one if that would help.

## Testing

I added two tests to the OIDC provider unit tests, one which tests when there are no groups in the OIDC_GROUPS_CLAIM and one that tests when there is an unrelated group in the claim.


michael-genson:
Thank you for your work on this! There are two open PRs solving this issue, so I'm closing this in favor of https://github.com/mealie-recipes/mealie/pull/6054. Both are pretty similar in scope

PR #5985, opened 2025-08-18
fix: Update variable name in RecipeCard.vue to enable household ratings to appear on recipes

### LINKED ISSUE #5863: [BUG] - Average household rating not displayed where user has not rated a recipe
### First Check

When a user has not rated a recipe, if the recipe has a rating from other members in the same household then the average rating should appear in greyscale (according to logic in [this pull request](https://github.com/mealie-recipes/mealie/pull/3345)). 

This does not appear to be happening and instead a blank rating is displayed.

### Steps to Reproduce

1. Set up two users in a household.
2. Login as user 1.
3. Create a recipe in the household.
4. Rate the recipe.
5. Login as user 2.
6. View the recipe - it should have no rating from the user 2 however it does have a household rating so should display that in greyscale.

No relevant item in log. For the recipe I tried the following was the only element in the log:

[INFO|httptools_impl|L476] 2025-08-01T09:02:55: 172.17.3.204:0 - "GET /api/households/self/recipes/garfagnana-sausage-ragu HTTP/1.1" 200

Note: I checked the recipe using the API and it does have a correct household rating so that part is working - it just appears to be the display of that rating on the front end.

### Mealie Version

Latest version - 3.0.2

### Deployment

Docker (Linux)

_No response_

ProperTeaMogul:

Fixes issue raised [here](https://github.com/mealie-recipes/mealie/issues/5863) where household ratings were not appearing when the user had not rated a recipe.

This seemed to be caused by an incorrect variable name in the recipecard.vue. This PR updates the variable to enable household ratings to show for recipes where the user has not rated the recipe themselves.

Fixes #5863 where household ratings were not appearing when the user had not rated a recipe.

## Testing
Tested on a copy of my main server to ensure I had household ratings from other users.

### DIFF
--- a/frontend/components/Domain/Recipe/RecipeCard.vue
+++ b/frontend/components/Domain/Recipe/RecipeCard.vue
@@ -57,7 +57,7 @@
 
               <RecipeRating
                 class="ml-n2"
-                :value="rating"
+                :model-value="rating"
                 :recipe-id="recipeId"
                 :slug="slug"
                 small


michael-genson (approved):
Thanks!

Kuchenpirat (commented):
Hey, thanks for the fix, we are currently doing some maintenance, and one part of this is updating the components to script setup functions and moving modelValues into a const defined by a defineModel function. 
Apparently i missed the recipeRating component in the PR that should have included all recipe components. 
If you are willing to collaborate, i would push the update for the RecipeRating to this PR and we can use v-model instead of :model-value.

---
edit: 
i have to stop using github on mobile, i constantly miss comments. 
Missed this is already approved, will add the changes to another PR then.

PR #6013, opened 2025-08-22
fix: Can't add first shopping list item to shopping list

### LINKED ISSUE #6010: [BUG] - Can't manually add item to shopping list without selecting a food first
### First Check

I can't seem to add just an arbitrary item to a shopping list by just populating the Note field, UNLESS first interacting with the Food drop down.

### Steps to Reproduce

Steps to reproduce:
1. Navigate to any particular shopping list
2. Select the Add button
3. Enter anything you want in the default focused Note entry, for example: `test`
4. Select the Save button

Expectation:
1. A new item appears and remains in the list with the label `test`

Actual:
1. Very briefly, perhaps for 0.1 to 0.25 seconds, the `test` item appears in the list
2. After that time, the item is removed completely from the list, not completed, just gone

Strange Work Around:
1. Navigate to any particular shopping list
2. Select the Add button
3. Select the Food drop down menu
4. Click off the menu again, no need to actually enter any value or select anything there, just engage with it
5. Enter anything you want in the default focused Note entry, for example: `test`
6. Select the Save button
7. You will now see the item entered as expected in the list.  This effect persists for additional future items, so long as you don't navigate away from the shopping list.

The behavior is evident without any logs.

### Mealie Version

Mealie V3.1.1, Build: 12d38c89eaff16e498f7791679ce7dab0148203f

### Deployment

Docker (Linux)

_No response_

michael-genson:

_(REQUIRED)_

When populating the create item data on the shopping list, the shopping list id isn't populated yet so the factory returns an empty string for the shopping list id. This empty string, if not updated, is sent to the backend (which fails because it's required).

While it seems like selecting a food first fixes this, it actually doesn't, the item just fails to get added to your shopping list. However then the create data factory refreshes and populates the shopping list id as expected.

This PR ensures the shopping list id is set before sending the item to the backend.

_(REQUIRED)_

Fixes https://github.com/mealie-recipes/mealie/issues/6010

I also fixed an unrelated console warning on this component.

## Testing

Manually


frontend/components/global/InputQuantity.vue
```diff
@@ -10,8 +10,9 @@
       :min="min"
       :max="max"
       type="number"
-      size="small"
       variant="plain"
+      density="compact"
+      style="width: 60px;"
```
michael-genson, line 15: `size="small"` throws a warning in the console that it's invalid (even though it works??) and setting the density and width here mimics the same behavior. Note that `width: 60px;` copies the `max-width` setting a few lines above this.

PR #6043, opened 2025-08-27
feat: consolidate settings gui

Kuchenpirat:

Settings are somewhat hidden behind the user avatar in the side bar and take up a lot of space at the bottom for not frequently changed settings. 

This consolidates all settings behind an universal settings button with a menu. 

None, just some cleaning up. 

Had to update some test. Not really familliar with playwright but it seems to be working as expected now. 

## Testing

Manual

## Screenshot
| Before | After | Mobile (After) |
| -- | -- | -- |
| image  | image |  image |


michael-genson (changes requested):
Love this. Just a couple suggestions.

Can we change "settings" to something else, like "user settings"? Seems strange to click on "settings" and then click on "settings" again.
image

It also might be worth adding a "data management" option and linking directly to the data management page (which defaults to food management). Then we could also remove it from the user profile page. Honestly a button for user/household/group settings might not be a bad idea here and then we can split up the user profile settings page.

Also, would it be better to leave the language selector out of this menu? I'm thinking if it's a user's first time opening the site and they don't speak/read English they might have a hard time finding the language setting. Not sure if this is necessary but just a thought I had.

Kuchenpirat:
Yeah, the naming does make sense 👍
Direct link to data management does also make sense to me 👍
I have to make a version with all the options to have this visualized, might be a bit too much, but I'll try it out. Like the idea in principle. 

Mealie defaults to the browser language I think, so people should usually not need to change language and it's loads cleaner to just have one button. Also I think it's pretty standard to have the language button behind another button afaik.

michael-genson:
> I have to make a version with all the options to have this visualized, might be a bit too much

For sure, if it ends out being too cluttered then how you have it is fine (and we can always revisit if we come up with something better)

> Also I think it's pretty standard to have the language button behind another button afaik

Sounds good to me, I'm spoiled by English-default so it's not something I'm terribly familiar with.

Kuchenpirat:
Updated the language and added the manage data links. 

Seperating out the settings is a bit to much for this layout imho, so left it out for now. 

Also verified that language defaults to the system or browser language so having it behind a settings menu should be no problem, especially since the settings menu and the languge button have icons that should make it easy even if you dont understand the currently displayed language.

frontend/components/Layout/LayoutParts/AppSidebar.vue
```diff
@@ -82,20 +83,22 @@
...
+        <v-menu location="end bottom" :offset="15">
+          <template #activator="{ props }">
+            <v-list-item v-bind="props" :prepend-icon="$globals.icons.cog" :title="$t('general.settings')" />
+          </template>
+          <v-list density="comfortable" color="primary">
+            <v-list-item :prepend-icon="$globals.icons.translate" :title="$t('sidebar.language')" @click="languageDialog=true" />
+            <v-list-item :prepend-icon="$vuetify.theme.current.dark ? $globals.icons.weatherSunny : $globals.icons.weatherNight" :title="$vuetify.theme.current.dark ? $t('settings.theme.light-mode') : $t('settings.theme.dark-mode')" @click="toggleDark" />
+            <v-divider v-if="loggedIn" class="my-2" />
+            <v-list-item v-if="loggedIn" :prepend-icon="$globals.icons.cog" :title="$t('profile.user-settings')" to="/user/profile" />
+            <v-list-item v-if="canManage" :prepend-icon="$globals.icons.manageData" :title="$t('data-pages.data-management')" to="/user/profile" />
```
michael-genson:
```suggestion
            <v-list-item v-if="canManage" :prepend-icon="$globals.icons.manageData" :title="$t('data-pages.data-management')" to="/group/data" />
```

michael-genson (changes requested):
LGTM, save for one issue with the new data management link

michael-genson (approved):
LGTM

PR #6057, opened 2025-08-28
feat: Customize Ingredient Plural Handling

### LINKED ISSUE #5777: English plural handling grammatically incorrect with ingredient amounts enabled
### First Check

When using ingredient amounts, the plural handling could be improved. I can't speak for other languages, but in English if I have a unit of an ingredient other than whole ingredients (slice of onion vs onion), only the unit should be pluralized (i.e. 2 slices onion vs. 2 onions). I cannot think of a case where this isn't true.

You may introduce an example like bunches of scallions, but for cases like this, the singular still has the mass noun (i.e. 1 bunch of scallions is correct), so scallions has to be a separate ingredient entry anyway.

Again, I don't know how this works in other languages, but I don't think it should be hard to at least add a setting that pluralizes the unit if a unit is present, and otherwise pluralizes the ingredient. Otherwise we're left with the default behavior that yields 2 cups sugars.

### Steps to Reproduce

N/A

N/A

### Mealie Version

_No response_

### Deployment

Docker (Linux)

_No response_

michael-genson:

_(REQUIRED)_

Plural handling is very complicated, especially in a multi-lingual app. There's no single implementation that will make everyone happy, and at some point there's a line we need to draw where we hit an unreasonable amount of complexity. That said, our current implementation leaves much to be desired.

The current logic for plural handling is as follows:
- If the quantity is <= 1, and not zero, use the singular form
- If the quantity is > 1 or == 0, use the plural form

In English, at least, this isn't really good enough; consider "2 cups rotisserie chickens chopped" (should be "2 cups rotisserie chicken chopped").

This PR adds a simple group-level setting that controls how plurals should be handled for the group for quantities > 1 or == 0:
- Pluralize only if there is no unit (e.g. "2 cups onion chopped" | "2 onions chopped")
- Always pluralize even if there is a unit (e.g. "2 cups onions chopped" | "2 onions chopped") (this is the current behavior)
- Never pluralize anything (e.g. "2 cups onion chopped" "2 onion chopped").

My rationale for this:
- In English, at least, this is good enough using the first option. There are some cases where you can argue that the food should be plural even if there is a unit (e.g. "1,000 ounces onions") or singular even if there is a unit (e.g. "1/2 cup ice cubes", though I'd personally switch it to "ice") but these are a lot less common, and basically impossible to configure in a predictable way. For users who don't like this, it's perfectly acceptable in English to simply always use singular versions of foods, and this is an option now
- In some languages, such as German (thanks @Kuchenpirat) the existing behavior is desired, and thus it's still an option
- In many languages, particularly languages like Chinese, Korean, and Japanese, pluralization isn't even really a thing, or at least not with regards to recipe ingredients, so there's no reason to have a plural setting since it doesn't make any sense
- Above all else, if anyone is unhappy with how plurals work, they can just turn them off now

The new UI for group plural settings:
image
image

_(REQUIRED)_

Fixes https://github.com/mealie-recipes/mealie/issues/5777

This PR got a lot bigger than I was expecting, particularly because we didn't have a way of exposing group settings to a normal user/public user.

I decided to make the database default `pluralize_food_without_unit` because I think most languages prefer it this way. I'm open to suggestions here.

This is discussed elsewhere, but this is specifically a group setting because the food store belongs to the group, so it makes sense that the group defines how the foods are pluralized (as opposed to the household). In theory the data is set up such that it only makes sense for one language/plural setting.

## Testing

Tested everywhere ingredients are parsed (the recipe, adding ingredients to cook mode, copying ingredients to the clipboard, the shopping list). Updated frontend tests and added new backend tests.


rkram5424:
Looking forward to this PR. I might purge all plural fields in the database in the meantime.

michael-genson:
Replacing this with https://github.com/mealie-recipes/mealie/pull/7057

PR #6066, opened 2025-08-30
fix: Incorrect Usage of $vuetify.display

### LINKED ISSUE #6047: Toggle View button missing after v3.0.2
### First Check

In Mealie v3.0.2, the Recipes page has a “Toggle View” button that switches between the compact view (image left) and the card view (image top). After upgrading to v3.1.0 and newer (tested up through v3.1.2), the button is no longer present.

I don't see this change mentioned in the changelog, so I guess it was removed unintentionally?

Mealie v3.0.2:
Image

Mealie v3.1.2:
Image

### Steps to Reproduce

1.	Run Mealie v3.0.2 → open the Recipes view
2.	Click on the three dots on the right of the "Random" button and you'll see the Toggle View button
3.	Run Mealie v3.1.0 or later → open the Recipes view
4.	The "Random" button is present but not the three dots and the Toggle View button

No logs needed

### Mealie Version

Currently on Mealie v3.1.2

### Deployment

Docker (Linux)

michael-genson:

_(REQUIRED)_

Throughout the app we use `$vuetify.display` a lot. This was a Nuxt 2 pattern which isn't recommended in Nuxt 3. More importantly, though, is combined with the new `script setup` syntax it actually causes shadowing problems where `$vuetify` is exported as-is and you need to reference the `value` in the template, which broke some component behavior.

This PR replaces all instances of `$vuetify.display` with `display = useDisplay()` in the script section of the template. I also fixed all references in the HTML to not use `.value` anymore.

_(REQUIRED)_

Fixes https://github.com/mealie-recipes/mealie/issues/6047

## Testing

Manually checked where we were doing this


Kuchenpirat (approved):

Kuchenpirat (approved):
👍

PR #6070, opened 2025-08-31
fix: Refactor and Optimize Explore Page Search

michael-genson:

_(REQUIRED)_

Breaks apart the behemoth that is the recipe explorer page into a few different components and a composable. Also includes some minor performance enhancements for the filter.

The main page is down from 715 lines to 73. The longest file in this new layout is the composable which handles the search/query-filter logic, and that's 466 lines of pure JS. Everything else is small (100-200 lines or so).

_(REQUIRED)_

N/A

This does not solve the bulk of the performance issues on the explorer page. I was going to address that but got fed up with how complicated the explorer page was and tackled that instead. The main performance issues seem to be with rendering the recipe cards.

## Testing

Manually verified the filters work as expected.


Kuchenpirat (dismissed):

Kuchenpirat:
Looks good to go for me.
There are probably better solutions for this but perfect is the enemy of finished, so let's get this in 👍

Kuchenpirat (approved):
Looks Great, so nice to have this cleaned up a bunch

PR #6071, opened 2025-08-31
fix: Optimize Recipe Context Menu

michael-genson:

_(REQUIRED)_

The recipe context menu is horrendously slow. It's contributing easily 50% of the slowdown on the recipe explorer page. And we render it 60 times every time we open that page. It's _bad_.

I haven't dug into why this is happening or how to fix it, but this PR takes the lazy backend-developer-don't-understand-frontend way out and instead lazy loads the content. I created a thin wrapper that loads an empty menu, and when the user opens the menu we load the whole context menu component. Even that is kind of slow but a 0.5 second slowdown when you click on the menu is better than a 6 month hike up Mount Everest every time you load the main page.

_(REQUIRED)_

Largely addresses https://github.com/mealie-recipes/mealie/issues/5206 but we're not done yet, it's still quite slow. ~~I think it's the recipe favorite and recipe rating components causing most of the rest of it~~ https://github.com/mealie-recipes/mealie/pull/6075 should cover the rest of it, for the most part.

## Testing

Open this PR up, ~~comment out `RecipeFavoriteBadge` and `RecipeRating`~~ and don't blink.


Kuchenpirat (approved):
Looks good to go for me.
There are probably better solutions for this but perfect is the enemy of finished, so let's get this in 👍

michael-genson:
Yeah this is definitely the laziest solution I could possibly think of, but if it solves the problem for 95% of usage it's good enough for now.

I learned a bit more in the recipe ratings PR that we already merged which might help diagnose the actual problem here, but at least this downgrades it from a major performance hit to a minor one.

PR #6075, opened 2025-08-31
fix: Optimize Recipe Favorites/Ratings

michael-genson:

_(REQUIRED)_

Along with https://github.com/mealie-recipes/mealie/pull/6071 optimizes most of the slowdown on the recipe explorer page. This makes two key optimizations:
1. Creates a global `i18n` instance that can be shared amongst components. Importantly, since this is global, it can be created once in any setup script and reused anywhere else _even if it's not in the setup script_. This allows us to defer setting up user APIs, removing basically all slowdown from the recipe favorite component.
2. Removes interactivity from the recipe rating used on the recipe card. This means you can no longer rate a recipe through the recipe card; you have to go to the recipe page to do this. The `v-rating` component is very slow, and rendering 60 of them at once on the explorer page is what leads to the massive slowdown from this component. The new implementation uses basic JS and CSS.

_(REQUIRED)_

Addresses https://github.com/mealie-recipes/mealie/issues/5206. Along with https://github.com/mealie-recipes/mealie/pull/6071 I think this effectively removes the performance issues we're seeing. More production testing after these are both merged will be warranted before we close that issue though.

The new global i18n optimization should allow us to vastly simplify our API usage throughout the app, including additional performance improvements. Since i18n can be used outside of the setup script (and therefore APIs can be too) we have a lot more control on how we set up APIs.

## Testing

Played around with rating a lot with logged-in and logged-out users.


frontend/components/Domain/Recipe/RecipeCardRating.vue
```diff
@@ -0,0 +1,111 @@
...
+        class="star-empty"
+      >
+        ☆
+      </span>
+      <span
+        v-if="star === 'full' || star === 'half'"
+        class="star-full"
+      >
+        ★
+      </span>
```
michael-genson, line 25: If you're wondering why I took this layering approach instead of using the half-filled unicode star, it's because... there actually isn't one, surprisingly. I feel like I've seen it before but maybe it's non-standard?

Kuchenpirat (commented):
All in all this looks ready to go. 
One small comment. 

Can we use script setup for the new component, all the other recipe components were already updated and it makes sense to have this uniform.

michael-genson:
Oops yep, force of habit. Just changed it

Kuchenpirat (approved):
Looks Good, lets get these sweet performance gains 🚀

Issue #6064, opened 2025-08-30
[BUG] - Inconsistent date format across app functions


ThisIsTheOnlyUsernameAvailable:
### First Check

- [x] This is not a feature request.
- [x] I added a very descriptive title to this issue (title field is above this).
- [x] I used the GitHub search to find a similar issue and didn't find it.
- [x] I searched the Mealie documentation, with the integrated search.
- [x] I already read the docs and didn't find an answer.
- [x] This issue can be replicated on the demo site (https://demo.mealie.io/).

### What is the issue you are experiencing?

I understand that Mealie backend and API uses the date format of YYYY-MM-DD, however the frontend has an inconsistent mix of date formats.  For example:

<img width="889" height="406" alt="Image" src="https://github.com/user-attachments/assets/058edc68-e0a9-4bfa-827b-f77e2dfe23ab" />
<img width="502" height="397" alt="Image" src="https://github.com/user-attachments/assets/1a23be0b-a5a7-41a8-9e31-ae71dcdaa049" />
<img width="229" height="58" alt="Image" src="https://github.com/user-attachments/assets/65d6f5f7-41dc-4cb8-bf30-c0905fad8473" />

The US date format is confusing to an international audience.  Whilst it's obvious on the date of opening this ticket (30 August), it becomes far more confusing on dates like 5 September.  Short of building date parsing logic into the app and all of the pages where dates are handled, I'd propose a standardisation of YYYY-MM-DD which would be obvious for all audiences.

### Steps to Reproduce

Browse any date-related page

### Please provide relevant logs

N/A

### Mealie Version

894162a669cef9fefea89e786f40128b8c6e621c

### Deployment

Docker (Linux)

michael-genson:
This is configurable per-locale, but as you've demonstrated there are some places we don't localize and instead hardcode YYYY-MM-DD

PR #6079, opened 2025-09-01
fix: Fix RecipeLastMade dialog date picker being off by a day

### LINKED ISSUE #5818: [BUG] - 'I Made This'-Calender chooses wrong Date
### First Check

Choosing a date on the calender of 'I Made This', selected the day before the selected one.

### Steps to Reproduce

1) I navigate to a reciept and click on the the 'I Made This'-Button (hope I've translated it correctly, because it's in German for me)
2)  I click on the date and a calendar appears
3) I choose any date (e.g. 2025-07-17)
4) I the date field shows the day before the day I've selected (e.g. 2025-07-16)
optionl:
5) I click on the date filed again and the originally selected day (e.g. 2025-07-17) is shown as selected

No relevent logs here

### Mealie Version

3.0.1 on my side and 3.0.2 on demo page

### Deployment

Docker (Linux)

_No response_

lehnerpat:

_(REQUIRED)_

This PR fixes a bug in the "I Made This" (i.e. 'Add to Timeline') dialog on the recipe page, where selecting a date in the popup calendar would put an incorrect date string into the text field, as described in https://github.com/mealie-recipes/mealie/issues/5818.

The root cause was some date manipulation in `RecipeLastMade.vue`, in addition to time zones, as Michael supposed in https://github.com/mealie-recipes/mealie/issues/5818#issuecomment-3124729197.
I'll point out more details to the fixes with inline comments.

Here are two short videos showing the buggy behavior before and the fixed behavior in this PR:

_Before (tested on Mealie Demo at 2025-09-01 20:45 JST):_ 

https://github.com/user-attachments/assets/8c4868b5-8b84-4851-acb2-3255d283ff61

_After:_

https://github.com/user-attachments/assets/3962b207-b092-48e8-8b03-368ea98253b1

_(REQUIRED)_

Fixed https://github.com/mealie-recipes/mealie/issues/5818

## Testing

I tested this change manually by running the dev server and using the dialog, after having first reproduced the issue (as shown in the video above).

### DIFF
--- a/frontend/components/Domain/Recipe/RecipeLastMade.vue
+++ b/frontend/components/Domain/Recipe/RecipeLastMade.vue
@@ -119,6 +119,7 @@
 
 <script setup lang="ts">
 import { whenever } from "@vueuse/core";
+import { formatISO } from "date-fns";
 import { useUserApi } from "~/composables/api";
 import { alert } from "~/composables/use-toast";
 import { useHouseholdSelf } from "~/composables/use-households";
@@ -148,7 +149,7 @@ const newTimelineEventImageName = ref<string>("");
 const newTimelineEventImagePreviewUrl = ref<string>();
 const newTimelineEventTimestamp = ref<Date>(new Date());
 const newTimelineEventTimestampString = computed(() => {
-  return newTimelineEventTimestamp.value.toISOString().substring(0, 10);
+  return formatISO(newTimelineEventTimestamp.value, { representation: "date" });
 });
 
 const lastMade = ref(props.recipe.lastMade);
@@ -169,7 +170,7 @@ whenever(
   () => madeThisDialog.value,
   () => {
     // Set timestamp to now
-    newTimelineEventTimestamp.value = new Date(Date.now() - new Date().getTimezoneOffset() * 60000);
+    newTimelineEventTimestamp.value = new Date();
   },
 );
 


frontend/components/Domain/Recipe/RecipeLastMade.vue
```diff
@@ -148,7 +149,7 @@ const newTimelineEventImageName = ref<string>("");
 const newTimelineEventImagePreviewUrl = ref<string>();
 const newTimelineEventTimestamp = ref<Date>(new Date());
 const newTimelineEventTimestampString = computed(() => {
-  return newTimelineEventTimestamp.value.toISOString().substring(0, 10);
+  return formatISO(newTimelineEventTimestamp.value, { representation: "date" });
```
lehnerpat, line 152: comment:

The main semantic change here is that previously, it seems `newTimelineEventTimestamp` was intended to be in UTC, but there were some issues where it wasn't (when it was set by `<v-date-picker>`).
`Date`'s `.toISOString()` method returns a string representation in UTC always, so if the `Date` object itself isn't, then the date can be off.

After this change, we let `newTimelineEventTimestamp` be in the local time zone, and use `date-fns`'s `formatISO` to do the (date-only) ISO formatting, which does this using the local time zone.

frontend/components/Domain/Recipe/RecipeLastMade.vue
```diff
@@ -169,7 +170,7 @@ whenever(
   () => madeThisDialog.value,
   () => {
     // Set timestamp to now
-    newTimelineEventTimestamp.value = new Date(Date.now() - new Date().getTimezoneOffset() * 60000);
+    newTimelineEventTimestamp.value = new Date();
```
lehnerpat, line 173: comment:
It seems the old code here tried to manually convert the `Date` to UTC.
This behavior is different from what I observed coming out of the `<v-date-picker>` component (this seems to return "start of the selected day in the local time zone").

As mentioned above, we now leave the time zone logic to `date-fns`, and just use the `Date` as-is in the local time zone here.

mariodz95:
I started working on the same code to change the format of this date, can you include it in your PR #6064 ?

michael-genson (approved):
Looks great, thanks!
I think behavior changed between Vuetify 2/3 which is why this broke the way it did. Whatever the case this is much better

lehnerpat:
@mariodz95 Sorry, I didn't get around to adjusting this PR to include proper date formatting.
Feel free to include it in your changes now that it's merged, or I can send a follow up PR soon :)

PR #6089, opened 2025-09-03
fix: Make Frontend Respect TOKEN_TIME

### LINKED ISSUE #6087: [BUG] - I get logged out every few days
### First Check

After years without any problems I now get logged out every few days or week.
I don't know how to troubleshoot it so please help 🙂
I get the feeling it is after a automatic update by watchtower. But I have created volumes for storage. So where are the active sessions stored? 

### Steps to Reproduce

Dunno

Don't think there are relevant log entries. 

### Mealie Version

Versie
v3.1.2
Build
894162a669cef9fefea89e786f40128b8c6e621c

### Deployment

Docker (Linux)

Docker compose

michael-genson:

_(REQUIRED)_

In the Nuxt 3 upgrade an arbitrary auth timeout of 7 days was introduced. This PR instead tells the frontend to read the `TOKEN_TIME` variable.

We don't really need to set an auth timeout since the backend handles it, but it's probably better this way so it doesn't sit in the user's application data. Plus it was a 2-second fix.

_(REQUIRED)_

Fixes https://github.com/mealie-recipes/mealie/issues/6087

Looking into the Nuxt 3 auth configuration there's a lot of room to improve how we handle auth (e.g. we could set a short auth time and implement a refresh token). Something worth looking into eventually, probably.

## Testing

Set this var to "1" and logging-in immediately logs me out, just to confirm this is the right var.


Kuchenpirat (approved):
LGTM 👍

PR #6092, opened 2025-09-03
fix: set touchless on AppSidebar

p0lycarpio:

_(REQUIRED)_

When browsing on mobile, side-swipe gestures (for back) unexpectedly open the context menu. Added the `touchless` attribute on AppSidebar to prevent this.

_(REQUIRED)_

N/A


Kuchenpirat (approved):
Thanks :)
LGTM 👍

PR #6106, opened 2025-09-05
feat: Improve first time setup ux

p0lycarpio:

_(REQUIRED)_

This PR aims to improve the user interface of the first Mealie configuration. It notably adds:
- modern navigation path using Vuetify's steppers
- responsive interface that works on smaller screens
-  page was switched to  `setup script`

Here some screenshots (on desktop and mobile view with dark-mode) :

Capture d’écran du 2025-09-05 16-28-40
Capture d’écran du 2025-09-05 16-29-56
Capture d’écran du 2025-09-05 16-30-24
Capture d’écran du 2025-09-05 16-32-16
Capture d’écran du 2025-09-05 16-33-32
Capture d’écran du 2025-09-05 16-33-59

_(REQUIRED)_

N/A but a warning was fixed when validating the v-model. It now accept object.

The strings are already translated but are not in "admin.setup" object. Should we copy them into?

## Testing

Process manually tested by validating that all API calls works by running `task dev:clean`


michael-genson:
Haven't had a chance to pull this one yet, just wanted to say looks great! Breaks my heart that we got rid of my custom `BaseWizard` component but I definitely had no idea ~~Vuetify has this built in (and seems much better)~~ ah, seems it was added quite recently in v3.4.0, that makes sense.

Would you be able to add some screenshots to the new flow? It should help reviewers.

> The strings are already translated but are not in "admin.setup" object. Should we copy them into?

No it's fine if you're using strings from other sections, copying them would trigger a new translation in Crowdin. Mealie crosses strings all over the place so it's not that big of a deal.

p0lycarpio:
I edited the description to add images. The `BaseWizard` component was already very good, thank you very much for your work!

frontend/components/global/AutoForm.vue
```diff
@@ -33,8 +33,9 @@
         <!-- Check Box -->
         <v-checkbox
           v-if="inputField.type === fieldTypes.BOOLEAN"
-          v-model="modelValue[inputField.varName]"
+          v-model="model[inputField.varName]"
           :name="inputField.varName"
+          :readonly="(inputField.disableUpdate && updateMode) || (!updateMode && inputField.disableCreate) || (readonlyFields && readonlyFields.includes(inputField.varName))"
           :disabled="(inputField.disableUpdate && updateMode) || (!updateMode && inputField.disableCreate) || (disabledFields && disabledFields.includes(inputField.varName))"
```
michael-genson: Is there a way to not have to duplicate this? Maybe make "disabled" a computed property?

frontend/pages/admin/setup.vue
```diff
@@ -1,479 +1,581 @@
...
-          </v-btn>
-          <v-card-text class="subtitle px-0 py-2">
-            {{ link.description }}
-          </v-card-text>
-        </div>
-      </v-container>
-    </BaseWizard>
-  </v-container>
+            <path
+              d="M8.1,13.34L3.91,9.16C2.35,7.59 2.35,5.06 3.91,3.5L10.93,10.5L8.1,13.34M13.41,13L20.29,19.88L18.88,21.29L12,14.41L5.12,21.29L3.71,19.88L13.36,10.22L13.16,10C12.38,9.23 12.38,7.97 13.16,7.19L17.5,2.82L18.43,3.74L15.19,7L16.15,7.94L19.39,4.69L20.31,5.61L17.06,8.85L18,9.81L21.26,6.56L22.18,7.5L17.81,11.84C17.03,12.62 15.77,12.62 15,11.84L14.78,11.64L13.41,13Z"
```
michael-genson: I know we use this in a few places, can we make this a component?

michael-genson (changes requested):
Looks great! A few small things below, and a couple more here:
- The submit button on the final confirmation is grey. Can we make it green to make it consistent with the rest of the app? If you're using `BaseButton` there's a prop for submit buttons
- On the post-confirmation page the "Home" button is also grey, that should be the primary theme color (and also if we could restore the "home" icon that would be great, but optional)
image

frontend/components/global/AutoForm.vue
p0lycarpio: I created a computed property `fieldState` that allows you to manage both disabled and readonly states, as the conditions are similar...

frontend/pages/admin/setup.vue
p0lycarpio: The `AppLogo.vue` component was created and used on pages that used the hard-coded icon. On the login page, the size is now 75px, consistent with register and setup pages.

p0lycarpio:
Thanks for your feedback ! I've restored the appearance of the buttons and implemented the other suggestions

frontend/pages/login.vue
```diff
@@ -236,6 +217,7 @@ import { usePasswordField } from "~/composables/use-passwords";
 import { alert } from "~/composables/use-toast";
 import { useAsyncKey } from "~/composables/use-utils";
 import type { AppStartupInfo } from "~/lib/api/types/admin";
+import AppLogo from "~/components/global/AppLogo.vue";
```
michael-genson: Since this is a global component you don't need to import it (it throws a lint error if you do)

frontend/components/global/AppLogo.vue
```diff
@@ -0,0 +1,45 @@
...
+        </svg>
+      </slot>
+    </v-avatar>
+  </div>
+</template>
+
+<style scoped>
+.icon-white {
+  height: 75px;
+  width: 75px;
```
michael-genson: Can we parametrize this with props? a `size` prop that defaults to `75` will work.

frontend/components/global/AppLogo.vue
```diff
@@ -0,0 +1,45 @@
+<template>
+  <div class="icon-container">
+    <v-divider class="icon-divider" />
+    <v-avatar
+      :class="['pa-2', 'icon-avatar']"
+      color="primary"
+      size="75"
```
michael-genson: This will also be the `size` prop

michael-genson (changes requested):
Just a few more small things, then I think this is good to go!

frontend/pages/login.vue
```diff
@@ -52,26 +52,7 @@
...
-      <div class="icon-container">
-        <v-divider class="icon-divider" />
-        <v-avatar
-          class="pa-2 icon-avatar"
-          color="primary"
-          size="100"
-        >
-          <svg
-            class="icon-white"
-            style="width: 100px; height: 100px"
```
michael-genson, line 65: We can switch this back to 100 if we parametrize the size

michael-genson (approved):
LGTM, thank you!

PR #6107, opened 2025-09-05
fix: Don't open the sidebar drawer by default on medium screens

noxhirsch:
When you open Mealie in a window with a width between 760 and 1280 pixels (or change the browser width to this size), the sidebar drawer always opens automatically and overlays the UI. This also happens everytime I switch between portrait and landscape on my smartphone. This is quite annoying.
I have adjusted it so that below 1280px, the sidebar drawer must be opened manually, as was previously the case below 760px. After simply adjusting the pixels in the code did not lead to a good result, I decided to use useDisplay() as with the sidebar in `DefaultLayout.vue`.

Before (on reload or browser width change):
image
After (on reload or browser width change)
{F810D913-480C-429E-951B-F614752099E5}

_(REQUIRED)_

_(REQUIRED)_

Please take a close look during the review - I'm just getting started with Vue and trying to understand everything ;)

EDIT: of course tested on my local machine

### DIFF
--- a/frontend/layouts/admin.vue
+++ b/frontend/layouts/admin.vue
@@ -43,7 +43,7 @@ const { $globals } = useNuxtApp();
 
 const sidebar = ref<boolean>(false);
 onMounted(() => {
-  sidebar.value = !display.lgAndUp.value;
+  sidebar.value = display.lgAndUp.value;
 });
 
 const topLinks: SidebarLinks = [
--- a/frontend/layouts/admin.vue
+++ b/frontend/layouts/admin.vue
@@ -43,7 +43,7 @@ const { $globals } = useNuxtApp();
 
 const sidebar = ref<boolean>(false);
 onMounted(() => {
-  sidebar.value = !display.md.value;
+  sidebar.value = !display.lgAndUp.value;
 });
 
 const topLinks: SidebarLinks = [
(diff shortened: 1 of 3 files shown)


michael-genson (changes requested):
Thanks for this! The change is great. I think the implementation would be cleaner if we adjust where we change this value, since it's a prop. Everywhere the `AppSidebar` component is used (only two places: the layouts `admin.vue` and `DefaultLayout.vue`) we have similar logic except it's set to `mdAndUp` and `md`. Can we change the value there instead?

So:
- On `AppSidebar.vue` completely remove the `onMounted` hook (and `useDisplay`)
- On `admin.vue` change `sidebar.value` to `display.lgAndUp.value`
- On `DefaultLayout.vue` do the same

That way we're not overriding the prop in the component.

noxhirsch:
Thank you for the very helpful review! I've implemented the requested changes :)

frontend/layouts/admin.vue
```diff
@@ -43,7 +43,7 @@ const { $globals } = useNuxtApp();
 
 const sidebar = ref<boolean>(false);
 onMounted(() => {
-  sidebar.value = !display.md.value;
+  sidebar.value = !display.lgAndUp.value;
```
michael-genson: I think this should be

```suggestion
  sidebar.value = display.lgAndUp.value;
```

michael-genson (changes requested):
Double check me but I think we need to invert the logic on the admin layout

noxhirsch:
I'm sorry, you're absolutely right. Fixed and tested.

michael-genson (approved):
Thank you!

PR #6108, opened 2025-09-05
fix: Context Menu Dialogs Not Working

michael-genson:

_(REQUIRED)_

https://github.com/mealie-recipes/mealie/pull/6071 fixed some major performance issues with the recipe context menu, but broke the dialogs. Since they're wrapped in a `v-menu`, when the menu closes, the dialogs unmount.

This PR conditionally adds the `eager` attribute to the content which mounts the menu content even if the menu isn't open. It only switches to eager if the menu is opened once, maintaining the performance improvements from https://github.com/mealie-recipes/mealie/pull/6071.

Also fixed an unrelated console warning for an emit we're not even using. Leaving the functionality in case we add it back in (there are several discussions around it).

_(REQUIRED)_

Reported by user in comment on performance issue: https://github.com/mealie-recipes/mealie/issues/5206#issuecomment-3258126232

## Testing

Manually verified it works as expected.

### DIFF
--- a/frontend/components/Domain/Recipe/RecipeContextMenu/RecipeContextMenu.vue
+++ b/frontend/components/Domain/Recipe/RecipeContextMenu/RecipeContextMenu.vue
@@ -3,6 +3,7 @@
     <v-menu
       offset-y
       start
+      :eager="isMenuContentLoaded"
       :bottom="!menuTop"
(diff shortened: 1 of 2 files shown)


noxhirsch:
This fixed it for me 👍 Thank you!

Kuchenpirat (approved):

PR #6110, opened 2025-09-05
feat: Put calendar directly in the date picker dialogs

noxhirsch:
If you want to add recipes to the meal plan, you first have to click on a read-only text field to open the calendar, even though there is enough space for the calendar even on small smartphones. I therefore implemented the calendar directly into the dialog and removed the unnecessary date field (I can add it again, if you prefer to show both, but it looks cleaner and less technical without).
In the process, I also applied the primary color to the calendar to make it look a little more lively and enabled the adjacent months, which makes it much easier to plan recipes for the new month at the end of a month.

Before:
image

After:
image

How it looks on a virtual iPhone SE (the smallest virtual device in Edge Dev Tools):
image

_(REQUIRED)_

None
_(REQUIRED)_

## Testing
Tested on my local machine in a dev container


noxhirsch:
I've also updated the calendar of this dialog:
image

michael-genson (approved):
LGTM, thanks!

PR #6127, opened 2025-09-08
fix: Upgrade Vuetify, fix Dev Dependencies, and fix Migration Tree View

### LINKED ISSUE #5844: [BUG] - time picker & treeview components not availlable in currently used vuetify version

michael-genson:

_(REQUIRED)_

Pins/upgrades Vuetify to the newest version, giving us back `v-treeview`. The API changed slightly so I fixed it and also simplified the id generation.

I noticed a bunch of dev deps that should be in dev deps and not the production deps, so I moved those.

_(REQUIRED)_

Fixes https://github.com/mealie-recipes/mealie/issues/5844

## Testing

Built the prod image to double check moving deps from deps to devDeps didn't break anything.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -1,6 +1,6 @@
 {
   "name": "mealie",
-"version": "3.1.2",
+  "version": "3.1.2",
   "private": true,
   "scripts": {
     "dev": "nuxt dev",
@@ -19,10 +19,8 @@
   },
   "dependencies": {
     "@mdi/js": "^7.4.47",
-    "@nuxt/eslint": "1.2.0",
     "@nuxt/fonts": "^0.11.4",
     "@nuxtjs/i18n": "^9.2.1",
-    "@nuxtjs/proxy": "^2.1.0",
     "@sidebase/nuxt-auth": "0.10.0",
     "@vite-pwa/nuxt": "0.10.6",
     "@vueuse/core": "^12.7.0",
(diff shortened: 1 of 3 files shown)


frontend/package.json
```diff
@@ -19,10 +19,8 @@
   },
   "dependencies": {
     "@mdi/js": "^7.4.47",
-    "@nuxt/eslint": "1.2.0",
     "@nuxt/fonts": "^0.11.4",
     "@nuxtjs/i18n": "^9.2.1",
-    "@nuxtjs/proxy": "^2.1.0",
```
michael-genson, line 25: We just don't use this so it's completely removed. Everything else I removed here was moved to `devDependencies`

p0lycarpio:
Thanks for that!

Other dependencies should be updated, such as `nuxt-auth`, `vite-pwa`, `vueuse/core`, which have released new major versions with no major changes. I can take care of creating a PR once it has been merged.

After that, we can also look into updating `nuxt`, `marked`, `i18n`, or `vite`, but we'll have to see if a migration is necessary.

michael-genson:
Yeah this is more to fix a specific issue with missing the tree component. In general we should turn on auto updates now that we're on Nuxt 3 but we need to do some manual updates first (e.g. node)

Kuchenpirat (approved):

PR #6130, opened 2025-09-08
dev: list availlable frontend updates on renovate dependency dashboard

Kuchenpirat:

With many things in the frontend being more up to date it might be nice to have an overview over what can be upgraded, what has new major versions etc...
So this will show availlable updates in a non intrusive way to the rest of the repo.

None

This will only show the updates in the dependecy dashboard, where a corresponding checkbox can be ticked which will then create a normal renove pr with the updates. 

## Testing

Not yet, we will see what happens in the removate dashboard i guess.


michael-genson (approved):
LGTM. At some point I think it will make sense to automatically open PRs like we do with the backend, but for now this is great

Kuchenpirat:
Yeah, I agree.
But without solid tests we might want to do less frequent merging

michael-genson:
Yeah that makes sense

PR #6149, opened 2025-09-10
feat: Add snack, drink, and dessert

cashpw:
## What?

1. Add new meal plan entry types: snack, drink, and dessert

## Why?

Fixes #5648.

This is easier than the full CRUD proposal, database and UI updates, etc which would be involved in #2301.

We'll need translations for "snack", "drink", and "dessert". I'm not sure what our process is for adding new translation strings. I've added the English translations. Should I add placeholder values to the other language files?

## Testing

I've omitted new tests because this feels like adding to an enum and new change-detector tests aren't good for the project, and there's already test coverage for the meal planner.


noxhirsch:
I'm not sure what I think about this ... "prepare" and "leftovers" would also be useful entries for some (including me). But I don't want to scroll through a list of meal types I don't use (that would at least be breakfast, side, drink & dessert) everytime I add a recipe to the meal planner.
If we want to go an easy route: What about just adding something like "miscellaneous" for everything thats not breakfast, lunch & dinner (I personally would replace "side" with "miscellaneous", but I understand that probably many users don't want to loose that).

cashpw:
I had meant for this pull request to be exactly that: an easy route.

Miscellaneous wouldn't address my use case, which would specifically be
aided by the inclusion of snack and drink. However, perhaps that is okay
for now.

To look at the non-easy route for a moment, is there anything non-obvious
you can think of which would be required to convert the hard-coded list of
types here into a dynamic list populated by user choice? And, are you
interested in a pull request which would change this list to be dynamic and
defined by user choice?

 As far as I can tell, we would need:

- A new UI element for adding a type (possibly the same one as is used for
tags, categories, and tools) and a decision on where we'd place it in the
meal planner view
- An inclusion in the data management view
- (?) A new table in the db (I haven't looked at how tags and such are
stored. Maybe it wouldn't require a full table)
- An update to all code sites which currently use the explicit list of
breakfast, lunch, and dinner
- A decision on whether breakfast, lunch and dinner can be deleted or
edited by users. Right now they are assumed to be present in tests and UI
elements. Perhaps those three are not made editable through the UI.

noxhirsch:
I think having customizable meal types would be the way better solution. However, I can't speak for the project in any way - I'm just a relatively new user who has only opened two basic frontend PRs so far (I also don't know anything about the backend at the moment). But michael-genson & Kuchenpirat also seem to share this opinion in a similar PR: https://github.com/mealie-recipes/mealie/pull/5649#issuecomment-3164736487

cashpw:
Great catch! I should have scrolled through the existing pull requests.
I'll take a look at the schema and see what I can find.

michael-genson:
Yeah as @noxhirsch pointed out I think I'd prefer to generalize this and let the user come up with values on their own via the data management page (similar to how we do foods and food aliases). If you're not up for that I'm fine merging this PR as-is, knowing we will remove it later.

One note: PRs should only contain changes to the `en-US` translation file, Crowdin handles the rest.

cashpw:
Great points!

Yeah, I would like to get this in as is possible and then have a follow-up
come through eventually for setting custom values. This seems like it'll
meet most people's needs until the full solution can be developed.

michael-genson (changes requested):
Looks good, though a few issues I ran into:
1. When selecting a "random" new recipe the new options are missing:
image

2. It doesn't look like these are usable in the meal plan rules:
image

3. We'll need to remove the translations from all files except `en-US`

cashpw:
Sorry for the delay.

>     1. When selecting a "random" new recipe the new options are missing:

Done.

>     2. It doesn't look like these are usable in the meal plan rules

Done.

>     3. We'll need to remove the translations from all files except `en-US`

Done.

cashpw:
@michael-genson Do you see any other issues with this pull or are we good to submit it?

michael-genson:
Apologies, been working my way through other reviews. I will get to this eventually, I promise!

michael-genson (approved):
LGTM, thanks for your work on this!

PR #6169, opened 2025-09-14
feat: frontend autocomplete is diacritics/ligatures insensitive

Noneangel:
In the UI, editing `food `in  `group/data/foods/` is now accent/diacritics ligatures insensitive. This change is frontend only.

It was normalized in backend for searches results if I understood this test (maybe not for ligature like `œ` and `æ`) : https://github.com/mealie-recipes/mealie/blob/9a0525c3a0c795a3d902dda620ce8b7860f8ab11/tests/unit_tests/repository_tests/test_recipe_repository.py#L98

Fixes https://github.com/mealie-recipes/mealie/discussions/5267

This is my first contribution to this repo, and my first time using Vue.js.

I put the function `normalizeFilter` in [use-utils.ts](https://github.com/mealie-recipes/mealie/blob/mealie-next/frontend/composables/use-utils.ts). Should I move it some-where else?

There are two files containing autocomplete that I did not change (I don't know if this change should be applied to them):
- [RecipeOrganizerSelector.vue](https://github.com/mealie-recipes/mealie/blob/mealie-next/frontend/components/Domain/Recipe/RecipeOrganizerSelector.vue)
- [LanguageDialog.vue](https://github.com/mealie-recipes/mealie/blob/mealie-next/frontend/components/global/LanguageDialog.vue)

This PR is based on this SO: https://stackoverflow.com/a/37511463

I use two functions/features:
- [Regex diacritic](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Regular_expressions/Unicode_character_class_escape#browser_compatibility)
- [String.prototype.normalize](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/normalize#browser_compatibility)

Note: I am French. I don't know how other letter should be converted.  
For instance, the German for `ä` is mapped to `a` with my code, but this [SO](https://stackoverflow.com/questions/1909815/regex-to-compare-strings-with-umlaut-and-non-umlaut-variations) suggest that it could be mapped to `ae`.

I did not notice performance impact, but if requested, I can precompute the normalization of `food`.

## Testing
- I added automated tests for the filter and normalization.
- I tested manually using a local instance.


michael-genson (approved):
LGTM, great work on this! Had to resolve a couple merge issues and tweaked a few style things. I also added normalization to the two components you mentioned (I think it's valuable in both).

PR #6192, opened 2025-09-19
dev: Partition Backend Tests In GH Actions

michael-genson:

_(REQUIRED)_

Partitions our backend tests into separate actions in GH. This should allow them to run in parallel making them much faster.
I also added a test to make sure our GH actions partitioning covers all tests (so if a new directory is added and we don't add it to GH actions our tests will fail).

_(REQUIRED)_

N/A

TBD if this actually helps but it probably will.

## Testing

👀


michael-genson:
It's barely faster 🙄 Our `user_recipe_tests` are most of the bottleneck

PR #6194, opened 2025-09-19
fix: multiple RecipeRating backend calls

p0lycarpio:

_(REQUIRED)_

This PR fixes the double call that is made when a user rates a recipe as well as the submission of a non-numeric value that results in an error 'Input should be a valid number'

_(REQUIRED)_

N/A

## Testing

Manually, by looking at network requests in the browser's developer tools

### DIFF
--- a/frontend/components/Domain/Recipe/RecipeRating.vue
+++ b/frontend/components/Domain/Recipe/RecipeRating.vue
@@ -13,7 +13,6 @@
         hover
         clearable
         @update:model-value="updateRating(+$event)"
-        @click="updateRating"
       />
       <!-- Group Rating -->
       <v-rating v-else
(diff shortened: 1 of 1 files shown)


frontend/components/Domain/Recipe/RecipeRating.vue
```diff
@@ -83,7 +82,7 @@ export default defineNuxtComponent({
     });
 
     function updateRating(val?: number) {
-      if (!isOwnGroup.value) {
+      if (!isOwnGroup.value || typeof val !== "number" || val === 0) {
```
michael-genson: Since this is either a number or undefined, could we simplify this?
```suggestion
      if (!isOwnGroup.value || !val) {
```

michael-genson (approved):
Thank you!

PR #6215, opened 2025-09-21
fix: Actually Fix Token Time

michael-genson:

_(REQUIRED)_

While debugging the [broken token time](https://github.com/mealie-recipes/mealie/pull/6089) I noticed we have two access tokens in our cookies: one for the frontend and one for the backend. This doesn't make any sense.

This PR removes the frontend-specific token (`mealie.auth.token`) and relies solely on the backend token (`mealie.access_token`). Previously the backend token was HTTP only for security, but that security is moot if we also have a non-HTTP-only token. In order for the frontend to use the backend token I set http only to `False` which isn't an _increased_ security risk because we already had one.

Also fixes and refactors a couple related things to get this to work properly.

_(REQUIRED)_

N/A, thought to be fixed by https://github.com/mealie-recipes/mealie/pull/6089 but it wasn't, only in dev.

## Testing

Checked the prod docker image and made sure it actually works in prod this time.


jonathanjeurissen-quantum:
@michael-genson First of all, thank you so much for fixing this bug, it’s been driving me crazy for the past couple of weeks, and the behavior has significantly improved!

I’ve been trying to set the token time to 2 years (17520 hours) on version 3.1.1, but according to my browser’s developer tools, the `mealie.access_token` cookie is still set to expire in just over 1 year (currently maxing out at Sun, 08 Nov 2026). Increasing the value doesn’t seem to extend the expiration date beyond that.

Additionally, I noticed that even when setting TOKEN_TIME to the minimum of 1 hour, the cookie still gets created with a lifespan of around 2 weeks. This might be problematic for others who rely on shorter-lived tokens for security or session management.

PR #6235, opened 2025-09-24
feat: Add user QueryFilter and improve UI on mobile

p0lycarpio:

_(REQUIRED)_

The need stems from a basic idea: creating a recipe book or collecting recipes that belong to a user. Here is what has been added:
- Added "users" as a query filter so that recipes or cookbooks can be filtered using one more criteria, just like households.
- Updated the `QueryFilterBuilder` component for column display on mobile screens (see below).

Before/after :

Capture d’écran 2025-09-24 à 22 54 41

Capture d’écran 2025-09-24 à 22 55 54

_(REQUIRED)_

N/A

This PR has no impact on the backend. The API correctly takes the `user_id` field and logical operators as parameters to perform queries.
Users who use the “users” filter should only see other users in their group, as this is scoped in the store and the backend would not allow otherwise.

## Testing

Manually via cookbooks and recipe finder pages and inspecting networking requests


frontend/components/Domain/Recipe/RecipeOrganizerSelector.vue
```diff
@@ -19,21 +20,6 @@
...
-        color="accent"
-        variant="flat"
-        label
-
-        closable
-        @click:close="removeByIndex(index)"
-      >
-        {{ item.value }}
-      </v-chip>
-    </template>
```
michael-genson, line 35: Is there a reason our custom chip style was removed?

michael-genson (changes requested):
Generally LGTM, one note below. Additionally, can we add a separator between elements on mobile? Collapsing them into multiple rows makes sense since there's not much horizontal space, but it gets hard to see each element visually:
image

Or maybe staggered colors or a border or something, but a line is probably simplest.

p0lycarpio:
Of course! I've added offset colors for mobile readability and restored the custom chip (and item-value). A few adjustments have been made to the spacing.

frontend/components/Domain/QueryFilterBuilder.vue
```diff
@@ -17,14 +17,15 @@
         <v-row
           v-for="(field, index) in fields"
           :key="field.id"
-          class="d-flex flex-row flex-wrap mx-auto"
+          class="d-flex flex-row flex-wrap mx-auto pb-2"
+          :class="$vuetify.display.xs ? (Math.floor(index / 1) % 2 === 0 ? 'bg-dark' : 'bg-light') : ''"
```
michael-genson, line 21: nice

michael-genson (approved):
Looks great, thanks!

PR #6240, opened 2025-09-25
chore: Upgrade Node and Nuxt

michael-genson:

_(REQUIRED)_

Upgrades from Node 20 -> Node 22 and upgrades Nuxt to 3.19. Had to update a few other deps/subdeps as well due to the Node upgrade (by re-locking).

_(REQUIRED)_

Replaces https://github.com/mealie-recipes/mealie/pull/6184

## Testing

Built a prod image and it seems to work fine


frontend/package.json
```diff
@@ -21,17 +21,17 @@
     "@mdi/js": "^7.4.47",
     "@nuxt/fonts": "^0.11.4",
     "@nuxtjs/i18n": "^9.2.1",
-    "@sidebase/nuxt-auth": "0.10.0",
+    "@sidebase/nuxt-auth": "1.1.0",
```
p0lycarpio: Can you unpin this library to automatically use the latest minor or patch version?

michael-genson: Yeah this PR is kind of a mess because it's causing E2E issues

PR #6248, opened 2025-09-26
fix: Update the random button flow

aliyyanWijaya:

- Fixes an issue where the Random button would not trigger again when clicked repeatedly. Previously, the user had to click another button (e.g., Updated or Alphabetical) before Random could work again.
- Added a refresh icon that appears after the Random button is clicked for the first time. This icon indicates that the Random button is active and can be used again.
- Attached before/after demo videos to illustrate the fix and new behavior.

N/A (feature improvement / bug fix not linked to GitHub issue)

- Please share any feedback regarding the added function and the icon behavior.
- Let me know if you think the icon placement/appearance should be adjusted or if an alternative indicator would be clearer for users.

## Testing

Manually tested by repeatedly clicking the Random button:
- Verified that it triggers new results every time.
- Confirmed that switching to other sort buttons (Updated, Alphabetical) still works as expected.
- Verified the refresh icon appears after the first click and remains as an indicator.

Before:
[Screencast from 2025-09-16 22-49-55.webm](https://github.com/user-attachments/assets/56e84f95-5340-4bce-afb5-54f14d7fd266)

After:
[Screencast from 2025-09-16 22-51-38.webm](https://github.com/user-attachments/assets/9b100104-837e-450e-b142-f56ed4643f60)


michael-genson (approved):
LGTM, thanks!

PR #6259, opened 2025-09-27
feat: Add `DELETE /{slug}/image`

chollinger93:

_(REQUIRED)_

- Adds `DELETE /{slug}/image`
- Frontend: Adds a button for the same
- Frontend: Adds a generic `AppButtonDelete` with a confirmation dialog element

I couldn't find this functionality and since I often copy similar recipes (e.g. permutations for protein ice cream), I am stuck with an old picture until I actually make the new recipe and snap a pic.

_(REQUIRED)_

Fixes #6258

(I might have skipped the feature request step and just built this since it bothers me on the daily, hope that's OK)

## Testing

[output_small.webm](https://github.com/user-attachments/assets/12c2a560-5d52-4ad5-8225-be1c7fc3bba1)

## Caveats
I'm not 100% sure if images are _always_ just the ones defined in `RecipeImageTypes`. Could somebody confirm this?

Fair warning, I also usually do backend, so my frontend-skills are only so-and-so.


michael-genson (approved):
LGTM, thanks for your work on this! I pushed a few changes/fixes

PR #6266, opened 2025-09-27
fix: Stores Not Populating Sometimes

michael-genson:

_(REQUIRED)_

Sometimes our stores don't populate correctly due to our async usage (especially on first load). This PR fixes some of the hydration logic and eliminates duplicate requests with smarter async keys.

_(REQUIRED)_

N/A

There's an opportunity for caching here with how the async data stores work, but I didn't want to implement that in this PR since it would require a lot of testing to make sure nothing breaks.

## Testing

Clicked around a lot to make sure things are populated correctly. Also did some basic CRUD to make sure refreshes work as expected.

### DIFF
--- a/frontend/composables/partials/use-actions-factory.ts
+++ b/frontend/composables/partials/use-actions-factory.ts
@@ -1,10 +1,11 @@
 import { useAsyncKey } from "../use-utils";
+import type { AsyncData, NuxtError } from "#app";
 import type { BoundT } from "./types";
 import type { BaseCRUDAPI, BaseCRUDAPIReadOnly } from "~/lib/api/base/base-clients";
 import type { QueryValue } from "~/lib/api/base/route";
 
 interface ReadOnlyStoreActions<T extends BoundT> {
-  getAll(page?: number, perPage?: number, params?: any): Ref<T[] | null>;
+  getAll(page?: number, perPage?: number, params?: any): AsyncData<T[] | null, NuxtError<unknown> | null>;
   refresh(page?: number, perPage?: number, params?: any): Promise<void>;
 }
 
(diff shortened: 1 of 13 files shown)


frontend/composables/partials/use-actions-factory.ts
```diff
@@ -29,7 +30,7 @@ export function useReadOnlyActions<T extends BoundT>(
     params.orderBy ??= "name";
     params.orderDirection ??= "asc";
 
-    const allItems = useAsyncData(useAsyncKey(), async () => {
+    const allItems = useAsyncData(storeKey, async () => {
```
michael-genson, line 33: Not 100% sure we needed this change but `useAsyncKey` was always dubious to me and seemed to cache things weirdly.

frontend/composables/partials/use-store-factory.ts
```diff
@@ -29,21 +30,22 @@ export const useReadOnlyStore = function <T extends BoundT>(
     },
   };
 
+  // initial hydration
   if (!loading.value && !store.value.length) {
-    const result = actions.getAll(1, -1, params);
-    store.value = result.value || [];
+    actions.refresh();
   }
```
michael-genson, line 36: Part of the actual fix. This logic didn't work right due to the return type of `getAll` being wrong.

frontend/composables/use-utils.ts
```diff
@@ -8,7 +8,7 @@ export const useToggleDarkMode = () => {
 };
 
 export const useAsyncKey = function () {
-  return String(Date.now());
+  return `${Date.now()}-${Math.random().toString(36).slice(2)}`;
 };
```
michael-genson, line 12: The other part of the fix. Requests fired quickly can have the same async key if we just rely on the date, causing some requests not to fire.

PR #6274, opened 2025-09-28
fix: Keep Linked Recipes After Being Checked In Shopping Lists Page

### LINKED ISSUE #5213: [BUG] - Shopping list items dont keep their linked recipies after being checked then unchecked
### First Check

When an item is sent from a recipe to the shopping list, it adds a button to see what recipe/s added that item to the list. However, since on mobile tapping anywhere on the left side of the screen sends an item to the checked item box, we find ourselves accidentally checking off items more frequently than we would like. When this happens, we lose the ability to see what recipe this item came from. 

### Steps to Reproduce

1. open a recipe,
2. add the ingredients from the recipe to a shopping list
3. check the item in the shopping list
4. uncheck the same item from the shopping lists checked items
5. find that the button to see the recipe is gone

N/A

### Mealie Version

2.7.1

### Deployment

Docker (Linux)

docker compose

aliyyanWijaya:

This PR fixes the issue where shopping list items lose their linked recipe references after being checked and then unchecked.
Currently, once an item is unchecked, the button to view the linked recipe(s) disappears, which makes it harder for users to trace where the item originated. This change ensures that recipe links remain intact throughout the check/uncheck process.

Fixes #5213

I didn't delete the code that remove the reference recipe in the backend, only comment it out, please let me know what is the best way to make this work.

## Testing

Verified that after applying the fix:

- Adding recipe ingredients to a shopping list still works.
- Checking and unchecking items does not remove recipe link buttons.
- Toggle recipe icons remain visible and functional.

[Screencast from 2025-09-28 22-55-58.webm](https://github.com/user-attachments/assets/46ea0ec3-2364-46c3-bdb8-1f879df29ae2)


boc-the-git:
Sorry, I meant to approve workflows and accidentally un-drafted the PR. Have set it to draft again.
What are you waiting for before setting this ready for review?

aliyyanWijaya:
All good now, it is ready to be reviewed

michael-genson:
You can go ahead and just delete the code instead of commenting it out. You'll also need to update the `test_shopping_list_ref_removes_itself` test, if you could rename it to `test_shopping_list_ref_persists_when_checked` (or something similar) and update the test logic that would be great.

michael-genson:
One problem I can see is the recipe refs at the shopping list level; those should be dynamically updated based on when an item is checked/unchecked. They should not include recipes for items that are checked off, and they need to be restored when the item is unchecked.

aliyyanWijaya:
Thank you for the feedback, I am still working to update the test and keep the recipe on the bottom goes on and off based on the item list.

michael-genson:
Hey there, just checking in, are you still working on this?

aliyyanWijaya:
Hi @michael-genson, I have updated the PR following your suggestions:

1. Backend: Reverted the permanent deletion of recipe references in shopping_lists.py.
2. Frontend: Implemented dynamic filtering in [id].vue. Linked recipes now disappear when all items are checked and reappear when unchecked.
3. Tests: Renamed and updated test_shopping_list_ref_removes_itself to test_shopping_list_ref_persists_when_checked.

Thank you for the guidance!

michael-genson (changes requested):
Looks better, though I ran into a couple of new issues:

---

Now that checked off items keep their recipe references, removing a recipe will update the checked-off items (which we don't want). For example:
1. Have a recipe with 1 egg and 1 slice bacon
2. Add two recipes to your shopping list
3. Check off 2 eggs
4. Remove one recipe
5. Observe your checked off item now says "1 egg" (was: 2)

---

Since recipe references are not deleted, they will eventually accumulate indefinitely. When we delete shopping list items we should also delete unused references somehow.

You may want to just build this into a daily task similar to `delete_old_checked_list_items`, instead of trying to figure it out when an item is deleted, though either way works.

---

It may be easier to instead keep the old behavior of deleting recipe references when no unchecked items reference them, and instead opt to create new ones when items are unchecked. That could potentially solve both of the above issues and be a bit simpler.

i.e.
- When a shopping list item is checked off, use existing logic which decrements or deletes list references (but keep item references)
- When a shopping list is unchecked, calculate if we need to create a new recipe reference (or increase the amount on an existing one)

tommyalatalo:
> Now that checked off items keep their recipe references, removing a recipe will update the checked-off items (which we don't want). For example:
> 
>     1. Have a recipe with 1 egg and 1 slice bacon
> 
>     2. Add two recipes to your shopping list
> 
>     3. Check off 2 eggs
> 
>     4. Remove one recipe
> 
>     5. Observe your checked off item now says "1 egg" (was: 2)

I thought this makes sense? What behavior is it you're saying should be there instead?
Doesn't it make sense that if you have 2 recipes, you check the eggs off, and then you change your mind and remove one recipe, you would now be shopping for one batch of that recipe and the checked off list should reflect that, i.e. show 1. 

Or rather, if I ask the inverse question; why should the checked off list keep showing 2 eggs when the recipe quantities no longer require that amount?

I feel like I'm not thinking about this the same way that you are, or I'm just overlooking something.

michael-genson:
Once I check off an item from the shopping list I'm done with it. That's the item I bought. If I want to look back and say "how many eggs did I buy?" I don't want it to have changed without me knowing. It's basically performing hidden changes which is unexpected.

You also can't edit shopping list items in the UI after they're checked off for this reason. Similarly, checked-off items aren't considered when merging items, etc. That's the intent behind the original design of deleting shopping list references after checking off items (though I definitely see the value in persisting them, in case you uncheck the items).

Most shopping lists I've used (e.g. Alexa) don't let you edit checked off items, they're considered "archived" or similar (unless you uncheck them of course).

tommyalatalo:
> Once I check off an item from the shopping list I'm done with it. That's the item I bought. If I want to look back and say "how many eggs did I buy?" I don't want it to have changed without me knowing. It's basically performing hidden changes which is unexpected.
> 
> You also can't edit shopping list items in the UI after they're checked off for this reason. Similarly, checked-off items aren't considered when merging items, etc. That's the intent behind the original design of deleting shopping list references after checking off items (though I definitely see the value in persisting them, in case you uncheck the items).
> 
> Most shopping lists I've used (e.g. Alexa) don't let you edit checked off items, they're considered "archived" or similar (unless you uncheck them of course).

Okay, I think there might be yet another angle to consider. So in our example we were getting 2 eggs because we had two instances of that recipe in the shopping list. If this was then reduced to 1 I would on some level expect the checked off items list to reflect that, otherwise we still have the count of 2 which now doesn't map correctly to the recipes linked to the shopping list. But more importantly it poses the question of what will happen if you first check off the 2 eggs, then reduce the recipe count to 1, followed by unchecking the 2 eggs - is that checked off quantity of 2 now going to be suddenly changed to 1 when it goes back to the shopping list as an unchecked item? To me that seems like it would be very confusing.

This is obviously not the place to go on with a lengthy discussion, but I think it would be worthwhile to consider if both unchecked and checked ingredients should try to follow the recipe quantities added to the list and if there should be some additional user interaction to manage this - i.e. a popup dialog appearing when changes are made that would alter checked off items so that the user is made aware of this and confirm? This doesn't seem like a trivial thing to solve UX-wise to make it logical, because I would say you both want the shopping list to dynamically follow the recipe quantities, but at the same time I agree with your point when having single batches of ingredients, like if you have 1 apple linked to a recipe and checked off and if you then remove the recipe containing that ingredient from the shopping list the apple should not just automatically disappear from the list if you already picked it up in the store and want to remove it from your cart. This is where I'm thinking a dialog to choose action might make sense - one to stop the user to make a choice by showing the affected ingredients and ask whether these should be removed from the shopping list or not, or something along those lines.

michael-genson:
For now let's keep it simple:
- when an item is checked off, delete the list-level references (if no other items contain references)
- when an item is unchecked, create new list-level references (if they don't exist)
- when a list-level recipe is adjusted, don't modify checked-off items

We can always revisit this later. I don't want to put too much complexity into what's otherwise likely a very niche feature.

michael-genson:

aliyyanWijaya:
> Hey there, are you still working on this one?

I tried to make sure checked items keep their references. Could you have another look when you get a chance?

michael-genson:
I'm still not convinced of this approach, since it doesn't resolve the issue of infinitely-accumulating recipe references since they're not being deleted anymore. This was my previous suggestion:
- when an item is checked off, delete the list-level references (if no other items contain references)
- when an item is unchecked, create new list-level references (if they don't exist)
- when a list-level recipe is adjusted, don't modify checked-off items

This preserves existing behavior of cleaning up recipe references, but also enables restoring them if items are unchecked.

aliyyanWijaya:
Hi, thanks everyone for the feedback.

I’ve tried a few times but still couldn’t find a clean solution without introducing other issues. I think this approach isn’t quite right.

I’ll close this PR for now. Thanks!

PR #6308, opened 2025-10-03
feat: Added url to current version release

Choromanski:

Adds a link to the version release page:

image

Due to this not being an official release it tries to take me to a "develop" release that doesn't exist


boc-the-git:
Your description mentions a place where this fails. It also won't work for `nightly` which we know is a commonly used image tag.
I think there needs to be a solution to handle these common cases before we consider merging this. I don't have a specific solution in mind.

The 'Build' link already takes you to a commit that would link to a release, I believe.

Choromanski:
@boc-the-git I've added exemptions for nightly and develop those will show up without anchors

boc-the-git (approved):
Looks good mate, thanks!

PR #6323, opened 2025-10-06
feat: Add footer nav bar for mobile

miah120:
This PR adds a footer navbar for mobile. 
This makes it easier for users to quickly navigate between the most common activities. 

The new menu bar:
image

N/A

In the future I want to make the buttons configurable, but I wanted to submit this now that it's usable. 

## Testing
Tested in Firefox.

### DIFF
--- a/frontend/components/Layout/DefaultLayout.vue
+++ b/frontend/components/Layout/DefaultLayout.vue
@@ -91,6 +91,7 @@
         </div>
       </v-scroll-x-transition>
     </v-main>
+    <FooterNav v-if="$vuetify.display.xs" />
   </v-app>
 </template>
 
--- a/frontend/components/Layout/DefaultLayout.vue
+++ b/frontend/components/Layout/DefaultLayout.vue
@@ -91,7 +91,7 @@
         </div>
       </v-scroll-x-transition>
     </v-main>
-    <FooterNav v-if="$vuetify.display.xs" />
+    <FooterNav v-if="$vuetify.display.xs && activityPreferences.showFooter" />
   </v-app>
 </template>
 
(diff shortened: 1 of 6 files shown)


noxhirsch:
While I like the idea of having shortcuts, I'm not a fan of having both a header *and* a footer in the UI, as this reduces the available space for the more important content (shopping lists, recipes, etc.) on small devices like smartphones.
Obviously, it’s up to the maintainers of Mealie to decide whether or not to implement an additional footer. However, if they want to add something like this, I’d like to suggest an alternative approach: placing these shortcuts in the header (instead of the logo on mobile) - similar to how Home Assistant does it. Quick mockup:
image

From my point of view, this has several advantages:
- More space for the main content
- Reduced chance of accidentally navigating away, for example while shopping
- Optionally: The icons could remain visible even on larger screens (for example, centered between the logo and the search bar), which makes sense if they are configurable and can link to elements not present in the sidebar (e.g., a specific shopping list)

Anyway: I've tested this PR on my machine (Windows+Edge) and it seems to be working as expected.

miah120:
I think having a header + footer is a common enough pattern that the loss of space wouldn't really be felt, but I did make it so that the header will collapse on scroll, which gives us back that space. 

As for header buttons: The major benefit to a footer is the ergonomics. Most phone users will use their phones resting in their hand or on their pinky; with their thumb near the bottom. By keeping navigation where the thumb is, users can navigate to different activities without needing to stretch or change the positioning of their phone. With the current nav, I end up needing to use two hands just to navigate to my shopping list. 

I also added a preference for users to turn off this feature.

michael-genson:
Hey there, we've discussed this one internally and decided against it. While having a bottom nav is definitely better than a side nav for mobile use, having both adds too much clutter.

We also are guessing users also probably aren't switching tabs super often in Mealie compared to, for instance, a social media app, and typically only use one feature at a time (at least on mobile). I think [your default activities PR](https://github.com/mealie-recipes/mealie/pull/5125) does a really good job customizing this scenario for different usecases (without adding additional UI elements).

I'm going to go ahead and close this one, but feel free to re-open if you have some other ideas / want us to take another look!

PR #6328, opened 2025-10-06
fix: Remove unused next-auth dependency

p0lycarpio:

_(REQUIRED)_

Since the introduction of #6322, the `next-auth` dependency is no longer needed.

N/A


michael-genson (approved):
Thanks! Didn't realize we had this pinned explicitly along with sidebase/nuxt-auth

PR #6345, opened 2025-10-10
feat: Improve shopping list label sections

miah120:
This PR updates the ingredient section labels for the shopping list to use [Vuetify's expansion panels](https://v2.vuetifyjs.com/en/components/expansion-panels/). This allows us to simplify the code a bit while improving the UI.

### What changed
This PR only makes UI changes and has no functional changes. 
* Uses Vuetify expansion panels to manage disclosure state for shopping list label sections and the deleted items section
* Tweaks the placement in code for the reorder labels dialogue to match the other dialogues
* Spacing tweaks

### Before
image

### After
image

NA

I updated the color handling to default to white if no color is set. This is an opinionated change, so I'm willing to change it, but I think grey tends to indicate that the control is disabled. 

## Testing
Tested from docker container in firefox in light and dark mode with a variety of label colors set


frontend/pages/shopping-lists/[id].vue
```diff
@@ -154,27 +179,15 @@
...
-          </v-icon>
-          {{ key }}
-        </v-btn>
-        <v-divider />
-        <v-expand-transition>
-          <div v-if="labelOpenState[key]">
+      <v-expansion-panels v-model="openedSections" multiple>
+        <v-expansion-panel v-for="(value, key) in itemsByLabel" :key="key">
+          <v-expansion-panel-title
+            :color="getLabelColor(value[0]) ? getLabelColor(value[0]) + '40' : '#ffffff40'"
```
Comment (michael-genson) [inline frontend/pages/shopping-lists/[id].vue]: Some of our colors already supply the opacity (it's supported in the color picker) so this can throw an error. Let's just leave this part off.
```suggestion
            :color="getLabelColor(value[0]) ? getLabelColor(value[0]) : '#ffffff40'"
```

frontend/pages/shopping-lists/[id].vue
```diff
@@ -403,6 +366,10 @@ export default defineNuxtComponent({
     const { store: allUnits } = useUnitStore();
     const { store: allFoods } = useFoodStore();
 
+    // opens the first 10 sections by default;
+    // the 11th and on will be closed by default to prevent excessive scrolling
+    const openedSections = ref([0, 1, 2, 3, 4, 5, 6, 7, 8, 9]);
```
Comment (michael-genson) [inline frontend/pages/shopping-lists/[id].vue]: I was skeptical about this but looks like after items get checked off the lower sections (i.e. 10+) automatically expand which is quite nice.

frontend/pages/shopping-lists/[id].vue
Comment (michael-genson) [reply frontend/pages/shopping-lists/[id].vue]: Also, can we modify this to just return the default in `getLabelColor`? That way we can get rid of this ternary.

michael-genson (changes requested):
I like the new look! Takes a second to get used to but it grew on me pretty quickly. Just one issue I ran into.

noxhirsch:
### Categories:
For two reasons, I'd suggest to set the padding/height of the opened categories the same as the closed ones:

1. The extra height adds a lot of wasted space if you have multiple/all categories open on a smartphone
2. The height-animation starts as soon as you open a category, while the content opens a moment later, which gives you a "double animation" that looks stuttery and weird

Screenshot 2025-10-11 114845


### Ingredient lists:
I would also strongly suggest reducing (or better removing) the padding of the content boxes, as they currently add a significant amount of unnecessary blank space - especially on the left and right sides (a total of 80 px instead of 32 px which is a lot on small devices). This results in less room for the ingredients, causing more line breaks and thus a less clear overall appearance and more scrolling needed.
Screenshot 2025-10-11 114822

frontend/pages/shopping-lists/[id].vue
Comment (noxhirsch) [reply frontend/pages/shopping-lists/[id].vue]: @michael-genson I'm not a huge fan of this, since it's unexpected for the user (they don't know why some sections are closed and some aren't - and it resets every reload). Also fast scrolling is a lot easier than opening the rest of the sections ;) 

_As a sidenote to this topic: If we want to reduce excessive scrolling, there is potential to improve the compactness of the shopping list. For example, here is a screenshot of my shopping list with numerous CSS modifications (though not all of them would necessarily be suitable for the general user base):_
image

miah120:
@noxhirsch I like the changes to padding, particularly in the expansion panel content. 
One small thing: the stuttering in the animation wasn't due to the height of the category, but rather in how the content of the panels gets rendered to the dom. 
I added the eager flag which loads the content in the dom when the panel is closed, and that fixed the issue.

frontend/pages/shopping-lists/[id].vue
Comment (miah120) [reply frontend/pages/shopping-lists/[id].vue]: I opted to just have remove the ternary and not set a color if one isn't present. 

I think this works as it gives us a more theme-neutral default. (i.e., black in dark mode and white in light mode). That way if someone doesn't use labels, the section blends into the ui a bit more. 

If you feel strongly about adding the default color in the `getLabelColor` I can make the change but I figure I'd provide this option first.

noxhirsch:
@miah120 Thank you very much - I like the changes :)

> the stuttering in the animation wasn't due to the height of the category, but rather in how the content of the panels gets rendered to the dom.

I didn't mean that the height is the cause - it just made the slow loading/rendering more obvious ;)
The problem also still seems to occur with categories that are closed when the page loads.
As you can see here, the first entry looks fine (it was open on the initial load), while the other two still take a moment to appear:

https://github.com/user-attachments/assets/4fb21372-89f6-4f63-a0fc-ebfbca8815f9


However, as mentioned here, I’d prefer not to close categories on load for the reasons discussed: https://github.com/mealie-recipes/mealie/pull/6345#discussion_r2422744675 
If we load all categories opened, the problem should be irrelevant. Let’s wait and see what Michael thinks.

frontend/pages/shopping-lists/[id].vue
Comment (michael-genson) [reply frontend/pages/shopping-lists/[id].vue]: Ah perfect, I like that better

frontend/pages/shopping-lists/[id].vue
Comment (michael-genson) [inline frontend/pages/shopping-lists/[id].vue]: I'm getting some weird behavior with this when unchecking items which belong to labels not previously loaded. Seems like they're closed by default (even if they're only the 5th label, for instance).

For simplicity let's just leave this off for now and keep them all open by default, we can always change this later if it becomes a problem

michael-genson (changes requested):
LGTM, one issue I ran into regarding the auto-collapse sections, otherwise I think this is good to go

frontend/pages/shopping-lists/[id].vue
Comment (miah120) [reply frontend/pages/shopping-lists/[id].vue]: Unfortunately there doesn't seem to be an easy way to make them be open by default :( 

I'm trying out using a watch on shoppingList.itemsByLabel, but that doesn't play well with the polling the page does every ~3s

Comment (michael-genson) [reply frontend/pages/shopping-lists/[id].vue]: Could you keep a separate variable that tracks open section by name? Something like:

```
const sectionIsOpenByName = { ... loop through all names and open them }
const openSections = [ ... for each label currently rendered by index, check the state in sectionIsOpenByName ]
```

And then add a watcher that keeps it in sync. You'd probably have to override the `@click` event on the expansion panel to update the `sectionIsOpenByName` ref.

michael-genson (changes requested):
Great, thank you! I ran into one small-ish issue, otherwise this looks good.
Say you have this list:
- [Produce]
- [Meat]

If you collapse meat, then re-order labels and move meat above produce, then produce will be collapsed, not meat.

Here's a demo:
2025-10-14_11h08_18

Should hopefully be as simple as re-calculating open/close state after a label re-order (since you're tracking open state by label name). If it's more complicated than that honestly just clearing open/close state seems like it would be fine to me since this would be a rare order of operations anyway.

miah120:
State tracking was getting to be a bit tedious so I factored it out into it's own Disclosure component so we can leave out any instance specific tracking. It takes a _tiny_ UI hit as the sections no longer collapse into a nice single block when multiple are closed, but I don't think anyone would notice/care.

michael-genson (approved):
Thank you looks great! I'm a little sad about removing the grouping but I agree it was getting really messy. I pushed a few small changes and will get this merged.

PR #6518, opened 2025-11-06
fix: Improve recipe ingredient selection

p0lycarpio:

_(REQUIRED)_

This PR improves the selection of recipes as ingredients by displaying a default list of recent recipes and adding an internationalized label and placeholder. The style of the added button has also been revised.

_(REQUIRED)_

N/A


frontend/components/Domain/Recipe/RecipeIngredientEditor.vue
```diff
@@ -173,9 +173,11 @@
           :items="search.data.value || []"
           item-title="name"
           class="mx-1 py-0"
-          placeholder="Choose Recipe"
+          :placeholder="$t('search.type-search')"
```
michael-genson: Oh wow I missed this in the original PR, good catch

michael-genson (approved):
LGTM thanks! I tweaked the translation string internal name

PR #6523, opened 2025-11-08
feat: Added Option to Import Recipe Category During Recipe Import

Gtt1229:

Adds the ability to import category during recipe import. Enables initial categorization where otherwise none is specified.

##### Before:
image

##### After:
image

##### Before:
image

##### After:
image

No direct issue, but quality of life and this request in discussion: https://github.com/mealie-recipes/mealie/discussions/4779

## Testing

Imported multiple recipes via URL and JSON where category is present, and verified categorization occurred, and did not occur when option is disabled.


frontend/lang/messages/en-US.json
```diff
@@ -634,6 +634,7 @@
     "scrape-recipe-have-raw-html-or-json-data": "Have raw HTML or JSON data?",
     "scrape-recipe-you-can-import-from-raw-data-directly": "You can import from raw data directly",
     "import-original-keywords-as-tags": "Import original keywords as tags",
+    "import-categories": "Import original category as tag",
```
p0lycarpio:
```suggestion
    "import-categories": "Import original categories",
```

p0lycarpio: PR tested and working for me ! Here is a clarification on the string

Gtt1229: Thank you for the suggestion!

I actually had it labeled as such originally.

Looking back, I should correct the variable itself. The schema standard has category as a single string, and by the standard, should only be a single string.

p0lycarpio: Oh yes! That explains why only one category was created during my tests.
Can we adapt the logic as well? `recipe_scrapers` always seems to return a string.

Gtt1229: So we're in a bit of a pickle. Upon further testing, there are some sites that improperly use the recipeCategory object as a list/tags.

For example: https://www.thepioneerwoman.com/food-cooking/recipes/a46587435/marry-me-chicken-recipe/ uses: `"recipeCategory":["30-minute meals","date night","dinner party","Midweek meals","Valentine's Day","weeknight meals","comfort food","dinner","main dish"]`

I have made a slight change to separate each of the categories into their own category to accommodate, but a part of me is wondering if only the first category should be used to prevent excessive category creation. Another part of me feels it may be infrequent enough that separating them isn't too big of a deal.

What are your thoughts?

michael-genson: We could instead just merge categories and tags (i.e. if importing tags, combine both categories and tags and add them all as tags).

But I'd also be fine with just leaving it; adding data automatically is always going to have edgecases like this, there's no way around it.

p0lycarpio: I like the idea of merging categories with keywords to form a tag. In my current usage, I systematically import keywords as tags, but I prefer to manage categories manually. I don't know about you?

michael-genson: Yeah that's what I do as well

Gtt1229: Welp, I messed up my branch....

mealie/services/scraper/cleaner.py
```diff
@@ -521,8 +521,8 @@ def clean_categories(category: str | list) -> list[str]:
         case str(category):
             if not category.strip():
                 return []
-
-            return [category]
+            # Split comma-separated categories
+            return [cat.strip().title() for cat in category.split(",") if cat.strip()]
```
Gtt1229, line 525: This fixes sites that do not adhere to the Schema format for categories, where they may use a list rather than a singular category.

frontend/lang/messages/en-US.json
Gtt1229: Fixed it.

I have added some logic for sites that improperly use a list for the category, so the categories are separated before import. https://github.com/mealie-recipes/mealie/pull/6523/files/5ffbaf00bf0ff8bcd520cd55c682e1343c7bd502#diff-40abdd43e31c58619af7498d55c4603d3a81eab5fa7d5f5430b2d6a4356d544e

Looking at most of my imports, it appears that the tags/keywords generally contain the category as well, so I think combining them isn't very useful due to the duplicity of it.

michael-genson:
LGTM, thanks! I went and updated the variable names and text names since we changed it to import multiple categories (vs one category).

PR #6569, opened 2025-11-19
feat: Animate shopping list and increase touch target

miah120:
Improves the UX of the shopping list by adding an animation that runs when an item is checked off. 
This provides an additional piece of feedback to confirm to the user which item was just taken off. 

On top of that, it also increases the touch target of the shopping list items from just the checkbox to the area shaded below: 
image

NA

NA

## Testing

Tested from docker container in firefox


p0lycarpio:
LGTM

michael-genson (approved):
LGTM, thanks!

PR #6574, opened 2025-11-20
feat: enable recipe deletion from recipes listing via context menu

tidzzz:
What this PR does / why we need it:

This PR allows users (recipe owner or admin) to delete a recipe directly from the recipes listing using the existing three‑dots  context menu on each recipe card. The UI updates immediately by removing the deleted recipe without a full page reload, improving workflow efficiency and user experience.

Changes (file-by-file):

- RecipeCard.vue
  - Enabled delete action in `RecipeContextMenu` by setting `delete: true` in `:use-items`.
  - Propagates deletion upward via `@deleted="$emit('delete', slug)"`.

- RecipeCardMobile.vue
  - Same change as desktop card: activates delete action and emits delete event.

- RecipeCardSection.vue
  - Listens for `@delete` on child cards.
  - Adds `onRecipeDeleted(slug)` which filters out the removed recipe and emits `replaceRecipes` to refresh visible list.
  - Hooks into existing replace/append pagination pattern (no extra API fetch needed after deletion).

- RecipeContextMenuContent.vue
  - Refactored menu item construction into `buildMenuItems()`.
  - Ensures the recipe is loaded (fetch if missing) before deciding whether to show “Supprimer”.
  - Hides delete unless current user is admin or owner (`canDelete`).
  - Adds reactive rebuild (watch on `recipeRef.value?.userId` and `isOwnGroup`) to keep menu accurate.
  - Emits `deleted` after successful API call to `api.recipes.deleteOne(slug)`.

Key decisions:

- Local list mutation instead of global refresh for responsiveness.
- Centralized authorization logic using existing `canDelete` pattern.
- Non-invasive changes: reuse existing dialogs, API methods, and events.
- Deferred adding toast feedback (can be a follow-up; dialog already provides confirmation).
- Refactored menu build for future extensibility (other conditional actions can plug in easily).

Before vs After (UI summary):

- Before: The three-dots menu on recipe cards did not offer deletion in the listing views; deletion required entering the recipe page or using bulk operations.
- After: Hover → click three dots → “Delete” (if permitted) → confirm dialog → card disappears instantly.

Testing

Manual:
- Logged in as recipe owner: menu shows “Delete” → recipe removed from grid.
- Logged in as non-owner, non-admin: “Delete” not shown.
- Logged in as admin (not owner): “Delete” shown with admin confirmation text.
- Mobile card behaves identically.


michael-genson:
I'm not sure we want to add this, I believe in the past we had issues with users accidentally deleting recipes by misclicking them. Is there a compelling reason to be able to delete recipes from the context menu?

We already have two ways of deleting recipes:
1. During recipe edit
2. From the data management page (which supports both single and bulk deletes)

PrettyPlot:
This is a feature I’ve really been hoping for, so I wanted to share my use case. I migrated my entire Paprika database into Mealie — thousands of recipes collected over about 10 years, plus recipes from way before then. A lot of them were added quickly without any review, so I’m now doing a big clean-up to make my recipe collection useable and curated.

When I search for something like “brownies” and see 20 results, I often can see straight away that several can be deleted without needing to open each recipe individually. Being able to delete recipes directly from the search results screen would make this tidy-up process so much faster.

michael-genson:
This sounds more like a usecase to add a search to the data management page to me. I actually didn't realize there wasn't one until just now.

PrettyPlot:
> This sounds more like a usecase to add a search to the data management page to me. I actually didn't realize there wasn't one until just now.

Yes - that would work, kind of, except that the data management page doesn't let you see the description text or photo, just the name.  So that doesn't necessarily help if there are, say, six different recipes with very similar names.  

(Speaking of data management, while not seeking to derail this PR - Another issue I'm facing is that I often want to bulk tag items - let's say add a "Lemon" tag to all of the recipes that have a lemon ingredient.  Currently, I can't easily do that.  I can search for ingredient, but I can't search for "is not tagged" to see which already have the tag.  But even then, it has to be one at a time.

So I thought the data management section would be the place to do it, but because you can't search in there either, when you have a list of thousands of recipes having to scroll to select what you want to bulk edit or bulk delete or anything like that is super time-consuming.)

michael-genson:
Going to go ahead and close this one, I definitely welcome a PR to improve the data management page to address the above concerns, but since we already had similar user discussions around this I don't want to re-introduce those problems.

More info on this old PR: https://github.com/mealie-recipes/mealie/pull/4842

PR #6579, opened 2025-11-21
feat: add mealplan nutrition filters and daily totals

parumpum:

Added the ability to create mealplan rules based on nutrition values in recipes.

https://github.com/mealie-recipes/mealie/discussions/3743

image


parumpum:
Looking at something more comprehensive. Will reopen later.

PR #6581, opened 2025-11-22
dev: Allow dev server to be accessed on local network

miah120:

This PR allows the dev server to be accessed from the local network when running from the dev container. 
This can be helpful for testing mobile workflows where previously to test those devs would either need to build a docker image or use browser emulation. 

NA

NA

## Testing
* Rebuild container
* Start up dev server with task py/task ui
* On a separate device go to `<host machine IP address>:3000` 
* Verify nothing hinky is going on


michael-genson (approved):
Thank you!

PR #6587, opened 2025-11-22
feat: add OpenAI fetch nutrition based on ingredients

parumpum:

Adds functionality to fetch nutrition information through OpenAI by processing the list of ingredients.

- Added button to retrieve the information on RecipeNutrition.vue if OpenAI is enabled
- RecipePageOrganizers.vue handles calling the fetch and display loading activity
- Added nutrition parser types and endpoints
- Added prompt for OpenAI to generate a proper JSON structure of nutrition data

image

This currently only sends the ingredients list in the request to OpenAI. Given the nature of LLMs, running the fetch multiple times for the same recipe may yield slightly different results.

The values returned will be for the total recipe yield, not per serving.

https://github.com/mealie-recipes/mealie/discussions/6051

## Testing

Tested retrieving nutritional data for recipes. Tested against recipes that already contained some nutritional data, and some that had no existing nutritional data.

Created a basic test case with mock response.

### DIFF
--- a/frontend/components/Domain/Recipe/RecipeNutrition.vue
+++ b/frontend/components/Domain/Recipe/RecipeNutrition.vue
@@ -5,42 +5,61 @@
         {{ $t("recipe.nutrition") }}
       </v-card-title>
       <v-divider class="mx-2 my-1" />
-      <v-card-text v-if="edit">
-        <div
-          v-for="(item, key, index) in modelValue"
-          :key="index"
+      <div v-if="loading">
+        <AppLoader
+          v-if="loading"
+          :loading="loading"
+          waiting-text=""
+        />
+      </div>
+      <div v-else>
+        <v-card-text v-if="edit">
+          <div
+            v-for="(item, key, index) in modelValue"
+            :key="index"
+          >
+            <v-text-field
+              density="compact"
+              :model-value="modelValue[key]"
+              :label="labels[key].label"
+              :suffix="labels[key].suffix"
+              type="number"
+              autocomplete="off"
+              variant="underlined"
+              @update:model-value="updateValue(key, $event)"
+            />
+          </div>
+        </v-card-text>
+        <v-list
+          v-if="showViewer"
+          density="compact"
+          class="mt-0 pt-0"
         >
-          <v-text-field
-            density="compact"
-            :model-value="modelValue[key]"
-            :label="labels[key].label"
-            :suffix="labels[key].suffix"
-            type="number"
-            autocomplete="off"
-            variant="underlined"
-            @update:model-value="updateValue(key, $event)"
-          />
-        </div>
-      </v-card-text>
-      <v-list
-        v-if="showViewer"
-        density="compact"
-        class="mt-0 pt-0"
-      >
-        <v-list-item
-          v-for="(item, key, index) in renderedList"
-          :key="index"
-          style="min-height: 25px"
+          <v-list-item
+            v-for="(item, key, index) in renderedList"
+            :key="index"
+            style="min-height: 25px"
+          >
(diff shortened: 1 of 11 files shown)


frontend/components/Domain/Recipe/RecipeNutrition.vue
```diff
@@ -5,42 +5,61 @@
...
+            </v-list-item-title>
+          </v-list-item>
+        </v-list>
+      </div>
+      <div class="d-flex justify-end pa-2">
+        <BaseButton
+          v-if="appInfo && appInfo.enableOpenai && edit"
+          :icon="$globals.icons.refresh"
+          :disabled="loading"
+          @click="emit('fetch-nutrition-info')"
```
michael-genson, line 58: Would we be able to change the verbiage to something like "Generate nutrition info (AI)"? I worry that users won't understand where the nutrition info comes from (and will assume it's accurate).

Could we also change the icon to match? Something like `auto-fix` or `robot`.

mealie/schema/openai/recipe_nutrition.py
```diff
@@ -0,0 +1,20 @@
...
+    nutrition: Nutrition | None = Field(
+        None,
+        description=dedent(
+            """
+            A dictionary of the nutritional information for the recipe. The dictionary should contain the following:
+            Calories, Carbohydrates, Cholesterol, Fat, Fiber, Protein, Saturated fat, Sodium, Sugar, trans-fat,
+            and unsaturated fat.
+            """
+        ),
+    )
```
michael-genson, line 20: You'll want to make `OpenAINutrition` independent of Mealie's models. Something like:
```python
class OpenAINutrition(OpenAIBase):
    calories: str | None = None
    carbohydrate_content: str | None = None
    cholesterol_content: str | None = None
    fat_content: str | None = None
    fiber_content: str | None = None
    protein_content: str | None = None
    saturated_fat_content: str | None = None
    sodium_content: str | None = None
    sugar_content: str | None = None
    trans_fat_content: str | None = None
    unsaturated_fat_content: str | None = None
```

mealie/services/openai/prompts/recipes/fetch-recipe-nutrition.txt
```diff
@@ -0,0 +1,43 @@
...
+
+Saturated fat
+
+Sodium
+
+Sugar
+
+Trans fat
+
+Unsaturated fat
```
michael-genson, line 25: You can get rid of the explicit field listing, this is handled by injecting the OpenAI model into the prompt

mealie/services/openai/prompts/recipes/fetch-recipe-nutrition.txt
```diff
@@ -0,0 +1,43 @@
...
+
+Do not include units in the nutrient values.
+
+Your response must be valid JSON and must follow the exact JSON schema provided below.
+
+Do not include any text outside of the JSON object.
+
+Ensure all required fields are present and formatted correctly.
+
+If you cannot determine a nutrient value, set that field to null. Do not omit any fields from the JSON object.
```
michael-genson, line 39: I'm fine with keeping this if you think it's helpful, but we will default to `None` if it's not included

michael-genson (changes requested):
Thanks for this, looks good so far! A few comments

michael-genson:
Upon more discussion internally we decided we're going to close this one. LLMs aren't really all that accurate at generating nutritional info so we're concerned the lack of accuracy makes having this moot.

Mealie fully support integrations via the API so users are more than welcome to spin up a better nutrition integration (potentially using AI) to populate nutrition data, if desired.

PR #6588, opened 2025-11-22
feat: Customizable OpenAI prompts (#5146)

ImanuelBertrand:

This PR adds a new backend configuration key for OpenAi: `OPENAI_CUSTOM_PROMPT_DIR`
Through usage of this setting, users (technically admins, not frontend users) can customize the prompts Mealie uses for OpenAI.
This has been requested in [this discussion](https://github.com/mealie-recipes/mealie/discussions/5146).
Using this feature will cut the user off from updates to the core prompts, so it is not intended for broad user adoption.
However, it does make it possible to work around edge cases and will make testing and improving the prompts easier in the future.

**Changed files:**

- **mealie/core/settings/settings.py**
  Added the new key `OPENAI_CUSTOM_PROMPT_DIR` to the settings class
- **mealie/services/openai/openai.py**
  Extracted the part that was loading the core prompt file into a separate method.
  It which first checks if the user has a custom non-empty prompt file. 
  If yes, the content of that file is returned, otherwise the core prompt is returned.
- **docs/docs/documentation/getting-started/installation/backend-config.md**
  Updated the documentation for the backend configuration key
- **tests/unit_tests/services_tests/test_openai_service.py**
  Added unit tests for the prompt loading with different scenarios

## Testing

The new unit tests pass, and I tested the functionality in a working instance by adding "Translate everything into french, no matter what" in my custom prompt.


KathiD303:
👍

pitonik:
I think this is desperately needed, please merge it and release it! :)

ImanuelBertrand:
@pitonik thanks! :) I'm also looking forward to have this merged.
The devs have stated in the discord that they are aware of the PR backlog, but they've been crunched at work for several weeks.
This is small and easy to review, so I'm hoping it will be a quick merge once they get to it.

michael-genson (approved):
LGTM, pushed a couple minor tweaks but no functional changes

pitonik:
glad to see this released! I was testing it today but unfortunately did not work for me... Of course, my main use case is automatically converting from imperial to metric during scrape/parse. I added the instruction to custom prompt, set everything in docker so it should be accessible by container but conversion did not happen.

I looked at docs and found this sentence: "When importing a recipe via URL, if the default recipe scraper is unable to read the recipe data from a webpage, the webpage contents will be parsed by OpenAI ( v1.9.0)".
-> Is this true? OpenAI parser is not used all the time to parse a webpage? That limits usefulness of this PR, I think. Is it possible to force it to use OpenAI? Or did I make a mistake in the setup?

michael-genson:
Currently OpenAI web scraping is only a fallback. This is because, as implemented in Mealie, the OpenAI scraper is just designed to work around the site not supporting the standard recipe schema.

It sounds like you're looking to customize the parsing, not the scraping. You can modify the parsing prompt and use the OpenAI parser.

To clarify, this is the workflow for a new recipe:
1. Scrape the website and convert it into a recipe
2. [Optional] parse the recipe ingredients, which are strings, into structured data

In step 1 we only use OpenAI as a fallback. In step 2 it's up to you to choose which parser you want (and it will always use your selected parser).

PR #6592, opened 2025-11-23
feat: adjust linked recipe qty display and load recipe with scale set

parumpum:

Makes the quantity of a linked recipe more clear by adding "batch"/"batches" after the number.

Added scale to the recipe link so that it will load with the scale set appropriately.

image

image

Helps resolve confusion noted on #4800 

## Testing

Tested by setting linked recipe quantities to different amounts to verify display of "batch"/"batches"


michael-genson (changes requested):
Vue I18n has a standard way of handling plurals with a single string, something like:

`|batch|batches`

And called:
`i18n.t("the-string-path", count)`

(note the leading pipe which displaying nothing for zero)

However, would it be better to just allow users to include a unit? They could create a batch/batches unit manually and also allow them to use any arbitrary unit. I know it won't post to the shopping list according to the unit but I think giving the option is still better than not.

michael-genson:
Closing in favor of: https://github.com/mealie-recipes/mealie/pull/7260, per above feedback

PR #6593, opened 2025-11-23
feat: Implement diacritic-insensitive search

niklas152:

Implements diacritic-insensitive search across the application. Searches now match text regardless of diacritics (e.g., "Käse" matches "kase" and vice versa).

**Backend changes:**

- Added `name_normalized` columns to `categories`, `tags`, and `tools` tables via migration `95f60087812f`

- Migration populates normalized fields for existing data using `unidecode`

- Added SQLAlchemy event listeners to auto-populate normalized fields when names change

- Updated `MultiPurposeLabelSummary` schema to enable search normalization (`_normalize_search = True`) and use `name_normalized` for searching

- Updated `RecipeTag` schema (affecting `RecipeCategory` and `RecipeTool` via inheritance) to use `name_normalized` for searching instead of `name`

- Search queries are normalized through the existing `SearchFilter` mechanism

**Frontend changes:**

- Added `use-text-normalize.ts` utility using Unicode NFD normalization to match backend behavior

- Updated `SearchFilter.vue` to normalize both search terms and item names before comparison

- Updated `RecipeOrganizerPage.vue` to use normalized text in Fuse.js searches via custom getter functions

This ensures consistent diacritic-insensitive search across recipes, tags, categories, tools, and other searchable entities.

The migration uses a local `normalize()` function that matches `SqlAlchemyBase.normalize()` behavior. The migration attempts to drop old indexes on `name` fields but handles cases where they may not exist. All changes are backward compatible and the migration runs automatically on application startup.

### Backend API Search

- Recipe search: works with and without diacritics

- Tag search: works with and without diacritics

- Category search: works with and without diacritics

- Tool search: works with and without diacritics

### Auto-population of Normalized Fields

- Created a new tag with diacritics via API

- Verified `name_normalized` is populated on creation

### Database Migration

- Migration `95f60087812f` applied successfully

- Verified `name_normalized` columns exist in `categories`, `tags`, and `tools` tables

- Verified existing data was normalized

- Verified indexes created for normalized fields

### Frontend UI

- Tested recipe search in UI (searched recipes, ingredients, tools, tags, and categories)

- All searches work with and without diacritics


frontend/composables/use-text-normalize.ts
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * Checks if a search term matches text, ignoring diacritics.
+ */
+export function matchesNormalized(searchTerm: string, text: string): boolean {
+  const normalizedSearch = normalizeText(searchTerm);
+  const normalizedText = normalizeText(text);
+  return normalizedText.includes(normalizedSearch);
+}
+
```
michael-genson: While looking at this PR I noticed this file didn't get linted for some reason, so the frontend check will fail. Can you run `task ui:lint`?

mealie/alembic/versions/2025-11-23-15.24.17_95f60087812f_add_normalized_fields_to_categories_tags_tools.py
```diff
@@ -0,0 +1,148 @@
...
+    tools = session.execute(select(Tool)).scalars().all()
+    for tool in tools:
+        if tool.name is not None:
+            session.execute(
+                sa.text(
+                    f"UPDATE {Tool.__tablename__} SET name_normalized=:name_normalized WHERE id=:id"
+                ).bindparams(name_normalized=normalize(tool.name), id=tool.id)
+            )
+
+    session.commit()
```
michael-genson: Can we use bulk operations here? Some instances may have a lot of tags. Something like:
```
session.execute(
    update(Category).values(name_normalized=normalize(Category.name))
)
```

michael-genson (changes requested):
Generally looks good, just a couple changes. Thanks for your work on this!

frontend/composables/use-text-normalize.ts
niklas152: Oh right, I think that should be fixed (for all files) now!

mealie/alembic/versions/2025-11-23-15.24.17_95f60087812f_add_normalized_fields_to_categories_tags_tools.py
niklas152: I am not super familiar with alembic and bulk operations yet, but I think I found a way (that is also batched to avoid to large operations at once, hopefully)

niklas152:
Also @michael-genson, I am not entirely sure whether this interferes with the earlier merged diacritics PR (that I missed when checking if there was already someone working on this before I started :upside_down_face:), IIRC that was purely in the frontend?

michael-genson:
Ah yes I think we did have a PR which did some of this on the frontend, but only for some elements not all: https://github.com/mealie-recipes/mealie/pull/6169

Would you be able to merge your changes in with these?

michael-genson:
Hey there looks like this is mostly covered by the existing implementation. We're also in the process of migrating over to `fuse.js` which includes normalization, so I'm going to go ahead and close this one. Appreciate your work on this one, even though we didn't get it in!

FYI: https://github.com/mealie-recipes/mealie/pull/7063

PR #6646, opened 2025-12-02
fix: Remove Auth Refresh

michael-genson:

_(REQUIRED)_

We check every five minutes to see if our auth state changed, except this five minute interval gets created a ton of times so it actually refreshes way too often. We don't really need this since we already have 401 handling, so I removed it.

_(REQUIRED)_

Reported on Discord


p0lycarpio:
Thanks!

PR #6662, opened 2025-12-06
fix: improve password manager autofill compatibility on login page

### LINKED ISSUE #5735: [BUG] - Login fields don't get picked up by password auto-fill
### First Check

With Mealie v2 I was able to use auto-fill for the credentials on the login page (using the Chrome KeePass Extension).
This is not working any more: The extension does not recognize the username and password field any more. I assume that in the previous version these fields were marked as such in the HTML and are not any more.

### Steps to Reproduce

1. Logout
2. Try to use auto-fill on the login page

### Mealie Version

Version
v3.0.1

Build
9ec15994276fb6784c59e83b92ab9a12b8c7ce8d

### Deployment

Docker (Linux)

_No response_

henricook:
## What type of PR is this?
- bugfix

Fixes password manager autofill not working correctly on the login page. Users reported that password managers like Bitwarden, KeePassXC, and Chrome's built-in autofill only fill one field at a time, requiring manual interaction with the second field.

### Root Cause

The issue has two parts:

1. **Vuetify's opacity animation**: Vuetify's `v-text-field` applies `opacity: 0` to inputs during initial render/animation. Password managers check for visible inputs and skip those with opacity=0, causing them to miss the fields entirely or behave inconsistently.

2. **Non-standard form attributes**: The username field had `name="login"` instead of the standard `name="username"`, and was missing an `id` attribute (while the password field had `id="password"`). This asymmetry confused some password managers' field detection.

### Changes

- Add `id="username"` to the email/username field for proper field identification
- Change `name="login"` to `name="username"` for standard form semantics
- Add scoped CSS to force `opacity: 1` on Vuetify's `.v-field__input`, preventing the animation from hiding inputs from password managers

Fixes https://github.com/mealie-recipes/mealie/issues/5735

## Before

https://github.com/user-attachments/assets/214b1334-24e2-4b64-81d0-c501d53694f4

## After

https://github.com/user-attachments/assets/24cecc17-a29e-4cb8-9308-60275ebb1aa2

## Testing

1. Navigate to the login page
2. Use a password manager (Bitwarden, KeePassXC, Chrome autofill, etc.)
3. Both username and password fields should now autofill simultaneously


henricook:
Sorry I've fixed the PR title but don't think I can rerun actions

michael-genson (approved):
Thanks!

PR #6665, opened 2025-12-06
fix: clear cached store data on logout to prevent user data leakage

### LINKED ISSUE #5513: [BUG] - Categories from different account/user showing after login
### First Check

When logging into different user accounts, the categories from the previously viewed account persist and display instead of the correct user's categories. A hard refresh is required to display the correct categories for the current user. This even happens on devices that have never signed in the to same device, which makes me think it isn't a browser/cache issue. 

- Fresh Mealie installation restored from backup
- Users on different networks/devices
- User B has never logged into User A's account on their device
- Issue persists across different browsers/devices

### Steps to Reproduce

1. User A logs into their account → sees their categories correctly
2. User B logs into their account on a different device/network → **always sees User A's categories at first click** instead of their own
3. Hard refresh required to display User B's correct categories
4. When User A logs back in → sees User B's categories initially
5. Hard refresh required again to display correct categories
6. This pattern repeats consistently - User B **always** sees User A's categories on initial login

_User A is the admin, if relevant_

No relevant logs. I can provide the backup if need be

### Mealie Version

v2.8.0

### Deployment

Unraid

Using docker compose on unraid

hay-kot:

When users log out and a different user logs in, category and other user-specific data from the previous user was persisting and displaying incorrectly. This was caused by:

1. Module-level refs in store composables that persist across the entire application lifecycle
2. Nuxt's `useAsyncData` cache using static keys without user context
3. The existing `flushStore()` function had a bug - it created a new ref instead of clearing the existing one

**Changes:**
- Fixed `flushStore()` in `use-store-factory.ts` to properly clear the ref value (`store.value = []` instead of `store = ref([])`)
- Added `resetXxxStore()` functions to all 9 store modules (category, tag, food, label, tool, cookbook, household, unit, user)
- Created `clearAllStores()` function in `store/index.ts` that calls all reset functions
- Updated `signOut()` in `use-auth-backend.ts` to call `clearAllStores()` and `clearNuxtData()` on logout

Fixes #5513

## Testing

To test this fix:
1. Log in as User A and navigate to view categories
2. Log out
3. Log in as User B (different group/household)
4. Verify User B sees their own categories immediately without requiring a hard refresh
5. Repeat switching between users to confirm data is properly cleared each time

### DIFF
--- a/frontend/composables/partials/use-store-factory.ts
+++ b/frontend/composables/partials/use-store-factory.ts
@@ -52,7 +52,7 @@ export const useStore = function <T extends BoundT>(
       return await storeActions.refresh(1, -1, params);
     },
     flushStore() {
-      store = ref([]);
+      store.value = [];
     },
   };
 
--- a/frontend/composables/store/index.ts
+++ b/frontend/composables/store/index.ts
@@ -1,7 +1,31 @@
-export { useCategoryStore, usePublicCategoryStore, useCategoryData } from "./use-category-store";
-export { useFoodStore, usePublicFoodStore, useFoodData } from "./use-food-store";
-export { useHouseholdStore, usePublicHouseholdStore } from "./use-household-store";
-export { useLabelStore, useLabelData } from "./use-label-store";
-export { useTagStore, usePublicTagStore, useTagData } from "./use-tag-store";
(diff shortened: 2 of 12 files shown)


michael-genson (approved):
LGTM

PR #6704, opened 2025-12-12
feat: Reprocess image user script

michael-genson:

_(REQUIRED)_

As a followup to https://github.com/mealie-recipes/mealie/pull/5883 this PR adds a user script to reprocess images. https://github.com/mealie-recipes/mealie/pull/5883 improved image processing a lot, and surely many users will want to take advantage of this on old recipes.

We detect whether or not the recipe images are "old" by checking the tiny image dimensions; if it's 300x300 it's an older recipe image, otherwise it's not (this will miss smaller images, i.e. the original image is <300, but this is good enough). If it's an older image we:
- Delete everything but the original image
- Re-process the original image

We also check all timeline images for these recipes with the same logic. We don't check timeline images for _all_ recipes, because if a newer recipe does not need to be processed, then we can assume the timeline images don't need to be either. If a recipe _does_ need to be reprocessed, then timeline images _may_ need to be reprocessed (it's possible users made an old recipe after this change, in which case the recipe must be reprocessed, but the timeline image does not).

_(REQUIRED)_

N/A

I opted to make this a dev script, not something that runs automatically, because image processing is quite slow for a lot of recipes. Even my medium-sized instance with about 350 recipes takes about a minute.

## Testing

Verified with the prod image.

### DIFF
--- a/docs/docs/documentation/getting-started/updating.md
+++ b/docs/docs/documentation/getting-started/updating.md
@@ -9,6 +9,23 @@
      - Create a Backup and Download from the UI
      - Upgrade
 
+!!! info "Improved Image Processing"
+    Starting with :octicons-tag-24: v3.7.0, we updated our image processing algorithm to improve image quality and compression. New image processing can be up to 40%-50% smaller on disk while providing higher resolution thumbnails. To take advantage of these improvements on older recipes, you can run our image-processing script:
+
+    ```shell
+    docker exec -it mealie bash
+    python /opt/mealie/lib64/python3.12/site-packages/mealie/scripts/reprocess_images.py
+    ```
+
+    ### Options
+     - `--workers N`: Number of worker threads (default: 2, safe for low-powered devices)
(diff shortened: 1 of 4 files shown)


michael-genson:
@1-tempest FYI

let92:
Hi, if I launch the script on a Synology NAS, this error occurs:
```
the input device is not a TTY
python: can't open file '/opt/mealie/lib64/python3.12/site-packages/mealie/scripts/reprocess_images.py': [Errno 2] No such file or directory
```

How can I use this script on a Synology NAS? Thanks in advance.

michael-genson:
Did you enter the docker container first? e.g. `docker exec -it mealie bash`
https://docs.mealie.io/documentation/getting-started/updating/

let92:
> Did you enter the docker container first? e.g. `docker exec -it mealie bash` https://docs.mealie.io/documentation/getting-started/updating/

Sorry, I was using Portainer, launching the commands on NAS via SSH worked, thank you very much!

PR #6719, opened 2025-12-14
fix: localize text validators message

### LINKED ISSUE #6715: Some explainations text does not use translation
A few different fields explaination text do not use the translation but show as english. To name some of them. In "recipe action" in Data Management when making new action. In Title and in URL the text text field "validation.this-field-is-required"="This Field is Required" never shows translation. 
Same with Invitation link, the same field "validation.this-field-is-required"="This Field is Required" show in english and not in translated language and also the email field: "user.e-mail-must-be-valid"="E-mail must be valid" (I actually think that this field is using plain text not calling from language file.
Hope this is enough info to find it.

p0lycarpio:

_(REQUIRED)_

Text field validators are not localized in the user's language. Use the global instance and the `useGlobalI18n()` composable to use the corresponding strings.

_(REQUIRED)_

Fixes #6715

### DIFF
--- a/frontend/lang/messages/en-US.json
+++ b/frontend/lang/messages/en-US.json
@@ -1428,5 +1428,13 @@
       "is-like": "is like",
       "is-not-like": "is not like"
     }
+  },
+  "validators": {
+    "required": "This Field is Required",
+    "invalid-email": "Email Must Be Valid",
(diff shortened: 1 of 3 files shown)


michael-genson (approved):
Thank you!

PR #6731, opened 2025-12-16
fix: PWA maskable android icons & enctype shared_target

p0lycarpio:

_(REQUIRED)_

The PWA icon on Android may appear on a black or other colored background depending on the browser when the application is installed. Maskable images must have additional margins and a background, unlike *any* icons. A white background has therefore been added to these images only.

The `enctype` property has been set to `share_target` to resolve the warning on Chrome and allow installation on mobile devices.

_(REQUIRED)_

N/A


michael-genson (approved):
Thank you!

PR #6747, opened 2025-12-19
feat: Add FDC nutrition information per ingredient

EinToni:
This PR comes from the discussion [2748](https://github.com/mealie-recipes/mealie/discussions/2748) and tries to implement a first step into the direction of adding nutritional data per ingredient. This would allow users to auto calculate nutritional data of recipes accurately form the ingretients.

I tried to keep the changes very simple for now as to not overload this pull request. The nutritional data is saved but unused. To be usable it would need a connection between the currently existing units and some internal "base units" like "gram", "kilogram" etc. If going that way the user could then select a base unit for each of the individualy created units. All weight units are then convertable and usable. For volume units like "cup" conversion per ingredient would need to be added or could even be pulled from FDC. But this is probably all for a different issue (i would be happy to work on).

Changes in this PR:
- Added an FDC_API_KEY to the settings for access to the FDC API.
- Added a backend API `/recipe/nutrition/fetch_fda/{fda_id}` that uses the FDC_API_KEY to call the API and return the data as nutrition_id/value pairs.
- Changed the database model of the ingredients to include a list of nutrition data which consists of the FDC nutrition id and the respecting value.
- Adapted `foods.vue` to allow for input of an FDC ID when creating an ingredient and when editing an ingredient. When the ID got edited the nutrition data is fetched and stored from the FDC API upon confirm of the dialog.
- Added a button next to the FDC ID entry. If the ID field is empty, it opens the FDC search page. If the field contains an ID, it opens the details page of the FDC with that ID.

<details> 
  <summary>Overview table image</summary>
   image
</details>

<details> 
  <summary>Create/edit food widthout FDC ID image</summary>
   image
</details>

<details> 
  <summary>Create/edit food with existing FDC ID image</summary>
   image
</details>

No issue exists yet, the PR arrises from discussion [2748](https://github.com/mealie-recipes/mealie/discussions/2748).

- All texts are currently hard coded and should instead be done via i18n before merging.
- I think it would be better to completely handle the update of the nutrition data in the backend instead of passing the data to the frontend only to store it from there in the db (something like `/recipe/ingredient/update_nutrition/{ingredient_id}`). But I was not sure how to implement that, so I chose the current solution instead.
- I am storing all nutrition information per ingredient. This is probably a bit too much as most data will never be used (like for example "SFA 14:0" id 1264, I have no clue what that is), but as it is just an int for the id and a float for the value it shouldn't matter too much.

## Testing

- Created a free FDC API key and added in the .dev.env file.
- Added a valid FDC ID to an ingredient.
- Displayed the stored data in a log I removed again as this was just for testing.


michael-genson:
I think this is definitely a step in the right direction for better nutritional data support. I wonder if we should hold off on this until we have better unit handling with a base unit/conversions, etc. This is something in the backlog to support other areas of the app too.

Do you think it would make sense to wait for that first? I worry adding this before we have a better handle on units/conversions may cause a lot of headache. I haven't looked too closely at your implementation yet but I read your PR description mentioning this.

SvenErik1968:
In addition to what @michael-genson suggested, would it be possible to make this feature more food composition database agnostic? As a user in Norway, I would be more interested in getting these values from the database that the _Norwegian Food Safety Authority_ [publishes](https://www.matvaretabellen.no/en/api/). An alternative would also be the European Food Safety Authority's forthcoming open-access [European Food Composition Database](https://www.efsa.europa.eu/en/data-report/food-composition) that is expected to be published in mid-2026.

EinToni:
It's probably a wise idea to first do proper unit handling and then get back to this, I simply wanted to start somewhere. Do you know if there is currently and work done regarding this @michael-genson? I will put this on hold until there is some better unit handling.

Regarding your more general idea of implementing this with different food database backends, I guess it is definitely possible, but will take way more architectural work to pull of cleanly, because every database has different identifiers for the inredients.

michael-genson:
> Do you know if there is currently and work done regarding this

It's in my (soft) backlog but I'm not currently working on it.

Regarding different db backends I'll have to dig into this PR, but I'm fine with picking one standard for now (like you did here). As long as we make it reasonably extensible it should be pretty easy to add on support later.

michael-genson:
@EinToni following up on this, I implemented some unit standardization here: https://github.com/mealie-recipes/mealie/pull/7121

`quantity + standard unit + FDC food` should hopefully be enough to generate reasonably accurate nutrition info. Let me know if the unit standardization using pint is sufficient.

mealie/services/nutrition/nutrition_service.py
```diff
@@ -0,0 +1,38 @@
...
+    def fdc(self, fdc_id: str) -> list[FdcNutritionItem]:
+        nutrition_items: list[FdcNutritionItem] = []
+
+        if not self.settings.FDC_API_KEY or self.settings.FDC_API_KEY == "":
+            self.logger.error("No API Key set for FDC access.")
+            return nutrition_items
+
+        response = requests.get(
+            f"https://api.nal.usda.gov/fdc/v1/food/{fdc_id}?api_key={self.settings.FDC_API_KEY}", timeout=10
+        )
```
michael-genson, line 19: Two things here:
1. Does this API support passing the API key in the header instead of the URL? There are places this might leak.
2. Can we use httpx instead of requests? We have a safe HTTPX transport to prevent against injection, etc. And then we can switch to async.

mealie/routes/recipe/nutrition.py
```diff
@@ -0,0 +1,15 @@
...
+from mealie.routes._base import BaseUserController, controller
+from mealie.schema.recipe.recipe_nutrition import FdcNutritionItem
+from mealie.services.nutrition.nutrition_service import NutritionService
+
+router = APIRouter(prefix="/nutrition")
+
+
+@controller(router)
+class NutritionController(BaseUserController):
+    @router.get("/fetch_fda/{fdc_id}", response_model=list[FdcNutritionItem])
```
michael-genson, line 12: Should this be `fetch_fdc`?

frontend/pages/group/data/foods.vue
```diff
@@ -116,6 +116,35 @@
...
+                icon
+                target="_blank"
+                href="https://fdc.nal.usda.gov/"
+                title="Open search FDC page"
+              >
+                <v-icon>{{ $globals.icons.search }}</v-icon>
+              </v-btn>
+            </template>
+          </v-text-field>
           <v-checkbox
```
michael-genson, line 148: We'll need to translate these strings using `i18n`. You can check out how other components do this (just add a message to `en-US.json` and reference it, don't do any other languages), but if you run into trouble happy to help.

mealie/services/nutrition/nutrition_service.py
```diff
@@ -0,0 +1,38 @@
...
+
+from mealie.schema.recipe.recipe_nutrition import FdcNutritionItem
+from mealie.services._base_service import BaseService
+
+
+class NutritionService(BaseService):
+    def fdc(self, fdc_id: str) -> list[FdcNutritionItem]:
+        nutrition_items: list[FdcNutritionItem] = []
+
+        if not self.settings.FDC_API_KEY or self.settings.FDC_API_KEY == "":
```
michael-genson, line 13: nit: just `if not self.settings.FDC_API_KEY` is fine here

mealie/db/models/recipe/ingredient.py
```diff
@@ -173,6 +173,7 @@ class IngredientFoodModel(SqlAlchemyBase, BaseMixins):
 
     label_id: Mapped[GUID | None] = mapped_column(GUID, ForeignKey("multi_purpose_labels.id"), index=True)
     label: Mapped[MultiPurposeLabel | None] = orm.relationship(MultiPurposeLabel, uselist=False, back_populates="foods")
+    fdc_id: Mapped[int | None] = mapped_column(sa.Integer, nullable=True)
```
michael-genson, line 176: Can we make this a string instead? A lot of other databases use strings rather than ints, and if we want to extend this in the future it could save us some trouble.

mealie/db/models/recipe/ingredient.py
```diff
@@ -337,6 +344,30 @@ def __init__(self, session: Session, name: str, **_) -> None:
...
 
 
+class IngredientFoodNutritionModel(SqlAlchemyBase, BaseMixins):
+    __tablename__ = "ingredient_foods_nutrition"
+    id: Mapped[GUID] = mapped_column(GUID, primary_key=True, default=GUID.generate)
+
+    food_id: Mapped[GUID] = mapped_column(GUID, ForeignKey("ingredient_foods.id"), index=True)
+    food: Mapped["IngredientFoodModel"] = orm.relationship("IngredientFoodModel", back_populates="nutrition")
+
+    fdc_nutrition_id: Mapped[int] = mapped_column(Integer, index=True)
```
michael-genson, line 354: Same here, can this be a string?

michael-genson (changes requested):
Sorry it took so long to get to this one, though I think it's much more valuable now that we have standardized units too.

Overall the approach looks good, I have some feedback and a couple more things:
- Can we abort saving the FDC ID if we can't successfully pull the data? i.e. if I type an invalid id we shouldn't save it.
- Can we hide the FDC fields if we don't have an FDC API key enabled? Look at how our other services (e.g. OpenAI) do this, it should be pretty straightforward.
- Can we add backend tests for the new service? You can mock the FDC response (check out how the OpenAI tests work).

Schr0ding3rscat:
This may be a bit late as feedback, but one possible alternative to API calls would be to ship Mealie with a built in db, such as  FooDB as part of the default ingredient data.

For example:
- Seed Mealie with a baseline ingredient-to-nutrient mapping for common ingredients.
- Link parsed recipe ingredients to that curated ingredient database using normalized names, aliases, and basic unit conversions.
- For ingredients that are already measurable by weight or volume, derive nutrition directly from the stored database values.
- For ingredients that are not easily represented by standard units, such as “1 onion,” “1 bunch cilantro,” or “1 clove garlic,” use the same OpenAI API workflow already used for recipe parsing to infer a normalized quantity or edible-weight estimate.
- Store that resolved mapping back into Mealie’s curated ingredient DB so the system improves over time and avoids repeating the same inference work.

That would reduce dependence on external nutrition APIs during normal use, while still allowing the parser to handle unstructured ingredient formats. 

Something to consider, either way looking forward to the release of this feature!

Pebkac03:
Wanted to join this since I've been thinking extensively about the external integration idea for the last while. Might be a bit late so do with it what you will.

I'd suggest not polluting any existing tables with columns for external ID:s, since this would require a migration anytime you'd want to add a new integration.

Instead I'd suggest creating a new table for external connections like so:
| name_of_connection |  linked_food_item (FK) | ext_id | url |
| ----------------------- | ------------------------ | ------ | ---- |
| 'FDC'                          | 'id_of_food'                  | 'ext_id' | localhost/api/fdc |

You could then (manually or cron) go through that table and fetch new data from the url and update the food item. You'd just have to define a schema for the response body and then you could include some integrations with mealie (an api gateway to map the response from the external service to the required format for mealie), or allow users to create their own integration gateways where instead of a url pointing to mealie it points to the other service.

I can easily see the number of integrations growing over time with different countries having nutritional databases and you might even need another another external database to get other information, like price for example.

MarcelBigger:
I've read up on your discussion. Allow me to insert myself. The associated PRs conflate different issues:

### 🧬 Nutritional Data:

**Macro- and micro-nutritional information is always expressed in relation to a reference. Nearly all national databases, including the USDA FDC in the United States, take the two International System of Units (SI) units, 100 grams and 100 milliliters, as the reference.** Example: Egg, white, dried, as published in the USDA FDC, has 79.9g of protein per 100g, 959 milligram of potassium per 100g and 34 microgram of iodine per 100g.

**... versus ...**

### 🍽️ Recipe information, servings and planned meals:

**Luckily most ingredients in recipes are also expressed in reference to 100 grams.** Example: 100g of egg, white, dried is used for 1 serving of 3 high-protein pancakes. The calculation of nutritional values per serving is then straight-forward.

Sometimes, individual ingredients are expressed as counts. Example: 3 Eggs for 3 high-protein pancakes. The calculation of nutritional values per serving must then rely on an average weight. Example: 33 grams of egg, white, dried in an average large egg.

If you live in a former British colony, e.g. the United States, you also, unfortunately, have to deal with volumetric measures, which are not International System of Units (SI) units of measurement. Example: 1 Cup of egg, white, dried. The calculation of nutritional values per serving will then require density conversion to get to grams.

The nutritional values for each planned meal is in all cases trivial by setting the number of servings.

### What does this mean for a proper implementation?

**As a first step**, you would expand the database schema to hold nutritional data. The nutrient child table design in the PR (one row per nutrient per ingredient is the correct normalized approach. The ingredients record should store nutrients against two possible reference bases, per 100g and per 100ml, or alternatively store the ingredient's density to allow conversion between them. The database addition would like something like:

**units ──< ingredients ──< ingredients-nutrients>── nutrients >── units**

Reading it left to right:
- units one-to-many nutrients (many ingredients share the same unit, e.g. either g as in 100g or ml as in 100ml)
- ingredients one-to-many ingredients-nutrients (one ingredient has many nutrients)
- ingredients-nutrients many-to-one nutrients (many ingredients-nutrients rows point to the same nutrient)
- nutrients many-to-one units (many nutrients share the same unit, e.g. both potassium and magnesium are expressed in milligram, both iodine and cholecalciferol are expressed in microgram)

Only if you have achieved that, you can work on the calculation of nutritional values for each recipe and for each planned meal **as a second step**.

To draw in nutritional data from external databases, e.g. the USDA FDC, the dozens of other national databases and/or the much larger and more accurate NCCDB (Nutrition Coordinating Center Food & Nutrient Database at the University of Minnesota), **is nice to have but really not critical**. The typical user will only use a few hundred ingredients for which nutritional information can be entered by hand in 1 to 2 hours. **Many open-source projects are over-engineered**, an avoidable affliction.

In any case, I would make the nutritional data schema **source-agnostic** as, for example, users in Germany, India and China will obviously not want to use USDA FDC data.

Pebkac03:
> I've read up on your discussion. Allow me to insert myself. The associated PRs conflate different issues:
> 
> ### 🧬 Nutritional Data:
> 
> **Macro- and micro-nutritional information is always expressed in relation to a reference. Nearly all national databases, including the USDA FDC in the United States, take the two International System of Units (SI) units, 100 grams and 100 milliliters, as the reference.** Example: Egg, white, dried, as published in the USDA FDC, has 79.9g of protein per 100g, 959 milligram of potassium per 100g and 34 microgram of iodine per 100g.
> 
> **... versus ...**
> 
> ### 🍽️ Recipe information, servings and planned meals:
> 
> **Luckily most ingredients in recipes are also expressed in reference to 100 grams.** Example: 100g of egg, white, dried is used for 1 serving of 3 high-protein pancakes. The calculation of nutritional values per serving is then straight-forward.
> 
> Sometimes, individual ingredients are expressed as counts. Example: 3 Eggs for 3 high-protein pancakes. The calculation of nutritional values per serving must then rely on an average weight. Example: 33 grams of egg, white, dried in an average large egg.
> 
> If you live in a former British colony, e.g. the United States, you also, unfortunately, have to deal with volumetric measures, which are not International System of Units (SI) units of measurement. Example: 1 Cup of egg, white, dried. The calculation of nutritional values per serving will then require density conversion to get to grams.
> 
> The nutritional values for each planned meal is in all cases trivial by setting the number of servings.
> 
> ### What does this mean for a proper implementation?
> 
> **As a first step**, you would expand the database schema to hold nutritional data. The nutrient child table design in the PR (one row per nutrient per ingredient is the correct normalized approach. The ingredients record should store nutrients against two possible reference bases, per 100g and per 100ml, or alternatively store the ingredient's density to allow conversion between them. The database addition would like something like:
> 
> **units ──< ingredients ──< ingredients-nutrients>── nutrients >── units**
> 
> Reading it left to right:
> - units one-to-many nutrients (many ingredients share the same unit, e.g. either g as in 100g or ml as in 100ml)
> - ingredients one-to-many ingredients-nutrients (one ingredient has many nutrients)
> - ingredients-nutrients many-to-one nutrients (many ingredients-nutrients rows point to the same nutrient)
> - nutrients many-to-one units (many nutrients share the same unit, e.g. both potassium and magnesium are expressed in milligram, both iodine and cholecalciferol are expressed in microgram)
> 
> Only if you have achieved that, you can work on the calculation of nutritional values for each recipe and for each planned meal **as a second step**.
> 
> To draw in nutritional data from external databases, e.g. the USDA FDC, the dozens of other national databases and/or the much larger and more accurate NCCDB (Nutrition Coordinating Center Food & Nutrient Database at the University of Minnesota), **is nice to have but really not critical**. The typical user will only use a few hundred ingredients for which nutritional information can be entered by hand in 1 to 2 hours. **Many open-source projects are over-engineered**, an avoidable affliction.
> 
> In any case, I would make the nutritional data schema **source-agnostic** as, for example, users in Germany, India and China will obviously not want to use USDA FDC data.
> 
> 

Hi, could you elaborate on why to use the schema of one row per nutrient per ingredient and not to have a table with each nutrient type having a column and the ingredient_table and ingredient_nutritional_information having a one-to-one relationship?

In my mind the types of nutrients to exist (carbohydrates, proteins, vitamins, etc.) belong as columns and not rows.

MarcelBigger:
Is the current schema already normalized?

Compared to a flat schema, which defines units by convention (for instance, iodine is in micrograms), a normalized schema allows units and other per-nutrient metadata to be first-class FK constraints, resistant to data entry errors.

Nutrient sets evolve. A flat schema would required table migrations whenever a new nutrient is added, including users defining other nutrients important to them.

The leading nutrient database store over 100 nutrients, however, most ingredients have far fewer nutrients recorded. With a flat schema, most of the cells would be empty.

michael-genson:
Going to close this since it's been a while without any activity. Feel free to re-open if you have interest in picking it back up!

1jamesthompson1:
Not sure if this is the best way to add nutritional information to Mealie. Yet for my usecase (personal) this is massive value add.
Any form of auto calculating nutritional intake based off menu where it is auto calculated from recipe ingredients would be great.

PR #6763, opened 2025-12-22
fix: add loader for create backup button

p0lycarpio:

_(REQUIRED)_

The “Create backup” button does not change state when pressed, which can be confusing for the user. Adding the “loading” state prevents the task from being submitted twice and provides visual feedback.

_(REQUIRED)_

N/A


michael-genson (approved):
Thanks!

PR #6780, opened 2025-12-27
feat: add proxy authentication provider

morgan-dgk:

This pull-request adds support for [proxy authorization](https://www.authelia.com/reference/guides/proxy-authorization/) / [ForwardAuth](https://docs.goauthentik.io/add-secure-apps/providers/proxy/forward_auth/). This authentication method relies on a trusted proxy appropriately setting the REMOTE-USER header in forwarded request.

This is a frequently requested feature and would provide a streamlined user experience for users running a centralised identity-provider with reverse proxy to manage access to private services. 

The approach taken here is based on #2040 but updates the code to use the newer AuthProvider interface. It also focuses on implementing **header-based** auth only. This simplifies the solution since Mealie need not be responsible for, e.g. redirecting users to the IdP if the session has expired. 

**N.B. because mealie [currently runs with --forwaded-allow-ips="*"](https://github.com/mealie-recipes/mealie/blob/7792f0504d02e0030eb5f3f6adab2126429551ea/mealie/app.py#L178), mealie allows any host to set the sensitive headers. We should consider also changing this to be configurable via app settings. Otherwise, users enabling forward auth will need to ensure that the unicorn server is only accessible via a trusted proxy.**

Related discussions [#801](https://github.com/mealie-recipes/mealie/discussions/801), [#3364](https://github.com/mealie-recipes/mealie/discussions/3664), [#284](https://github.com/mealie-recipes/mealie/discussions/284).

Forward Auth requires access to the headers sent with the request to the /auth/token/ endpoint. However, request headers are not part of the AuthProvider interface, nor are they provided to the get_auth_provider function.  For this reason, to avoid modifying the existing interface I have chosen to simply set these on the AuthProvider instance inside the /auth/token route handler iff the AuthProvider in use is the ProxyAuth provider.

## Testing

I have started adding a new e2e test for this feature. However, I am unsure on how to create an internal user
for use in testing during this e2e test.

As such, testing with this setup currently requires completing the registration process at least once, registering a user with username 'john'. This matches the static Remote-User header set in `traefik/dynamic_config.yml`

Once the stack defined in `docker-compose-forward-auth.yml` is up, mealie should be reachable at mealie.localhost:8080. Traefik should then add the static Remote-User header with value john to each request.

I have tested this manually (i.e. without using GitHub Actions) or another automated runner.

### DIFF
--- a/frontend/app/pages/login.vue
+++ b/frontend/app/pages/login.vue
@@ -276,6 +276,14 @@ whenever(
   { immediate: true },
 );
 
+whenever(
+  () => $appInfo.proxyAuthEnabled && !isDirectLogin(),
+  () => {
+      authenticate({ isproxyAuthRequest: true });
+    },
+  { immediate: true },
+);
+
 const loggingIn = ref(false);
 const oidcLoggingIn = ref(false);
 
@@ -320,16 +328,17 @@ async function oidcAuthenticate(callback = false) {
   }
 }
 
-async function authenticate() {
-  if (form.email.length === 0 || form.password.length === 0) {
-    alert.error(i18n.t("user.please-enter-your-email-and-password"));
-    return;
-  }
+async function authenticate({ isproxyAuthRequest } = { isproxyAuthRequest: false}) {
+  if (!isproxyAuthRequest)
+    if (form.email.length === 0 || form.password.length === 0) {
+      alert.error(i18n.t("user.please-enter-your-email-and-password"));
+      return;
+    }
 
   loggingIn.value = true;
   const formData = new FormData();
-  formData.append("username", form.email);
-  formData.append("password", form.password);
+  formData.append("username", isproxyAuthRequest ? "proxy" : form.email);
+  formData.append("password", isproxyAuthRequest ? "proxy" : form.password);
   formData.append("remember_me", String(form.remember));
 
   try {
--- a/mealie/core/security/providers/proxy_provider.py
+++ b/mealie/core/security/providers/proxy_provider.py
@@ -1,0 +1,51 @@
+from datetime import timedelta
+
+from sqlalchemy.orm.session import Session
+from starlette.datastructures import Headers
+
+from mealie.core import root_logger
+from mealie.core.config import get_app_settings
+from mealie.core.security.providers.credentials_provider import CredentialsProvider
+from mealie.schema.user.auth import CredentialsRequest
+
+
+class ProxyProvider(CredentialsProvider):
+    """Authentication provider that authenticates a user using the header(s) forwarded from a trusted proxy"""
+
+    _logger = root_logger.get_logger("proxy_provider")
+
+    def __init__(self, session: Session, data: CredentialsRequest) -> None:
+        # Extra attributes necessary for forward auth that are not included in AuthProvider protocol nor passed
+        # to get_auth_provider. Instead, we assign these later in the /auth/token route where we have access to
+        # the Request object.
+        self.headers: Headers
+        super().__init__(session, data)
+
+    def authenticate(self) -> tuple[str, timedelta] | None:
+        """Attempt to authenticate a user with the username supplied in the
+        `REMOTE-USER` header (or the header specified by settings.REMOTE_USER_HEADER)."""
+        # When proxy auth is enabled, we need to still also support authentication with the Mealie backend.
+        # First we look to see if the required user header is set. If the required header is not set,
+        # we fallback to authenticating using the credentials provided in the login form. If the required
+        # header IS set, we attempt to match the value of this header with an existing mealie user.
+        # If we fail to find an existing user, we fallback to authenticating with the Mealie backend.
+
+        settings = get_app_settings()
+        remote_user = self.headers.get(settings.REMOTE_USER_HEADER, None)
+
+        if not remote_user:
+            self._logger.debug(
+                f"Required header {settings.REMOTE_USER_HEADER} not found. Falling back to internal auth."
+            )
+            return super().authenticate()
+
+        user = self.try_get_user(remote_user)
+
+        if not user:
+            # For now, there is no option to automatically provision new users, i.e.
+            # users who have not been previously registered as internal users. Though
+            # this could be added fairly trivially.
+            self._logger.debug(f"Failed to find existing user for remote user: {remote_user}")
+            return super().authenticate()
+
+        return self.get_access_token(user, False)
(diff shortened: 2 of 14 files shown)


morgan-dgk:
Running this on my own deployment now to test - so far so good.

michael-genson:
Hey there, is this still a WIP, or is it ready for a review?

morgan-dgk:
> Hey there, is this still a WIP, or is it ready for a review?

Hey Michael, there's probably some outstanding work related to service worker pre-caching as described in this [comment](https://github.com/mealie-recipes/mealie/pull/2040#issuecomment-1447235457) on #2040.  That said, I think this should be fairly trivial once the service-worker is properly fixed (#6815 is my attempt at this).

Have been running mealie stably for ~2 weeks now with both of the above changes applied without issue.

morgan-dgk:
So, as far as I can tell there's no runtime caching happening currently with the service worker. So there shouldn't be any caching related issue with the `/app/info` endpoint that indicates whether proxy auth is enabled or not.

Only thing I'm still considering is whether it's a problem to show the generic Invalid Credentials error when landing on the login screen if proxy auth fails. Otherwise, this seems to works reasonably well with my deployment. 

**Auth failure**

https://github.com/user-attachments/assets/31210932-78d9-4a2b-82b7-debf9d9b9109

This only fires once if Proxy Auth fails when loading the page and the logger will log when the provider fails to find a user matching the Remote-User header if the DEBUG log level is set.

Is much harder to get a recording of a successful login since this is typically pretty fast. Occasionally there is a brief flash between when the login page loads and auth completes and the main app shell is loaded but I think the service worker pre-caching helps reduce this.

Do you think this once is ready for nightly @michael-genson ?

michael-genson:
I'm fine with it as-is, we can always improve it down the line if needed. I haven't fully reviewed since it's in draft, but go ahead and mark it as ready and I can take a look!

> whether it's a problem to show the generic Invalid Credentials error when landing on the login screen if proxy auth fails

I don't see a problem with this. To the user it's a credentials error, doesn't matter how it's implemented. Server owners can debug via the logs.

morgan-dgk:
Mucked up something rebasing, let me fix.

morgan-dgk:
Assume this is failing because of something to do with trusted hosts setting - I don't see this failure locally.

michael-genson:
Aside from the failing test is this ready? If so, can you mark it as ready for review?
I can take a look at the test if you're not sure why it's failing, but I don't want to if you're going to update it 😁

morgan-dgk:
> Aside from the failing test is this ready? If so, can you mark it as ready for review? I can take a look at the test if you're not sure why it's failing, but I don't want to if you're going to update it 😁

I think I've figured out the cause of failing tests in CI. The backend tests seem to mostly target the LDAP Auth provider, so I needed to ensure that LDAP was not enabled for tests targeting the proxy provider.

I'd say this is ready as an _experimental_ feature. Does mealie use this designation, a la something like actualbudget?

michael-genson:
No not really, typically we will merge into nightly, and then after some time it'll be released. We don't really have an experimental branch or anything like that.

While it's in nightly we can gather user feedback and fix/revert as needed.

morgan-dgk:
~~Gonna turn this one back into a draft, slightly more complicated than I anticipated to ensure that the header is securely set.~~

It turns out uvicorn will by default attempt to set the client address using the `X-Forwarded-For` header as long as the connecting ip is in the `forwarded-allow-ips` configuration. For this reason, there is no obvious way to actually get at connecting IP via the requests object.  

In consequence, I have removed the `TRUSTED_PROXIES` setting as well as the related check in the ProxyAuth provider. 

As noted above, mealie currently runs with `--forwaded-allow-ips="*"`.

For this reason, it is imperative that users who enable this feature ensure that they can trust all values within the forwarded headers (e.g. because their proxies remove any existing headers before setting their own.)

mealie/core/security/providers/proxy_provider.py
```diff
@@ -0,0 +1,51 @@
...
+            )
+            return super().authenticate()
+
+        user = self.try_get_user(remote_user)
+
+        if not user:
+            # For now, there is no option to automatically provision new users, i.e.
+            # users who have not been previously registered as internal users. Though
+            # this could be added fairly trivially.
+            self._logger.debug(f"Failed to find existing user for remote user: {remote_user}")
```
michael-genson, line 48: Other debug logs should show the request info already, and code scanning alerts is complaining about this.

```suggestion
            self._logger.debug("Failed to find existing user for remote user")
```

frontend/app/pages/login.vue
```diff
@@ -322,16 +330,17 @@ export default defineNuxtComponent({
...
+          alert.error(i18n.t("user.please-enter-your-email-and-password"));
+          return;
+        }
 
       loggingIn.value = true;
       const formData = new FormData();
-      formData.append("username", form.email);
-      formData.append("password", form.password);
+      formData.append("username", isproxyAuthRequest ? "proxy" : form.email);
+      formData.append("password", isproxyAuthRequest ? "proxy" : form.password);
```
michael-genson: Would it be better to just send the email/pw and figure out on the backend whether we should/shouldn't use proxy auth?

michael-genson (changes requested):
Overall looks good, a few notes:

> I have removed the `TRUSTED_PROXIES` setting

I'm nervous about this, since if a user misconfigures their proxy they're silently opened up to an attack. Could we instead have a `PROXY_TRUSTED_IPS` or similar which is required, and we can validate? That way users are forced to configure it on the Mealie side, preventing accidentally screwing that up.

Can we also add documentation for the new env vars, and maybe a page specifically for proxy auth?

morgan-dgk:
> Overall looks good, a few notes:
> 
> > I have removed the `TRUSTED_PROXIES` setting
> 
> I'm nervous about this, since if a user misconfigures their proxy they're silently opened up to an attack. Could we instead have a `PROXY_TRUSTED_IPS` or similar which is required, and we can validate? That way users are forced to configure it on the Mealie side, preventing accidentally screwing that up.
> 
> Can we also add documentation for the new env vars, and maybe a page specifically for proxy auth?

Thanks @michael-genson , I've done so more research on how this can be made more secure and will take another pass at this. I'll also include some documentation around the new env vars and Proxy Auth.

Some good info from [discussion on uvicorn](https://github.com/Kludex/uvicorn/pull/1611) and [oauth2 proxy project](https://github.com/oauth2-proxy/oauth2-proxy/issues/1095#issuecomment-803323889).

michael-genson:

ScottESanDiego:
Tagging @morgan-dgk in case he missed the earlier Github ping on the issue. :-P

morgan-dgk:
Sorry folks, I've been busy at work. I'll try and pick this one backup again ~~but from what I recall I ran into some kind of issue with session expiration that I was worried about~~

michael-genson:
Going to close this since it's been a while without any activity. Feel free to re-open if you have interest in picking it back up!

PR #6793, opened 2025-12-30
feat: User profile + settings menu in top right (AppHeader) instead of top/bottom left (AppSidebar)

lehnerpat:

_(REQUIRED)_

This PR consolidates the link to the current user's profile and the "settings" submenu into a single profile/account menu, represented by the user's avatar in the top right (i.e. at the end of the AppHeader).
This aligns Mealie's UI more with common conventions found in other apps that have user accounts (e.g. Github, Google, etc.).
This also prevents accidental logout, especially on mobile, where the logout button on the top right was easy to hit by mistake (and there was no confirmation dialog before logging out).

In more detail:
* before this change, there was:
  * the user avatar shown at the top of the left sidebar (link to user profile page and to user favorite recipe's)
  * the settings submenu shown at the bottom of the left sidebar (introduced in https://github.com/mealie-recipes/mealie/pull/6043)
  * a login/logout button shown at the right of the top bar (AppHeader)
* after this change, all three of these are consolidated to a "user menu" at the top right (except for the link to favorite recipes, which is moved to the nav links in the left sidebar)
  * the menu is represented by the user's avatar and user name
  * when clicked, a submenu opens, showing:
    * a link to the profile page (previous on the left sidebar on the top)
    * all items from the "settings" submenu (previously in the left sidebar on the bottom), such as theme and language switching, admin settings, etc.
      * except for the "User Settings" menu, because this linked to the "user profile" page which is now at the top of this menu
    * a logout button (previously directly in the app bar)

### Screenshots

| | Before | After<br>(menu closed) | After<br>(menu open) |
|---|---|---|---|
| Desktop | CleanShot 2026-01-05 at 13 56 11@2x | CleanShot 2026-01-05 at 13 56 44@2x | CleanShot 2026-01-05 at 13 56 52@2x |
| Mobile | CleanShot 2026-01-05 at 13 58 02@2x | CleanShot 2026-01-05 at 13 58 26@2x | CleanShot 2026-01-05 at 13 58 33@2x |

_(REQUIRED)_

Addresses discussion https://github.com/mealie-recipes/mealie/discussions/5528

There are three things in particular I'd like reviewers to take a look at and/or let me know:
* The sidebar had specific handling for when a user isn't logged in; for example, the language and theme switcher menu items were also shown to logged-out users. I'm not sure what the use cases for this are.
  * Right now, the user menu on the top right is _only_ shown to logged-in users. If logged-out usage is important, please let me know how we should address this here (e.g. use a placeholder icon instead of user avatar for the menu button)
* The AppHeader shows a "login" button when it's viewed in a logged-out state. I've kept this for now, but I'm not sure when this would be actionable (similar to the logged-out sidebar). Shouldn't the UI just redirect to the login page right away when the user isn't logged in? 🤔 
* I have applied PR feedback in additional commits within this PR, partially so that they're easy to adjust or revert based on pending feedback by the core maintainers. If the PR doesn't get merged with a "squash" strategy, I'm happy to squash these manually before merging (or for one of the maintainers to do this for me).

Two minor things I noticed but didn't address in this PR yet (i think these can be addressed in follow-up PRs if desired):
* ~~Inside the new profile menu, the user avatar and the other icons are not horizontally centered (they're left-aligned). This is the default behavior of Vuetify's `v-list`, and I didn't find an obvious way to change this. In fact, `v-list-item` uses pretty hard-coded spacing between the prepend slot and the item's text, making customization of this a bit tricky (and probably not easy to maintain).~~ This is now addressed -- the user avatar is no longer showing inside of the menu
* When you click the theme switcher (dark/light mode), the menu closes (default behavior of the popup menu component). This is the same as it currently does in the "settings" menu in the sidebar. I think it would be nicer if the menu stayed open, to make it easier to toggle back if you don't like the other theme or if you clicked it accidentally (imagine blinding yourself by accidentally switching to light mode at night 🙈).

## Testing

Manually, running the dev server locally.

### DIFF
--- a/frontend/components/Layout/DefaultLayout.vue
+++ b/frontend/components/Layout/DefaultLayout.vue
@@ -233,6 +233,13 @@ export default defineNuxtComponent({
         restricted: false,
       },
       {
+        icon: $globals.icons.heart,
+        // TODO: how to hide this for non-logged-in view?
+        to: `/user/${$auth.user.value.id}/favorites`,
+        title: i18n.t("user.favorite-recipes"),
+        restricted: false,
+      },
+      {
         icon: $globals.icons.search,
         to: `/g/${groupSlug.value}/recipes/finder`,
         title: i18n.t("recipe-finder.recipe-finder"),
--- a/frontend/components/Layout/DefaultLayout.vue
+++ b/frontend/components/Layout/DefaultLayout.vue
@@ -235,7 +235,7 @@ export default defineNuxtComponent({
       {
         icon: $globals.icons.heart,
         // TODO: how to hide this for non-logged-in view?
-        to: `/user/${$auth.user.value.id}/favorites`,
+        to: `/user/${$auth.user.value?.id ?? ""}/favorites`,
         title: i18n.t("user.favorite-recipes"),
         restricted: false,
       },
--- a/frontend/components/Layout/LayoutParts/AppHeader.vue
+++ b/frontend/components/Layout/LayoutParts/AppHeader.vue
@@ -79,7 +79,6 @@
             <v-list-item :prepend-icon="$vuetify.theme.current.dark ? $globals.icons.weatherSunny : $globals.icons.weatherNight" :title="$vuetify.theme.current.dark ? $t('settings.theme.light-mode') : $t('settings.theme.dark-mode')" @click="toggleDark" />
             <!-- TODO: what was the use case for sidebar settings in logged-out state? -->
             <v-divider v-if="loggedIn" class="my-2" />
-            <v-list-item v-if="loggedIn" :prepend-icon="$globals.icons.cog" :title="$t('profile.user-settings')" to="/user/profile" />
             <v-list-item v-if="canManage" :prepend-icon="$globals.icons.manageData" :title="$t('data-pages.data-management')" to="/group/data" />
             <v-divider v-if="isAdmin" class="my-2" />
             <v-list-item v-if="isAdmin" :prepend-icon="$globals.icons.wrench" :title="$t('settings.admin-settings')" to="/admin/site-settings" />
--- a/frontend/components/Layout/LayoutParts/AppHeader.vue
+++ b/frontend/components/Layout/LayoutParts/AppHeader.vue
@@ -62,17 +62,31 @@
         <LanguageDialog v-model="languageDialog" />
         <v-menu z-index="2020">
           <template #activator="{ props }">
-            <v-btn v-bind="props" icon data-testid="user-menu-button">
+            <v-btn
+              v-if="!xs"
+              v-bind="props"
+              rounded="xl"
+              data-testid="user-menu-button"
+              class="ps-0"
+              height="40px"
+            >
+              <div class="d-flex align-center ga-2">
+                <UserAvatar list :user-id="sessionUser.id" :tooltip="false" />
+                <div>
+                  {{ sessionUser.fullName }}
+                </div>
+              </div>
+            </v-btn>
+            <v-btn v-else v-bind="props" icon data-testid="user-menu-button">
               <UserAvatar list :user-id="sessionUser.id" :tooltip="false" />
             </v-btn>
           </template>
           <v-list density="comfortable" color="primary" data-testid="user-menu">
-            <v-list-item :title="sessionUser.fullName ?? undefined" :to="userProfileLink">
-              <template #prepend>
-                <!-- TODO: horizontally center avatar with other icons in menu (why isn't it automatically centered?) -->
-                <UserAvatar list :user-id="sessionUser.id" :tooltip="false" />
-              </template>
-            </v-list-item>
+            <template v-if="xs">
+              <v-list-subheader :title="sessionUser.fullName ?? undefined" class="justify-center" />
+              <v-divider class="my-2" />
+            </template>
+            <v-list-item :prepend-icon="$globals.icons.user" :title="$t('settings.profile')" :to="userProfileLink" />
             <v-divider class="my-2" />
             <v-list-item :prepend-icon="$globals.icons.translate" :title="$t('sidebar.language')" @click="languageDialog=true" />
             <!-- TODO: prevent menu from closing when toggling light/dark mode -->
--- a/frontend/components/Layout/LayoutParts/AppHeader.vue
+++ b/frontend/components/Layout/LayoutParts/AppHeader.vue
@@ -73,7 +73,6 @@
                 <UserAvatar list :user-id="sessionUser.id" :tooltip="false" />
               </template>
             </v-list-item>
-            <v-list-item :prepend-icon="$globals.icons.heart" :title="$t('user.favorite-recipes')" :to="userFavoritesLink" />
             <v-divider class="my-2" />
             <v-list-item :prepend-icon="$globals.icons.translate" :title="$t('sidebar.language')" @click="languageDialog=true" />
             <!-- TODO: prevent menu from closing when toggling light/dark mode -->
@@ -125,7 +124,6 @@ export default defineNuxtComponent({
     const { loggedIn } = useLoggedInState();
     const isAdmin = computed(() => $auth.user.value?.admin);
     const canManage = computed(() => $auth.user.value?.canManage);
-    const userFavoritesLink = computed(() => $auth.user.value ? `/user/${$auth.user.value.id}/favorites` : undefined);
     const userProfileLink = computed(() => $auth.user.value ? "/user/profile" : undefined);
     const route = useRoute();
     const groupSlug = computed(() => route.params.groupSlug as string || $auth.user.value?.groupSlug || "");
@@ -181,7 +179,6 @@ export default defineNuxtComponent({
       sessionUser: $auth.user,
       smAndUp,
       toggleDark,
-      userFavoritesLink,
       userProfileLink,
       xs,
     };
--- a/frontend/components/Layout/LayoutParts/AppHeader.vue
+++ b/frontend/components/Layout/LayoutParts/AppHeader.vue
@@ -87,13 +87,12 @@
               <v-divider class="my-2" />
             </template>
             <v-list-item :prepend-icon="$globals.icons.user" :title="$t('settings.profile')" :to="userProfileLink" />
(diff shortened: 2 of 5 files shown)


frontend/layouts/admin.vue
```diff
@@ -15,7 +15,6 @@
       v-model="sidebar"
       absolute
       :top-link="topLinks"
-      :user="{ data: true }"
```
lehnerpat, line 18: comment: it seems this `user` prop of the sidebar wasn't even used anymore before my changes in this PR

frontend/components/Layout/LayoutParts/AppHeader.vue
```diff
@@ -58,17 +58,37 @@
...
+              </template>
+            </v-list-item>
+            <v-list-item :prepend-icon="$globals.icons.heart" :title="$t('user.favorite-recipes')" :to="userFavoritesLink" />
+            <v-divider class="my-2" />
+            <v-list-item :prepend-icon="$globals.icons.translate" :title="$t('sidebar.language')" @click="languageDialog=true" />
+            <!-- TODO: prevent menu from closing when toggling light/dark mode -->
+            <v-list-item :prepend-icon="$vuetify.theme.current.dark ? $globals.icons.weatherSunny : $globals.icons.weatherNight" :title="$vuetify.theme.current.dark ? $t('settings.theme.light-mode') : $t('settings.theme.dark-mode')" @click="toggleDark" />
+            <!-- TODO: what was the use case for sidebar settings in logged-out state? -->
+            <v-divider v-if="loggedIn" class="my-2" />
+            <v-list-item v-if="loggedIn" :prepend-icon="$globals.icons.cog" :title="$t('profile.user-settings')" to="/user/profile" />
```
lehnerpat: comment: right now, these `v-if="loggedIn"` conditionals here are redundant, because the entire user menu is only rendered when a user is logged in.

If we want to show a menu for logged-out users too (e.g. for switching language and theme), this needs to be reviews. If we don't want to support that use-case, these redundant conditionals should be removed. See PR description for more details.

lehnerpat:
The E2E tests are currently failing because they expect the user name to appear in the sidebar (which it now no longer does).

I'm looking into how to adjust them to find the user name in the user menu instead.

lehnerpat:
I've updated the E2E tests to work locally.

Edit: and they work in CI as well 👌

p0lycarpio:
That sounds good! Here are a few suggestions:
- To correct the avatar's misalignment with the other items on the list, could we use a generic "account" icon with the title and subtitle "Profile" + username?
- The entry to "User Settings" seems to be a slight duplication to me, and I don't see the added value it brings other than saving a click. Perhaps it could be removed if the previous suggestion is implemented?
- I would keep the shortcut to favorite recipes in the hamburger menu, below the "Create" button, as it is more common to navigate via this menu than the menu at the top right. This is equivalent to the "Favorites" menus found in other cloud storage services, for example.
- The language change button is not used very often, and I think we could move it to the user settings page instead, but that could be done in another PR.

lehnerpat:
@p0lycarpio, thanks for your feedback! 🙌 



> * To correct the avatar's misalignment with the other items on the list, could we use a generic "account" icon with the title and subtitle "Profile" + username?

Ah yes, good point. While looking around at other examples of this pattern, I originally thought some other sites put the profile image into the popover menu, and use that to link to the profile page. Looking around again now, that doesn't seem to be the case.

Some examples:

| Github | Paperless-ngx<br>(Desktop view) | Paperless-ngx<br>(Mobile view) |
| --- | --- | --- |
| CleanShot 2026-01-05 at 10 46 30@2x<br>This shows the profile image inside the popover, but it's not an active link. There's another "Profile" link below. Also, the icons don't try to be aligned with the profile image. | CleanShot 2026-01-05 at 10 48 02@2x<br>This does not replicate the profile image inside the menu (caveat: Paperless-ngx doesn't currently support profile images). It shows the user name in the app bar in Desktop mode. | CleanShot 2026-01-05 at 10 50 37@2x<br>In mobile mode, Paperless-ngx shows the user name inside the menu, which I find a useful idea. |


> * The entry to "User Settings" seems to be a slight duplication to me, and I don't see the added value it brings other than saving a click. Perhaps it could be removed if the previous suggestion is implemented?

Oh, good catch! I didn't even realize that "User Settings" links to the same page as the "Profile" link (what used to be the link of the profile image on the top left). This is clearly redundant.

In Mealie, the so-called "profile" page (at `/user/profile`) isn't really what you'd consider a user profile based on other apps; in my understanding, a user profile page is a place showing information about that user (profile image, name, contact info, bio, etc.).
Mealie instead shows more of an account overview or account management screen here (the subtitle text it shows is "Manage your profile, recipes, and group settings."). Confusingly, the subsection there titled "User Settings" is where you actually edit your profile information, such as your display name and email address.

Combining this with your first suggestion, I think I'll go for something like "Account" or "My Account" (though this may introduce a new translation string, I'll have to check if there is already a suitable one).



> * I would keep the shortcut to favorite recipes in the hamburger menu, below the "Create" button, as it is more common to navigate via this menu than the menu at the top right. This is equivalent to the "Favorites" menus found in other cloud storage services, for example.

Fair point, I didn't really reflect on this too much (since I never use this).
Leaving it in the left-side menu raises the question of its position in the order of the links; I expect users likely have some muscle memory about the links' current order, especially "Recipes" being the topmost one.

But adjusting the order based on further feedback should be easy, so I'll move it over there for the time being.



> * The language change button is not used very often, and I think we could move it to the user settings page instead, but that could be done in another PR.

Agreed that this doesn't need to be here (similar to the old Logout button in the top bar, the language switcher is likely too prominent here).
But I don't see a natural place where the language switcher would fit immediately in the user account page, so let's do this  in a separate PR indeed.

frontend/components/Layout/DefaultLayout.vue
```diff
@@ -232,6 +232,13 @@ export default defineNuxtComponent({
         title: i18n.t("general.recipes"),
         restricted: false,
       },
+      {
+        icon: $globals.icons.heart,
+        // TODO: how to hide this for non-logged-in view?
+        to: `/user/${$auth.user.value.id}/favorites`,
+        title: i18n.t("user.favorite-recipes"),
+        restricted: false,
+      },
```
lehnerpat, line 241: comment: this change now introduced a dependency of this file to the logged-in state of the user; how should we handle this?

p0lycarpio: You can set `restricted` to true for this.

frontend/components/Layout/LayoutParts/AppHeader.vue
```diff
@@ -58,17 +58,48 @@
...
-        :variant="smAndUp ? 'text' : undefined"
-        :icon="xs"
-        @click="logout()"
-      >
-        <v-icon :start="smAndUp">
-          {{ $globals.icons.logout }}
-        </v-icon>
-        {{ smAndUp ? $t("user.logout") : "" }}
-      </v-btn>
+      <div v-if="loggedIn && sessionUser" class="mx-1">
```
p0lycarpio, line 61: <div v-if="loggedIn && sessionUser" :class="xs ? 'pr-2' : 'px-2'">

frontend/components/Layout/LayoutParts/AppHeader.vue
```diff
@@ -58,17 +58,48 @@
...
+            </v-btn>
+            <v-btn v-else v-bind="props" icon data-testid="user-menu-button">
+              <UserAvatar list :user-id="sessionUser.id" :tooltip="false" />
+            </v-btn>
+          </template>
+          <v-list density="comfortable" color="primary" data-testid="user-menu">
+            <template v-if="xs">
+              <v-list-subheader :title="sessionUser.fullName ?? undefined" class="justify-center" />
+              <v-divider class="my-2" />
+            </template>
```
p0lycarpio, line 88: Instead of the block template above, here is another way to do this on mobile: 
```vue
 <v-list-item
    :prepend-icon="$globals.icons.user"
    :title="$t('settings.profile')"
    :subtitle="xs ? sessionUser.fullName || undefined : undefined"
    :to="userProfileLink"
    :lines="xs ? 'two' : 'one'"
  />
```

p0lycarpio (commented):
Thanks for the changes, it looks good! Here are a few more suggestions for the layout and spacing.

michael-genson (changes requested):
Generally like this change, I think it makes sense and solves the log out issue. A few notes:
- You can browse Mealie even if you aren't logged-in, so keeping language and theme settings available is important
- I don't think we should get rid of the setting menu, putting it in the lower left seems natural to me. Can we instead limit the new profile button to just "Profile" and "Logout" and keep the rest in settings?
- This might just be personal opinion but the button being rounded looks a little odd to me since the profile picture is already rounded and right up against the border, maybe we can make it less round and add some padding? Open to criticism on this, I'm more of a backend person
Image

michael-genson:

lehnerpat:
@michael-genson, apologies for not responding to your previous comment. I haven't been able to get back to this, and unfortunately I don't think I'll get to it soon.

Please feel free to take over this PR or adopt individual changes if you like.

Thanks for providing background on the logged-out usage! I've never used Mealie that way (and my instance isn't set up to make that meaningful), so that would've hard for me to test.

Regarding the menu vs settings;

> I don't think we should get rid of the setting menu, putting it in the lower left seems natural to me. Can we instead limit the new profile button to just "Profile" and "Logout" and keep the rest in settings?

I can see your perspective, but I think that's personal preference. For me, having settings in a menu in the bottom of the sidebar is quite unintuitive. I'm used to this kind of navigation being in the user menu (often at the top right or left, iconified by the user avatar and/or user name), so that feels most natural to me.

This is of course your decision as the maintainer team in the end. I'd encourage you to gather feedback about this from users to base the decision on that, in the absence of an overall design system for Mealie.

Thanks as always for your ongoing work on Mealie! ❤️

michael-genson:
No worries, appreciate your work on this!

We definitely need to keep the language selector and theme settings available, so either way we still need something accessible for non-logged-in users. If we get rid of the settings menu we need to relocate all of the settings to the profile button, and somehow make it accessible even if you're not logged-in. That seems pretty complicated, so I still think keeping the settings button makes more sense here.

Kuchenpirat:
Thanks for your work on this one @lehnerpat.
I think i will close this PR out for now, but feel free to reopen once this is again ready to review.

PR #6796, opened 2025-12-30
dev: Small .gitignore fixes

lehnerpat:

_(REQUIRED)_

This PR makes two small fixes to the .gitignore file (see inline comments for details).

_(REQUIRED)_

None.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -70,8 +70,11 @@ wheels/
 .installed.cfg
 *.egg
 
+# packaged output - temporarily written here by `uv build`
+/mealie-*
+
 # frontend copied into Python module for packaging purposes
-/mealie/frontend/
+/mealie/frontend
 
 # PyInstaller
 #  Usually these files are written by a python script from a template


.gitignore
```diff
@@ -70,8 +70,11 @@ wheels/
 .installed.cfg
 *.egg
 
+# packaged output - temporarily written here by `uv build`
+/mealie-*
```
lehnerpat, line 74: comment: The backend's packaging process includes a call to `uv build` (which in turn uses `setup-tools`), which temporarily writes the "source distribution" (I think?) to the directory `mealie-<version>` in the workspace root.

This directory is deleted after the build creates the `dist/mealie-<version>.tar.gz` file (successfully). It may be left around if the build command fails or is manually interrupted. It may also exist for a while if the build step is slow. In either case, this folder might confuse other tooling if it exists and is unignored, so ignoring it seems prudent.

(A quick LLM-based search and manual glance at the `uv` manual indicated that there's no standard way to tell the python build toolchain to use a different temporary folder that's easier to ignore.)

.gitignore
```diff
@@ -70,8 +70,11 @@ wheels/
...
+
 # frontend copied into Python module for packaging purposes
-/mealie/frontend/
+/mealie/frontend
```
lehnerpat, line 77: comment: In my devcontainer, the frontend's package command (`nuxt generate`) writes all its output files to `frontend/.output`, and then symlinks the actual static output files to `frontend/dist`. When the overall packaging process copies the frontend dist folder to `mealie/frontend`, that also is a symlink, which the ignore pattern `/mealie/frontend/` didn't catch for me.

`/mealie/frontend` catches it as intended even if it's a symlink.

michael-genson (approved):

PR #6797, opened 2025-12-30
dev: Add tasks for e2e tests

lehnerpat:
## What this PR does / why we need it:

_(REQUIRED)_

This PR adds some tasks (to `Taskfile.yml`) for running e2e tests.
This improves developer experience by making it easier to run the e2e tests locally, shortening the feedback cycle when any UI / features covered by E2E tests are changed.

Note: #6798 further improves the DX by improving `task`'s caching of several tasks.

Here's a dependency graph of all the new tasks (all tasks except for `py:package` are new):

```mermaid
graph BT
    py_package["py:package"]
    e2e_build_image["e2e:build-image"]
    e2e_start_server["e2e:start-server"]
    setup_e2e["setup:e2e"]
    e2e_test["e2e:test"]
    e2e["e2e"]
    e2e_stop_server["e2e:stop-server"]
    
    %% Dependencies flow
    e2e_build_image --> py_package
    e2e_start_server --> e2e_build_image
    e2e_test --> setup_e2e

    
    e2e --> e2e_start_server
    e2e --> e2e_test
    e2e --> e2e_stop_server
    
```

## Which issue(s) this PR fixes:

_(REQUIRED)_

Fixes https://github.com/mealie-recipes/mealie/issues/6943

## Special notes for your reviewer:

It would probably be a good idea for the reviewer to test these new tasks :)

## Testing

I ran all these tasks (often).

PR #6798, opened 2025-12-30
dev: Improve caching in taskfile

lehnerpat:

_(REQUIRED)_

This PR improves developer experience by adding more caching (["preventing unnecessary work"](https://taskfile.dev/docs/guide#prevent-unnecessary-work)) to some tasks in `Taskfile.yml`.

This is mainly done by adding `sources:` definitions to more tasks, so `task` can tell if their inputs changed (and thus the tasks don't need to be re-run).

This is especially helpful for tasks that take a long time, such as building the frontend (`ui:generate`) or packaging the whole app (`py:package`). 

It makes running the e2e tests locally much faster, since the tooling can just skip re-doing a lot of unnecessary steps (related to #6797).

_(REQUIRED)_

None

It would probably be a good idea for the reviewer to test these new tasks :)

## Testing

I tried all these tasks to check that they won't be re-run if there are no changes to their inputs.

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -47,8 +47,6 @@ tasks:
     sources:
       - package.json
       - yarn.lock
-    generates:
-      - node_modules/**
 
   setup:py:
     desc: setup python dependencies
@@ -179,12 +177,21 @@ tasks:
     status:
       - '{{ .SKIP_PACKAGE_DEPS | default "false"}}'
 
-  py:package:
-    desc: builds Python packages (sdist and wheel) in top-level dist directory
+  py:package:build:
+    internal: true
     deps:
       - py:package:deps
     cmds:
       - uv build --out-dir dist
+    sources:
+      - uv.lock
+      - pyproject.toml
+      - mealie/**
+
+  py:package:
+    desc: builds Python packages (sdist and wheel) in top-level dist directory
+    cmds:
+      - task: py:package:build
       - task: py:package:generate-requirements
 
   py:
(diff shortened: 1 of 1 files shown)


Taskfile.yml
```diff
@@ -47,8 +47,6 @@ tasks:
     sources:
       - package.json
       - yarn.lock
-    generates:
-      - node_modules/**
```
lehnerpat, line 51: comment: I don't know if this is as-intended by `task` or a bug, but if `generates:` is present here, it always re-runs the commands and doesn't cache as expected.
It seems like nothing currently relies on this `generates` specification, so I removed it to allow `task` to cache the frontend's `yarn install`.

Taskfile.yml
```diff
@@ -179,12 +177,21 @@ tasks:
...
+  py:package:build:
+    internal: true
     deps:
       - py:package:deps
     cmds:
       - uv build --out-dir dist
+    sources:
+      - uv.lock
+      - pyproject.toml
+      - mealie/**
```
lehnerpat, line 189: comment: splitting the `uv build` call into a separate internal task ensures that the `sources` definition is closely scoped

Taskfile.yml
```diff
@@ -223,18 +236,36 @@ tasks:
...
     desc: runs the frontend linter
     dir: frontend
     cmds:
       - yarn lint --max-warnings=0
+    sources:
+      - "**"
+      - exclude: .nuxt/**
+      - exclude: .output/**
+      - exclude: dist/**
+      - exclude: node_modules/.cache/**
```
lehnerpat, line 256: comment: the sources definitions are probably overly broad for some of these frontend tasks, but this works and should reduce some unnecessary re-runs.

The downside of too-broad sources lists is that `task` takes longer to compute the checksum for the change detection. 

Happy for other people to finetune these later.

michael-genson (approved):
LGTM! There are instances where I can imagine devs want to re-run frontend tests without changing any files, but you can use `task -f ui:test` and bypass this so I don't see harm in merging it as-is

PR #6803, opened 2025-12-30
fix: #6802 prevent 500 internal server error when patching recipe tags

### LINKED ISSUE #6802: [BUG] - 500 Internal Server Error when Patching Recipe Tags
### First Check

A 500 Internal Server Error is returned by the Mealie API when attempting to PATCH a recipe's tags using the /api/recipes/{slug} endpoint. This occurs when the payload includes new tags that do not yet exist in the database.

It appears related to the RecipeTag schema validation missing the group_id field during the update process, causing a TypeError in the backend when it attempts to initialize the tag object.

### Steps to Reproduce

1. Have a running instance of Mealie.
2. Identify an existing recipe slug.
3. Send a PATCH request to /api/recipes/{slug} with a payload that includes a new tag name that does not exist in the database.

```

{
  "tags": [
    {
      "name": "New Tag",
      "slug": "new-tag"
    }
  ]
}
```

4. Observe the 500 Internal Server Error response.

`TypeError: init() missing 1 required positional argument: 'group_id'
`
(Note: This is the expected traceback based on similar reports; please confirm with your actual server logs if available).

### Mealie Version

Build ID: 4fbace36-29e6-4cd6-94f8-f7445fd74855

### Deployment

Other (please specify below)

Proxmox LXC Container

Interacting with Mealie via Python script using urllib. The issue is reproducible consistently via API.

SimeonSouttar:
### What this PR does / why we need it:
This PR fixes a 500 Internal Server Error `(TypeError: init() missing 1 required positional argument: 'group_id')` that occurs when patching a recipe with new tags via the API.

The issue was caused by `RecipeService.patch_one` not injecting the current user's group_id into new tags before attempting to update the recipe. The Tag ORM model requires group_id for initialization, but this field is typically excluded from the RecipeTag Pydantic model during patch operations (where `exclude_unset=True` is used).

### Changes:

- mealie/services/recipe/recipe_service.py: Updated patch_one to iterate over `patch_data.tags` (if present) and explicitly inject `self.user.group_id` into any tag that needs it.

### Which issue(s) this PR fixes:

- Fixes #6802

### Testing
I have added a new integration test case to reproduce the bug and verify the fix.

- New Test: tests/integration_tests/user_recipe_tests/test_issue_6802_patch_tags.py

- Verification: Ran the test locally with Python 3.12.
- Command: `pytest tests/integration_tests/user_recipe_tests/test_issue_6802_patch_tags.py`

- Result: PASSED


michael-genson (changes requested):
Thanks for this! A couple small things:
-  Is it possible to move this to the recipe repo instead of the service? That way we can fix this for updates as well as patches
- Can we do this for categories/tools too? I'm guessing it's the same issue.
- Can we move the test to `tests/unit_tests/repository_tests/test_recipe_repository.py`? (assuming we move the functionality to the repo)

SimeonSouttar (commented):
@michael-genson updated inline with your previous comments

michael-genson (dismissed):
Thank you!

michael-genson:
@SimeonSouttar it looks like there are lint errors, can you check on those?

PR #6805, opened 2025-12-31
`feat:` Add language translation to OpenAI recipe parser

lcs-crr:

Translation is currently only available for image parsing. This PR allows the OpenAI parser to translate its inputs. The original idea was proposed in discussion #6784 with @michael-genson. 
The changes made include:

- `frontend/app/components/Domain/Recipe/RecipePage/RecipePageParts/RecipePageParseDialog.vue`: Added a `shouldTranslate` check box with default of `false` to minimise disruption and wired the checkbox + user's UI locale (i18n.locale) into the `parseIngredients` call as `translateLanguage`. This checkbox only appears when the OpenAI parser is selected. 
Before: 
image
After:
image

- `frontend/app/lib/api/user/recipes/recipe.ts`: Added `translateLanguage` as a parameter to the `parseIngredients` and `parseIngredient` methods.
- `mealie/schema/recipe/recipe_ingredient.py`: Added `translate_language` as a field to the `IngredientsRequest` and `IngredientRequest` classes.
- `mealie/services/parser_services/openai/parser.py`: Added `translate_language` as a parameter to the `_parse`, `parse_one` and `parse` methods. Added if statement which injects command to translate to the user message.
- `mealie/services/parser_services/_base.py`: Added `translate_language` as a parameter to the `parse_one` and `parse` abstract methods.
- `mealie/services/parser_services/ingredient_parser.py`: Added `translate_language` as a parameter to the `parse_one` and `parse` methods to the other parser types for consistency. 
- `tests/unit_tests/services_tests/ingredient_parser/test_openai_parser.py`: Added test to ensure translation command is injected into user message if flag is provided.

## Note

This PR only adds translation for ingredients, not recipe title, description or steps.

No related issue, but feature discussed in #6784.

## Testing

Tested manually through UI and ran `task py:check`; all passing.


mealie/services/parser_services/openai/parser.py
```diff
@@ -175,12 +177,12 @@ async def _parse(self, ingredients: list[str]) -> OpenAIIngredients:
             ingredients=[ingredient for response in responses for ingredient in response.ingredients]
         )
 
-    async def parse_one(self, ingredient_string: str) -> ParsedIngredient:
-        items = await self.parse([ingredient_string])
+    async def parse_one(self, ingredient_string: str, translate_language: str | None = None) -> ParsedIngredient:
```
lcs-crr, line 180: Could it be that this method is never called? I'm assuming it has to exist because of the abstract method in _base.py, but imo it is a bit confusing to have any logic within it beyond a `raise NotImplementedError("This method is not supported.")`

lcs-crr:
No capacity to work on it :/

lcs-crr:
Hi @michael-genson :) Do you reckon this makes sense to merge? I'll close the PR otherwise

michael-genson:
In general I like this idea, but I think this might be better suited as a post processing step:
1. Import the recipe however you want
2. AI post-processing (like translation)
3. Parsing

This isn't something we support today but is something I want to have implemented in the near future. There are a few things messy about putting it all together (e.g. how the other parsers currently receive the "translate" flag but ignore it).

michael-genson:
With the new [AI workflow](https://github.com/mealie-recipes/mealie/pull/8043) approach I don't think this is needed anymore. You can now translate the recipe for any input type through the AI import page.

In the (probably near) future I want to build parsing into the importing process rather than as a separate stage, which should also help with translations etc.

Appreciate your work on this, I'm going to close it since I think the new approach handles this more naturally for users.

PR #6810, opened 2026-01-01
fix: Make quantity input in shopping list item editor visually consistent with other inputs

lehnerpat:

_(REQUIRED)_

This PR makes the styling of the quantity input field in the item edit view on the shopping list more consistent with the other inputs in that form.

In particular on mobile, the quantity input was quite small and hard to tap.

Screenshots:

| | Desktop | Mobile |
|---|---|---|
| Before | CleanShot 2026-01-01 at 13 30 06@2x | CleanShot 2026-01-01 at 13 30 33@2x |
| After | CleanShot 2026-01-01 at 17 36 15@2x | CleanShot 2026-01-01 at 17 35 32@2x |

_(REQUIRED)_

None.

### DIFF
--- a/frontend/components/Domain/ShoppingList/ShoppingListItemEditor.vue
+++ b/frontend/components/Domain/ShoppingList/ShoppingListItemEditor.vue
@@ -10,11 +10,8 @@
               :label="$t('form.quantity-label-abbreviated')"
               :min="0"
               :precision="null"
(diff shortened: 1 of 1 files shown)


michael-genson (approved):
Thanks! I don't remember why I did it this way but this looks much better

PR #6815, opened 2026-01-02
fix: service-worker precache manifest entries not generated correctly

morgan-dgk:
The precache manifest generated by workbox-build is not currently populated correctly. The manifest does not include the app shell nor almost any of the assets produced during the build process.

Importantly, this results in a failure to install the handler for the navigateFallback route which should point to the app shell at `/`.

The cause of this issue appears to be related tohow the glob package used by workbox-build is transpiled from typescript. Upgrading @vite-pwa/nuxt appears to resolve this issue.

Additionally, the empty 200 and 400 pages generated by nuxt also need to be excluded from the service worker pre-cache to avoid `bad-precaching-response, status: 404` errors.

Application assets are not cached correctly and the application shell is not correctly installed as the navigationFallback route. This results in failures when attempting to navigate while offline. 

## Testing

Tested manually locally using Docker and chrome developer tools

### DIFF
--- a/frontend/nuxt.config.ts
+++ b/frontend/nuxt.config.ts
@@ -24,8 +24,9 @@ export default defineNuxtConfig({
   devtools: {
     enabled: false,
   },
+
   app: {
-    baseURL: process.env.SUB_PATH || "",
+    baseURL: process.env.SUB_PATH || "/",
 
     head: {
       title: "Mealie",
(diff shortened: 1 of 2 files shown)


frontend/nuxt.config.ts
```diff
@@ -24,8 +24,9 @@ export default defineNuxtConfig({
   devtools: {
     enabled: false,
   },
+
   app: {
-    baseURL: process.env.SUB_PATH || "",
+    baseURL: process.env.SUB_PATH || "/",
```
morgan-dgk, line 29: It appears that it is necessary to set this value since nuxt injects a ManifestTransform that replaces `index.html` with  this value in the precache manifest.

michael-genson (approved):
LGTM, thanks!

PR #6857, opened 2026-01-09
feat: Add days in the past selector on meal planner

arnassavickas:
### Summary
Adding `Default Days in the Past` selector in the Meal Planner. It allows to set how many days in the past meal planner should open by default.

### Why
I found this feature useful when a meal is planned one day, but not made on that day. By default, that meal is then hidden on the previous day.

Having an option to show 1 or 2 days in the past by default, allows to easily see and move those meals if needed.

### Screenshots
| Before | After |
|--------|--------|
| image | image |
| image | image |

### Changes
- `frontend/composables/use-users/preferences.ts`
  - added `numberOfDaysPast` to the user preferences.
- `frontend/lang/messages/en-US.json`
  - added new translation for the days in the past input.
- `frontend/pages/household/mealplan/planner.vue`
  - added the new days in the past input and handling for it.
  - fixed `adjustForToday` to correctly adjust days in the past.
- `frontend/pages/household/mealplan/planner/view.vue`
  - added highlight to the current day's title.


michael-genson (approved):
LGTM thanks! I think we can do better with the meal planner ranges but this is great for now.
I fixed the merge conflicts and fixed a small bug.

PR #6885, opened 2026-01-16
fix: prevent XSS via javascript: URIs in recipe actions

hay-kot:
## Summary

- Add URL scheme validation to recipe actions to only allow http/https schemes
- This prevents stored XSS attacks where malicious users could create recipe actions with `javascript:` URIs that execute when other users click them
- Also blocks other unsafe schemes: `data:`, `file:`, `ftp:`, protocol-relative URLs

## Test plan

- [x] Added parametrized tests covering valid schemes (http, https) and invalid schemes (javascript, data, file, ftp, protocol-relative, schemeless)
- [x] All existing tests pass
- [x] Linter passes

### DIFF
--- a/mealie/schema/household/group_recipe_action.py
+++ b/mealie/schema/household/group_recipe_action.py
@@ -1,7 +1,7 @@
 from enum import Enum
 from typing import Any
 
-from pydantic import UUID4, ConfigDict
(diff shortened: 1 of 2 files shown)


michael-genson (approved):
LGTM

Issue #4773, opened 2024-12-25
[BUG] - Cook Mode ignores ingredient sections


sckilcoyne:
### First Check

- [X] This is not a feature request.
- [X] I added a very descriptive title to this issue (title field is above this).
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Mealie documentation, with the integrated search.
- [X] I already read the docs and didn't find an answer.
- [X] This issue can be replicated on the demo site (https://demo.mealie.io/).

### What is the issue you are experiencing?

You can create sections of ingredients in a recipe, but in Cook Mode, all ingredients are grouped together without the section separations. Sections of the instructions remain in Cook Mode.

### Steps to Reproduce

1. Create a recipe with ingredient sections
2. Go to Cook Mode and see the sections are ignored

### Please provide relevant logs

n/a

### Deployment

Docker (Synology)

ChipSkylark37:
Somewhat related, but not the same issue:

If you have sections for ingredients, and also linked ingredients to steps:
1. If you print the recipe and have linked ingredients to steps, those don't show on cook mode. 
2. If you enter cook mode in the app, you lose the ingredient sections, but linked ingredients show.

I forget where I had posted this, whether it was in a Feature Request / Discussion or perhaps in one of the Survey - Feedback areas.

sckilcoyne:
this is still a problem I'd like to see addressed

Kuchenpirat:
This was a design decision we made when we implemented the new cook mode, but i think we can do better here. 
I will mark this as task so it doesn't get marked as stale.

codefaux:
> If you print the recipe and have linked ingredients to steps, those don't show on cook mode.

@ChipSkylark37 -- Since I was in the code I decided to take a look. This is not a related issue, and you should file a separate one for it. Specifically, RecipePrintView.vue is responsible for assembling the recipe page into a printable view, and its function is beyond me and I can't approach it. It makes no attempt to find or assemble linked recipe objects, but I don't understand them well enough to approach it. Lines 63 thru 95 (and supporting function calls below) would essentially need to be replicated for LinkedIngredients using their no doubt different object layout.

michael-genson:
This has since been resolved

PR #6946, opened 2026-01-28
fix: Keep ingredient headers in cook mode

codefaux:

Simply; (Opinion) Cook Mode should not remove ingredient group headers as it degrades ease of use.

In a bit more detail; I provide Mealie to several users. Today discussing fixes and changes, the reply was "Did they fix ingredient group headers in Cook Mode yet?"
This led to a discussion of exactly what that meant, and it struck me as negative design. I asked the other users I know, and they agreed, so here we are.

I considered implementing a toggle or tickbox or some such but that's above my head even more than this was.

What do you folks think?

[#4773](https://github.com/mealie-recipes/mealie/issues/4773)

(That's funny, I didn't check if anyone else asked for it before, lol.)

I thought about setting it to use a smaller text size for the ingredient group header in cook mode, but I genuinely have no idea where to start. If you'd rather see this, please let me know which class to assign it.

## Testing

Deployed on my (semi-public) personal server, with a few users. I've asked them to "just make sure most of the things work the same" and reports are positive.


codefaux:
Also in case you want to see it without the whole build and deploy process, here's a screenshot of part of one we transcribed from a Youtube channel;
image


EDIT: Collapsable ingredient groups might be neat? That's a bit beyond me at the moment but if you want it, I'll try.

hay-kot (approved):
LGTM, thanks for contributing!

hay-kot (approved):

Issue #6263, opened 2025-09-27
[UX] On profile picture upload, no preview


6543:
If you upload a own custom prfile picture, you only get the feedback of an successfull upload after you hit the save (update) button. 

I suggest to show the new avatar as preview as soon as the upload is successfull and save it permanently if the update button is hit

Also while it is uploading you could add an animation to let users notice that there is something going on (optional).

<img width="1713" height="758" alt="Image" src="https://github.com/user-attachments/assets/50b59556-b3d0-4343-9def-96031b5b2e5b" />

6543:
witch makes it not irrelevant

6543:
@michael-genson could you make it so that it is not going to be closed?

PS: i think this could be labled "good first issue" ?

michael-genson:
We don't auto-close issues anymore, but I went ahead and removed the stale tag!

6543:
thanks :)

PR #6985, opened 2026-02-02
fix: #6263 Force user avatar refresh after upload

Sim-sat:

This pr addresses an issue where the uploaded avatar is only visible after reloading the page due to browser caching.

## Changes made:
- added a key to force a rerender of the UserAvatar component. This is the change with the least impact on other components. 
- added a handler function that increments this key.

#6263 

- this is my first contribution, so I picked something small. I'm happy about any feedback or guidance. I choose to increment the key instead of using a UUID or the current Date to make the code easier to debug and read.

## Testing

I ran all tests and verified the change actually works.

### DIFF
--- a/frontend/pages/user/profile/edit.vue
+++ b/frontend/pages/user/profile/edit.vue
@@ -4,6 +4,7 @@
       <template #header>
         <div class="d-flex flex-column align-center justify-center">
           <UserAvatar
+            :key="avatarKey"
             :tooltip="false"
             size="96"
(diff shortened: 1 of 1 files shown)


michael-genson:
Thanks for this, though this happens due to a bug with using a reserved variable name (see: https://github.com/mealie-recipes/mealie/pull/7033).

Going to close this in favor of https://github.com/mealie-recipes/mealie/pull/7033

PR #6995, opened 2026-02-03
fix: use BASE_URL config for redirect_url  if available

### LINKED ISSUE #6038: [BUG] - Mealie redirect link for OIDC (authentik) is stuck to http when everything is set to https
### First Check

I have on my home server authentik as a authentication service  and mealie and i'm trying to link the both of them 
my current problem is that when clicking on "Login with Authentik", the reirect url send by mealie to my authetik service uses http instead of https.
My server is hidden behind cloudflare's tunnel and nginx proxy manager.
Other services (affine, outline) are already using authentik with absolutely no error so i don't think that is the problem.

here is what i did:

by reading the [official OIDC documentation V1] (https://docs.mealie.io/documentation/getting-started/authentication/oidc/) it is written that
>If you are hosting Mealie behind a reverse proxy (nginx, Caddy, ...) to terminate TLS, make sure to start Mealie's Gunicorn server with --forwarded-allow-ips=<ip-of-proxy>, otherwise the X-Forwarded-* headers will be ignored and the generated OIDC redirect URI will use the wrong scheme (http instead of https)
Howeaver if i understood right mealie currently uses uvicorn instead of Gunicorn.
For that reason i added `entrypoint: uvicorn mealie.app:app --host 0.0.0.0 --port 9000 --forwarded-allow-ips="*"`  to my docker-compose file. 

I hard after that my error could be linked to my reverse proxy making me add lines to my Custom Nginx Configuration for the mealie host.

I did nothing regarding Cloudflare since i don't know what i can do.

I have an http acces to mealie but refuse myself from using it since i don't know how much it would reduce security.
If someone could give me an answer.

### Steps to Reproduce

i'm gonna write here my full mealie setup.

docker compose:

```yml
services:
  mealie:
    image: ghcr.io/mealie-recipes/mealie:nightly
    container_name: mealie
    restart: unless-stopped
    ports:
      - "${PORT:-9925}:9000"
    expose:
      - 9000
    env_file:
      - .env
    deploy:
      resources:
        limits:
          memory: 1000M
    volumes:
      - mealie-data:/app/data/
    entrypoint: uvicorn mealie.app:app --host 0.0.0.0 --port 9000 --forwarded-allow-ips="*" # Trust NPM headers for HTTPS
    environment:
      ALLOW_SIGNUP: "false"
      PUID: 1000
      PGID: 1000
      BASE_URL: ${BASE_URL}
      # Database Settings
    depends_on:
      postgres:
        condition: service_healthy
    networks:
        - mealie_net
        - intra

  postgres:
    container_name: postgres
    image: postgres:15
    restart: unless-stopped
    env_file:
      - .env
    volumes:
      - mealie-pgdata:/var/lib/postgresql/data
    environment:
      PGUSER: ${POSTGRES_USER}
    healthcheck:
      test: ["CMD", "pg_isready"]
      interval: 30s
      timeout: 20s
      retries: 3
    networks:
        - intra

volumes:
  mealie-data:
  mealie-pgdata:

networks:
  intra:
    internal: true
  mealie_net:
    external: true
```

.env 

```
BASE_URL=https://<FQDN>
PREFERRED_URL_SCHEME=https
TZ=Europe/Paris
PORT=9925

#postgres
DB_ENGINE=postgres
POSTGRES_USER=<User>
POSTGRES_DB=mealie
POSTGRES_PASSWORD=<Pasword>
POSTGRES_SERVER=postgres

#OIDC
OIDC_AUTH_ENABLED=true
OIDC_PROVIDER_NAME=authentik
OIDC_CONFIGURATION_URL=https://<FQDN>/application/o/mealie/.well-known/openid-configuration
OIDC_CLIENT_ID=<Cliend ID>
OIDC_CLIENT_SECRET=<Client Secret>
OIDC_SIGNUP_ENABLED=true
OIDC_USER_GROUP=Mealie_Users
OIDC_ADMIN_GROUP=Mealie_Admins
OIDC_AUTO_REDIRECT=false
```

Nginx proxy manager do not uses the local server IP for reverse roxy but the name of the docker image.

Image

Nginx proxy manager's Custom Nginx Configuration:

```
proxy_set_header Host $host;
proxy_set_header X-Real-IP $remote_addr;
proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
proxy_set_header X-Forwarded-Proto https;

proxy_set_header Upgrade $http_upgrade;
proxy_set_header Connection "upgrade";
```

mealie login when clicking "Login with Authentik" 
`INFO     2025-08-26T18:09:36 - [2a01:cb19:f86:e500:c493:1886:4e2d:9f7:0] 302 Found "GET /api/auth/oauth HTTP/1.1"`

authentik login when clicking "Login with Authentik"
```
{"auth_via": "unauthenticated", "domain_url": "<FQDN>", "event": "The request fails due to a missing, invalid, or mismatching redirection URI (redirect_uri).", "host": "<FQDN>", "level": "warning", "logger": "authentik.providers.oauth2.views.authorize", "pid": <PID>"request_id": "<ID>", "schema_name": "public", "timestamp": "2025-08-26T16:11:29.220991"}
{"auth_via": "unauthenticated", "domain_url": "<FQDN>", "event": "/application/o/authorize/?response_type=code&client_id=kb6LeXxxcnGVb2jWwHNeM0qgiDOlygMBBn7CCKfe&redirect_uri=http%3A%2F%2F<FQDN>%2Flogin&scope=openid+email+profile+groups&state=EHLUUmdsYZNsNL1xH04TislwRnTMsT&nonce=fTjDlKcXFCVqkUb0GKht&code_challenge=qvZ0tMnXR8qbI9Uc1O5xuszUZrYeSH7N-NoS-_JmWZY&code_challenge_method=S256", "host": "<FQDN>", "level": "info", "logger": "authentik.asgi", "method": "GET", "pid": 92575, "remote": "<IP ?>", "request_id": "66a8807eda1044e49357f517e85e91e6", "runtime": 16, "schema_name": "public", "scheme": "https", "status": 400, "timestamp": "2025-08-26T16:11:29.224788", "user": "", "user_agent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:142.0) Gecko/20100101 Firefox/142.0"}
```

### Mealie Version

_No response_

### Deployment

Docker (Linux)

i am on a server running proxmox. the docker is on a VM where all my other services are running (still learning how to use any of it).

whattheschnell:

_(REQUIRED)_

Currently, the OIDC `oauth_login` route calculates the` redirect_url` using `request.base_url`. This relies entirely on the incoming Host or X-Forwarded-Host headers from a reverse proxy.

In many standard reverse proxy setups (like Nginx Proxy Manager default configs), these headers may not match the external FQDN, causing the generated `redirect_url` to default to an internal docker container IP or incorrect scheme (http vs https). This results in a 404 or a "Redirect URI mismatch" from the OIDC provider.

The Fix This PR updates the logic to prioritize the user-configured BASE_URL environment variable.

If `settings.BASE_URL` is set, it uses that as the source of truth for the redirect.

If not, it falls back to the existing `request.base_url` dynamic detection.

Related Issue Fixes OIDC redirect 404s behind proxies that do not pass X-Forwarded-Host.

_(REQUIRED)_

Fixes https://github.com/mealie-recipes/mealie/issues/6038


mealie/routes/auth/auth.py
```diff
@@ -99,7 +99,9 @@ async def oauth_login(request: Request):
         # in development, we want to redirect to the frontend
         redirect_url = "http://localhost:3000/login"
     else:
-        redirect_url = URLPath("/login").make_absolute_url(request.base_url)
+        # Prioritize User Configuration over Request Headers
+        base = settings.BASE_URL if settings.BASE_URL else request.base_url
```
michael-genson: FYI you can do

```python
base = settings.BASE_URL or request.base_url
```

and achieve the same thing. Nice syntax shortcut which returns the first truthy value.

michael-genson (dismissed):
LGTM

mealie/routes/auth/auth.py
whattheschnell: Nitpicking, but probably have to do `request.base_url` otherwise FastAPI could throw an `AttributeError`

michael-genson: Oh yeah oops that's a typo

michael-genson:
Might need to take a look at the E2E tests, seems like they're hanging

whattheschnell:
> Might need to take a look at the E2E tests, seems like they're hanging

there is no BASE_URL env variable in the /tests/e2e/docker/docker-compose.yml

There must be some issue in the E2E between mealie running on localhost:9000 and the OIDC server running on localhost:8080.  

Another reason why this needs to be included somehow.  Because Mealie was using the `request.base_url` instead of its own internal BASE_URL, it reconstructs the redirect URIs base on what it was passed from the OIDC server.

michael-genson:
Since we set a default `BASE_URL` is there risk in breaking existing instances with this change? For instance should we handle `localhost` base urls differently? Or maybe check if we're using the default?

Might be a dumb question, not too familiar with OIDC implementations.

whattheschnell:
It still has the mechanism in place to use the request.base_url if there isn't anything set for settings.BASE_URL.

If a user explcitly sets BASE_URL=localhost:9000 but use it publicly, that a user misconfiguration, not a bug.  Its still valid for localized testing or deployments.  If they only set BASE_URL=localhost, OIDC would correctly fail

We're only setting BASE_URL in the E2E test, not forcing it as a required env variable. I didn't notice the BASE_URL default in /mealie/core/settings/settings.py though.

```
class AppSettings(AppLoggingSettings):
    theme: Theme = Theme()

    BASE_URL: str = "http://localhost:8080"

```

So we probably need to check in auth.py if settings.BASE_URL differs from the default (ie a env was passed to overwrite the default in settings.py) I'm not sure the most elegant way to do this, but what about something like this:

I don't love hardcoding the check, but this is what I'd want to accomplish.  

```
if not settings.PRODUCTION:
        redirect_url = "http://localhost:3000/login"
else:
        # Only use config if it's set AND different from the default value from settings.py
        # This prevents breaking existing instances that rely on dynamic headers
        if settings.BASE_URL and settings.BASE_URL != "http://localhost:8080":
            base = settings.BASE_URL
        else:
            base = request.base_url
            
        redirect_url = URLPath("/login").make_absolute_url(base)

```

whattheschnell:
An architectural alternative is whether we want to require users to set a BASE_URL, rather than provide a default.


```
BASE_URL: str | None = None
```

michael-genson:
In the interest of breaking as little as possible I'd be fine with a hardcoded check. Let's add the logic to settings though instead of in the OIDC module. Something like:


```python
@property
def is_default_base_url(self):
    return self.BASE_URL == "..."
```

And then reference `settings.is_default_base_url` in the OIDC logic

(on my phone or I'd be a bit more specific, let me know if that doesn't make sense)

whattheschnell:
> In the interest of breaking as little as possible I'd be fine with a hardcoded check. Let's add the logic to settings though instead of in the OIDC module. Something like:
> 
> ```python
> @property
> def is_default_base_url(self):
>     return self.BASE_URL == "..."
> ```
> 
> And then reference `settings.is_default_base_url` in the OIDC logic
> 
> (on my phone or I'd be a bit more specific, let me know if that doesn't make sense)

Yea, i got the idea. I think this should address it that way

michael-genson (dismissed):
Thank you!

whattheschnell:
Not sure what the formatting failure is

michael-genson:
I went ahead and fixed it, no worries. Thank you!

DjCoke:
Thank you guys for fixing this! I was pulling my hairs out with this one...

Issue #6887, opened 2026-01-16
[BUG] - JSON Import Ignores Instruction Titles, Categories, Summary, Notes, and Tools


mswdev:
### First Check

- [x] This is not a feature request.
- [x] I added a very descriptive title to this issue (title field is above this).
- [x] I used the GitHub search to find a similar issue and didn't find it.
- [x] I searched the Mealie documentation, with the integrated search.
- [x] I already read the docs and didn't find an answer.
- [x] This issue can be replicated on the demo site (https://demo.mealie.io/).

### Preface: Addressing Schema Validity

This bug report documents fields that **Mealie's internal schema supports** but the JSON importer silently discards. 

I want to address a pattern I've seen in previous discussions (#4490, #6406) where similar reports were dismissed with "that's not valid schema.org." **This defense doesn't hold for two reasons:**

1. **Several of these fields ARE valid schema.org** (see table below)
2. **Even for fields that aren't schema.org, if Mealie supports them internally, the importer should too.** At minimum, Mealie should be able to import its own export format.

| Field | Mealie Supports? | Schema.org? |
|-------|-----------------|-------------|
| `recipeInstructions[].title` | ✅ Section header | ✅ via `HowToSection.name` |
| `recipeInstructions[].summary` | ✅ Step title | ❌ Mealie-specific |
| `recipeIngredient[].title` | ✅ Section header | ❌ Mealie-specific |
| `recipeIngredient[].display` | ✅ Ingredient notes | ❌ Mealie-specific |
| `notes[]` | ✅ Recipe notes | ❌ Mealie-specific |
| `recipeCategory` | ✅ Yes | ✅ `recipeCategory` |
| `tools` | ✅ Yes | ✅ `tool` (from HowTo) |
| `tags` | ✅ Yes | ✅ via `keywords` |

### Describe the Bug

When importing recipes via JSON (API endpoint `/api/recipes/create/html-or-json` or UI "Import from HTML or JSON"), several fields that Mealie's data model supports are silently discarded:

#### 1. Instructions: `title` and `summary` ignored

Mealie's `RecipeStep` schema has three fields:
- **`title`** - Section header (toggleable in UI, groups steps like "Make the Sauce")
- **`summary`** - Individual step title (appears above the step text)
- **`text`** - The instruction content

**Bug:** The importer hardcodes `title=""` and never extracts `summary`.

**Code location:** `mealie/services/scraper/scraper_strategies.py` (~line 127)
```python
[...]

PR #7017, opened 2026-02-06
fix: restore recipe notes during JSON import

materialsportler:

This PR fixes silent data loss where the `notes` field is dropped during recipe import operations via JSON import and URL scraping.

**Problem:**
When importing recipes through `/api/recipes/create/html-or-json` or `/api/recipes/create/url`, the `notes[]` array is silently ignored and not persisted to the database, even though the API returns success (201 Created).

**Root Cause:**
The `RecipeScraperPackage.clean_scraper()` method in `mealie/services/scraper/scraper_strategies.py` did not map the `notes` field from schema.org JSON to the Mealie `Recipe` object.

**Changes:**

**File: `mealie/services/scraper/scraper_strategies.py`**
- Added `RecipeNote` import (line 20)
- Added `get_notes()` function (lines 192-207) to extract and validate notes from JSON
- Added `notes=get_notes()` parameter to Recipe constructor (line 240)

**File: `mealie/schema/recipe/recipe.py`**
- Added `validate_notes()` field validator (lines 299-314) for defense-in-depth
- Handles dict-to-RecipeNote transformation
- Works for direct API calls without scraper

**Impact:**
- ✅ Backwards compatible (no breaking changes)
- ✅ No database migration needed (field exists, just wasn't populated)
- ✅ Fixes data loss for all JSON import users

Partially addresses #6887 (specifically fixes item 6: recipe notes field)

The issue reports seven different fields being ignored during import. This PR focuses on fixing the `notes[]` field. The other fields (instruction titles, categories, tools, ingredient sections, markdown preservation) would require separate work and are outside the scope of this PR.

- I'm new to contributing to Mealie - feedback on code style/patterns welcome!
- Tested on self-hosted Docker instance (v3.10.2) with successful results
- If unit tests are required, I'd appreciate guidance on where to add them and what test cases would be most valuable

## Testing

**Test Environment:**
- Self-hosted Mealie v3.10.2 (Docker)
- Tested both JSON import UI and API endpoint

**Test Case 1: JSON Import with notes**
```json
{
  "@context": "https://schema.org",
  "@type": "Recipe",
  "name": "Test Recipe",
  "notes": [
    {"title": "Storage Tip", "text": "Keep refrigerated up to 3 days"},
    {"title": "Variation", "text": "Add chili flakes for extra heat"}
  ]
}
```

**Results:**
- **Before fix:** `notes: []` ❌
- **After fix:** `notes: [{"title": "Storage Tip", "text": "..."}, ...]` ✅

**Test Case 2: Export → Re-import workflow**
1. Exported existing recipe with notes via GET `/api/recipes/{slug}`
2. Re-imported using POST `/api/recipes/create/html-or-json`
3. Verified notes were preserved

**Results:**
- **Before fix:** Notes lost during re-import ❌
- **After fix:** Notes preserved correctly ✅

**Test Case 3: UI Import**
- Used "Import from JSON" feature in UI
- Notes displayed correctly in recipe view
- Notes editable via UI after import


michael-genson (approved):
Thank you! I removed the validator on the Pydantic model since those are only used _after_ data is clean. I also added a few tests

PR #7033, opened 2026-02-09
fix: #6263 remove reserved prefix 

### LINKED ISSUE #6263: [UX] On profile picture upload, no preview

ZacharySchaffter:

- removes the `$`-prefix from the `auth` property within all vue components.
- fixes bug related to `$`-prefixed state var causing user image upload to not properly refresh

### Additional Info
- `$`-prefixed state vars returned in `setup()` trigger a warning in dev mode, but then are silently omitted/inaccessible to the UI.  This was leading to an issue in  `/user/profile/edit.vue` component where a call to `$auth.refresh()` was throwing an 'accessing 'refresh' on undefined' error.
- Note: #6985  addresses this same issue with an iterative key to force a re-render, though I think the root issue can be fixed by renaming this prop.  This can be considered an alternative, let me know if that strategy is preferable and I'll close this!
 
#### Before
[error-before.webm](https://github.com/user-attachments/assets/8e2b8704-0312-415b-b627-c630ffbf24dd)

#### After
[error-after.webm](https://github.com/user-attachments/assets/e49d935f-a0a3-4bd4-82ae-d4a2ce442e39)

Fixes #6263

## Testing

- Tested this manually on both a local dev and local production build, using the provided dev containers.  
- The behavior I'm seeing doesn't quite match the original report (the undefined $auth call means the AppButtonUpload's loader never goes away, which isnt visible in the original screenshot).

### DIFF
--- a/frontend/components/Domain/Admin/Setup/EndPageContent.vue
+++ b/frontend/components/Domain/Admin/Setup/EndPageContent.vue
@@ -41,8 +41,8 @@
 export default defineNuxtComponent({
   setup() {
     const i18n = useI18n();
-    const $auth = useMealieAuth();
-    const groupSlug = computed(() => $auth.user.value?.groupSlug);
+    const auth = useMealieAuth();
+    const groupSlug = computed(() => auth.user.value?.groupSlug);
     const { $globals } = useNuxtApp();
 
     const sections = ref([
--- a/frontend/components/Domain/Cookbook/CookbookPage.vue
+++ b/frontend/components/Domain/Cookbook/CookbookPage.vue
@@ -73,11 +73,11 @@ import { useLoggedInState } from "~/composables/use-logged-in-state";
 import type { ReadCookBook } from "~/lib/api/types/cookbook";
 import CookbookEditor from "~/components/Domain/Cookbook/CookbookEditor.vue";
 
-const $auth = useMealieAuth();
+const auth = useMealieAuth();
 const { isOwnGroup } = useLoggedInState();
 
 const route = useRoute();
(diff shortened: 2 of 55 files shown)


michael-genson:
Thanks! Would you be able to fix other instances of `$auth --> auth` as well? The `$auth` pattern was used in Vue 2 but is no longer needed in Vue 3 (as you pointed out).

ZacharySchaffter:
@michael-genson happy to, but to clarify - is the ask to rename the existing `$auth` vars across the vue components even if they're not an issue?

Reason I ask is that I think(!) the use of `$`-prefixes is fine within setup's closure, Vue only warns/omits when they're returned by setup. Ex: I tested leaving the assignment alone and that seemed to work as expected, ala:

```
 const $auth = useMealieAuth()
 ...
 return {
   ...
   auth: $auth
 }
```

(I don't think there are any more instances of `$`-prefixed vars being returned, though there are many being assigned)

That said, renaming the existing `$auth` vars could def. help prevent a similar issue in the future, but wanted to make sure before I touch a lot of the vue components.

michael-genson:
Yeah just a global find/replace specifically for `$auth`. Double check that there aren't any conflicts but I can't imagine we're using both `$auth` and `auth` in any file.

michael-genson (approved):
Thank you!

PR #7034, opened 2026-02-09
feat: Dynamic Placeholders UI

michael-genson:

_(REQUIRED)_

Follow up to https://github.com/mealie-recipes/mealie/pull/6984 which enabled query filters using relative dates. This adds limited support to the UI.

There's a new internal `FieldType` of `relativeDate`. Currently only `last_made` uses this field type. This is what the UI looks like:

image

Where your only two operators are:
image

"days ago" is translatable, and supports zero/one/multiple (i.e. `1 day ago`):

image
image

This was added to both meal plan rules and the recipe finder. I didn't add this to cookbooks because I don't think it makes sense for cookbooks to be dynamic.

There are several intentional limitations of this UI:
- Only "negative" displacements are supported (e.g. you cannot set a value of "30 days in the future"; you can _only_ set X days ago, including "0 days ago")
- Only intervals of days are supported (so if you want to do months/years you must convert to days. You cannot do hours/minutes/seconds)
- "older than" includes the day (e.g. "older than 30 days" translates to `date <= $NOW-30d`)
- "newer than" _also_ includes the day (e.g. "newer than 30 days" translates to `date >= $NOW-30d`)
- Fields configured as relative dates _cannot_ use absolute dates, and vice versa (e.g. `created_at` can _only_ reference an _absolute_ date; `last_made` can _only_ reference a _relative_ date)

I tried to support a few more configurations but implementation of this got extremely complicated, so I came up with a simpler approach (which is still kind of complicated under the hood but the UX is _very_ simple). The complexity of the UX was not at all worth the added value of slightly more configuration when 99.99% of users aren't going to care. Users always have the option of setting custom queries via the backend if they really want to.

_(REQUIRED)_

N/A

Would love some feedback on whether or not this implementation is intuitive. I think it is, but I wrote it 😄

I also made a few layout tweaks to support larger queries. Not really relevant for the new feature, but it was helpful during development and I think it's a marginal improvement so I left it in.

## Testing

Created simple Meal Plan rules which were properly respected, including compound queries with multiple date rules. This is mostly taken care of in [the backend PR](https://github.com/mealie-recipes/mealie/pull/6984), though.

### DIFF
--- a/frontend/components/Domain/Household/GroupMealPlanRuleForm.vue
+++ b/frontend/components/Domain/Household/GroupMealPlanRuleForm.vue
@@ -76,7 +76,6 @@ const MEAL_DAY_OPTIONS = [
 ];
 
 function handleQueryFilterInput(value: string | undefined) {
-  console.warn("handleQueryFilterInput called with value:", value);
   queryFilterString.value = value || "";
 }
 
--- a/frontend/components/Domain/Household/GroupMealPlanRuleForm.vue
+++ b/frontend/components/Domain/Household/GroupMealPlanRuleForm.vue
@@ -113,7 +113,7 @@ const fieldDefs: FieldDefinition[] = [
   {
     name: "last_made",
     label: i18n.t("general.last-made"),
-    type: "date",
+    type: "relativeDate",
   },
   {
     name: "created_at",
--- a/frontend/components/Domain/QueryFilterBuilder.vue
+++ b/frontend/components/Domain/QueryFilterBuilder.vue
@@ -158,34 +158,63 @@
               :model-value="field.value"
               @update:model-value="setFieldValue(field, index, $event!)"
             />
-            <v-menu
+            <div
               v-else-if="field.type === 'date'"
-              v-model="datePickers[index]"
-              :close-on-content-click="false"
-              transition="scale-transition"
-              offset-y
-              max-width="290px"
-              min-width="auto"
+              class="d-flex gap-2 flex-wrap align-end"
+              style="width: 100%;"
             >
-              <template #activator="{ props: activatorProps }">
-                <v-text-field
-                  :model-value="field.value ? $d(new Date(field.value + 'T00:00:00')) : null"
-                  persistent-hint
-                  :prepend-icon="$globals.icons.calendar"
-                  variant="underlined"
-                  color="primary"
-                  v-bind="activatorProps"
-                  readonly
+              <v-switch
+                :model-value="field.fieldConfig.absoluteDate"
+                class="mr-4"
+                hide-details
+                inset
+                color="primary"
+                @update:model-value="toggleAbsoluteDate(field, index, $event)"
+              >
+                <template #label>
+                  <v-icon size="24">
+                    {{ $globals.icons.calendar }}
+                  </v-icon>
+                </template>
+              </v-switch>
+              <v-menu
+                v-if="field.fieldConfig.absoluteDate"
+                v-model="datePickers[index]"
+                :close-on-content-click="false"
+                transition="scale-transition"
+                offset-y
+                max-width="290px"
+                min-width="auto"
+              >
+                <template #activator="{ props: activatorProps }">
+                  <v-text-field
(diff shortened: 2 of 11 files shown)


frontend/components/Domain/Household/GroupMealPlanRuleForm.vue
```diff
@@ -114,7 +113,7 @@ const fieldDefs: FieldDefinition[] = [
   {
     name: "last_made",
     label: i18n.t("general.last-made"),
-    type: "date",
+    type: "relativeDate",
   },
```
michael-genson, line 117: Since this used to be `date` and I converted it to `relativeDate` existing rules will fail to load in the UI and be converted to the default value of `$NOW-30d`. However the rules still work internally, so it won't break anything until the user attempts to edit it. Not the best UX, but pretty minor.

frontend/components/Domain/QueryFilterBuilder.vue
```diff
@@ -129,9 +129,9 @@
             :class="config.col.class"
           >
             <v-select
-              v-if="field.fieldOptions"
+              v-if="field.fieldChoices"
               :model-value="field.values"
-              :items="field.fieldOptions"
+              :items="field.fieldChoices"
```
michael-genson, line 134: This was renamed because I originally introduced a new property `fieldConfig` which is very confusing next to `fieldOptions`. I ended out removing `fieldConfig` but I think choices makes more sense anyway so I left it.

frontend/components/Domain/QueryFilterBuilder.vue
```diff
@@ -396,16 +425,29 @@ function setField(index: number, fieldLabel: string) {
...
+    case "date":
+      fields.value[index].value = safeNewDate("");
+      break;
+    case "relativeDate":
+      fields.value[index].value = "$NOW-30d";
+      break;
+
+    default:
+      break;
+  }
```
michael-genson, line 450: New date fields will default to today, and new relative date fields will default to "30 days ago". This isn't strictly necessary but I think it makes the UX a little easier to understand.

mealie/services/query_filter/builder.py
```diff
@@ -138,7 +144,7 @@ def as_json_model(self) -> QueryFilterJSONPart:
             logical_operator=None,
             attribute_name=self.attribute_name,
             relational_operator=self.relationship,
-            value=self.value,
+            value=self.raw_value,  # we use the raw value to preserve placeholder keywords
         )
```
michael-genson, line 148: My previous PR didn't account for the produced JSON filling in placeholder values, which completely breaks the frontend parsing. This is fixed here by storing the unparsed value. I also added a test for this.

hay-kot (approved):
LGTM, I think we can always get some feedback on UX. If you _really_ wanted to, we could put a little blurb at the bottom to comment on a Github Discussion or something to provide UX feedback.

bpmartin20:
These kinds of enhancements are really extend the functionality of Mealie,  so thank you @michael-genson for undertaking it.

I do disagree with you on one bit. You said `I didn't add this to cookbooks because I don't think it makes sense for cookbooks to be dynamic.`. I think it does make sense on two points:

1. Aren't cookbooks already dynamic? Suppose I create a cookbook selecting on category = Dessert, and let's say I get 10 recipes in it as a resultHu. If I create a new Dessert recipe,  and then revisit my cookbook, don't I now find 11 recipes?
2. More important, though, is that cookbooks provide the only way we have in the UI to save reoccurring searches. My single most heavily required search is for recipes in category "Main Course" that we haven't made in the last 60 days. That's a perfect candidate for a saved search, but without having Last Made as a search option, that's not possible. A similar argument can be made for the ability to have rating as a search term, though I suspect implementing rating as a search term might be difficult. Still, having even one of those would be very helpful.

Thanks for creating Mealie. We've been on it for about 3 weeks now, and my wife just spontaneously stated "I love Mealie" because she needed to edit a recipe and she could do it right there on her phone easily.

PR #7076, opened 2026-02-15
feat: improve BaseDialog on mobile and use it globally

p0lycarpio:

_(REQUIRED)_

This pull request refactors several dialog components in the recipe management frontend to the unified `BaseDialog` component instead of directly using Vuetify's `v-dialog` and `v-card` components. This change streamlines dialog appearance and behavior, improves code maintainability, and introduces consistent support for dialog icons, titles, loading states, and actions. Additionally, some UI improvements and minor logic updates were made ;

- Dialog boxes now always display in full screen on small screens, and action buttons are always placed in the same location at the bottom. This improves screen consistency within the application and predictability on mobile devices.

- The dialog box for adding ingredients, steps, or recipes in bulk has also been slightly redesigned to display simpler action buttons.

**Before/after:**

 

 

_(REQUIRED)_

N/A


Kuchenpirat (commented):
Hey, thanks for the contribution :)
Havent had time to look at the code much so this is just based on visuals. 

Regarding the Bulk Add Dialog:
I know we are usually not super consistent about this sort of thing but most buttons are colored, so i would suggest sticking with that. 
Also i think they could be a tad smaller.

p0lycarpio:
Thanks for your feedback!
I've restored the color to the button. Here are two more tests with smaller icon and button sizes (using Vuetify's “small” sizes instead of “default”). I'm afraid the last one might be a little less accessible on desktop.

**Default size / reduced icon / reduced button**

miah120:
Imo the first and third options both look good. I'd lean towards the first.

Kuchenpirat (approved):
Sorry for taking so long on this one. 
Changes look good to me and on taking a better look i agree with  @miah120 that the first one is the best option.
Thanks :)

PR #7091, opened 2026-02-18
feat: Improve add shopping list item form

miah120:

This PR updates the add item form in the shopping list with a few changes: 
* Update ordering and positioning of entry fields to improve scannability
* Moved label icons into the input field
* In mobile, moved the add item form to be a bottom drawer and styled the food entry to match
    *  In desktop, the UI remains the same

Before:
Screen Shot 2026-02-17 at 22 50 00

Screen Shot 2026-02-17 at 22 50 33

Screen Shot 2026-02-17 at 22 51 05

After: 
Screen Shot 2026-02-17 at 22 37 19

Screen Shot 2026-02-17 at 22 38 14

Screen Shot 2026-02-17 at 22 40 30

_(REQUIRED)_

NA

NA

## Testing

Tested in Firefox and Android


miah120:
Looks like I made the quantity field a bit too small. I was focused on the add item form, but w the edit item form all you can see of the label is Q... 

I'll fix that tonight

p0lycarpio:
It looks good! The spacing is better managed. The mobile view is interesting and more intuitive.
On desktop, what would you think about replacing the "Add" button with a field similar to the one on mobile that would expand the card when clicked (same behavior as the bottom sheet)? As I write this, I realize that it could also be used for the mobile version... What do you think?

A quick note about the field displayed at the bottom on mobile devices: the two search/food icons can clutter up the interface, so I would choose just one on the left.

michael-genson (changes requested):
Overall looks great, definitely a huge improvement for mobile and a good improvement for desktop (other than the broken quantity field ;) )

This might be an issue with [the auto focus logic I just merged](https://github.com/mealie-recipes/mealie/pull/7096), but on mobile when I tap into the food field to add a new item the food menu appears before the drawer opens up, so it covers the dialog:
image

Tapping out and back in it's fine:
image

Similar to what @p0lycarpio said I don't think we need both the magnifying glass and the apple icon, maybe just use the apple icon where the magnifying glass is? I recognize that it's a search field but I think the food icon conveys that you're searching for a food well enough on its own.

The "Add item" text works when it's by itself at the bottom, but once the full dialog is open it's weird to see  "Add item" (I am already adding an item). Could we switch it back to "Food" (or whatever) once we open it up?

Also, and I don't know if this is actually a good suggestion, but could we adjust the style of the food input field once the drawer opens up? When it's by itself at the bottom I like how it looks, but once the dialog opens up it's weird to see it to be a different shape than the rest of the fields. Would it make sense to change the shape of the field back?

miah120:
Okay here's the followups enumerated: 
* ✅ Fix the quantity field sizing
* ✅ Show only a single icon: 
    * I opted to pick the apple over the search icon since it'll link the bottom bar to the food field; this agrees with Michael and it feels a bit more fun/whimsical :3
* ✅ Add the "Add item" bar to the desktop view (but in the same location as the add button)
* ❌ Fix the menu anchoring issue
    * This problem occurs because the anchor position is determined as soon as the field is clicked/tapped, then the bottom sheet animates causing the field to go behind the menu.  
    * I'm not sure how to go about fixing this... however
    * Testing on a mobile device (Android) this doesn't happen. It's possible that when a virtual keyboard is in use (which would always be the case for this feature), that this is non-existent. 
    * I don't have an iOS device to test with, and it's possible that this could happen depending on what OS/keyboard/browser you're using, but if we're only seeing this on desktop browsers I'm hesitant to put a bunch of time into trying to fix something that may not be a real issue. 
    * Let me know if you think these are appropriately mitigating factors on this issue or if you want me to find a way around it.  
* ✅ Adjust text and style of the field depending on whether the add form is open
    * I was skeptical of this one but after testing it I think it looks nice and adds cohesion 

Some additional tweaks I made while addressing these: 
* Edit item forms now have a colorful left border to more clearly mark the edit forms in a way that looks nice 
* Significant tweaks to the edit item form (sorry!) to remove unnecessary wrapping divs to keep things inline between the two forms
* padding tweaks

Before: 
Screen Shot 2026-02-18 at 20 57 37

After: 
Screen Shot 2026-02-18 at 20 42 28

(after writing this I feel like this has AI tells in it, I swear I just use a lot of emojis)

Kuchenpirat:
Hi, sorry for taking literaly over a month to get to this. Especially because i feel this is a really good idea. 

Could you update this to
1. resolve the merge conflicts
2. use a script setup component for ShoppingListAddItemForm (you can find a bunch of examples on how to convert in this PR #7299)

I would want to experiment a bit with the design of the add container and i am still not sold on the order of the input (food first, then quantity and unit) but those are small tweaks

miah120:
Updated the branch and switched to a setup script. 

My reasoning for the order mostly relies on the idea that users should only need to scan as far into the form as necessary. In my experience at least, I find that in cases where units/quantity end up being filled out, it's coming from recipes and not being entered manually; I typically use the add form for one off items where I'm just filling out the food.

frontend/components/Domain/ShoppingList/ShoppingListAddItemForm.vue
```diff
@@ -0,0 +1,211 @@
...
+  <v-navigation-drawer
+    permanent
+    rounded="xl"
+    location="bottom"
+    class="pa-4 pt-2 my-0"
+    style="margin: 0 !important; border-radius: 1.5rem 1.5rem 0 0 !important;"
+    width="300"
+    rail-width="85"
+    :rail="rail"
+    elevation="4"
```
Kuchenpirat:
```suggestion
    rounded="t-xl"
    permanent
    location="bottom"
    class="pa-4 pt-2 mb-0"
    width="300"
    rail-width="85"
    :rail="rail"
    elevation="4"
```

We can simplify this a bit

frontend/components/Domain/ShoppingList/ShoppingListItemEditor.vue
```diff
@@ -1,105 +1,96 @@
...
-            {
-              icon: $globals.icons.save,
-              text: $t('general.save'),
-              event: 'save',
-            },
-          ]"
-          @save="$emit('save')"
-          @cancel="$emit('cancel')"
-          @delete="$emit('delete')"
+  <v-card variant="elevated" class="pa-2" border="s-lg" style="border-color: rgb(var(--v-theme-primary)) !important;">
```
Kuchenpirat: You can use the border prop to set color and opacity

```suggestion
  <v-card variant="elevated" class="pa-2" border="primary s-lg opacity-100">
```

frontend/components/Domain/ShoppingList/ShoppingListAddItemForm.vue
Kuchenpirat, line 1: This has a lot of redundant code to from the ShoppingListItemEditor component. 

Why not clean this up a bit and use it here?

frontend/components/Domain/ShoppingList/ShoppingListItemEditor.vue
miah120: 😯 Didn't know you could do it that way! Fixed!

frontend/components/Domain/ShoppingList/ShoppingListAddItemForm.vue
miah120: fixed!

frontend/components/Domain/ShoppingList/ShoppingListAddItemForm.vue
miah120: Created a new composable and internal component shared between the two forms. Let me know if that fixes it for u!

michael-genson (approved):
Looks great, thanks! I fixed/cleaned-up a few things but no functionality changes.

RightiousNewb:
a little late here, i know, but my parents are having trouble on iphones with the new ui (I have an android). When they try to create a new food the keyboard covers up the create button, and scrolling doesn't seem to fix it. seems like it loses focus and the button disappears when they scroll. any suggestions? - They're using the PWA.

PR #7107, opened 2026-02-20
chore: refactor data management pages

Kuchenpirat:

The Data Management Pages have quite a lot of very redundant code. 
This is an attempt in solving that by introducing the GroupDataPage Component. 

This encompasses all pages besides the recipe page which uses a custom CRUD Table

None 

First PR after being back, so i might be out of it a bit. 

There are also updates for these components in here: 
AutoForm:
- Updated some out of date code and removed some stuff we were not using at all. 
- made it expose an _isValid_ prop that can be used to track validity in the parent components. 
- Made it validate on input instead of lazy. if there was a reason for it we can default to lazy but i think imediate feedback on inputs is quite valuable.
- updated to use our existing color picker component

Crud Table
- emits a bulk action event with the event type as string instead of having a custom non typed event 

Validators
- adds a new startsWithHttp validator that checks if a string is starting with http or https. Before a user could hit create on a recipe action that had an inproper URL and was rejected by the backend. 

## Testing

Manual testing of the new flow and the pages that use the updated AutoForm component


Kuchenpirat:
This should be good to go now

frontend/lang/messages/en-US.json
```diff
@@ -1445,6 +1445,7 @@
     "invalid-url": "Must Be A Valid URL",
     "no-whitespace": "No Whitespace Allowed",
     "min-length": "Must Be At Least {min} Characters",
-    "max-length": "Must Be At Most {max} Characters"
+    "max-length": "Must Be At Most {max} Characters",
```
michael-genson: This is probably silly, but would it be worth making this pluralize with a max of "1"?
`"Must Be At Most {max} Character|Must Be At Most {max} Characters",`

frontend/lib/validators/inputs.ts
```diff
@@ -29,6 +29,11 @@ export function urlOptional(v: string | undefined | null) {
   return v ? url(v) : true;
 }
 
+export function startsWithHttp(v: string | undefined | null) {
+  const i18n = useGlobalI18n();
+  return (!!v && (v.startsWith("http://") || v.startsWith("https://"))) || i18n.t("validators.http");
+}
+
```
michael-genson: Is there a reason not to use the existing URL validator? I know we added the extra enforcement on http/s for security on recipe actions on the backend, but we could probably modify the url validator instead of adding a new one.

michael-genson (dismissed):
Looks great! Soooo much cleaner.
Two potential changes but neither are blockers so I'm approving as-is.

frontend/lib/validators/inputs.ts
Kuchenpirat: yeah, this would probably make more sense, i was just too lazy to look up where we were using the thing. 

Will update

michael-genson: I'm pretty sure it's only for importing a recipe via URL, so requiring http/s seems fine

Kuchenpirat: Updated. Only thing is that now we loose the separate hints for valid url and the https

frontend/pages/group/data/recipe-actions.vue
```diff
@@ -1,286 +1,121 @@
...
+    label: i18n.t("general.title"),
+    varName: "title",
+    type: fieldTypes.TEXT,
+    rules: [validators.required],
+  },
+  {
+    label: i18n.t("general.url"),
+    varName: "url",
+    type: fieldTypes.TEXT,
+    rules: [validators.required, validators.url, validators.startsWithHttp],
```
michael-genson: I think we need to remove `validators.startsWithHttp` now

Kuchenpirat: already on it, you are too fast ;)

michael-genson: 😅 I sent that before CI failed, my bad

michael-genson (approved):
Sweeeet

PR #7118, opened 2026-02-22
feat: Shopping list / Swipe to check off

miah120:

This PR allows shopping list item to be swiped off when using gesture controls. 
We will also now display a toast when items are checked off, allowing users to quickly undo an action if it was an accident. 

[Screencast_20260222_071622.webm](https://github.com/user-attachments/assets/cf51eed7-3d8f-4cdc-a399-9c846d22f17d)

This feature has seen a lot of tweaks recently:
* [This issue](https://github.com/mealie-recipes/mealie/issues/5423) originally reported that the touch target was too large resulting in accidentally checking items off the list
* [This PR](https://github.com/mealie-recipes/mealie/pull/6569) then undid the fix for that bug in an effort to make things easier to check off
* [This PR](https://github.com/mealie-recipes/mealie/pull/6794) then re-implemented the fix for the original issue

After a discussion in discord, people seemed to be leaning towards swipe controls with a toast to undo being the best option. This allows for mobile users to check items off without needing to reposition their thumbs while still requiring a more precise gesture than if the whole shopping list item was a touch target; the toast allows for immediate error correction if the item was checked off by mistake.  

NA

NA

## Testing

Tested in Firefox on Android and Desktop Firefox.


Kuchenpirat (commented):
I really like the UX of this. 

The rate of people to discover this will probably be pretty low but besides maybe adding something to the docs or highlighting it in the release notes i am not sure it would make sense to add something to the UI to point this out. 

Also is there a reason why we wouldn't use our existing toast/snackbar system? The visuals of the Toast are quite different from what we usually use.

miah120:
Oops, the reason was that I didn't realize it was available. Just fixed that.

I had to expand on it's capabilities a lil bit. I opted not to include options on the loader toasts; I can't imagine a scenario where those would be helpful since closing a loader toast isn't exactly going to make the content load any faster, and timeouts should always be -1.

michael-genson (approved):
Love this, thanks so much! I fixed the merge conflict and made a couple tweaks. I also added support for RTL languages (so those users can swipe from right to left instead of left to right).

PR #7121, opened 2026-02-22
feat: Unit standardization / conversion

michael-genson:

_(REQUIRED)_

Enables standardization of units, allowing for things such as automatic unit conversion. This PR specifically implements:
- Two new unit fields: standard quantity & standard unit
- A bunch of unit conversion tools levering these fields and the [pint](https://github.com/hgrecco/pint) library (which is already used by our NLP library)
- Automatic standardization of new/existing units (where possible)
- Automatic merging of shopping list items for compatible units

To the user the only new useful feature of this PR is automatically merging shopping list units:
2026-02-21_16h30_44

Going forward, unit standardization will enable several features, such as:
- Automatic conversion of recipe ingredients on the fly (we can add an "Imperial" vs "Metric" toggle)
- Unit parsing post-processing (e.g. define "pint" but always convert into "cup" or "liter")
- Automatic nutrition calculation (not fully enabled by this PR, but is a required step)

How is the data stored?
---

At the data layer standard units are just strings. The understanding is that anything stored as a "standard unit" is something understood by pint, but we don't actually validate this (the unit conversion processing has error handling if we get bad data, so it doesn't matter). On the frontend we have an arbitrary hardcoded list:
- fluid ounce
- cup
- ounce
- pound
- milliliter
- liter
- gram
- kilogram

Pint supports hundreds of units so this is obviously not comprehensive, but I think this covers everything a Mealie user would need (includes a few different volume and mass units, in both imperial and metric systems). We could always add/remove options, too, since this list is not stored in the db.

How is this data used?
---

The new `UnitConverter` class handles conversions between units. It's a wrapper around pint and enables parsing units as strings or `pint.Unit` objects, as well as associated quantities. When we receive a Mealie unit we tell pint what this unit means:

```python
uc.ureg.define(f"mealie-unit-1 = {unit_1.standard_quantity} * {unit_1.standard_unit}")
uc.ureg.define(f"mealie-unit-2 = {unit_2.standard_quantity} * {unit_2.standard_unit}")
```

...and now pint understands what `mealie-unit-1` and `mealie-unit-2` are. This is done at runtime so we don't have to manage any external definitions file.

How is this data managed?
---

Since units, unlike foods, are pretty universally standardized in context of recipes, I've implemented several automatic standardization methods:
1. When a unit is created (either manually or via seeding) and standardization data is not provided, if it matches one of several hardcoded units (teaspoon, cup, liter, pound, kilogram, etc.) we automatically add the standard definitions to them. This hardcoded mapping is _locale aware_ and leverages our seed data translations (e.g. "taza", which is "cup" in Spanish, will be standardized if the user has their language set to Spanish or seeds Spanish data).
2. Using similar logic, all existing units will be standardized at migration-time. Unlike the local-aware logic, the migration has no idea what language your units are in, so we check all supported languages (which is kind of slow, but it's a migration so it doesn't matter).

So existing users don't have to do anything to standardize their data for the most part, unless they have unexpected units they want to standardize. Similarly, new users using seed data also don't have to do this manually.

For users with unexpected units, or users who want to tweak things for some reason, these fields are added to the data management page:
image

Where the second field is a dropdown of our arbitrary known units:
image

_(REQUIRED)_

Several discussions either ask for or otherwise depend on this:
- https://github.com/mealie-recipes/mealie/discussions/1659
- https://github.com/mealie-recipes/mealie/discussions/2748
- https://github.com/mealie-recipes/mealie/discussions/2766
- https://github.com/mealie-recipes/mealie/discussions/3336
- https://github.com/mealie-recipes/mealie/discussions/3881
- https://github.com/mealie-recipes/mealie/discussions/5095
- https://github.com/mealie-recipes/mealie/discussions/5978

I added some gross-looking special handling for `ounce`. Ounce is a weird one because often times (at least here in the US) recipes use "ounce" for both mass and volume (e.g. "1 cup = 8 ounces"). This is _technically_ incorrect because ounces are a mass unit, whereas we should be using "fluid ounces" (e.g. "1 cup = 8 fluid ounces"). The special handling looks at the two units it wants to combine, and if one is literally "ounce" and the other is a volume unit, assume we actually meant "fluid ounce".

Users can, of course, use "ounce" and "fluid ounce" correctly by defining both and diligently updating their recipes to use the correct unit... but I'm certainly not going to do that (and I'm sure most users agree). AFAIK there are no other cases where we want this kind of logic other than this exact use case.

## Testing

Comprehensive backend tests. I also manually tested a few edgecases around the data management page.


frontend/lib/api/types/non-generated.ts
```diff
@@ -58,3 +58,13 @@ export interface QueryFilterJSONPart {
...
+
+export type StandardizedUnitType
+  = | "fluid_ounce"
+    | "cup"
+    | "ounce"
+    | "pound"
+    | "milliliter"
+    | "liter"
+    | "gram"
+    | "kilogram";
```
michael-genson, line 70: Our arbitrary list of unit types. Pint understands all of these.
There's no difference in defining a unit as `1 * liter` vs `1000 * milliliter` other than user preference.

mealie/alembic/versions/2026-02-21-17.59.01_a39c7f1826e3_add_unit_standardization_fields.py
```diff
@@ -0,0 +1,106 @@
...
+        standardized_data = repo._add_standardized_unit(unit_data)
+        std_q = standardized_data.get("standard_quantity")
+        std_u = standardized_data.get("standard_unit")
+        if std_q and std_u:
+            logger.info(f"Found unit '{unit.name}', which is standardized as '{std_q} * {std_u}'")
+            unit.standard_quantity = std_q
+            unit.standard_unit = std_u
+
+    session.commit()
+    session.close()
```
michael-genson, line 82: Check all units against all languages and automatically standardize known ones. This assumes there isn't an identical string between languages with different definitions. I think this is a fair assumption (and users can manually fix if needed).

mealie/repos/repository_units.py
```diff
@@ -1,17 +1,119 @@
...
+            self._standardized_unit_map = {}
+            locale_file = IngredientUnitsSeeder.get_file(locale=locale)
+            for unit_key, unit in IngredientUnitsSeeder.load_file(locale_file).items():
+                for prop in ["name", "plural_name", "abbreviation"]:
+                    val = unit.get(prop)
+                    if val and isinstance(val, str):
+                        self._standardized_unit_map[val.strip().lower()] = unit_key
+
+        return self._standardized_unit_map
+
```
michael-genson, line 38: For a given locale, check values against "name", "plural_name" and "abbreviation" from our seed data (our seed data doesn't define a plural abbreviation for some reason).

mealie/repos/repository_units.py
```diff
@@ -1,17 +1,119 @@
...
+                    data["standard_quantity"] = 1
+                    data["standard_unit"] = StandardizedUnitType.GRAM
+                case "kilogram":
+                    data["standard_quantity"] = 1
+                    data["standard_unit"] = StandardizedUnitType.KILOGRAM
+                case "milligram":
+                    data["standard_quantity"] = 1 / 1000
+                    data["standard_unit"] = StandardizedUnitType.GRAM
+                case _:
+                    continue
```
michael-genson, line 105: Big hardcoded standardization logic. These are ultimately the units we automatically standardize.

mealie/schema/recipe/recipe_ingredient.py
```diff
@@ -130,7 +150,21 @@ class CreateIngredientUnit(UnitFoodBase):
...
+    @model_validator(mode="after")
+    def validate_standardization_fields(self):
+        # If one is set, the other must be set.
+        # If quantity is <= 0, it's considered not set.
+        if not self.standard_unit:
+            self.standard_quantity = self.standard_unit = None
+        elif not ((self.standard_quantity or 0) > 0):
+            self.standard_quantity = self.standard_unit = None
+
+        return self
```
michael-genson, line 167: A standard quantity without a unit is useless, and vice versa, so we drop partial definitions.

mealie/services/household_services/shopping_lists.py
```diff
@@ -69,7 +81,20 @@ def merge_items(
...
+                from_item.quantity or 0, from_item_unit, to_item.quantity or 0, to_item_unit
+            )
+            to_item.quantity = merged_qty
+            to_item.unit_id = merged_unit.id
+            to_item.unit = merged_unit
+
+        else:
+            # No conversion needed, just sum the quantities
+            to_item.quantity += from_item.quantity
+
```
michael-genson, line 97: The only user-facing part of this PR, aside from data management, which merges shopping list items

mealie/services/parser_services/parser_utils/unit_utils.py
```diff
@@ -0,0 +1,146 @@
...
+
+    # Convert to the bigger unit if quantity >= 1, else the smaller unit
+    merged_q, merged_u = uc.convert(merged_q, merged_u, max(pint_unit_1, pint_unit_2))
+    if abs(merged_q) < 1:
+        merged_q, merged_u = uc.convert(merged_q, merged_u, min(pint_unit_1, pint_unit_2))
+
+    if str(merged_u) == PINT_UNIT_1_TXT:
+        return merged_q, unit_1
+    else:
+        return merged_q, unit_2
```
michael-genson, line 146: The meat of the PR, which handles all the actual unit conversions

mealie/services/parser_services/parser_utils/unit_utils.py
```diff
@@ -0,0 +1,146 @@
...
+        FL_OUNCE = self.ureg("fluid_ounce")
+        VOLUME = "[length] ** 3"
+
+        if unit_1 == OUNCE and unit_2.dimensionality == VOLUME:
+            return FL_OUNCE, unit_2
+        if unit_2 == OUNCE and unit_1.dimensionality == VOLUME:
+            return unit_1, FL_OUNCE
+
+        return unit_1, unit_2
+
```
michael-genson, line 41: The magic "ounce -> fluid ounce" converter (more info in the PR body)

michael-genson:
Switched to the new auto form layout, doesn't look as fancy but we can pretty it up in another PR (will need to overhaul the auto form to allow side-by-side fields)
image

Kuchenpirat:
> will need to overhaul the auto form to allow side-by-side fields

YES, we absolutely do. Also its wasting huge amounts of vertical space because it adds space for potential hints, labels etc that are usually hidden

michael-genson:
Yeah something as simple as putting it in a grid and specifying the cols (defaulting to 12) would probably work quite well. Definitely something I can take up ~~in the future~~ now https://github.com/mealie-recipes/mealie/pull/7150

michael-genson:
The updated form now that https://github.com/mealie-recipes/mealie/pull/7150 is merged:
image

Kuchenpirat (approved):
Finally had time to look at this and played around with it a bunch (in german as well) without finding a way to break it. So LGTM

PR #7138, opened 2026-02-24
fix: Added mealie.log rotations to excluded files for backup

Choromanski:

Stops the log rotations of mealie.log from being included in the backup.

I see that mealie.log is only configured for a max of 3 rotations.
If you'd rather me do some regex to determine if the file is "mealie.log.\d+" let me know as that might be a better approach incase the rotation count changes.

## Testing

Did a build and tested the backup and the log rotations are no longer included.


michael-genson (changes requested):
Yeah could we switch this to a regex? I fear we'll forget about this if we ever change logging.
Otherwise LGTM

Choromanski:
@michael-genson I've updated the PR to include a check using regular expressions

michael-genson (approved):
Thanks!

PR #7161, opened 2026-03-01
Add dietary restriction labels to default seed data

stuckvgn:
## Summary

Adds five dietary restriction labels to the en-US seed data:

- Vegan
- Vegetarian
- Plant-Based
- Dairy-Free
- Gluten-Free

## Motivation

These are among the most frequently-used shopping list label categories for users who track dietary restrictions. They are absent from the current default label set, which means every user who wants them must create them manually before they can begin labeling items.

Adding them to the seed data reduces friction for new installations and aligns with the existing set of general grocery categories (Produce, Dairy Products, Health Foods, etc.).

## Changes

- `mealie/repos/seed/resources/labels/locales/en-US.json` — five new label entries appended after "Other"

This is a purely additive change to a JSON seed file. No code paths, database schema, or existing labels are affected.

Translations for other locales can follow through the existing Crowdin workflow once this is merged.

### DIFF
--- a/mealie/repos/seed/resources/labels/locales/en-US.json
+++ b/mealie/repos/seed/resources/labels/locales/en-US.json
@@ -61,5 +61,20 @@
     },
     {
         "name": "Other"
+    },
+    {
+        "name": "Vegan"
+    },
+    {
+        "name": "Vegetarian"
+    },
+    {
+        "name": "Plant-Based"
+    },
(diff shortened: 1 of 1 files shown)


stuckvgn:
Closing this PR — on reflection, this change may not be the right fit for this project. Thanks for your time!

PR #7163, opened 2026-03-01
dev: Improve support for front end unit tests

miah120:

Adds a unit test coverage report and other utils to make front end unit tests easier to report on, develop, and standardize.

Adds the following: 
* Opens the vitest ui when running frontend tests
* Adds a "mtest" (short for mealie test) snippet to quickly populate the boilerplate of a new unit test file
* Stubs the i18n composable globally 
* Provides a wrapper util to make testing composables easier
* Adds some test coverage to the shopping list composables

NA

This is a followup based on a conversation in the discord server

* The coverage report now opens when running task ui:test. Let me know if it should be a separate command. 
* I put each test under a __tests__ directory. This breaks from the pattern of including the unit tests directly next to the file under test, but I think it makes for a cleaner pattern as we work towards more coverage. If you want I can move the other test files into similar directories. Let me know if another pattern here would be preferred. 

## Testing

use `task ui:test` to run the coverage report. 
See composables/shopping-list-page/sub-composables for the new coverage. 

## AI / LLM Assistance

Used Proton's Lumo to help configure vitest, all actual test content was written by me.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -44,6 +44,7 @@
     8000, // used by mkdocs
     9000,
     9091, // used by docker production
+    51204, // used for test coverage report
     24678 // used by nuxt when hot-reloading using polling
   ],
   // Use 'onCreateCommand' to run commands at the end of container creation.
--- a/.vscode/test-block.code-snippets
+++ b/.vscode/test-block.code-snippets
@@ -1,0 +1,30 @@
+{
+  "Test Block": {
+    "prefix": "mtest",
+    "body": [
+      "import { mount } from \"@vue/test-utils\";",
+      "import { describe, expect, test, vi } from \"vitest\";",
+      "import { makeWrapper } from \"~/tests/utils\";",
+      "",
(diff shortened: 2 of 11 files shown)


Kuchenpirat (commented):
Don't have the time to look at the code closely right now but as a quick FYI, we already have e2e tests setup and integrated into the github actions 

You can find them at `/tests/e2e`.

michael-genson:
@Kuchenpirat thoughts on moving frontend unit tests into their own dedicated directory a la this PR? I think I'm for it since this is how we do the backend tests, wanted to get your input though.

@miah120 if we do decide to restructure the unit tests let's move all the existing ones here too. I don't want to straddle two different test structures.

Changes broadly look good, need to take a closer look at the implementation but we should probably figure out the test structure first :)

Kuchenpirat:
I am happy with both solutions. 
Only thing i would like to avoid is to have one set of frontend tests in `/tests/e2e` and another in `/frontend/tests`. 

Also, we should have them integrated into the github workflows

hay-kot (approved):

PR #7167, opened 2026-03-02
chore: Improve PT-BR language

thiagodsti:
Improve the language for Brazilian Portuguese, created with Claude (AI), but I tested locally, and it's much better than just English, and of course it will never be as good as a person translating manually


Kuchenpirat:
We only accept translations through our crowdin which you can contribute to over here: https://crowdin.com/project/mealie

I go ahead an close this PR.

PR #7201, opened 2026-03-07
feat: Add enviroment variable to set default locale

fhp:

This PR adds the ability to set a default language. As all my recipes are in Dutch, and I wanted to enable public access, I wanted to have Mealie in Dutch as well by default.

I didn't find an issue, but there was this discussion:
https://github.com/mealie-recipes/mealie/discussions/3898

I am no Nuxt developer, and use a bit of AI to help me here. I did a manual review, and it made sense to me, but please check if this is the right way.

## Testing

I tested by locally building and starting the docker. It opened in my preferred language in private browsing.

## AI / LLM Assistance

I used Claude Code for the coding, I did the reviewing and testing.

### DIFF
--- a/docs/docs/documentation/getting-started/installation/backend-config.md
+++ b/docs/docs/documentation/getting-started/installation/backend-config.md
@@ -17,7 +17,7 @@
 | TZ                            |          UTC          | Must be set to get correct date/time on the server                                                                                                      |
 | ALLOW_SIGNUP<super>\*</super> |         false         | Allow user sign-up without token                                                                                                                        |
 | ALLOW_PASSWORD_LOGIN          |         true          | Whether or not to display the username+password input fields. Keep set to true unless you use OIDC authentication                                       |
-| DEFAULT_LOCALE                |         en-US         | The default locale/language for the frontend when no user preference is set (e.g. `en-US`, `de-DE`, `fr-FR`, `nl-NL`)                                   |
+| DEFAULT_LOCALE                |         auto          | The default locale/language for the frontend. Set to `auto` to detect from browser language, or a specific locale (e.g. `en-US`, `de-DE`, `fr-FR`, `nl-NL`) |
 | LOG_CONFIG_OVERRIDE           |                       | Override the config for logging with a custom path                                                                                                      |
 | LOG_LEVEL                     |         info          | Logging level (e.g. critical, error, warning, info, debug)                                                                                              |
 | DAILY_SCHEDULE_TIME           |         23:45         | The time of day to run daily server tasks, in HH:MM format. Use the server's local time, *not* UTC                                                      |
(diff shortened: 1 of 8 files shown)


Kuchenpirat (commented):
Hi, thanks for the contribution.

While I agree with the idea, I’m strongly opposed to removing browser language detection. In most cases it correctly identifies the user’s preferred language and provides more flexibility. Forcing a single default language on the entire server could be problematic, especially if multiple groups with different language preferences are using it.

alanmilinovic:
> Hi, thanks for the contribution.
> 
> While I agree with the idea, I’m strongly opposed to removing browser language detection. In most cases it correctly identifies the user’s preferred language and provides more flexibility. Forcing a single default language on the entire server could be problematic, especially if multiple groups with different language preferences are using it.

So it will not be merged?

Kuchenpirat:
Not with the current implementation. 
I could think of two ways to clear that hurdle

1. Just set the fallback locale to the language specified
2. Instead of having a default value, only force the language when the env has been explicitly set which is i think in line with what you wanted to achieve

fhp:
I just pushed a commit which should do that, but didn't properly test it yet. I will do that later today.

alanmilinovic:
@fhp did you managed to test it. @Kuchenpirat can you please review it? I would really like to have this feature.

fhp:
Yes, I just tested and it works as expected.
When a `DEFAULT_LOCALE` is set, that locale is used unless the user makes a manual override. If it is not set, it keeps the old behavior: the browser language is used.

alanmilinovic:
@Kuchenpirat can you please merge it?

Kuchenpirat (changes requested):
Hey, sorry for the delay on this i am pretty busy currently. 

I still would like to see a few things updated. 

We can still rely on the detectBrowserLanguage if we do something like this: `detectBrowserLanguage: defaultLocale !== auto? false : { useCookie: true, alwaysRedirect: true, fallbackLocale: "en-US", cookieKey: "mealie_locale" },`

Also i think it would make sense to move all the cookie logic into the useLocale composable instead of the app-info client file

alanmilinovic:
> Hey, sorry for the delay on this i am pretty busy currently.
> 
> I still would like to see a few things updated.
> 
> We can still rely on the detectBrowserLanguage if we do something like this: `detectBrowserLanguage: defaultLocale !== auto? false : { useCookie: true, alwaysRedirect: true, fallbackLocale: "en-US", cookieKey: "mealie_locale" },`
> 
> Also i think it would make sense to move all the cookie logic into the useLocale composable instead of the app-info client file

@fhp maybe you can check it?

alanmilinovic:
@fhp do you have time to correct it? I would really love to have this feature and it is almost done. Much appriciated, thanks.

alanmilinovic:
Hey I did a new PR, @Kuchenpirat can you please check?

michael-genson:
Closing in favor of https://github.com/mealie-recipes/mealie/pull/7387

PR #7203, opened 2026-03-07
fix: migrate RecipeRating component to script setup and defineModel

### LINKED ISSUE #6874: [BUG] - Fix to recipe ratings was not completed in correct way
This is an issue raised to properly fix the recipe rating issue in line with the rest of the Mealie architecture.

See comments within pull request below for information:

https://github.com/mealie-recipes/mealie/pull/5985

ProperTeaMogul:
This PR completes the migration of the `RecipeRating` component to the `script setup` syntax and `defineModel`, which was missed in a previous PR that updated all other recipe components (and my initial patch to fix the ratings).

Fixes https://github.com/mealie-recipes/mealie/issues/6874

I've checked for any other references to '<RecipeRating' in the code base but reviewers may also want to search for any other components that reference this in case they need updating (or if they know if might be referred to as something else). If Kuchenpirat is back from sabbatical then they might be the best placed to review this request.

The logic that appears to be working at the moment is this: IF user has rated the recipe THEN show that rating ELSEIF household rating exists then show that ELSE show null/blank. Hopefully that is the intention from the developers?

## Testing
Tested locally by building and running a Docker container from this branch against a copy of my actual Mealie SQLite database. Verified that:
- The star rating display renders correctly for both user and group ratings
- Selecting and clearing a star rating updates
- The component behaves consistently on both the recipe card and recipe page views

## AI / LLM Assistance
This was created with assistance from Claude (Anthropic). I don't have any Vue experience and used Claude to understand what changes were required, why they were needed, and to assist with code writing - this was very much writing it side by side together and not delegated to an LLM entirely (otherwise what is the point, we won't learn!).


Kuchenpirat (approved):
Looks Good. 
Thanks for the contribution :)

PR #7215, opened 2026-03-09
fix: shopping list UI improvements

p0lycarpio:

_(REQUIRED)_

This PR improves the lists overall. It introduces several UI improvements and code cleanups for shopping list and recipe components, focusing on icon updates, responsive layouts, and removing unused code.

**Icon and UI Improvements:**

* Updated shopping list and recipe-related icons to use multi-checkbox variants for "check all" and "uncheck all" actions across multiple components, improving clarity and consistency.
* Changed recipe reference icon to `silverwareForkKnife` for better representation in shopping list items and linked recipes section.
* Added display of image or icon on recipes linked to items in the list.

**Responsive Layout and Recipe List Enhancements:**

* Recipes related to an ingredient are now displayed side by side on larger screens.
* Adjusted shopping list item container to ensure that the line being edited is not hidden.
* Updated shopping list page header layout for better alignment and responsiveness, including displaying the list name as a header only on larger screens. Remove the illustration in the list title to save space.

**Cleanup/removal:**

* Removed the unused `getTextColor` function and its references from shopping list label components and shopping list page because Vuetify automatically handles this.

**Before/after:**

 

_(REQUIRED)_

N/A

## AI / LLM Assistance

_(REQUIRED)_

Find the appropriate flexbox classes for displaying related recipe tiles and generating the pull request description.


Kuchenpirat (changes requested):
Hi, thanks for the PR. 
Looks pretty good so far 👍 

Only thing i would like to see changed is showing the ingredient while editing. 
I don't think it can agree with that design decision. I feel its more natural to show each element only once and toggle between an editing and viewing state. 

Any reason in particular why you went for this change?

p0lycarpio:
Thanks for your feedback!

Yes, I intentionally left the ingredient name visible while editing because I find it helpful to see both the old reference value and the item being edited. This can be useful when modifying multiple items in the list at once and makes it easier to keep track of everything.

However, upon retesting it, I noticed that the original value is sometimes temporarily overwritten during editing. I can fix this if you tell me to keep the line; otherwise, I can revert to the original behavior.

p0lycarpio:
Hi @Kuchenpirat,

I've resolved the various merge conflicts and restored the original behavior regarding the display of the ingredient being edited. Can you take a look ?

michael-genson (approved):
LGTM, thanks!

PR #7232, opened 2026-03-12
chore: run Ruff with --fix in Pre-commit to auto-fix fixable errors

deanmalan:

Run [Ruff](https://github.com/astral-sh/ruff-pre-commit) with `--fix` in Pre-commit to auto-fix fixable errors. It makes it easier and quicker to fix Ruff linting errors.

For example, the screenshot below shows how Ruff found 1 error and fixed it:

Screenshot 2026-03-12 at 12 10 34

While before, it would only show an error was detected:

Screenshot 2026-03-12 at 12 12 43

Also changed the [legacy](https://github.com/astral-sh/ruff-pre-commit/blob/main/.pre-commit-hooks.yaml#L24) `ruff` alias for the linter to the new `ruff-check` one. It works exactly the same.

None.

## Testing

Adding incorrectly linted files and running `git commit` to trigger pre-commit.

## AI / LLM Assistance

None

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -14,5 +14,8 @@ repos:
     # Ruff version.
     rev: v0.15.5
     hooks:
-      - id: ruff
+        # Linter
+      - id: ruff-check
+        args: [ --fix ]
+        # Formatter
       - id: ruff-format


michael-genson (approved):
Thanks!

PR #7238, opened 2026-03-13
feat: Optional Category filtering when adding a recipe to a mealplan

torbenvanassche:

Currently, any recipe can be registered as any type of food/drink/snack item. This PR aims to allow filtering of this information based on matching category with the entry type. By doing this, it would allow easier searching of items in the database to schedule them.

Code changes are kept to a minimum, and are as follows:
- Added a toggle on the UI, so the user can choose to remove the filter entirely.
- Added a translation string in line with the other elements on the page.
- Added a (private) function to the element to filter and return the recipes.
iWEQZgF4RG

- This is my first ever open-source contribution and pull request, I will probably make mistakes! 
Please let me know if you have tips, improvement or advice.

- I was unsure where to place the toggle for this setting, so I decided to add it in the create dialog. 
This should minimize users being required to move to the settings and back each time.
Unless this should be a global setting and we don't want to "pollute" the creation dialog.

- I am unsure how to retain the value when the window is closed and re-opened, which I think for this particular feature would be desired. I see value in defaulting the toggle to true, but I am perhaps biased to solving this problem. Keeping the behavior the way it is and allowing the user to set it globally seems like the best course of action.

## Manual test

Tested on Zen, Arc, Chrome and Firefox manually to verify the current case and my desired outcome as well as verify the integrity of the code before my change.


michael-genson:
Does this PR assume recipes have categories for each of the different meal types? This isn't something we really enforce in the app, so it seems like it would be confusing for users.

torbenvanassche:
> Does this PR assume recipes have categories for each of the different meal types? This isn't something we really enforce in the app, so it seems like it would be confusing for users.

The PR assumes categories may exist for meal types, but the app doesn’t enforce this, so the toggle could appear even when users don’t use categories and would have no effect, which might be confusing. That said, most users in that case likely wouldn’t use the feature.

Edit: I updated the logic to keep current behavior: if the filter is enabled but no recipes have categories, it falls back to showing all entries. Recipes without categories are always included.

To address your concern: the feature doesn’t add a dependency on categories, it will simply return the entire list if no categories are defined. 
If needed, we can hide the toggle unless category data actually exists. Alternatively, I could look to create an element outside categories, tags and required tools that just allows a mealtype selection. Something like this on the recipe page: 
image

and add preset values based on the dropdown provided here to be an exact match: 
image

I want to note that this would allow adding a recipe to multiple mealtypes ofcourse.

torbenvanassche:
Made some changes to allow category filters rather than the exact solution I need to allow more broad application. 
Having a category filter set will show all recipes containing the category. 
if no category is specified, shows every recipe. (current behaviour is default)
image

torbenvanassche:
@michael-genson can you give this another look with the current changes. this seems like a better way to handle the usecase!

michael-genson:
I like the change you made, but while reviewing this PR it exposes how poor the existing editor is so I created a PR to completely revamp it: https://github.com/mealie-recipes/mealie/pull/8152. This PR builds in the changes you added in here using the existing filter components.

Going to go ahead and close this one, appreciate your work on this!

Issue #7249, opened 2026-03-15
[SCRAPER] - eatingwell.com failing to import from URL


HerbCSO:
### First Check

- [x] I used the GitHub search to find a similar issue and didn't find it.
- [x] I have verified that this issue _is not_ related to the underlying library
[hhyrsev/recipe-scrapers](https://github.com/hhursev/recipe-scrapers) by **1)** checking
the [debugger](https://demo.mealie.io/g/home/r/create/debug) and data is returned, **2)**
verifying that there _are_ errors in the log related to application level code, or
**3)** verified that the site provides recipe data, or is otherwise supported by
[hhyrsev/recipe-scrapers](https://github.com/hhursev/recipe-scrapers)

- [x] This issue can be replicated on the demo site (https://demo.mealie.io/)

### Please provide 1-5 example URLs that are having errors

https://www.eatingwell.com/creamy-garlic-chicken-broccoli-skillet-11900291

I lied a little - I did find a similar issue in that I suspect this is similar to #7243 in that import via HTML works. It's just that I can't really ask my wife to copy-paste HTML all the time... ;] So I don't know if we can maybe change the user agent somehow to get around the blocking that they apparently do?

### Please provide your logs for the Mealie container `docker logs <container-id> > mealie.logs`

mealie    | INFO     2026-03-15T11:44:15 - HTTP Request: GET https://www.eatingwell.com/creamy-garlic-chicken-broccoli-skillet-11900291 "HTTP/1.1 403 Forbidden"
mealie    | INFO     2026-03-15T11:44:15 - HTTP Request: GET https://www.eatingwell.com/creamy-garlic-chicken-broccoli-skillet-11900291 "HTTP/1.1 403 Forbidden"
mealie    | INFO     2026-03-15T11:44:16 - HTTP Request: GET https://www.eatingwell.com/creamy-garlic-chicken-broccoli-skillet-11900291 "HTTP/1.1 403 Forbidden"
mealie    | INFO     2026-03-15T11:44:16 - HTTP Request: GET https://www.eatingwell.com/creamy-garlic-chicken-broccoli-skillet-11900291 "HTTP/1.1 403 Forbidden"

### Deployment

Docker (Linux)

HerbCSO:
Awesome, can confirm the new version works again - thank you so much for the quick fix! Renewed my sponsorship. ;]

Issue #7243, opened 2026-03-14
[SCRAPER] - seriouseats.com recipe imports fail


SeeThisIsMe:
### First Check

- [x] I used the GitHub search to find a similar issue and didn't find it.
- [x] I have verified that this issue _is not_ related to the underlying library
[hhyrsev/recipe-scrapers](https://github.com/hhursev/recipe-scrapers) by **1)** checking
the [debugger](https://demo.mealie.io/g/home/r/create/debug) and data is returned, **2)**
verifying that there _are_ errors in the log related to application level code, or
**3)** verified that the site provides recipe data, or is otherwise supported by
[hhyrsev/recipe-scrapers](https://github.com/hhursev/recipe-scrapers)

- [x] This issue can be replicated on the demo site (https://demo.mealie.io/)

### Please provide 1-5 example URLs that are having errors

I have used the seriouseats.com website for years to scrape recipes into Mealie without problems, it always used to work fine until a couple of months ago. I think they might have changed something with the website or formatting of recipes (possibly to try and prevent external scrapers importing recipes as they recently rolled out their own recipe app service).

When now trying to scrape serious eats recipes into Mealie the following error occurs for all recipes:

<img width="1346" height="776" alt="Image" src="https://github.com/user-attachments/assets/30f8f616-80d9-4069-b763-fbcba40a1221" />

This error also happens on the demo site, and seriouseats.com is on the supported sites list. 

However, when copy/pasting the raw HTML of a recipe page as the import method into Mealie, it does work and the complete recipe is imported.

The error can be reproduced with any URL, here are some examples:
https://www.seriouseats.com/crispy-pork-and-cabbage-egg-rolls
https://www.seriouseats.com/30-minute-creamy-lemony-chickpea-tahini-soup-11924951
https://www.seriouseats.com/chicken-enchiladas-negras-recipe-11827466

### Please provide your logs for the Mealie container `docker logs <container-id> > mealie.logs`

I'm unsure how to do this. But the issue can be reproduced on the demo site.

### Deployment

Other

michael-genson:
They seem to be blocking our scrape traffic, have you tried the HTML importer? Seems to work fine on the demo: https://demo.mealie.io/g/home/r/crispy-pork-and-cabbage-egg-rolls-recipe

SeeThisIsMe:
Hi, 

[...]

Issue #6180, opened 2025-09-16
[SCRAPER] - All Recipes


odaty:
### First Check

- [x] I used the GitHub search to find a similar issue and didn't find it.
- [x] I have verified that this issue _is not_ related to the underlying library
[hhyrsev/recipe-scrapers](https://github.com/hhursev/recipe-scrapers) by **1)** checking
the [debugger](https://demo.mealie.io/g/home/r/create/debug) and data is returned, **2)**
verifying that there _are_ errors in the log related to application level code, or
**3)** verified that the site provides recipe data, or is otherwise supported by
[hhyrsev/recipe-scrapers](https://github.com/hhursev/recipe-scrapers)

- [ ] This issue can be replicated on the demo site (https://demo.mealie.io/)

### Please provide 1-5 example URLs that are having errors

I'm trying to import this:
https://www.allrecipes.com/recipe/268249/italian-baked-meatballs/

### Please provide your logs for the Mealie container `docker logs <container-id> > mealie.logs`

In demo.mealie.io it works.

 Looks Like We Couldn't Find Anything
Only websites containing ld+json or microdata can be imported by Mealie. Most major recipe websites support this data structure. If your site cannot be imported but there is json data in the log, please submit a github issue with the URL and data.

**Details**
Version: v3.2.1
Build: db2c14093da1dc2ea8825ef6dea7b86bdd6ae2c2
Application Mode: Production
Demo Status: Not Demo
API Port: 9000
API Docs: Enabled
Database Type: sqlite
Default Household: Family
Recipe Scraper Version: 15.9.0

**Checks**
Secure Site: No
Server Side Base URL: No
LDAP Ready: No
OIDC Ready: No
OpenAI Ready: No
Email Configured: No

The debug scraper just says this: "recipe_scrapers was unable to scrape this URL"

### Deployment

Docker (Synology)

michael-genson:
Can you provide your backend logs? Likely what's happening is your request is getting blocked and you're not able to fetch the HTML. You can also try using the HTML importer and manually copy the source that way (right click webpage -> view source -> copy/paste the whole page into the HTML importer).

odaty:
Sorry for not knowing where are the logs on the backend?

michael-genson:
They'll be in the container logs
`docker logs mealie` (assuming your container is named `mealie`)

odaty:
[...]

PR #7250, opened 2026-03-15
fix: Update user agents in user-agents.txt

RealMochaFox:
Provide fresh user agents to Mealie, fixing scraper 403 errors on certain websites.

An example of one of the default user agents being blocked:
scraper failing

That same request passing with a new user agent:
image

Potentially #6180
Potentially #7243
Potentially #7249

2 of 3 user agents are getting 403'd. I kept the single one that still works. It may be worth creating a larger list or considering pulling in a dependency that generates valid, randomized user agents on the fly in the future if you haven't considered these routes.

## Testing
Local testing with Postman to recreate 403 issues and remedy them.


RealMochaFox:
Unlike the linked issue, note that this can actually be reproduced now on demo.mealie.io.

michael-genson (approved):
Thanks! I think I'm going to look into a library that handles this better in the future, but for now this will do

RealMochaFox:
I am generally a TypeScript person, however Python is not entirely out of my wheelhouse.

Depending on how the next week or so goes, I might take the time to work something up if that is cool!

Thank you for the approval!

PR #7255, opened 2026-03-15
fix: exclude index.html from workbox precache to prevent stale 401 app shell

### LINKED ISSUE #7226: [BUG]: PWA Service Worker traps user in 401 JSON view due to stale `navigateFallback` cache
### First Check

**I have noticed this only happens in my OIDC configured setup, it did not happen in https://demo.mealie.io**

When a user session expires, the PWA Service Worker continues to serve a cached "App Shell" (`index.html`) via the `navigateFallback` configuration.
Instead of the Nuxt frontend recognizing the unauthenticated state and redirecting to `/login`, the cached shell attempts to hydrate and fetch data from `/api/app/about`.
The API correctly returns a `401 Unauthenticated` JSON response, but because the Service Worker has intercepted the navigation, the browser displays the Nuxt 401 page instead of the login page.

The root cause is the `workbox` configuration in `nuxt.config.ts`:
*   `navigateFallback: "/"`: Causes the Service Worker to serve the cached root `index.html` for all non-API navigation requests.
*   `ssr: false`: The app relies entirely on client-side hydration. If the cached shell is "stale," it attempts to run as an authenticated app when it is not.
*   **The Conflict:** The `navigateFallbackAllowlist: [/^(?!\/api)/]` prevents the Service Worker from caching the API, but it **forces** the Service Worker to serve the cached frontend for the root URL. This "traps" the user in a client-side environment that cannot escape to the login page without a Hard Refresh (`Ctrl+Shift+R`).

### Steps to Reproduce

1. Log into Mealie (verified on a root subdomain deployment).
2. Manually delete the `mealie.access_token` cookie and all `localStorage` items in DevTools **without refreshing**.
3. Perform a **standard refresh** (`Ctrl+R` / F5).
4. **Observed Result:** The browser gets trapped in Nuxt's 401 page:

Image

6. **Expected Result:** The app should detect the 401/missing token and redirect to the `/login` page.

I don't think are relevant for this report

### Mealie Version

3.12.0

### Deployment

Docker (Linux)

_No response_

api2062:

Excludes `index.html` from the PWA Workbox precache list in `frontend/nuxt.config.ts`.

When a user session expires, a stale cached app shell can be served and the app can get stuck on a 401 page instead of cleanly returning the user to `/login`. By excluding `index.html` from precache, this avoids serving that stale shell in the reported scenario.

Fixes #7226

- Minimal one-line Workbox config change only.
- Approach follows maintainer guidance in the issue discussion.

## Testing

- Manual repro path from #7226:
  1. Login to Mealie
  2. Remove auth token/cookies
  3. Perform normal refresh
  4. Verify app no longer gets trapped in stale 401 shell behavior

- In this environment, full frontend dependency install/check commands were not completed due package manager/network instability.

## AI / LLM Assistance

Used AI assistance to help draft and refine the config change and PR text. Final change selection and issue mapping were reviewed manually.


michael-genson (approved):
LGTM, thanks!

PR #7276, opened 2026-03-19
fix: Use latest python image as base

dswd:

Use the latest python 3.12-slim as base image to reduce the number of vulnerabilities and get rid of critical ones.

image

## AI / LLM Assistance

Pure hard human work


michael-genson (approved):
Thanks!

PR #7320, opened 2026-03-25
feat: floating save button when editing recipe

DeepReef11:

When editing a recipe and scrolling down past the toolbar, a floating save button appears at the bottom-right corner of the screen. It disappears when the toolbar scrolls back into view.

This solves the UX issue of having to scroll all the way back up to save after editing ingredients or instructions further down the page.

- Uses `IntersectionObserver` on the recipe header toolbar component
- When the toolbar exits the viewport, a fixed-position save button (green, with save icon) appears at bottom-right
- Clicking it triggers the same `saveRecipe()` function as the toolbar save button
- Only visible in edit mode, hidden in view/cook mode and print
- Tooltip shows "Save" on hover

**Files changed:**
- `frontend/components/Domain/Recipe/RecipePage/RecipePage.vue` — Added floating button, IntersectionObserver logic, CSS

No specific issue — general UX improvement for recipe editing.

## Testing

- Tested in production
- Edit a recipe → scroll down → floating save button appears
- Click floating button → recipe saves, exits edit mode
- Scroll back up → button disappears
- Verified not visible in view mode or cook mode

## AI / LLM Assistance

This PR was developed with assistance from Claude Code (Claude Opus 4.6). Claude implemented the IntersectionObserver logic and floating button styling. All code was reviewed and tested by the PR author.


michael-genson:
Please update your PR description to use the GitHub PR template.

DeepReef11:
I have updated PR description

p0lycarpio:
It looks good! One small suggestion ; could you use the [v-fab](https://vuetifyjs.com/en/components/floating-action-buttons/) component instead of a button and a custom class?

DeepReef11:
Done — switched to `v-fab` component with `location="bottom end"`, `app`, and `appear` props. Removes the custom `.floating-save-btn` CSS class entirely. Tested and working in prod.

p0lycarpio:
LGTM

DeepReef11:
Been using for a while without issue. The only thing I found irritating is that it hides behind the keyboard on mobile . I have to check if there would be a solution for that so that it would stay in screen when keyboard is on. I will try investigate soon (or do another pr later if its taking too long)

p0lycarpio:
Not stale

DeepReef11:
I did not check for preventing btn to hide behind keyboard and did not plan investigating. Otherwise, current floating save btn works

p0lycarpio:
That seems sufficient as it is. Can you resolve the conflicts, and then we'll submit it to the maintainers?

DeepReef11:
CI fixed

michael-genson (commented):
Generally looks good, is there a reason for only having the "save" icon and not the text? (like the button).

In regards to the overlap with the keyboard, maybe having it in the top right would work better? I'm fine with it as-is though, we can always address the overlap issue later

DeepReef11:
I use Mealie in French. The icon only is much smaller than "Sauvegarder" or "Enregistrer". But perhaps I should add text as well for uniformity?

michael-genson:
Ah fair enough, let's go ahead and get this in as-is, if it bugs me (or someone else) enough we can always change it. Thanks for your work on this!

michael-genson (approved):

PR #7322, opened 2026-03-26
feat: Added sitewide statistics

Choromanski:

Adds sitewide statistics to /admin/site-settings

image

N/A

N/A

## Testing

Built and tested counts

## AI / LLM Assistance

None


Choromanski:
Let me know if this is the correct page to add this to, if it is preferred that this is on a different page I will change it.

PrettyPlot:
Do you think it could also include "unparsed" ?

Presumably, then, you can click on the box and be taken to a filtered view that lets you edit the ones missing tags/categories?

Choromanski:
@PrettyPlot what do you mean by unparsed?

I'm assuming you're looking for unparsed ingredients?

PrettyPlot:
> @PrettyPlot what do you mean by unparsed?
> 
> I'm assuming you're looking for unparsed ingredients?

Sorry, I mean it could be a useful way to see which recipes still need to have their ingredients parsed.  Helpful to find rogue recipes that slipped through the net, as it were!

Choromanski:
Yeah idk how what the maintainers goals are for the project. Seems like this could very quickly turn into a whole administration dashboard. Which I suppose isn't the goal of the project. 

I have seen some AI projects that try to make these dashboards for mealie.

Perhaps those would be a better option for you?

michael-genson (approved):
LGTM, thank you!

PR #7331, opened 2026-03-27
feat: Migrate PWA manifest to backend

### LINKED ISSUE #4563: [BUG] - Setting theme primary colour doesn't change theme-color meta
### First Check

I've recently customised the `THEME_DARK_ACCENT` and `THEME_LIGHT_PRIMARY` on my Mealie instance. After figuring out the caching issues, I've noticed that the `theme-color` meta tag isn't affected by this, leading to what I'd consider a UI bug:

ima_83684e1

On iOS for instance, the orange banner appears as you start scrolling, which is rather jarring difference from the selected primary colour used in the banner.

### Steps to Reproduce

- Customise your mealie colours
- Observe

n/a

### Mealie Version

v2.2.0

### Deployment

Other (please specify below)

On NixOS

Choromanski:

The PWA manifest was previously created at build and didn;t take any of the environment variables.

This lead to issues with the border color when using a custom theme:

Ex:
image

Inspecting the manifest file you can see it was using orange:
image

With the fix it now uses the environment variables:

image

Fixes #4563 

## Testing

Built and uploaded screenshots

## AI / LLM Assistance

Used to confirm PWA configuration


michael-genson:
Amazing. Does this work even after installing the PWA? In our older PWA with Nuxt 2 the manifest was statically installed on devices, so changes wouldn't work without a reinstall.

Choromanski:
@michael-genson Is this the scenario you're considering:

1. Mealie running (default env variables)
2. Install PWA
3. Stop Mealie, update env variables and re-start mealie
4. Open PWA, does color change?

michael-genson:
Yup!

michael-genson:
Seems like it will work with how our current PWA works with updates, but I know this was a limitation previously.

Choromanski:
It should It should. If you go to chrome://web-app-internals/ there is a "manifest_update_time" value (set to "1601-01-01" at install)

[This article](https://web.dev/articles/manifest-updates) says that it updates every day or so

michael-genson:
Yup seems right. Even if it's not it's still better than what we had before

mealie/routes/spa/__init__.py
```diff
@@ -234,6 +234,125 @@ async def serve_shared_recipe_with_meta(group_slug: str, token_id: str, session:
...
+def serve_manifest(resp: Response):
+    settings = get_app_settings()
+    sub_path = os.environ.get("SUB_PATH", "/") or "/"
+
+    manifest = {
+        "name": "Mealie",
+        "short_name": "Mealie",
+        "id": sub_path,
+        "start_url": sub_path,
+        "scope": sub_path,
```
michael-genson: Was there a reason to update these? The old manifest hardcoded these to `/`.

mealie/routes/spa/__init__.py
```diff
@@ -234,6 +234,125 @@ async def serve_shared_recipe_with_meta(group_slug: str, token_id: str, session:
         return response_404()
 
 
+def serve_manifest(resp: Response):
```
michael-genson: You can remove this arg

michael-genson (changes requested):
One question and one small thing. Also, can we move this into its own submodule? Something like `manifest.py`. This whole module could probably use a refactor but because we use global vars it won't be trivial, I can take that up later.

mealie/routes/spa/__init__.py
Choromanski: Removed

mealie/routes/spa/__init__.py
Choromanski: Just updated these:
- [id](https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Manifest/Reference/id): BASE_URL(has to be unique across all installed PWAs so this is probably best)
- [start_url](https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Manifest/Reference/start_url): relative path (defaults to '/' if no relative path in BASE_URL, needed if anyone has mealie installed at say www.example.com/mealie)
- [scope](https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Manifest/Reference/scope): same as start_url

Choromanski:
Added a bit of logic to see if light_primary or dark_primary is different from the default and if so set that as the theme_color (if both it takes light)

Also set the background color to #1E1E1E so that the white page on first load isn't as agressive.

Choromanski:
Yeah it looks like having the ID set to '/' was a bad ides:

https://developer.chrome.com/docs/capabilities/pwa-manifest-id

anyways I'm done and thus PR has been tested again and is ready.

Choromanski:
@michael-genson Is this still waiting on comments, I've resolved the one request and added a comment to the other one trying to explain it.

michael-genson:
No you're fine, apologies just haven't had time to build/test since you updated. Will try to get to this soon!

Choromanski:
You're all good, I know this one is kind of annoying to test since you have to wait for the PWA to update.

On android I know you can go to chrome://webapks to do a force update of the manifest file.

michael-genson (dismissed):
LGTM. I made two changes:
- I reverted the id to "/". While technically incorrect, changing this will break existing installs, forcing users to re-install the PWA manually. We can revisit this in the future but right now it's working fine as-is.
- I removed the custom theme color for light/dark, since it wouldn't work right if users only customized one but not the other. Instead we only use the light mode one

PR #7334, opened 2026-03-27
feat: Pass user defined units as custom units to parse_ingredient function.

### LINKED ISSUE #7108: [BUG] - Ingredient parser not detecting units & ingredients in database
### First Check

When I import à recipe, the ingredient parser is unable to detect units & ingredient even if it should be a perfect match from the database.

### Steps to Reproduce

1) Add the unit & ingredient in the database (Data management)
2) Parse a recipe that has those units & ingredients
3) The ingredient parser cannot detect the units & ingredients in both NLP and brute parser 

Image

Image

Image

### Mealie Version

[v3.11.0](https://github.com/mealie-recipes/mealie/releases/tag/v3.11.0)

### Deployment

Docker (Linux)

Using HTTP through a VPN.

strangetom:

Version 2.6.0 of ingredient-parser-nlp added a `custom_units` keyword argument to `parse_ingredient` that takes a dict of custom units. The changes in this PR construct a dict of custom units from the database that will includes any user defined units and passes it to the `custom_units` argument.

This means that any user defined units that weren't part of the training data used to train the model in ingredient-parser-nlp should be correctly identified as units.

Fixes #7108 (partially, this doesn't work if a unit has multiple words)
Fixes #7078 

## Testing

Tested manually:

Before:
Screenshot from 2026-03-27 14-14-38

Add new unit "barrel":
Screenshot from 2026-03-27 14-15-56

After:
Screenshot from 2026-03-27 14-17-18

## AI / LLM Assistance

No AI/LLM assistance used.


michael-genson (changes requested):
Awesome, I thought I saw this somewhere but couldn't find it in the docs.
Can we iterate through `units_by_id` instead? `units_by_alias` has a bunch of duplicates.

```python
for ingredient_unit in self.data_matcher.units_by_id.values():
    ...
```

Also can we guard and make sure `unit.name` isn't an empty string? It shouldn't be but we have similar guards elsewhere:

```python
if ingredient_unit.name:
    ...
```

Same for abbreviation (which is actually optional):

```python
if ingredient_unit.abbreviation:
    ...
```

strangetom:
Thanks for the feedback. I've made the suggested changes.

I've also done some additional (manual) testing with some simple non-English sentences and it behaves as expected once the units are defined in the database.

michael-genson (approved):
Wonderful, thank you! This should be a big help for non-english recipes

PR #7344, opened 2026-03-28
fix: prevent stale SPA shell after container rebuild

DeepReef11:

After rebuilding a Mealie container, Nuxt generates new content-hashed bundle filenames (e.g. `_nuxt/abc123.js` → `_nuxt/xyz789.js`). However, browsers (and Home Assistant iframe cards) serve the old cached `index.html` which references the old bundles that no longer exist on the server, resulting in a **blank white page** until the user hard-refreshes or clears their cache.

This is especially disruptive in Home Assistant dashboard iframes where users can't easily force-refresh.

Changes to `mealie/routes/spa/__init__.py`:
- `_nuxt/*` assets: `Cache-Control: public, max-age=31536000, immutable` (content-hashed filenames make this safe)
- HTML shell (everything else): `Cache-Control: no-cache, no-store, must-revalidate` so the browser always fetches fresh markup with correct bundle references

No existing issue — discovered in production when HA iframe consistently showed blank page after container rebuilds.

The fix is minimal — a single file change to the `SPAStaticFiles` class that serves the Nuxt frontend. The `_nuxt/*` immutable caching is safe because Nuxt uses content hashing for those filenames. The `no-cache` on HTML is standard practice for SPAs (e.g. this is the default behavior in nginx SPA configs).

## Testing

- [x] Rebuilt container (confirmed new bundle hashes generated: `bD9kXa78.js` → `BMf-Ov-T.js`)
- [x] Verified `Cache-Control: no-cache, no-store, must-revalidate` on HTML responses
- [x] Verified `Cache-Control: public, max-age=31536000, immutable` on `_nuxt/*` assets
- [x] Confirmed HA iframe loads correctly after rebuild without manual cache clear
- [x] Confirmed normal browsing still works (assets cached properly)

## AI / LLM Assistance

This PR was authored with Claude Code (Claude Opus 4.6). The problem was identified by the user from real-world HA iframe usage; Claude implemented the fix, tested it on a production instance, and drafted this PR.


michael-genson (changes requested):
This seems too aggressive, basically removing all caching, including static assets, just so the first load works properly.

I'd prefer to see a solution in the frontend rather than the backend. Is there a workbox setting that can address this?

DeepReef11:
I will keep working on it. I did not get the results I was hoping for unfortunately. It solves the iframe issues, that is pretty much it.

DeepReef11:
Thanks for the feedback on the aggressiveness. I've pushed a narrowed version that:
- Only targets HTML responses (not favicon, manifest, sw.js, etc.)
- Uses `no-cache` instead of `no-cache, no-store, must-revalidate`
- Drops the `Pragma` header

That said, I've also found that this can be handled entirely at the reverse proxy level (nginx) without any Mealie code changes, which is what I'm doing in my own deployment:

```nginx
map $upstream_http_content_type $cache_control_header {
    ~text/html  "no-cache";
    default     "";
}

location /_nuxt/ {
    proxy_pass http://mealie:9000;
    proxy_hide_header Cache-Control;
    add_header Cache-Control "public, max-age=31536000, immutable";
}

location / {
    proxy_pass http://mealie:9000;
    proxy_hide_header Cache-Control;
    add_header Cache-Control $cache_control_header;
}
```

Happy to keep this PR open if the team thinks it's worth having the fix built-in for users without a reverse proxy, or close it if the recommendation is to handle this at the proxy layer. Either way works for me.

michael-genson (approved):
Thanks! I removed a redundant `or` but otherwise left it as-is.

PR #7358, opened 2026-03-29
fix: eliminate white flash on page load for dark theme users

DeepReef11:

Adds a small inline `<script>` in the HTML `<head>` that sets `document.documentElement.style.backgroundColor` before the browser paints, matching the user's selected theme (dark: `#1E1E1E`, light: `#FFFFFF`).

Without this, there is a visible white flash during SPA hydration — the `<body>` defaults to a white background until Vuetify mounts and applies the theme. This is especially noticeable when Mealie is embedded in a Home Assistant iframe card with a dark dashboard.

The script reads `localStorage['vueuse-color-scheme']` (set by `@vueuse/useDark` in the existing `dark-mode.client.ts` plugin) and falls back to the OS `prefers-color-scheme` media query if no preference is stored.

Changes:
- `frontend/nuxt.config.ts`: added inline script to `app.head.script`

No existing issue — cosmetic improvement discovered when using Mealie in a Home Assistant dashboard iframe.

The inline script is ~200 bytes, runs synchronously before paint, and has zero performance impact (just a localStorage read and one style assignment). It's wrapped in a try/catch so it silently no-ops if localStorage is unavailable (e.g. incognito/restricted contexts).

This is the standard pattern used by Next.js, Nuxt, and other SPA frameworks to prevent theme flash (FOIT — Flash of Incorrect Theme).

## Testing

- [x] Verified dark theme: no white flash on reload
- [x] Verified light theme: background stays white (no dark flash)
- [x] Verified fallback: with no localStorage entry, uses OS preference
- [x] Verified in Home Assistant iframe: no flash on dashboard load
- [x] Verified script appears in generated HTML `<head>`

## AI / LLM Assistance

This PR was authored with Claude Code (Claude Opus 4.6). The problem was identified by the user from real-world HA iframe usage; Claude implemented the fix, tested it on a production instance, and drafted this PR.


michael-genson (approved):
Thanks!

PR #7363, opened 2026-03-29
feat: allow user to specify cookiefile for yt-dlp

space192:

Mealie yt-dlp integration currently doesn't allow the user to specify the cookiefile to use to browse the web, leading to rate limits or errors on websites (instagram for example).

The goal of this PR is to add that option by specifying the path to the cookiefile in the env of the docker container

## Testing

Tested that the feature did not create any issue when not provided, due to the default value set to None, also tested when specifying a file to be used by yt_dlp

## AI / LLM Assistance

No use


lehtomar:
I'm hitting this error as well. Hope this work could be merged

michael-genson (approved):
Thank you for this! I fixed the merge conflicts and tweaked the docs.

PR #7370, opened 2026-03-31
feat: Back button sets view to where you left page

### LINKED ISSUE #5350: [BUG] - UX problem: Back from the detail view, scroll position always at the start again

Choromanski:

Viewing a recipe and pressing back in the browser now takes you to where you left the page (instead of the top of the page).

Fixes #5350

New to writing composables so let me know if I missed anything.

## Testing

Built and ran.

## AI / LLM Assistance

A decent amount to figure out how composables work and to determine how to cache the scroll state.


frontend/composables/use-scroll-position.ts
```diff
@@ -0,0 +1,32 @@
+export function useScrollPosition() {
+  const router = useRouter();
+  const scrollPositions = new Map<string, number>();
+
+  router.beforeEach((to, from) => {
+    scrollPositions.set(from.path, document.documentElement.scrollTop);
+  });
+
+  router.afterEach((to, from, failure) => {
```
michael-genson: We need to clean these up on unmount otherwise there's a memory leak. Something like:
```ts
const unregister = router.beforeEach(...)
onUnmounted(() => unregister());
```

I forget exactly what the API looks like, but it's something like that.

frontend/composables/use-scroll-position.ts
```diff
@@ -0,0 +1,32 @@
...
+
+    let timeout: ReturnType<typeof setTimeout>;
+
+    const observer = new MutationObserver(() => {
+      clearTimeout(timeout);
+      timeout = setTimeout(() => {
+        observer.disconnect();
+        document.documentElement.scrollTop = savedPosition;
+      }, 100);
+    });
```
michael-genson: Let's add a catch to make sure this runs if no mutations occur, e.g.:
```ts
const observer = new MutationObserver(() => {
  clearTimeout(timeout);
  timeout = setTimeout(() => {
    clearTimeout(fallback);
    observer.disconnect();
    document.documentElement.scrollTop = savedPosition;
  }, 100);
});

const fallback = setTimeout(() => {
  clearTimeout(timeout);
  observer.disconnect();
  document.documentElement.scrollTop = savedPosition;
}, 500);
```

michael-genson (changes requested):
Nice! A few (fixable) issues with this noted below, and one bigger one:

This won't work if you scroll far enough to hit the third (or higher) page of recipes, since we only fetch the first two on load. Let's have `useScrollPosition` expose a ref which stores the page size and current page, then do something like:
```ts
const currentPage = ref<number>(1);  // we actually might start on page 2, I forget, check the implementation
useScrollPosition(currentPage);  // initialize currentPage if we have it saved (i.e. we hit back)
...  // load recipes and initialize scrolling
```

We will also need to load recipes _before_ restoring the scroll state, so you might need to do something like:
```ts
{ restorePosition } = useScrollPosition(...);
...  // load recipes
restorePosition();
```

frontend/composables/use-scroll-position.ts
Choromanski: Implemented

Choromanski:
@michael-genson, I've got it working with multiple pages, I only have 99 recipes in my test and it's working with all of them.

It me know if you'd like me to add more recipes and try again.

michael-genson:
Looks and works great! Just one more small improvement: the `AppLoader` component on the `RecipeCardSection` is nested under the `v-if="recipes && ready"` block, so it only appears _after_ the initial load. Can we move it out of that block so it always renders? (if `loading`).

This is a pre-existing issue if the initial fetch is slow, but now that we're potentially fetching a ton more data it's way more obvious than it used to be.

Choromanski:
@michael-genson, I've moved the AppLoader component. 
I've also set loading to true for the duration of onMounted so that the "Loading Recipes" loading widget pops up.

Let me know if you would like me to revert that change.

michael-genson (approved):
Wonderful, thanks for your work on this! This will make a lot of people very happy (including me!).
I wonder if there is a way to fix the long load upon returning to the main page using caching, but as-is this is a huge improvement.

I made a couple last-minute tweaks, mostly for code cleanliness nits.

Choromanski:
Appreciate it thanks for helping out

JoshuaArus:
It's already merged but I still wanted to thank both of you and let you know that I've been dying for this feature ! 

I would love that, as mentionned, recipes aren't reloaded when getting back to the list, but it's still already enjoyable :)

Choromanski:
> recipes aren't reloaded when getting back to the list

Could you elaborate on this?

JoshuaArus:
> > recipes aren't reloaded when getting back to the list
> 
> Could you elaborate on this? 
> 

I was refering to your "I wonder if there is a way to fix the long load upon returning to the main page using caching". Maybe I put that badly, sorry.

Choromanski:
Ahh that makes sense, I thought it was broken. This is something that I can look into

PR #7384, opened 2026-04-03
feat: Added scroll to top on all pages that have recipeCardSection

Choromanski:

With fix #7370 being implemented there may be a scenario where you go back to a recipe list, are put in the middle of the page when you actually wanted to be at the top of the page.

This adds a button to the bottom right to quickly take you to the top of the page.

This will only effect recipe list, favorite list, and cookbooks.

image

Mobile:
image

None

I picked [mdiArrowUp ](https://pictogrammers.com/library/mdi/icon/arrow-up/)as the icon but also think that [mdiChevronDoubleUp ](https://pictogrammers.com/library/mdi/icon/chevron-double-up/) would look nice (open to any other suggestions). 

I set the z-index of this to 999 (higer than everything else I could find, excpet AppHeader)m but don't know what you guys would prefer for this.

chevron double up:
image

## Testing

Built and manually tested.

## AI / LLM Assistance

None


michael-genson (approved):
Was just thinking about this this morning with the new back feature. Thanks!

Choromanski:
Great minds think alike 😉

PR #7389, opened 2026-04-03
fix: Display issues with data management pages on mobile

Choromanski:

The buttons on this page extend the page right past the mobile view window.

This also causes the header to get cut off when you scroll down.

Issue on mobile:
image

None

## Testing

Tested it by adding the class in devtools, but I need to do a proper build.

Just did a build and was able to test and confirm that it works.

## AI / LLM Assistance

None


michael-genson (approved):
Thanks!

PR #7390, opened 2026-04-04
feat: Improved handeling of duplicate backups

Choromanski:

For those who do regular backups.

From a users perspective backups function the same way, but when duplicate backup is created we use hard links which can significantly reduce disk usage over time.

- Implemented basic TorrentZip functionality
  - Overrides timestamps on files
  - Database json fle is sorted by keys before being added to archive
  - Added files to archive in sorted order
- zip is originally created as a .temp file and md5sum is calculated
  - if hash matches another backup, a hardlink is created to the matching zip
- Added sidecar `.zip.md5` file containing the md5sum of the `.zip` for all backups created/uploaded

None

- I set the file datetime to 1/1/1980 (earlest date zip supports)
- I set compression algorithim to ZIP_DEFLATED but ZIP_STORED will be faster (but larger)
- I used MD5 as the hash function (fastest "standard" algorithim in hashlib and I think its fine for file detecting duplicates)

## Testing

Built and tested api also works:

image

image

You can see that the inodes go up with each archive added but the storage usage stays constant:

image

Example of creating duplicate backup:
image

## AI / LLM Assistance

Used claude to help me come up with a torrentzip approach when I found out it wasn't supported in zipfile.


michael-genson (changes requested):
Generally looks good. A few things before merging:

1. `.temp_{name}.zip` is written directly into `BACKUP_DIR`, and pathlib's glob matches dotfiles — if the process is killed mid-backup before the except-handler cleanup runs, the stray file gets picked up by `get_all()`'s `glob("*.zip")` and shown as a real backup. Can we write it to the existing `.temp` subdir instead?
2. None of the new logic (hashing, duplicate detection, hardlinking, sidecar creation/cleanup) has test coverage. Can we add tests for the duplicate-detection path — two identical backups, second one returns `duplicate=True`, both share an inode, `_hash_zip`/`_find_duplicate` behave correctly with/without a sidecar present?
3. Same for the `create_one` duplicate-response branch (asserting `duplicate`/`duplicateOf` are set) and confirming `delete_one` also removes the `.zip.md5` sidecar.
4. `admin_backups.py:upload_one` calls `BackupV2._hash_zip()` across the module boundary — can we drop the leading underscore to make it public instead?

michael-genson:
Hey there, are you still working on this?

michael-genson:

PR #7392, opened 2026-04-04
dev: Bumped gh actions to support node 24

Choromanski:

To address warnings like this:

image

None

## Testing

Going to test with this PR

## AI / LLM Assistance

None


.github/workflows/release-drafter.yml
```diff
@@ -14,17 +14,22 @@ on:
...
+    runs-on: ubuntu-latest
     permissions:
-      # write permission is required to create a github release
       contents: write
-      # write permission is required for autolabeler
-      # otherwise, read permission is required at least
+    steps:
+      - uses: release-drafter/release-drafter@v7
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
michael-genson, line 25: Hopefully this works, if not might have to switch it to our custom Mealie bot token

michael-genson (approved):
Thanks!

PR #7401, opened 2026-04-05
feat: Add meal plan scale and shopping list quick add

sparkison:

Adds per-meal-plan-entry serving size scaling, allowing users to set a custom scale on each meal plan entry rather than always using the default recipe serving size. Also adds the ability to create a new shopping list on the fly from the "Add to Shopping List" dialog.

**Meal plan scale changes:**
- `mealie/db/models/household/mealplan.py` — added `recipe_scale` Float column to `GroupMealPlan`
- `mealie/schema/meal_plan/new_meal.py` — added `recipe_scale: float = 1.0` to `CreatePlanEntry`; all subclasses (`UpdatePlanEntry`, `ReadPlanEntry`, `SavePlanEntry`) inherit it automatically
- `mealie/alembic/versions/2026-04-05-*_add_recipe_scale_to_meal_plan.py` — migration adding the column with `server_default=1` for backward compatibility
- `frontend/lib/api/types/meal-plan.ts` — added `recipeScale?: number` to all four plan entry interfaces
- `frontend/pages/household/mealplan/planner/edit.vue` — added a "Servings Scale" number input to the create/edit meal dialog; shows a scale badge chip on cards when scale ≠ 1
- `frontend/pages/household/mealplan/planner.vue` — fixed `weekRecipesWithScales` to use `meal.recipeScale ?? 1` instead of hardcoded `1`, so "Add All to List" uses the correct scale per recipe
- `frontend/pages/household/mealplan/planner/view.vue` — added `MealPlanScaleControl` below each recipe card in the view tab; passes `recipeScale` to navigation so the recipe detail page opens pre-scaled
- `frontend/components/Domain/Household/MealPlanScaleControl.vue` — new component styled identically to the recipe detail page scale control (same `secondary-darken-1` badge, popover input, +/- buttons); auto-saves changes to the meal plan entry with a debounce
- `frontend/components/Domain/Household/GroupMealPlanDayContextMenu.vue` — updated to accept `ReadPlanEntry[]` and use stored `recipeScale` when adding a day's recipes to a shopping list
- `frontend/components/Domain/Recipe/RecipeCardMobile.vue` — added `scale` prop; passes it to `RecipeContextMenu` so individual "Add to Shopping List" from a meal plan card uses the correct scale; appends `?scale=` to recipe navigation route
- `frontend/components/Domain/Recipe/RecipePage/RecipePage.vue` — reads `?scale` query param on mount and initialises the scale state, so opening a recipe from the meal planner pre-sets the serving size
- `frontend/components/Domain/Recipe/RecipeContextMenu/RecipeContextMenuContent.vue` — passes `recipeScale: props.recipeScale` when adding a recipe to the meal plan, so the scale set on the detail page is captured

**Create shopping list on the fly:**
- `frontend/components/Domain/Recipe/RecipeDialogAddToShoppingList.vue` — added a "New List" button to the shopping list selection dialog; expands an inline form to name and create a list via the API, then immediately proceeds to the ingredient selection step for the new list
- `frontend/lang/messages/en-US.json` — added `new-list-name`, `failed-to-create-shopping-list`, and `servings-scale` i18n keys

_(no existing issue — new feature)_

- The `recipe_scale` field defaults to `1.0` at both the DB level (`server_default=sa.text("1")`) and the Pydantic schema level, so all existing meal plan entries are unaffected.
- `MealPlanScaleControl` is intentionally a separate component rather than reusing `RecipeScaleEditButton` directly, because it needs to own the save-to-API side effect while keeping the same visual appearance.
- The `RecipeDialogAddToShoppingList` change works for both the recipe detail page and the meal planner "Add All to List" flow since both go through the same component.

## Testing

- Added a recipe to the meal plan from the detail view with a non-default scale (e.g. 2×) — verified the meal plan entry stored the correct scale
- Edited scale via the +/- controls and popover in the meal planner view tab — verified changes saved and persisted on refresh
- Clicked a recipe card in the meal planner — verified the recipe detail page opened with the correct serving size pre-set
- Used "Add All to List" — verified ingredient quantities reflected the stored scales
- Used "Add to Shopping List" from the per-card context menu — verified it used the meal plan scale, not the default
- Tested "New List" button with no existing lists and with existing lists — verified the new list was created and the ingredient dialog opened immediately

## AI / LLM Assistance

This pull request was developed with significant AI assistance (Claude via Claude Code). The AI was used to explore the codebase, design the implementation plan, and write the majority of the code changes. All changes were reviewed and tested manually before submission.

## Screenshots

Meal planner serving size adjustment:
scale-in-menu-planner

Shopping list quick add:
on-the-fly-list-creation


sparkison:
This addresses discussion requests:
- https://github.com/mealie-recipes/mealie/discussions/5241
- https://github.com/mealie-recipes/mealie/discussions/2303

sparkison:
Just synced up with upstream to resolve conflicts. Should be ready to merge again! 😉

wlburks:
Hey does this just need someone to resolve the conflict.  The ability to adjust servings via the meal planner would be great to have!

Back2War13:
Great future feature! Adjusting servings for a big family, where each day there might be different number of people eating is a real game changer!

wlburks:
Someone with permissions need to resolve the conflicts! This would be a lifesaver!

wlburks:
Bump

PaeyMoopy:
bumping this... whats the hold up?

michael-genson (changes requested):
I like the feature, a few changes:

1. Can we add backend tests for `recipe_scale` on `GroupMealPlan`? Specifically, cover the default value, create/update via the API, and confirm the migration chain applies cleanly.
2. `createAndOpenList` pushes the new list into `filteredShoppingLists` (a plain ref, so that part's fine), but never updates `props.shoppingLists`, which is what the reset-on-reopen watcher re-filters from. The new list will disappear again the next time this dialog is opened, until the parent independently refetches. Should this emit an event so the parent can update its own list, or does something else already handle that?
3. The auto-save watcher in `RecipePage.vue` silently overwrites the meal plan entry's `recipeScale` whenever someone adjusts serving size while just previewing a recipe opened from a meal plan link. Is that intentional, or should it be gated behind an explicit save action?
4. Can we add frontend test coverage for `MealPlanScaleControl.vue`'s recalculate/debounce logic and the new-list creation flow? Neither has any tests currently.

Kipmo:
Has this PR been abandoned? Would have been really nice to have these features.

michael-genson:

danielemenni:
Too bad it was closed, as it is now it's impossible to make a shopping list from the recipes because most recipes are made for 4 people, but for now we are only two... I wouldn't want to scale all the recipes from 4 to 2 and then up to 3 or 4 when we have kids o guests...

Issue #7072, opened 2026-02-15
[BUG] - Failed recipe PATCH can persist ingredients with NULL `reference_id`, bricking recipe (500 ValidationError on GET/PATCH/DELETE)


Faryn:
### First Check

- [x] This is not a feature request.
- [x] I added a very descriptive title to this issue (title field is above this).
- [x] I used the GitHub search to find a similar issue and didn't find it.
- [x] I searched the Mealie documentation, with the integrated search.
- [x] I already read the docs and didn't find an answer.
- [ ] This issue can be replicated on the demo site (https://demo.mealie.io/).

### What is the issue you are experiencing?

# Bug: Failed recipe PATCH can persist ingredients with NULL `reference_id`, bricking recipe (500 ValidationError on GET/PATCH/DELETE)

## Summary
On Mealie **nightly**, creating a recipe via the API and then attempting to patch recipe ingredients with an **invalid** `referenceId` (non-UUID) correctly returns **422**, but appears to **partially persist** ingredient rows with `reference_id = NULL` in SQLite. After that, the recipe becomes **unopenable** in the UI and the API returns **500** on `GET /api/recipes/{slug}` (and also on PATCH/DELETE), with a Pydantic validation error.

This creates a “stuck” recipe that cannot be deleted or repaired through the API/UI because the backend crashes during model validation.

## Environment
- Image: `ghcr.io/mealie-recipes/mealie:nightly`
- Container name: `mealie`
- Database: SQLite
- Reverse proxy: nginx (serving HTTPS)

## What I was trying to do
Create a simple recipe via API (similar to what I already do for other recipes), using existing ingredient foods/units.

## Steps to Reproduce
1. Create a recipe stub:

   ```bash
   curl -X POST \
     -H "Authorization: Bearer $TOKEN" \
     -H "Content-Type: application/json" \
     -d '{"name":"Feta Creme"}' \
     https://<mealie-host>/api/recipes
   ```

   The API responds with the slug (e.g. `"feta-creme"`).

2. Patch the recipe with ingredients, but **mistakenly** send `referenceId` values that are **not UUIDs**, e.g. `"ing-feta"`:

   ```json
   {
     "recipeIngredient": [
       {
         "quantity": 200,
         "unit": {"id": "<unit-uuid>", "name": "gram"},
         "food": {"id": "<food-uuid>", "name": "feta cheese"},
         "display": "200 g feta cheese",
         "originalText": "200 g feta cheese",
[...]

PR #7426, opened 2026-04-08
chore: Nuxt 4 upgrade

Kuchenpirat:
## What this PR does / why we need it:

This upgrades the frontend to Nuxt v4. 
The most notable change is the new data structure. 
- static is renamed to public
- new app folder that houses most of the frontend application now

## Which issue(s) this PR fixes:

None

## Notes
I removed some dependencies that are no longer necessairy. Mostly old Typescript stuff from the nuxt 2 era. 

Updates regarding folder aliases. 
- `@` or `~` now point at `/frontent/app` 
-  `@@` or `~~` point towards `/frontend`

**Please don't merge too many frontend PRs before this is in as every single one of them will create merge conflicts.** 

## Testing

Have run this for about two weeks as prod in my household and we did not run into any issues during my testing. 

## AI / LLM Assistance

None

PR #7428, opened 2026-04-08
fix: BaseDialog padding

p0lycarpio:

This PR increases the spacing between the action buttons in the **BaseDialog** when displayed on mobile devices, following #7072.

The goal is to improve the accessibility of these buttons, which could previously be too close to the bottom of the screen and to the OS buttons.
The unused `below-actions` slot has also been removed.

N/A

## AI / LLM Assistance

N/A


Kuchenpirat:
Hi, we just merged the migration to nuxt 4 (#7426) which means that all frontend components live within the `frontend/app` folder.

Besides that there should be nothing else to consider so just migrate the changes to the new location.

p0lycarpio:
Indeed, great work! The merge was smooth.

Kuchenpirat:
Thanks, yeah luckily it was pretty straight forward 👌

Kuchenpirat (approved):
Looks good. 
Thanks :)

PR #7431, opened 2026-04-08
feat: Announcements

michael-genson:

_(REQUIRED)_

### Why?

This PR adds a mechanism for announcements, allowing us (the maintainers) to communicate directly with users in the app. There have been several instances in the past where we introduce something that requires action by the users, and simply including it in the release notes isn't sufficient for many users. One recent example is the [v3.7.0 release](https://github.com/mealie-recipes/mealie/releases/tag/v3.7.0) which includes some breaking changes to how images display on the home page. We received feedback from several users ranging from "I wish I knew about this earlier" and "I thought it was a bug that would eventually get fixed" (which are perfectly reasonable thoughts if you haven't seen the release notes).

It's important to note that these are _not_ a replacement for release notes. Users that care about release notes are probably already reading our releases. This is for things like "the healthcheck script has changed" and "run this user script to optimize your images".

### The Feature

And thus, announcements:
image

Announcements appear as notifications down near the settings menu:
image

Users can:
- Hit "Next" to go to the next unread announcement
- Mark all announcements as read
- Jump around to any past/present announcement

2026-04-08_16h45_43

### User Controls

Some users may not want to see announcements. Some admins may not want _any_ users to see announcements. This PR includes controls to disable announcements at various levels:
- Users can disable announcements for themselves
- Admins (or anyone with household management permissions) can disable announcements for the entire household
- Admins (or anyone with group management permissions) can disable announcements for the entire group

image

image

image

### Implementation

Aside from the feature toggles (which are checkboxes on existing preference models) this PR adds a single string field on the user model: `last_read_announcement`. Announcements are implemented as components prefixed with a timestamp. For instance, the first announcement (the only one in this PR) is `2026-03-27_1_welcome.vue`. This tells us:
- The announcement is dated `2026-03-27` (which is rendered to the user).
- This is the first announcement on that date (`1`). I assume most/all of the time there will always be exactly one announcement per date, but I figured I'd future proof it.
- The _internal_ name for this announcement is `welcome`

The format of announcements is a frontend unit test, to ensure we don't accidentally add an announcement in the wrong format (and our application logic assumes it can parse the date, etc.).

Since announcements are timestamped, we know that if we've seen announcement `2026-03-27_1_welcome` then we can automatically mark _older_ announcements as read, and _newer_ announcements as unread. Hence, we only need to store the last read announcement in the database. This does create a limitation: if a user skips ahead and reads the newest announcement, but doesn't read older ones, when they refresh the page all announcements up to the newest one will be marked as read. Application logic dictates that if a user views an older announcement, we _don't_ update their `last_read_announcement` value (otherwise announcements would constantly be considered "unread" if a user were to browse backwards). This implementation also lets us be flexible and delete announcements if we want to (there is no db relationship, it's just a timestamp).

Announcements are fully-fledged Vue components. That means we can fully leverage our components (as seen in the welcome example with the buttons), scripts, styles, router logic, etc. We export metadata on the announcement to set the title:
```html
<script lang="ts">
export const meta: AnnouncementMeta = {
  title: "Welcome to Mealie 🎉",
};
</script>
```

There is special logic for users who have not seen an announcement yet (i.e. new users, including brand new instances): regardless of how many announcements there are, users will _always_ be presented with exactly one announcement: the welcome announcement. All other announcements will be marked as read. This prevents the scenario where you launch Mealie for the first time and have 100 announcements to read (which are all likely useless anyway since you're not familiar with the app and they will be for older features).

Lastly, since these announcements are meant to be read once (around the time they come out) there's no real reason to enable translations on them. By the time they're translated everyone that needs to read them will have already read them. It would also be a _huge_ pain to maintain and take a ton of work for translators for ultimately little to no benefit.

_(REQUIRED)_

N/A, been wanting this for a while but hadn't thought of an implementation I was happy with.

Things to review (aside from the obvious):
- Does the UX make sense? Will users understand how to navigate announcements?
- Does the placement of the announcement button make sense? Should it be somewhere else? The top app bar?
- How does the welcome message look/read? This will _always_ be the first announcement a user sees, so it's important to get right.

## Testing

Added some tests to ensure newly-added announcements are formatted correctly (which could otherwise break the announcement logic). Played around with various workflows a lot, I generated some dummy announcements to test the UX.

## AI / LLM Assistance

_(REQUIRED)_

Built out some of the boring boilerplate and helped with the tests. Also generated the placeholder announcements (which are not in the PR).


Kuchenpirat (commented):
This is seriously pretty cool. 

UI/UX on desktop is already pretty good.

On mobile it is a bit tough tho. 
I would propably make the list of all anouncements a navigation drawer that one can toggle on smaller screens.

michael-genson:
@Kuchenpirat I've given mobile some love:
image
image

frontend/app/components/Domain/Announcement/Announcements/2026-03-27_1_welcome.vue
```diff
@@ -0,0 +1,62 @@
...
+    <div class="mb-2">
+      Announcements are reserved for things like:
+      <ul class="ml-6">
+        <li>Important new features</li>
+        <li>Major changes</li>
+        <li>Anything that might require additional user actions (such as migration scripts)</li>
+      </ul>
+    </div>
+    <p>
+      While we generally keep everything in our <a href="https://github.com/mealie-recipes/mealie/releases" target="_blank">GitHub release notes</a>,
```
Kuchenpirat: Just the most minor thing ever.
Since i had to remove the color from links for the vuetify 4 upgrade they will no longer be in the primary color. 

I'll see what can be done about this soon™️, to have links and links only automatically be in orange. 

```suggestion
      While we generally keep everything in our <a class="text-primary" href="https://github.com/mealie-recipes/mealie/releases" target="_blank">GitHub release notes</a>,
```

Kuchenpirat (dismissed):
Besides that it looks really good 🤩
Good work on the UI 👍

p0lycarpio:
Thanks for the PR ! Overall I really like the direction here.

What’s great:
- The UX looks clean and discoverable (sidebar entry + badge + dialog navigation).
- The "last read announcement" approach is lightweight and pragmatic (minimal DB/state, easy to reason about).
- I also appreciate the opt-out controls at user/household/group level because not everyone want to see them.

Right now announcements are fully bundled as static Vue components. That means the only way to publish a new announcement is to ship a new frontend release. For "release notes reinforcement" that’s fine, but it limits the feature as a real communication channel.

I suggest to consider switching to a *remote manifest approach* that allows:
- Keep the current UI/flow, but load a small announcements.json (or similar) from a configurable URL (e.g. official CDN/GitHub raw), describing available announcements + metadata (id/key, title, date, severity, etc.).
- The frontend could then either render remote content (Markdown) or use the manifest to decide which local components to show.

This would let maintainers communicate without requiring a new app release, which is especially valuable for time-sensitive messages like security vulnerabilities, bugs/incidents, behavior changes, or "please update to version X" prompts.

Not a blocker for me, but I think it would significantly increase the impact of this feature while keeping the implementation relatively simple.

Kuchenpirat:
Yeah, i have thought the same and noted something along those lines down in my todos. 
But wouldn't say this is a blocker to this PR, just something we would want to add in the future👍

michael-genson:
> load a small announcements.json (or similar) from a configurable URL (e.g. official CDN/GitHub raw), describing available announcements + metadata

Yeah I like this approach in general, I opted for this simpler approach mainly because I wanted to include Vue components 😅 

We could very easily integrate both approaches using the timestamp key I implemented here though, so definitely something to consider for the future!

PR #7432, opened 2026-04-09
chore: upgrade to vuetify v4

Kuchenpirat:

This is the next major upgrade of our frontend deps: Vuetify.

User facing changes in this are minimal. Mostly that [breakpoints](https://vuetifyjs.com/en/getting-started/upgrade-guide/#breakpoints) are changing a bit to better fit modern devices. We can set these ourselves if we feel some other breakpoints would better fit the application.

None

## Notes

- I had to remove the default color from links because otherwise the text / icon color of every button / link item etc would be set to primary. 
- I Updated some breaking Points in the Meal Planner to better fit the new Breaking Points

## Testing

Manual Testing. I am sure there are some minor UI things that slipped through the cracks but most stuff should be there. 

## AI / LLM Assistance

None

### DIFF
--- a/frontend/app/assets/main.css
+++ b/frontend/app/assets/main.css
@@ -61,10 +61,6 @@
   max-width: 100%;
 }
 
-a {
-  color: rgb(var(--v-theme-primary));
-}
-
 .fill-height {
   min-height: 100vh;
 }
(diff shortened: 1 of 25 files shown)


michael-genson (approved):
Wonderful

PR #7433, opened 2026-04-09
feat: add filter for recipes with unparsed ingredients

Xenov-X:
                                                                                                                                                                                                         
  Adds the ability to filter recipes by whether they have unparsed ingredients (ingredients missing a food or unit association). Unparsed ingredients degrade functionality like shopping lists, so users need a way to identify these recipes and proactively improve data quality.                                                                                                                            
                                                                                                                                                                                                         
  - **`mealie/db/models/recipe/recipe.py`** — Adds a `column_property` (`has_unparsed_ingredients`) to `RecipeModel` using a correlated EXISTS subquery. This is a SQL-computed attribute (no physical  column, no migration needed) that integrates with the existing `QueryFilterBuilder` system. An ingredient is considered unparsed if `food_id` or `unit_id` is NULL, excluding section headers and  recipe references.                                                                                                                                                                                     
  - **`frontend/app/composables/recipes/use-recipes.ts`** — Extends `refreshRecipes()` with an optional `queryFilterOverride` parameter to support dynamic re-fetching with different filters.
  - **`frontend/app/pages/group/data/recipes.vue`** — Adds a "Show only recipes with unparsed ingredients" checkbox to Settings > Data Management > Recipes. Toggling it re-fetches recipes server-side  with `hasUnparsedIngredients = true` in the query filter.                                                                                                                                              
  - **`frontend/app/lang/messages/en-US.json`** — Adds the i18n string for the checkbox label.                                                                                                           
                                                                                                                                                                                                         
  The filter is also available via the API for any use case that supports query filters:                                                                                
 `GET /api/recipes?queryFilter=hasUnparsedIngredients=true`                                                                                                                                               
                                                                                                                                                                                                         
 Feature Reqs:                                                                                                                                                                                                        
 #6762                                                                                                                                                                                        
 #2235
                                                                                                                                                                                                         
 The `column_property` approach was chosen over modifying `QueryFilterBuilder` to support virtual column aliases. Since `last_made` and `rating` are both real model columns whose aliases just override  the SQL expression, adding has_unparsed_ingredients` as a `column_property` follows the same pattern with zero changes to the filter infrastructure. The subquery only executes when the attribute is actually queried/filtered on.                                                                                                                                                                         
                                                                  
  ## Testing                                                                                                                                                                                             
   
  - Manual testing against a Docker deployment with seeded data and imported recipes                                                                                                                     
  - Verified the checkbox filters correctly on the Data Management > Recipes page
  - Verified the query filter works via API: `GET /api/recipes?queryFilter=hasUnparsedIngredients=true`                                                                                                  
  - Verified recipes with zero ingredients are not flagged              
  - - Verified recipes with all parsed ingredients are not flagged                                                                                                                                         
  - Verified section headers and recipe-reference ingredients are excluded from the unparsed check                                                                                                       
                                                                                                                                                                                                         
  ## AI / LLM Assistance                                                                                                                                                                                 
                                                                                                                                                                                                         
This PR was developed with significant assistance from an LLM. Used to identify implementation option, implementation and writing this PR. All changes were reviewed and tested by the author (testing completed noted above).

### DIFF
--- a/frontend/app/composables/recipes/use-recipes.ts
+++ b/frontend/app/composables/recipes/use-recipes.ts
@@ -136,8 +136,9 @@ export const useRecipes = (
     }
   })();
 
-  async function refreshRecipes() {
-    const { data } = await api.recipes.getAll(page, perPage, { loadFood, orderBy: "created_at", queryFilter });
+  async function refreshRecipes(queryFilterOverride?: string | null) {
+    const effectiveFilter = queryFilterOverride !== undefined ? queryFilterOverride : queryFilter;
+    const { data } = await api.recipes.getAll(page, perPage, { loadFood, orderBy: "created_at", queryFilter: effectiveFilter });
     if (data) {
       recipes.value = data.items;
     }
--- a/frontend/app/lang/messages/en-US.json
+++ b/frontend/app/lang/messages/en-US.json
@@ -1185,7 +1185,9 @@
       "export-recipes": "Export Recipes",
       "delete-recipes": "Delete Recipes",
       "source-unit-will-be-deleted": "Source Unit will be deleted",
-      "show-only-unparsed": "Show only recipes with unparsed ingredients"
+      "show-only-unparsed": "Show only recipes with unparsed ingredients",
+      "filter-missing-food": "Unparsed food",
+      "filter-missing-unit": "Unparsed unit"
     },
     "recipe-actions": {
       "recipe-actions-data": "Recipe Actions Data",
--- a/frontend/app/lang/messages/en-US.json
+++ b/frontend/app/lang/messages/en-US.json
@@ -1184,7 +1184,8 @@
       "categorize-recipes": "Categorize Recipes",
       "export-recipes": "Export Recipes",
       "delete-recipes": "Delete Recipes",
-      "source-unit-will-be-deleted": "Source Unit will be deleted"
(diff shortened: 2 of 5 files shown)


michael-genson:
This should be achievable with the existing API, rather than adding a database column which adds unnecessary overhead to every API call involving recipes. Check out[ the API docs](https://docs.mealie.io/documentation/getting-started/api-usage/#compound-filters) for more info on how the query filter works.

Xenov-X:
Good shout, hadn't clocked support for SQL expressions via the API. Updated to front end change only.

michael-genson (changes requested):
A few things:

1. The null-check filter (`recipeIngredient.foodId IS NULL` / `recipeIngredient.unitId IS NULL`) will also match section headers and recipe-reference ingredient rows, which legitimately have null `food_id`/`unit_id` by design. Can we exclude those so the filter only catches genuinely unparsed ingredients?
2. Related — since this joins on the one-to-many `recipe_ingredient` relationship, a recipe with multiple matching ingredient rows could show up as a duplicate in the results. Can we add `.distinct()` or confirm this isn't an issue?
3. Can we add test coverage for the new filter — normal parsed ingredients, unparsed ingredients, section headers, and recipe references?

michael-genson:
Moving this to draft until it can be picked back up

michael-genson:

PR #7444, opened 2026-04-11
fix: Update frontend refs

michael-genson:

_(REQUIRED)_

The Nuxt 4 upgrade changed some paths, and some of our hardcoded paths need to be changed as a result.

_(REQUIRED)_

N/A

## AI / LLM Assistance

_(REQUIRED)_

None


frontend/app/types/components.d.ts
```diff
@@ -16,7 +16,6 @@ import type BaseDivider from "@/components/global/BaseDivider.vue";
 import type BaseExpansionPanels from "@/components/global/BaseExpansionPanels.vue";
 import type BaseOverflowButton from "@/components/global/BaseOverflowButton.vue";
 import type BasePageTitle from "@/components/global/BasePageTitle.vue";
-import type BaseStatCard from "@/components/global/BaseStatCard.vue";
```
michael-genson, line 19: Unrelated, but I ran `dev:generate` to test and these got removed

frontend/app/types/components.d.ts
```diff
@@ -57,7 +56,6 @@ declare module "vue" {
     BaseExpansionPanels: typeof BaseExpansionPanels;
     BaseOverflowButton: typeof BaseOverflowButton;
     BasePageTitle: typeof BasePageTitle;
-    BaseStatCard: typeof BaseStatCard;
```
michael-genson, line 60: same here

Issue #7435, opened 2026-04-09
[BUG] ambigious usage of `BASE_URL` starting in `3.10.2` causes broken redirect URLs for OIDC


ToshY:
### First Check

- [x] This is not a feature request.
- [x] I added a very descriptive title to this issue (title field is above this).
- [x] I used the GitHub search to find a similar issue and didn't find it.
- [x] I searched the Mealie documentation, with the integrated search.
- [x] I already read the docs and didn't find an answer.
- [ ] This issue can be replicated on the demo site (https://demo.mealie.io/).

### What is the issue you are experiencing?

**Problem**

After upgrading from Mealie `3.9.2` to `3.14.0`, OIDC logins break if `BASE_URL` is set as recommended for notifications (e.g. `https://recipes.example.com`).

Introduced in https://github.com/mealie-recipes/mealie/pull/6995, when `BASE_URL` is set to the external application domain (e.g., `https://recipes.example.com`), the OIDC redirect URI generated during login by Mealie is malformed and **loses the subdomain**, resulting in OIDC requests like `redirect_uri=https://example.com/login` instead of `https://recipes.example.com/login`. The OIDC provider then returns an error or redirects to the wrong target, resulting in a failed login.

**Expected behavior**

When `BASE_URL` is set to the external origin (e.g., `https://recipes.example.com`) and the redirect URIs are configured in the OIDC provider as per the documentation, the OIDC redirect URI should use that value correctly, producing `redirect_uri=https://recipes.example.com/login` — including the full subdomain.

**Actual behavior**

The generated `redirect_uri` in the OIDC authorize request strips the subdomain:
```
https://auth.example.com/application/o/authorize/?...&redirect_uri=https%3A%2F%2Fexample.com%2Flogin&...
```

This does not match the redirect URIs configured in the OIDC provider per the documentation, causing the provider to reject the request.

**Screenshots / Logs**

OIDC authorize URL produced by Mealie:
```
https://auth.example.com/application/o/authorize/…&redirect_uri=https%3A%2F%2Fexample.com%2Flogin&…
```

OIDC provider error:
```
The provided redirect_uri is not valid for this client.
```

Mealie UI warning when `BASE_URL` is omitted:
```
BASE_URL is still the default value on API Server. This will cause issues with notifications links generated on the server for emails, etc.
```

## Root cause analysis

The bug was introduced in `3.10.2` in `mealie/routes/auth/auth.py`. The code now computes the OIDC redirect URL as follows:

[...]

PR #7449, opened 2026-04-12
docs: Enhance BASE_URL description in backend config

BadCo-NZ:
Updated BASE_URL description to include OIDC callback URL.

clarifies what the BASE_URL is used for

No issues

I recently ran into a problem with not being able to login with my OIDC provider due to invalid call back URL, turns out I did not have `https` on my `BASE_URL`. It took a while to diagnose because according to the docs, `BASE_URL` is unrelated to OIDC. Hopefully adding more to the description will help others in the future.

## AI / LLM Assistance

Copilot auto filled the commit title and message.


ToshY:
I have also encountered this problem in #7435 when updating Mealie. As mentioned there, the change to use `BASE_URL` also for OIDC was introduced in #6995, which is a breaking change (as you have experienced yourself). Updating the docs is a good first step, but it is not enough to resolve the root cause.

michael-genson:
Thank you!

PR #7468, opened 2026-04-15
fix: make PWA share target functional on Android Chrome

zdenek-stursa:

The `share_target` entry in the PWA manifest and the redirect middleware were already in place, but the feature has never worked end-to-end due to four independent bugs. This PR fixes all of them so that sharing a recipe URL from Chrome on Android directly opens Mealie and pre-fills the import form.

**Changes by file:**

- **`frontend/nuxt.config.ts`** — The manifest mapped the `text` Web Share field to `recipe_import_url`, but Chrome Android puts the shared page URL in the `url` field, not `text`. Both fields are now mapped to separate query params (`recipe_import_url` and `recipe_import_text`) so Chrome's behaviour is handled correctly and apps that share URLs as plain text keep working.

- **`frontend/app/middleware/pwa-share-target-redirect.global.ts`** — Changed `redirectCode` from `301` to `302` (the redirect depends on auth state and must not be cached). When the user is not logged in, the full path including the `recipe_import_url` query param is now preserved in a `?redirect=` parameter so the URL survives the login step.

- **`frontend/app/pages/login.vue`** — After a successful login the page now honours the `?redirect=` query param set by the middleware above, sending the user back to the import page with the recipe URL intact. For OIDC flows
  (where the provider reload clears the query string) the redirect target is saved to `sessionStorage` before navigating to the provider and restored after the callback. An open-redirect guard (`startsWith('/')`) is included.

- **`frontend/app/pages/g/[groupSlug]/r/create/url.vue`** — Two fixes:
  1. The `recipeUrl` computed getter now falls back to `recipe_import_text` when `recipe_import_url` is absent, but only after validating with `new URL()` that the value is actually an http/https URL (arbitrary shared text is ignored).
  2. Removed the auto-submit on mount. Previously the form would flash briefly and immediately fire the import, giving the user no chance to review or adjust import options (tags, categories, edit mode, ingredient parsing).  The URL is now pre-filled and the user confirms by clicking the button.  Legacy `use_keywords` and `edit` query params for the Bookmarklet are still applied on mount.

Closes https://github.com/mealie-recipes/mealie/discussions/7457

The `text` → `url` manifest change is the most impactful fix. The original 2023 commit (defdc82d) switched from `url` to `text` with a comment saying "If there are issues with sharing, uncommenting those lines might help solve the puzzle." — that puzzle is now solved: `url` is what Chrome populates for browser URL shares, while `text` covers app-to-app text sharing. Both are needed and mapped to different params.

The no-auto-submit change is a deliberate UX decision. The share target is the user's first real interaction with the import page, so they should see the options rather than have them hidden by an immediate import.

## Testing

Tested manually on Android (Chrome + Mealie installed as PWA):

1. **Happy path (logged in):** Share any recipe URL from Chrome → Mealie opens, import form appears with the URL pre-filled, import options visible, import starts after tapping the button. ✅

2. **Unauthenticated path:** Log out, share a URL → redirected to login, log in → redirected back to the import form with the URL still present. ✅

Platform note: PWA Share Target is not supported on iOS/Safari as of iOS 18. The Mealie icon simply will not appear in the iOS share sheet — no regression for existing users.

## AI / LLM Assistance

Claude (claude-sonnet-4-6) was used as a coding assistant throughout this PR. It helped with root-cause analysis of the four bugs, writing and reviewing the code changes, and structuring the commit messages. All code was reviewed, understood, and manually tested on a real Android device before submission.
The logic, decisions, and test results are my own.


Choromanski:
PR #7331 updates how the manifest is loaded.

I also have not been able to replicate the issue with sharing from Chrome on Android. It seems to import just fine.

Used [w3s demo](https://w3c.github.io/web-share-target/demos/sharetarget.html?utm_source=chatgpt.com) to validate that the URL is sent in the text field from chrome.

Choromanski:
I do like the idea of checking the URL first then falling back to text

zdenek-stursa:
Thanks for looking at this and for pointing to #7331 — that's important context.

**On PR #7331:** The new `manifest.py` carries over the same `"params": {"text": "recipe_import_url"}` mapping that this PR fixes. If #7331 lands first, the manifest field fix will need to be ported there (one line change). Happy to do that in a follow-up or rebase onto #7331 if that's preferred.

**On the `url` vs `text` field:** The W3C share-target demo at that link is a *sender* demo — it explicitly calls `navigator.share({ text: '…', title: '…' })` without populating the `url` field, so it validates the `text` path specifically. Chrome's native share sheet (the three-dot menu → Share while browsing a recipe page) does the opposite: it calls `navigator.share({ title: pageTitle, url: pageUrl })`, leaving `text` empty. You can verify this by checking the query string that arrives at `/r/create/url` — with the current `text`-only manifest, `recipe_import_url` will be empty for a native Chrome browser share.

Our fix maps both fields to separate params and falls back gracefully, so both scenarios work.

**On why "import just fine" with the original code:** There are two distinct failure modes depending on session state:

1. **Valid session (your test scenario):** The auto-submit fires immediately. The URL *does* arrive — but only when shared via the `text` path (e.g. from an app that shares URLs as plain text). A native Chrome browser share goes into `url`, which the original manifest ignores entirely, so `recipe_import_url` is blank and the import runs with an empty URL. Even when it does work, the user has no chance to configure tags, categories, or edit mode before the import starts.

2. **No valid session:** The middleware redirects to `/login`, but the `recipe_import_url` query param is silently dropped. After a successful login the user lands on the import page with a blank URL field. The import never starts and there is no indication why. This is the most common real-world path for a user who just installed Mealie as a PWA and immediately tries the share feature.

To reproduce situation 2: log out of Mealie completely, then share any recipe URL from Chrome to Mealie. The URL will be gone after login.

The no-auto-submit change is intentional and independent of the field mapping — the share target is often the user's first interaction with the import page, and silently firing an import without showing the options panel is a poor first impression.

Choromanski:
@zdenek-stursa Yes I understand, I'm just saying now that it has been merged you will need to resolve the merge conflicts

zdenek-stursa:
Conflicts resolved — the manifest field fix has been ported to `manifest.py` as part of the rebase onto `mealie-next`. All CI checks are passing. Thanks for the review and for pointing out #7331!

frontend/nuxt.config.ts
```diff
@@ -234,7 +234,155 @@ export default defineNuxtConfig({
       periodicSyncForUpdates: 120,
     },
     includeAssets: ["favicon.ico", "apple-touch-icon.png", "safari-pinned-tab.svg"],
-    manifest: false, // This is served via the backend, see mealie/routes/spa/manifest.py
```
Choromanski, line 237: This should be set back to false

Choromanski (changes requested):

zdenek-stursa:
Thanks for the catch! This slipped through during the rebase onto `mealie-next` — when we ported all the manifest fields to `manifest.py` to resolve the conflict with #7331, we forgot to revert `nuxt.config.ts` back to `false`. Fixed in the latest commit.

michael-genson (approved):
Thanks!

PR #7503, opened 2026-04-20
fix: pressing Enter in dialogs now confirms instead of silently closing

zdenek-stursa:

Pressing <kbd>Enter</kbd> inside a `BaseDialog` that uses the `can-confirm` prop would silently close the dialog without saving any data — indistinguishable from pressing <kbd>Escape</kbd>. This affected all create/edit dialogs built on `GroupDataPage` and `RecipeOrganizerPage`,including:

  - Labels, Units, Foods, Recipes (data pages) — create & edit dialogs
  - Recipe organizer pages (Categories, Tags, Tools) — update dialog
  - Recipe context menu — Duplicate recipe, Add to meal plan

**Root cause:** The `@keydown.enter` handler always called `submitEvent()` which emits `"submit"`. However, `can-confirm` dialogs listen for `@confirm`, not `@submit`. The dialog closed via an internal watcher (because `submitted` was set to `true`), but the `confirm` event was never fired, so no data was saved.

**Fix:** When `canConfirm` mode is active, Enter now behaves identically to clicking the Confirm button — emits `"confirm"` and closes the dialog. If the button is disabled (`submitDisabled`), Enter is also a no-op. Existing `canSubmit` dialogs are unaffected.

Changed file: `frontend/app/components/global/BaseDialog.vue`

No related issue.

## Testing

Manually tested create and edit dialogs for Labels, Units, Foods (data pages). Enter confirms, ESC still cancels, disabled Confirm button blocks Enter correctly.

## AI / LLM Assistance

This PR was developed with significant AI assistance (Claude). The bug was identified and analyzed by the LLM, which also proposed and implemented the fix. The fix was manually reviewed and tested by the human author.


michael-genson (approved):
Thanks! There was one edgecase which this fix broke, so I pushed a change, otherwise LGTM

PR #7521, opened 2026-04-22
fix: redirect to login and validate input on password reset flow

zdenek-stursa:

The password reset flow (`/forgot-password` and `/reset-password`) had two problems:

**No post-submission behavior on success:**
The submit button remained active after a successful submission, allowing the user to repeatedly click it and send multiple API requests in quick succession. The only feedback was a toast notification, which could easily be missed — leaving the user confused about whether the action succeeded.

**No input validation on `/forgot-password`:**
The form accepted any input including an empty field, silently sent the request and displayed a success toast regardless. While the backend correctly avoids revealing whether an email exists in the database, there is no reason to allow obviously invalid input through.

Changes:
- `frontend/app/pages/forgot-password.vue`: added client-side email format validation using the existing `validators.email` composable (no new translation keys). After a successful submission, navigates to `/login`.
- `frontend/app/pages/reset-password.vue`: after a successful password reset, navigates to `/login`.

No translation files were modified.

No existing issue. These are UX problems spotted during review of the password reset flow.

The redirect happens after `alert.success(...)` is called — Nuxt does not destroy the toast on navigation, so the user sees the confirmation message on the `/login` page.

The email validation uses the project's existing `validators.email` composable and `VForm` ref pattern (same approach as `pages/register/index.vue`). The backend anti-enumeration behavior is unaffected — it still returns 200 regardless of whether the email exists in the database.

## Testing

Manually tested:
- `/forgot-password`: empty submit → inline validation error, no request sent. Invalid email format → inline error. Valid email → toast appears on `/login`.
- `/reset-password`: submit with valid token → toast appears on `/login`. Button no longer clickable after submission.

## AI / LLM Assistance

This PR was developed with AI assistance (Claude Sonnet 4.6). The logic, approach, and validation strategy were designed and verified by the contributor.


michael-genson (approved):
Thanks! I pushed a couple minor changes

PR #7522, opened 2026-04-22
fix: redirect to new slug URL after recipe rename

zdenek-stursa:

When a user renames a recipe and saves it, the backend generates a new URL slug from the new name. However, the browser URL remained on the old (now invalid) slug after saving. Refreshing the page (F5) would result in a 404 error.

**Root cause:** A race condition between two concurrent Vue Router navigations:
1. `setMode(VIEW)` triggers the `isEditMode` watcher, which calls `paramsEdit.value = undefined` to remove `?edit=true` from the URL. The `useRouteQuery` setter schedules this via `nextTick(() => router.replace(...))` using the **current (old) path**.
2. `router.push(newSlug)` starts navigation to the new slug.
3. The watcher's `router.replace(oldSlug)` fires after the push and Vue Router cancels the in-progress navigation to the new slug.

Result: the URL reverts to the old slug without `?edit=true`, which is no longer valid after the rename.

**Changes in `RecipePage.vue`:**
- Move `recipe.value` and `originalRecipe.value` updates **before** navigation (was after), so `hasUnsavedChanges()` correctly returns false if any guard fires.
- Use `await nextTick(); await nextTick()` to let the `isEditMode` watcher's `router.replace` complete before our navigation runs.
- Replace `router.push` with `router.replace` so the old (now invalid) slug is not added to browser history — prevents a 404 when pressing the Back button.
- Only navigate when the slug actually changed (`data.slug !== route.params.slug`).

_(no existing issue found — this bug was discovered during testing)_

## Testing

Tested manually in Chrome:

- Rename a recipe and save → URL updates to the new slug immediately
- F5 after rename → page loads correctly at the new slug
- Back button after rename → navigates to the page before the recipe (not the old invalid slug)
- Save without changing the name → URL stays the same, no unnecessary navigation
- Open recipe with `?edit=true` URL, rename and save → URL updates to new slug without `?edit=true`

## AI / LLM Assistance

This PR was developed with assistance from Claude Code (claude-sonnet-4-6). The root cause analysis (race condition between Vue Router navigations) and the fix were identified and implemented with AI assistance. The fix was manually verified by the author in Chrome.

### DIFF
--- a/frontend/app/components/Domain/Recipe/RecipePage/RecipePage.vue
+++ b/frontend/app/components/Domain/Recipe/RecipePage/RecipePage.vue
@@ -358,10 +358,15 @@ async function saveRecipe() {
     setMode(PageMode.VIEW);
   }
   if (data?.slug) {
-    router.push(`/g/${groupSlug.value}/r/` + data.slug);
     recipe.value = data as NoUndefinedField<Recipe>;
-    // Update the snapshot after successful save
     originalRecipe.value = deepCopy(recipe.value);
+    if (data.slug !== route.params.slug) {
+      // Wait for the isEditMode watcher (which removes ?edit=true via router.replace)
+      // to run and complete before navigating, otherwise it would cancel our navigation.
+      await nextTick();
+      await nextTick();
+      router.replace(`/g/${groupSlug.value}/r/` + data.slug);
+    }
   }
 }
 
--- a/frontend/app/components/Domain/Recipe/RecipePage/RecipePage.vue
+++ b/frontend/app/components/Domain/Recipe/RecipePage/RecipePage.vue
@@ -336,8 +336,16 @@ onMounted(() => {
   }
 });
 
+// When set, the isEditMode watcher skips its URL cleanup because saveRecipe
+// is navigating to a new slug that naturally omits ?edit=true.
+const isNavigatingAfterRename = ref(false);
+
 watch(isEditMode, (newVal) => {
   if (!newVal) {
+    if (isNavigatingAfterRename.value) {
+      isNavigatingAfterRename.value = false;
+      return;
+    }
     paramsEdit.value = undefined;
   }
 });
@@ -355,16 +363,15 @@ watch(isParsing, () => {
 async function saveRecipe() {
   const { data, error } = await api.recipes.updateOne(recipe.value.slug, recipe.value);
   if (!error) {
+    if (data?.slug && data.slug !== route.params.slug) {
+      isNavigatingAfterRename.value = true;
+    }
     setMode(PageMode.VIEW);
   }
   if (data?.slug) {
     recipe.value = data as NoUndefinedField<Recipe>;
     originalRecipe.value = deepCopy(recipe.value);
     if (data.slug !== route.params.slug) {
-      // Wait for the isEditMode watcher (which removes ?edit=true via router.replace)
-      // to run and complete before navigating, otherwise it would cancel our navigation.
-      await nextTick();
-      await nextTick();
       router.replace(`/g/${groupSlug.value}/r/` + data.slug);
     }
   }


frontend/app/components/Domain/Recipe/RecipePage/RecipePage.vue
```diff
@@ -358,10 +358,15 @@ async function saveRecipe() {
...
   if (data?.slug) {
-    router.push(`/g/${groupSlug.value}/r/` + data.slug);
     recipe.value = data as NoUndefinedField<Recipe>;
-    // Update the snapshot after successful save
     originalRecipe.value = deepCopy(recipe.value);
+    if (data.slug !== route.params.slug) {
+      // Wait for the isEditMode watcher (which removes ?edit=true via router.replace)
+      // to run and complete before navigating, otherwise it would cancel our navigation.
+      await nextTick();
+      await nextTick();
```
michael-genson: Can we think of a better way to do this rather than relying on `nextTick`? I'm not convinced this is consistent enough.

michael-genson (changes requested):

frontend/app/components/Domain/Recipe/RecipePage/RecipePage.vue
zdenek-stursa: Thank you for the feedback! I'd like to explain why the double `nextTick` is actually deterministic in this context.

The root cause lies in how `useRouteQuery`'s setter is implemented — it internally schedules its `router.replace` call inside a `nextTick`:

```ts
set(v) {
  nextTick(() => {
    router.replace({ query: { ...route.query, [name]: v } });
  });
}
```

So when `setMode(PageMode.VIEW)` is called, the following happens in a precise, deterministic sequence:

1. `isEditMode` becomes `false` — Vue's scheduler queues the `watch(isEditMode)` callback.
2. **First `await nextTick()`** — Vue flushes its scheduler, which runs the watcher. The watcher sets `paramsEdit.value = undefined`, which in turn queues *another* `nextTick` (inside `useRouteQuery`'s setter) for a `router.replace` with the old slug.
3. **Second `await nextTick()`** — that inner `router.replace(oldSlug)` from `useRouteQuery` runs and completes.
4. Only *then* we call `router.replace(newSlug)`, so nothing can cancel it.

Each `await nextTick()` corresponds exactly to one level of async indirection in the chain. It's not a guess or a workaround — it's a direct consequence of how Vue's scheduler batches reactive updates.

That said, I completely understand the concern — two chained `nextTick` calls without context look fragile. Could you let me know what specifically worries you? For example:
- Is it readability / understandability of the timing?
- Is there a Vue version or environment where you think the ordering could differ?
- Or would you prefer to change `useRouteQuery`'s setter to call `router.replace` synchronously (dropping its internal `nextTick`), which would reduce this to a single `await nextTick()`?

Happy to iterate on whichever direction makes the most sense to you.

michael-genson: > Each await nextTick() corresponds exactly to one level of async indirection in the chain. It's not a guess or a workaround — it's a direct consequence of how Vue's scheduler batches reactive updates.

This assumes the underlying framework never changes, and that we never modify the existing logic around this. Sure, it works _now_, but it's fragile since any number of unrelated changes can break this. It would be better to come up with a direct fix for the race condition rather than relying on arbitrary framework implementation.

zdenek-stursa: Fair point — you're right that it creates an implicit dependency on implementation details that could silently break.

I've refactored to eliminate the timing dependency entirely (commit 384d4460). Instead of waiting for async side effects, the fix now uses an explicit guard flag:

- Before calling `setMode(PageMode.VIEW)`, `saveRecipe` sets `isNavigatingAfterRename = true` when the slug has changed.
- The `isEditMode` watcher checks the flag: if set, it resets the flag and returns early — the URL cleanup is skipped because the new slug URL naturally omits `?edit=true`.
- No `nextTick` involved at all.

This is a direct fix for the race condition: only one router navigation happens, and it's fully deterministic regardless of Vue scheduler internals or future changes to `useRouteQuery`.

michael-genson (approved):
Thank you!

PR #7537, opened 2026-04-25
fix: prevent double-scaling of sub-recipe ingredients in shopping list

### LINKED ISSUE #7518: [BUG] - Adding recipe to shopping list scales sub-recipes twice
### First Check

When adding a recipe to a shopping list, if it contains another recipe as an ingredient (a sub-recipe), then the sub-recipe will be scaled by the square of the quantity (it gets scaled twice).

### Steps to Reproduce

1) Create a recipe, e.g. "cheese sauce" with an ingredient amount, e.g. "1000 gram cheese".
2) Create another recipe, e.g. "cheesy pasta" using "cheese sauce" as an ingredient, and 0.5 as the amount.
3) Add "cheesy pasta" to a shopping list. 
4) The dialog will add "250 gram cheese" to the shopping list, not the expected "500 gram cheese"

None applicable.

### Mealie Version

v3.16.0 [5fc4851ef5261be3cee131ab17e5532fc4820f14](https://github.com/mealie-recipes/mealie/commit/5fc4851ef5261be3cee131ab17e5532fc4820f14)

### Deployment

Other (please specify below)

Podman on Linux

zdenek-stursa:

_(REQUIRED)_

When adding a recipe that references another recipe as an ingredient (sub-recipe) to a shopping list, the sub-recipe's ingredient quantities were being scaled twice:

1. Once manually in `addSubRecipeToMap` by pre-multiplying the quantity into the ingredient object (`quantity: (ing.quantity || 1) * (subIng.quantity || 1)`)
2. A second time automatically by the template via the `recipeScale` prop passed to `RecipeIngredientListItem`

**Fix:** Remove the manual pre-multiplication — the scale is already correctly applied via `recipeSection.recipeScale` in the template.

**Also fixed:** After saving a recipe whose slug changed, the router navigation was being cancelled by the `isEditMode` watcher running `router.replace` first. Fixed by awaiting two `nextTick()` calls to let the watcher complete before navigating.

Changed files:
- `frontend/app/components/Domain/Recipe/RecipeDialogAddToShoppingList.vue` — remove double-scale of sub-recipe ingredient quantity
- `frontend/app/components/Domain/Recipe/RecipePage/RecipePage.vue` — fix navigation after recipe save when slug changes
- `tests/integration_tests/user_household_tests/test_group_shopping_lists.py` — add regression test for the double-scaling bug

_(REQUIRED)_

Fixes #7518

## Testing

Regression test added: `test_shopping_lists_add_nested_recipe_ingredients_with_scale`

The test creates a sub-recipe with 1000g of an ingredient and a parent recipe that references it with quantity 0.5. It verifies that the resulting shopping list item has quantity 500 (1000 × 0.5), not 250 (double-scaled).

Manually tested in Docker with a real recipe containing a sub-recipe reference — quantities in the shopping list dialog now match the expected values.

## AI / LLM Assistance

_(REQUIRED)_

This fix was developed with the assistance of Claude Sonnet (claude-sonnet-4-6) via Claude Code CLI. The root cause analysis, code changes, and regression test were written with AI assistance. The fix was reviewed and manually tested by the author.

### DIFF
--- a/frontend/app/components/Domain/Recipe/RecipeDialogAddToShoppingList.vue
+++ b/frontend/app/components/Domain/Recipe/RecipeDialogAddToShoppingList.vue
@@ -321,7 +321,7 @@ async function consolidateRecipesIntoSections(recipes: RecipeWithScale[]) {
         const householdsWithFood = subIng.food?.householdsWithIngredientFood || [];
         ownIngs.push({
           checked: !householdsWithFood.includes(currentHouseholdSlug.value),
-          ingredient: { ...subIng, quantity: (ing.quantity || 1) * (subIng.quantity || 1) },
+          ingredient: subIng,
         });
       }
     }


frontend/app/components/Domain/Recipe/RecipePage/RecipePage.vue
```diff
@@ -358,10 +358,15 @@ async function saveRecipe() {
...
     originalRecipe.value = deepCopy(recipe.value);
+    if (data.slug !== route.params.slug) {
+      // Wait for the isEditMode watcher (which removes ?edit=true via router.replace)
+      // to run and complete before navigating, otherwise it would cancel our navigation.
+      await nextTick();
+      await nextTick();
+      router.replace(`/g/${groupSlug.value}/r/` + data.slug);
+    }
   }
 }
```
michael-genson, line 366: Is this needed for this change or is this leftover from your other slug update PR?

michael-genson (changes requested):

frontend/app/components/Domain/Recipe/RecipePage/RecipePage.vue
zdenek-stursa: You're right — that change belongs to #7522, not here. I've removed it from this PR (force-pushed). Sorry for the noise.

michael-genson (changes requested):
This looks good, only question - was the backend test for another issue, or was it intended for this bug? Because the fix is in the frontend the backend test doesn't serve any purpose here.

If there's a need for the backend test that's fine, but otherwise I'd rather remove it since I'm pretty sure this scenario is caught in existing tests already.

zdenek-stursa:
Good catch — you're right, the test was unnecessary. The bug is purely in the frontend (the Vue component was pre-multiplying quantities before the template applied `recipeScale` again), so a backend integration test doesn't validate it and the backend was never broken. I've removed the test and force-pushed. The PR now contains only the one-line frontend fix.

michael-genson (approved):
Thanks!

PR #7542, opened 2026-04-26
feat: show full recipe title in tooltip on recipe cards

zdenek-stursa:

Recipe card titles are often truncated (ending with `…`) because card width is limited. This is especially frustrating when multiple recipes share a common prefix and differ only in the part that gets hidden — making it impossible to distinguish them without clicking into each recipe.

This PR adds a `v-tooltip` to the recipe card title in both card view variants (large grid cards and compact list cards), so hovering over a truncated title reveals the full name after a short delay.

Changes:
- `frontend/app/components/Domain/Recipe/RecipeCard.vue` — wrapped `<v-card-title>` with Vuetify `v-tooltip` (large card variant)
- `frontend/app/components/Domain/Recipe/RecipeCardMobile.vue` — wrapped `<v-list-item-title>` with Vuetify `v-tooltip` (compact card variant)

Both tooltips use `open-delay="700"` to avoid the tooltip triggering while the user is simply moving the cursor across the list. The existing Vuetify `v-tooltip` component is used — no new dependencies introduced.

**Note for maintainers:** The `open-delay` is currently set per component. A cleaner long-term solution would be to add a global default in `nuxt.config.ts` under the `vuetify.vuetifyOptions.defaults` section — e.g. `VTooltip: { openDelay: 700 }` — so the delay is consistent across the whole app without repeating it at every usage site. We're happy to adjust this PR if that direction is preferred.

No existing issue. This is a small UX improvement.

## Result
Snímek obrazovky 2026-05-01 154605
Snímek obrazovky 2026-05-01 154630

## Testing

Manually tested on the group recipe listing page (`/g/<group>/`) in both card view modes:
- Tooltip appears after ~700 ms of hovering over a recipe title
- Tooltip shows the full recipe name regardless of truncation
- No visual regressions observed in card layout or hover effects

## AI / LLM Assistance

This PR was developed with assistance from Claude (Sonnet 4.6) via Claude Code. The implementation approach, component selection, and code changes were reviewed and guided by the human author. The use of the existing Vuetify `v-tooltip` pattern was intentional — matching conventions already present in the codebase.

### DIFF
--- a/frontend/app/components/Domain/Recipe/RecipeCard.vue
+++ b/frontend/app/components/Domain/Recipe/RecipeCard.vue
@@ -36,9 +36,13 @@
             </div>
           </v-expand-transition>
         </RecipeCardImage>
-        <v-card-title class="mb-n3 px-4" style="font-size: 1.25rem;">
-          {{ name }}
-        </v-card-title>
+        <v-tooltip location="bottom" :text="name">
+          <template #activator="{ props: tooltipProps }">
+            <v-card-title v-bind="tooltipProps" class="mb-n3 px-4" style="font-size: 1.25rem;">
+              {{ name }}
+            </v-card-title>
+          </template>
+        </v-tooltip>
 
         <slot name="actions">
           <v-card-actions
--- a/frontend/app/components/Domain/Recipe/RecipeCard.vue
+++ b/frontend/app/components/Domain/Recipe/RecipeCard.vue
@@ -36,7 +36,7 @@
             </div>
           </v-expand-transition>
         </RecipeCardImage>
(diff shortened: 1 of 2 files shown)


michael-genson (changes requested):
Can we only wrap in a tooltip when `scrollWidth > clientWidth`? Right now it shows on every hover even when the title isn't truncated.

Also, let's go ahead with moving `open-delay="700"` into a global `VTooltip` default in `nuxt.config.ts` per your note, rather than repeating it per component.

michael-genson:
Moving this to draft until it can be picked back up

michael-genson:

PR #7564, opened 2026-04-30
fix: restore create-item button in recipe dropdowns (categories, tags, tools)

### LINKED ISSUE #7557: [BUG] - Cannot create new categories/tools/keywords from within a recipe
### First Check

So i've updated my docker to the last version of mealie and in the meantime I've migrated from sqlite to postgres (the migration was a bliss). Everything works fine but one part is missing : 

When I create or edit a recipe, I'm no longer able to add new categories/tool/keyword from the recipe interface (I was able to do that before). I can still add new ingredients or units from the recipe, and i can still add new tools or categories from the dedicated menus.

What I expect is what i had before and similar to ingredients, a scrolling menu with a +new/create button : 

Image

What I get is a scroll menu without a +new/create button:

Image

 Idk if it's due to my switch to postgres...

### Steps to Reproduce

There is the same issue on the demo site

I have no logs

### Mealie Version

Version: v3.16.0
Build: 5fc4851ef5261be3cee131ab17e5532fc4820f14
Application mode : production
Database : postgres
Scrapper : 15.11.0

### Deployment

Other (please specify below)

Podman with quadlet for system-d (essentially docker linux)

zdenek-stursa:

_(REQUIRED)_

The \"Create\" button (green `+` button at the bottom of the dropdown list) was missing from the Categories, Tags, and Tools autocomplete fields on the recipe edit/create page. This was a regression introduced during the Nuxt 4 upgrade (#7426), where `RecipeOrganizerSelector` was migrated without the `#append-item` slot that renders the create button.

Changes per file:

- **`frontend/app/components/Domain/Recipe/RecipeOrganizerSelector.vue`**
  - Added `#append-item` slot with a `BaseButton` that creates a new item directly via the store (same pattern as `RecipeIngredientEditor`)
  - Added `#no-data` slot showing "Press Enter to Create" hint when no results are found
  - Added `@keyup.enter` handler that creates the item on Enter when no exact match exists
  - Added `handleEnter()` and `createItem()` functions
  - The create button is only shown when the search field is non-empty (prevents creating empty items)

- **`frontend/app/components/Domain/Recipe/RecipeIngredientEditor.vue`** _(bonus fix for pre-existing bug)_
  - The Create button for foods and units was always visible even with an empty search field or when the typed string exactly matched an existing item — both cases could silently create duplicate or empty records
  - Added `showCreateFood` and `showCreateUnit` computed properties that gate the button on: non-empty search AND no exact case-insensitive match in the filtered list

Before / after for organizer dropdowns:

| Before | After |
|--------|-------|
| No create button in Categories/Tags/Tools dropdowns | Green Create button appears when typing a term that doesn't exist |
| Typing a non-existent term showed an empty list with no way to create inline | "Press Enter to Create" hint shown; Enter key creates and selects the item |

Snímek obrazovky 2026-04-30 143414

_(REQUIRED)_

Fixes #7557

The `createItem()` function in `RecipeOrganizerSelector` uses `// @ts-expect-error` (same pattern already used in `RecipeOrganizerDialog`) because the shared `storeMap` covers organizer types with different required fields, and TypeScript cannot narrow the type in this context.

The Enter-to-create handler checks for an exact case-insensitive match against the full item list before creating, so it will not create a duplicate if the user types an exact existing name.

## Testing

Tested manually on the recipe edit page:
- Categories, Tags, Tools: typing an unknown term shows the Create button and "Press Enter to Create" hint; both the button click and Enter key create and immediately select the new item
- Categories, Tags, Tools: typing an existing term hides the Create button
- Ingredients (foods, units): Create button now hidden when field is empty or when typed term exactly matches an existing item

## AI / LLM Assistance

This pull request was developed with significant assistance from Claude (Anthropic). The bug was diagnosed and the fix was implemented using Claude Code. The human contributor reviewed, tested, and verified all changes before submission.

### DIFF
--- a/frontend/app/components/Domain/Recipe/RecipeIngredientEditor.vue
+++ b/frontend/app/components/Domain/Recipe/RecipeIngredientEditor.vue
@@ -88,7 +88,7 @@
             </div>
           </template>
           <template #append-item>
-            <div class="px-2">
+            <div v-if="unitSearch" class="px-2">
               <BaseButton
                 block
                 size="small"
@@ -147,7 +147,7 @@
             </div>
           </template>
           <template #append-item>
-            <div class="px-2">
+            <div v-if="foodSearch" class="px-2">
               <BaseButton
                 block
                 size="small"
--- a/frontend/app/components/Domain/Recipe/RecipeIngredientEditor.vue
+++ b/frontend/app/components/Domain/Recipe/RecipeIngredientEditor.vue
@@ -88,7 +88,7 @@
             </div>
           </template>
           <template #append-item>
-            <div v-if="unitSearch" class="px-2">
+            <div v-if="showCreateUnit" class="px-2">
               <BaseButton
                 block
                 size="small"
@@ -147,7 +147,7 @@
             </div>
           </template>
           <template #append-item>
-            <div v-if="foodSearch" class="px-2">
+            <div v-if="showCreateFood" class="px-2">
               <BaseButton
                 block
                 size="small"
@@ -344,6 +344,11 @@ const foodData = useFoodData();
 const foodAutocomplete = ref<HTMLInputElement>();
 const { search: foodSearch, filtered: filteredFoods } = useSearch(foodStore.store);
 
+const showCreateFood = computed(() =>
+  !!foodSearch.value
+  && !filteredFoods.value.some((f: any) => (f.name ?? "").toLowerCase() === foodSearch.value.toLowerCase()),
+);
+
 async function createAssignFood() {
   foodData.data.name = foodSearch.value;
   model.value.food = await foodStore.actions.createOne(foodData.data) || undefined;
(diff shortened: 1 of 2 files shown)


frontend/app/components/Domain/Recipe/RecipeIngredientEditor.vue
```diff
@@ -344,6 +344,11 @@ const foodData = useFoodData();
 const foodAutocomplete = ref<HTMLInputElement>();
 const { search: foodSearch, filtered: filteredFoods } = useSearch(foodStore.store);
 
+const showCreateFood = computed(() =>
+  !!foodSearch.value
+  && !filteredFoods.value.some((f: any) => (f.name ?? "").toLowerCase() === foodSearch.value.toLowerCase()),
```
michael-genson, line 349: Do we have a normalization function somewhere? Might be more effective than lowercase. If not, this is fine

frontend/app/components/Domain/Recipe/RecipeIngredientEditor.vue
```diff
@@ -376,6 +381,11 @@ const unitsData = useUnitData();
 const unitAutocomplete = ref<HTMLInputElement>();
 const { search: unitSearch, filtered: filteredUnits } = useSearch(unitStore.store);
 
+const showCreateUnit = computed(() =>
+  !!unitSearch.value
+  && !filteredUnits.value.some((u: any) => (u.name ?? "").toLowerCase() === unitSearch.value.toLowerCase()),
```
michael-genson, line 386: Same here

frontend/app/components/Domain/Recipe/RecipeIngredientEditor.vue
michael-genson: Nevermind I realize this isn't what I thought it was, it wouldn't make sense to normalize here

michael-genson (approved):
Thanks!

PR #7573, opened 2026-05-05
feat: add per-recipe unit system conversion (metric / imperial / US)

agorton:

Implements the unit-conversion request from discussion #3881. Recipes can be viewed in metric, imperial (UK), or US-customary units, and instruction temperatures are converted between °F and °C. **Conversion is display-only** — the database stores authored values verbatim. **Same-dimension only** (mass↔mass, volume↔volume); volume↔weight (cups↔grams) is explicitly out of scope, as is shopping-list aggregation.

### Backend
- New `UnitSystem` enum (`original | metric | imperial | us`) and a `RecipeUnitConverter` service. The picker switches on the existing `StandardizedUnitType` and selects a display unit per dimension + system + magnitude (e.g. `≥1000 g → kg`, `≥453.6 g → lb`). Math goes through the existing Pint-backed `UnitConverter`.
- Imperial volumes intentionally **skip `imperial_cup`** — Pint defines it as 284 ml, but UK/AU recipes commonly use 250 ml. Imperial volumes below an imperial pint render in ml, which matches UK home-cooking convention. Pint also lacks `imperial_tablespoon`/`imperial_teaspoon`.
- `convert_instruction_temperatures` rewrites °F↔°C in instruction text. The regex has a trailing word-boundary guard (`(?![A-Za-z])`) so things like \"vitamin C\" or \"350Free\" don't match. Ranges like \"350-375°F\" are supported.
- New endpoint `GET /api/recipes/{slug}/conversions?system=...` returns the converted ingredient list and instructions. Multi-tenant via existing `RecipeController`. No DB writes.
- Migration adds `users.preferred_unit_system` (nullable) and `household_preferences.default_unit_system` (default `'original'`). `UpdateHouseholdPreferences` and `UserBase` carry the new fields; the existing `/users/self` PUT and `/households/preferences` PUT round-trip them via `auto_init` / `model_dump`.

### Frontend
- `useUnitSystem()` composable resolves user pref → household pref → `original`.
- `RecipeUnitSystemToggle` mirrors the scale button and lives in `RecipePageScale`. Hidden in edit mode so authored values stay visible while editing.
- `RecipePage` introduces a `displayedRecipe` computed that swaps in converted ingredients/instructions for view-only sites (cook mode, print, ingredient list). Edit components keep the canonical recipe so writes still persist.
- New v-selects in `HouseholdPreferencesEditor` and the user profile page; en-US strings only (Crowdin manages other locales).

### Known limitations (called out, not fixed in v1)
- Mealie's standardization layer maps every \"cup\" string to `StandardizedUnitType.CUP`, which Pint resolves as US-customary (1/16 US gallon = 236.59 ml). A UK user authoring \"1 cup\" is canonically 236.59 ml, so converting that to imperial UK will be slightly off. Out of scope here; documented for follow-up.
- Display unit is picked from the unscaled magnitude. At very large scales (e.g. 10×) the chosen unit can feel awkward (\"1000 g\" instead of \"1 kg\"). Most scaling is ≤4×; acceptable for v1.

Closes #3881 (discussion).

- Sanity-check the picker rules in `mealie/services/recipe/recipe_unit_conversion.py` — particularly the imperial volume rule (skip `imperial_cup`, ml below an imperial pint).
- The temperature regex: feedback welcome on rounding (integer if input was integer, else 1 decimal) and on the imperial→°C convention (UK kitchens use °C).
- Frontend approach uses a `displayedRecipe` computed clone in `RecipePage.vue` so cook mode, print, and the ingredient list inherit converted values without per-component prop changes.

## Testing

- 59 unit tests in `tests/unit_tests/services_tests/recipe/` — every `StandardizedUnitType × UnitSystem` pair, threshold edges (999 g vs 1000 g, oz↔lb, ml↔l), `imperial_cup` skip guard, `°F↔°C` round-trips, range temperatures, word-boundary negatives (\"vitamin C\", \"350Free\"), passthrough for missing units.
- Integration tests in `tests/integration_tests/user_recipe_tests/test_recipe_convert_units.py` for the new endpoint (auth, scope, full round-trip per system, invalid system → 422).
- Extended `test_household_perferences.py` and `test_user_crud.py` for the new pref fields.
- Migration covered by the existing `tests/unit_tests/test_alembic.py` harness (SQLite + Postgres).

Run locally:
\`\`\`
task py:check
task ui:check
\`\`\`

Manual verification done: recipe with `2 cups flour`, `8 oz butter`, instruction \"Bake at 350°F for 30 minutes\". Toggling Metric → cups become ml, oz become g, instruction shows ~177°C. Toggling US, Imperial, Original works as expected. Toggle hidden in edit mode. Household default + per-user override flow works.

## AI / LLM Assistance

This PR was authored with substantial assistance from Claude Code (Anthropic Claude Opus 4.7). The model wrote the implementation, tests, and most of the prose in this description, working from a plan I co-designed and reviewed with it. I reviewed every change and ran the test suite locally before opening the PR.


Xyolyp:
Hi @agorton — I wanted this feature too and started looking at it, found your PR, and decided to build on it rather than beside it. Your picker design and the preference model are the shape I ended up keeping. I have a branch with a variant on top: [`Xyolyp:feat/unit-conversion-3881-fe`](https://github.com/Xyolyp/mealie/compare/mealie-next...Xyolyp:mealie:feat/unit-conversion-3881-fe). It is off current `mealie-next`, so a PR into your branch would show all 344 commits of drift — hence a comment rather than a PR. Happy to open one wherever is useful to you.

Two things I hit that affect your version as it stands:

**The temperature regex matches cups.** `(?P<lo>\d+...)\s*°?\s*(?P<unit>[FCfc])(?![A-Za-z])` accepts a bare lowercase `c`, and `c` is the seeded abbreviation for *cup*. `"Add 2 c water"` becomes `"Add 35.6°F water"`, and `"1/2 c sugar"` matches on the `2`. Requiring uppercase unless a degree sign is present fixes it while still accepting `350F`, `350 F`, `180°C` and `4°c`.

**The unscaled-magnitude limitation goes away if the conversion moves to the client.** You called this out as acceptable for v1 ("1000 g instead of 1 kg at 10x"). Since `standardQuantity`/`standardUnit` already ship with every ingredient in the recipe payload, the client has everything it needs — and it also knows the scale. Picking the display unit from `quantity * standardQuantity * scale` while returning the quantity *unscaled* means the existing `useParsedIngredientText(ingredient, scale)` call sites keep working untouched, and 10× a 100 g ingredient reads "1 kg". That also drops the endpoint, the `RecipeConversionResponse` schema, the shared Pint registry and the per-toggle round trip; cook mode and print inherit conversion for free. The eight canonical units are two dimensions of pure arithmetic, so Pint is not really earning its keep on this path. Your `_pick_mass`/`_pick_volume` rules ported over more or less verbatim; the one change is that US volumes run on cups all the way up ("8 1/2 cups" rather than "4 pints") with tbsp/tsp below a quarter cup.

One addition: temperature became its own preference (`system | celsius | fahrenheit`). It is the request that keeps coming back in #3881 — @Spinkman's "I like my temps in F and measurements in metric" — and it is cheap once the two are decoupled.

Kept from yours: display-only, same-dimension-only, `users.preferred_*` nullable with a household default, the toggle next to the serving scaler, hidden in edit mode, `imperial` skipping `imperial_cup` for the 284 vs 250 ml reason you documented.

State: 25 files, +1033/−16, eslint clean, 318 frontend tests (51 new, covering every canonical-unit × system pair, threshold edges, scale awareness, passthrough for unconvertible units, and the temperature false positives above), mypy clean, backend suite green. Note that `test_admin_household_actions` needs a touch — it looks every response preference key up in a hand-written payload, so any new preference field breaks it with a `KeyError`.

Would you rather fold this into #7573 yourself, or should I open a fresh PR that credits you as co-author? Your call — it is your groundwork and I would rather not fragment a thread that has been waiting since 2024.

*Disclosure: written with substantial assistance from Claude Code (Opus 5). I reviewed the change and ran the suites locally.*

agorton:
@Xyolyp happy for you to take over this one. I don’t think I have time to work on this much

Xyolyp:
@agorton thank you — both for the offer and for the groundwork. Your picker rules, the preference model and the toggle placement are what I built on. Working out where the thresholds should sit was the part that took the actual thinking, and I mostly got to inherit it. You are on the commit as co-author.

Opened as #8202. The one substantive departure is that the conversion moved to the client: the standardization data is already in the recipe payload, which removes the endpoint and, as a side effect, fixes the unscaled-magnitude limitation you documented — the converter gets handed the scale, so 10× a 100 g ingredient reads "1 kg".

Would you like to look it over before a maintainer does? You know this problem space better than anyone else on the thread, and I would rather have your eyes on the threshold choices than not — particularly US volumes, where I kept cups running all the way up ("8 ½ cups") instead of switching to pints above 473 ml the way you did. I am genuinely unsure which reads better to a US cook. No obligation at all if you are short on time; you have already done the hard part.

Feel free to close this one whenever suits you.

michael-genson:
Replaced by https://github.com/mealie-recipes/mealie/pull/8202

PR #7591, opened 2026-05-09
feat: add recipe note linking to recipe instructions

jallier:

Adds the ability to link recipe notes to specific recipe steps, similar to how the existing ingredient to step linking already works. When viewing a recipe, users can click an icon on the step to show the notes linked to a step via a slide up bottom sheet.

This makes it easier for users to reference their notes when they are relevant for a specific step, as well as showing that some steps have relevant notes that you might want to check. For an example, [this recipe](https://www.recipetineats.com/slow-cooked-shredded-beef-ragu-pasta/#recipe) has a bunch of notes which apply to specific steps, so having an easier way to show and reference them is nice.

### Screenshots:

Main recipe view (desktop)
image

Notes sheet opened on click
image

Note linking dialog on edit screen (also shows icon on card in the background)
image

Mobile example
image

### Changes:

  **Backend**                                                                                                                          
  - New `recipe_note_ref_link` table joining `recipe_instructions` to notes, with an Alembic migration.
  - New `reference_id` (GUID) column on `notes`, indexed and backfilled for existing rows. Notes are referenced from steps by            
  `reference_id` rather than by primary key, mirroring how `recipe_ingredients.reference_id` already works. This is to match how the ingredient link works to keep it consistent.
  - Added `note_references` relationship on `RecipeInstruction` with `cascade="all, delete-orphan".
  - Service-layer orphan cleanup (`_remove_non_existent_note_references`) on recipe update, matching the existing ingredient-reference cleanup.
  - `RecipeNote` schema switched to `MealieModel` so `reference_id` serializes as `referenceId` through requests.

  **Frontend**
  - `RecipeNoteLinkerDialog`: per-step dialog for selecting which notes to attach. Shows the step's text at the top so you know what you're linking to, same as the ingredient linker.
  - `RecipeNoteBottomSheet`: bottom sheet for viewing notes linked to a step.
  - `RecipePageInstructions` integrates both, plus a note-linker icon button on each step.
  - API types updated (`NoteReference`, `RecipeNote.referenceId`).   

No issue, new feature.

This my first try at Vue and python (familiar with React and php primarily), so there might be stuff i'm missing. I'm happy to fix anything needed to get it across the line.

## Testing

- Added a repository test covering note references being saved, retrieved, and orphan-cleaned on recipe update
  (`tests/unit_tests/repository_tests/test_recipe_repository.py`). 
- Updated existing tests to handle the extra reference field on the note
- Manually tested everything through the UI in dev: linked/unlinked notes from steps in the editor, saved, reopened, confirmed the linked notes render in the bottom sheet. Verified that deleting a note cleans up dangling references on next save.          
- Migration tested by running `alembic upgrade head` against an existing dev DB (notes get backfilled `reference_id`s) and `alembic  downgrade -1` (column and link table removed cleanly).
- Tested on mobile and desktop

## AI / LLM Assistance

I used Claude Code to assist with brainstorming and code writing. All code has been manually reviewed and tested by myself.


michael-genson (changes requested):
I like the idea in general, but I think the current approach is too cumbersome. Comparing to how ingredient linking already works in this file, a few things would simplify this:
- Drop the bottom sheet and render linked notes inline like ingredient references already do, as a side column in cook mode, instead of a new icon and popup.
- Move the note-linker dialog into the existing ingredient-linker dialog code at the top of RecipePageInstructions.vue instead of a new component, since it's the same checkbox and Save & Next pattern.
- See if you can find a way to merge "Link Ingredients" and "Link Notes" into a single "Link references" menu entry with two sections, rather than two separate menu items and dialogs.

jallier:
Thanks for the review! I appreciate you taking the time to do so.

I've updated as you have asked, and rebased against the current `mealie-next` branch, fixing the conflicts at the same time - hope that isn't too annoying to re-review for you.

I think combining the reference items dialogs and shifting the notes into the left side when using cook mode are good suggestions. The only thing I'm not a huge fan of is that it doesn't make it super obvious if there is a linked note when you aren't using cookmode (which I do often, but maybe i'm just holding it wrong haha). However, i'm not going to argue about it and will defer to your preference here.

Updates:
Removed the linked notes from regular mode. Here's an example showing the different states of steps with linked notes, ingredients and no linked references:
image

The dialogs have been combined into a single one that shows both the ingredients and notes together:
image

michael-genson:
Yeah you're right now you really can't see it unless you're in cook mode. I'd say go ahead and restore how you had it originally in view mode, it's less intrusive than I initially thought. I think the updated editor is much better!

frontend/app/lang/messages/en-US.json
```diff
@@ -480,11 +480,16 @@
...
+    "note-linker": "Note Linker",
     "unlinked": "Not linked yet",
     "linked-to-other-step": "Linked to other step",
     "auto": "Auto",
     "cook-mode": "Cook Mode",
     "link-ingredients": "Link Ingredients",
+    "link-notes": "Link Notes",
+    "link-references": "Link references",
+    "linked-notes": "Linked Notes",
+    "no-notes-to-link": "No notes to link",
```
michael-genson: Are all of these still used or did you drop any with the refactor?

mealie/alembic/versions/2026-05-19-00.00.00_a1b2c3d4e5f6_add_recipe_note_ref_link.py
```diff
@@ -0,0 +1,62 @@
+"""add recipe_note_ref_link
+
+Revision ID: a1b2c3d4e5f6
+Revises: 2187537c52b8
+Create Date: 2026-05-19 00:00:00.000000
```
michael-genson: It looks like this was written manually, can you delete this file and regenerate the migration? Be sure to keep the code for populating the new `reference_id` value.

mealie/alembic/versions/2026-05-19-00.00.00_a1b2c3d4e5f6_add_recipe_note_ref_link.py
```diff
@@ -0,0 +1,62 @@
...
+
+    bind = op.get_bind()
+    notes_table = sa.Table(
+        "notes",
+        sa.MetaData(),
+        sa.Column("id", sa.Integer, primary_key=True),
+        sa.Column("reference_id", mealie.db.migration_types.GUID()),
+    )
+    for row in bind.execute(sa.select(notes_table.c.id)).fetchall():
+        bind.execute(notes_table.update().where(notes_table.c.id == row.id).values(reference_id=str(uuid.uuid4())))
```
michael-genson: Can this be done as a batch-op instead of row by row with multiple executes?

mealie/db/models/recipe/note.py
```diff
@@ -11,7 +11,9 @@ class Note(SqlAlchemyBase):
     recipe_id: Mapped[GUID | None] = mapped_column(GUID, sa.ForeignKey("recipes.id"), index=True)
     title: Mapped[str | None] = mapped_column(sa.String)
     text: Mapped[str | None] = mapped_column(sa.String)
+    reference_id: Mapped[GUID | None] = mapped_column(GUID, index=True, default=GUID.generate)
 
-    def __init__(self, title, text) -> None:
+    def __init__(self, title, text, reference_id=None, **_) -> None:
         self.title = title
         self.text = text
+        self.reference_id = reference_id or GUID.generate()
```
michael-genson: I don't think this is needed since you already specified a default on the field

mealie/alembic/versions/2026-05-19-00.00.00_a1b2c3d4e5f6_add_recipe_note_ref_link.py
jallier: Ah yeah, sorry about that, I tried to just update the head revision to point to the newer one - have regenerated it properly and updated the uuid generation to use a single update.

The updated version will make a list of all the UUIDs first, which I guess could be a problem if users have a huge amount of notes. Not sure if its worth complicating it out by chunking it though. Happy to update if you think so

frontend/app/lib/api/types/recipe.ts
```diff
@@ -269,12 +269,17 @@ export interface RecipeTool {
   slug: string;
   householdsWithTool?: string[];
 }
+export interface NoteReference {
+  referenceId?: string | null;
+}
+
```
michael-genson: Was this regenerated with `task dev:generate`? Nothing else in this file has a blank line between interfaces, so it looks like it might have been added by hand.

frontend/app/components/Domain/Recipe/RecipePage/RecipePageParts/RecipePageInstructions.vue
```diff
@@ -327,6 +408,21 @@
...
+                  >
+                    <div
+                      v-for="(noteRef, i) in step.noteReferences"
+                      :key="noteRef.referenceId ?? i"
+                      class="mb-1 d-flex align-center text-body-2"
+                    >
+                      <v-icon size="14" class="mr-1" style="cursor: default;">
+                        {{ $globals.icons.noteTextOutline }}
+                      </v-icon>
+                      {{ noteRef.referenceId != null ? noteLookup[noteRef.referenceId] : '' }}
```
michael-genson: Can we fall back to `$t('recipe.note')` here the way the dialog and the bottom sheet do? A note with no title currently renders as just the icon with a blank label next to it.

frontend/app/components/Domain/Recipe/RecipePage/RecipePage.vue
```diff
@@ -262,6 +262,28 @@ const notLinkedIngredients = computed(() => {
   });
 });
 
+function pruneOrphanedNoteReferences() {
```
michael-genson: Do we need this on the frontend? `_remove_non_existent_note_references` already prunes these on save, and we don't do the equivalent for ingredient references.

michael-genson (changes requested):
Just a few more small things and then I think this is good to go.

frontend/app/lib/api/types/recipe.ts
jallier: It wasn't generated using that command, have done so now. When I run the generator I get a bunch of unrelated changes in the python schema files __init__.py that I won't include.
The .ts files also seem to add some unrelated fields to types that wouldn't be affected by the changes here and `[k: string]: unknown;` types in a few places. Not sure if all that should be included, but have left as is.

I'll commit the output from running the generator on its own commit to make it easier to inspect, please let me know if you'd like any changes with it 👍

frontend/app/components/Domain/Recipe/RecipePage/RecipePage.vue
jallier: This was just so that the note reference is immediately removed from any linked steps when you delete it without having to save. But you're right, we don't really need it twice so have removed and the note reference will be removed on save

jallier:
> Just a few more small things and then I think this is good to go.

Have updated. Thanks for your patience working with me through this

michael-genson:
Looks great! Can we drop the extra codegen churn? Our generated files seem to have drifted, I'll look into that separately.

frontend/app/lang/messages/en-US.json
```diff
@@ -494,6 +494,9 @@
     "auto": "Auto",
     "cook-mode": "Cook Mode",
     "link-ingredients": "Link Ingredients",
```
michael-genson, line 496: Can we clean up these unused i18n translations? Both `recipe.link-ingredients` (this line) and `recipe.ingredient-linker` (line 491) no longer have any usages now that the menu entry and dialog title both moved to `recipe.link-references`.

michael-genson (commented):
Looks great now thanks!

michael-genson (dismissed):
Thanks! Appreciate all your work on this one

jallier:
@michael-genson thanks for helping me through the changes. Looks like the CI failed because my migration filename is out of order. Am I right in thinking I can just rename it to be the most recent timestamp migration, while ensuring the `Revises:` and `down_revision` fields are still correct? Or safer to just regenerate it again?

Cheers

michael-genson:
Oh yup meant to check that, my mistake. Yes you have it right:
1. The filename must be a later date than the previous migration (just set it to today if you want)
2. The "Create Date" in the comment should match the date of the filename
3. The "Revises" id in the comment references the previous latest id (currently `4b91d3a7c0e2`, not likely to change by the time you do this)
4. The `down_revision` also matches that id

Whenever you update it feel free to tag me, I'll approve it again!

jallier:
@michael-genson have updated, thanks for the speedy reply!

michael-genson (approved):

PR #7623, opened 2026-05-13
fix: use locale for Recipe Created timeline event (#4497)

### LINKED ISSUE #4497: [BUG] - "Recipe created" Translation not used
### First Check

Hello,

I just installed the latest version and added my first recipe. 
The language of mealie is in french, and the message "Recipe created" appear at the end of the process : 

Capture d’écran du 2024-11-02 10-08-14

How can I help you to use the existing translation?

### Steps to Reproduce

1) Add a recipe
2) See the english text instead of the localized string

No logs

### Mealie Version

2.1.0

### Deployment

Docker (Linux)

_No response_

hay-kot:

_(REQUIRED)_

The `recipe-created` timeline event subject was always stored in the database as the pre-translated English string `"Recipe Created"`, ignoring the locale of the creating user. This happened because `self.t("recipe.recipe-created")` was called at creation time using the HTTP request's `Accept-Language` header, which often defaults to `en-US` even when the user has configured a different locale in the UI.

Changes:
- `mealie/services/recipe/recipe_service.py`: Store the bare i18n key `"recipe.recipe-created"` in the database instead of the translated string.
- `mealie/routes/recipe/timeline_events.py`: Translate `subject` for `system`-type events at serve time in both `GET /timeline/events` and `GET /timeline/events/{id}`, using the locale from the request's `Accept-Language` header. Non-system events (user comments/info) are returned unchanged.
- Old events already stored with a plain English string are backward-compatible — the translation lookup falls back to the key itself when no match is found.

_(REQUIRED)_

Fixes #4497

## Testing

New integration test `test_recipe_created_system_event_is_translated` in `tests/integration_tests/user_recipe_tests/test_recipe_timeline_events.py`:
- Creates a recipe, then fetches its timeline events with `Accept-Language: fr-FR` and asserts the subject is `"Recette créée"`.
- Fetches the same events with `Accept-Language: en-US` and asserts the subject is `"Recipe Created"`.
- Also verifies the single-event GET endpoint translates correctly.

All 19 existing timeline-event integration tests continue to pass.

## AI / LLM Assistance

_(REQUIRED)_

This PR was fully generated with Claude Code (claude-sonnet-4-6).

### DIFF
--- a/mealie/services/recipe/recipe_service.py
+++ b/mealie/services/recipe/recipe_service.py
@@ -224,7 +224,7 @@ class RecipeService(RecipeServiceBase):
         timeline_event_data = RecipeTimelineEventCreate(
             user_id=new_recipe.user_id,
             recipe_id=new_recipe.id,
-            subject=self.t("recipe.recipe-created"),
+            subject="recipe.recipe-created",
             event_type=TimelineEventType.system,
             timestamp=new_recipe.created_at or datetime.now(UTC),
         )
--- a/mealie/services/recipe/recipe_service.py
+++ b/mealie/services/recipe/recipe_service.py
@@ -38,6 +38,8 @@ from mealie.services.scraper import cleaner
 
 from .template_service import TemplateService
 
+RECIPE_CREATED_EVENT_SUBJECT = "recipe.recipe-created"
+
 
 class RecipeServiceBase(BaseService):
     def __init__(self, repos: AllRepositories, user: PrivateUser, household: HouseholdInDB, translator: Translator):
@@ -224,7 +226,7 @@ class RecipeService(RecipeServiceBase):
         timeline_event_data = RecipeTimelineEventCreate(
             user_id=new_recipe.user_id,
             recipe_id=new_recipe.id,
-            subject="recipe.recipe-created",
+            subject=RECIPE_CREATED_EVENT_SUBJECT,
             event_type=TimelineEventType.system,
             timestamp=new_recipe.created_at or datetime.now(UTC),
         )
--- a/mealie/routes/recipe/timeline_events.py
+++ b/mealie/routes/recipe/timeline_events.py
@@ -15,6 +15,7 @@ from mealie.schema.recipe.recipe_timeline_events import (
     RecipeTimelineEventPagination,
     RecipeTimelineEventUpdate,
     TimelineEventImage,
+    TimelineEventType,
 )
 from mealie.schema.recipe.request_helpers import UpdateImageResponse
 from mealie.schema.response.pagination import PaginationQuery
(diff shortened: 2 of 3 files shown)


mealie/services/recipe/recipe_service.py
```diff
@@ -224,7 +224,7 @@ def create_one(self, create_data: Recipe | CreateRecipe) -> Recipe:
         timeline_event_data = RecipeTimelineEventCreate(
             user_id=new_recipe.user_id,
             recipe_id=new_recipe.id,
-            subject=self.t("recipe.recipe-created"),
+            subject="recipe.recipe-created",
```
hay-kot: @michael-genson what do you think of just storing the translation key here for on the fly translations? It means we're tightly coupled to the keys. I can write a regression test so we fail if this key no longer exists.

michael-genson: Yeah I think this works! Since we fall back to the literal string if there's no match (e.g. a manually entered subject via the API or similar)

As long as we have some sort of regression test as you mentioned I don't see a problem. Only other consideration would be false positive translations but unless we introduce top level translations (i.e. without a ".") then we should be fine (and I don't see us ever doing that)

michael-genson: Another issue we'll run into if we adapt this pattern for other events is the "User Made This" events. Those we dynamically fill the user's name into the translation string.

Recipe timeline events have users associated with them so for this specific instance we should be okay (since we can assume the user who created the event is also the user we want to populate in the subject), though other similar patterns we might run into trouble later down the line. Not really something we need to worry about now though

michael-genson (approved):
LGTM, only question before merging - do we want to do this for our other subjects too?

hay-kot:
Good question. The only other auto-generated subjects are the meal-plan "made this" events in `create_timeline_events.py` (e.g. `"{user} made this for dinner"`). Those have a few complicating factors vs. the recipe-created event:

1. **Dynamic content** — they embed the user's full name and the entry-type value, so they can't be a bare i18n key; they'd need a parameterised template key.
2. **`info` not `system` type** — the current serve-time translation only applies to `system` events. Extending it to `info` events would risk translating user-authored event subjects.
3. **Deduplication query** — `create_timeline_events.py` checks for duplicate events by matching the exact subject string. If we change what's stored we'd need to update that query at the same time.
4. There's already a `# TODO: make this translatable` comment in that code acknowledging the gap.

Happy to tackle those in a follow-up PR if you'd like, but I'd rather keep that work separate to avoid scope creep here. Let me know how you'd like to proceed.

michael-genson:
Nope that sounds good to me for now then, thanks!

PR #7624, opened 2026-05-14
feat: use OIDC avatars

doenke:

This PR improves OIDC-based avatar handling so profile images are both sourced correctly and refreshed when the upstream identity provider updates them.

Changes made:
- `mealie/core/security/providers/openid_provider.py`
  - Added support for using the OIDC `picture` claim as the user avatar source.
  - Updated avatar update behavior to rotate the user cache key when avatar-related profile data changes.

Why:
- Some OIDC providers expose user profile images via the `picture` claim; this ensures Mealie can consume that standard claim directly.

N/A

(If no issue exists, replace with: `N/A`)

- Review focus: `openid_provider.py` avatar update flow.
- The change is intentionally scoped to OIDC provider logic only.

## Testing

- Verified OIDC login/sync path still succeeds with existing claims.
- Verified avatar is populated from `picture` claim when present.
- Verified avatar changes trigger cache-key rotation so updated images are served instead of stale cached content.

## AI / LLM Assistance

This PR was developed with AI assistance for drafting and refining implementation details, commit/PR text, and validation steps.  
All code changes and final behavior were reviewed and validated by the author.


saschabrockel:
I would love to see that!

kylmp:
+1 would like this feature

However, maybe adding a setting for it like `settings.OIDC_PICTURE_ENABLED` to have it optional

mealie/core/security/providers/openid_provider.py
```diff
@@ -132,3 +140,26 @@ def required_claims(self):
         if settings.OIDC_REQUIRES_GROUP_CLAIM and not self.use_default_groups:
             claims.add(settings.OIDC_GROUPS_CLAIM)
         return claims
+
+    def _update_profile_image_from_claim(self, user_id, claims: UserInfo) -> None:
+        picture = claims.get("picture")
```
michael-genson, line 239: This re-fetches and rewrites the avatar plus rotates cache_key on every login, even if the picture hasn't changed. Can we persist the last-seen picture value (or a hash) and skip the fetch/rotation when it's unchanged?

mealie/core/security/providers/openid_provider.py
```diff
@@ -132,3 +140,26 @@ def required_claims(self):
...
+        if not picture or not isinstance(picture, str):
+            return
+
+        try:
+            response = requests.get(picture, timeout=15)
```
michael-genson: This fetches a URL taken directly from an OIDC claim with no validation -- SSRF risk if a malicious/compromised provider points picture at an internal or metadata endpoint. Can we enforce https and reject private/loopback/link-local IPs before this request?

mealie/core/security/providers/openid_provider.py
```diff
@@ -132,3 +140,26 @@ def required_claims(self):
...
+            response.raise_for_status()
+
+            with get_temporary_path() as temp_path:
+                temp_img = Path(temp_path).joinpath(str(uuid4()))
+                with temp_img.open("wb") as file:
+                    file.write(response.content)
```
michael-genson: No size cap on the download before writing to disk. Can we stream this with a Content-Length/chunked max-byte check instead?

michael-genson (changes requested):
Generally looks good. I want to add some test coverage for the new path: unit tests for `_update_profile_image_from_claim` covering a missing/non-string picture claim, a successful fetch, and a failed download being swallowed without raising.

mealie/core/security/providers/openid_provider.py
```diff
@@ -132,3 +149,89 @@ def required_claims(self):
...
+        if user is None:
+            return
+
+        # Skip the download entirely when the claim still points at the image we already stored.
+        picture_hash = hashlib.sha256(picture.encode()).hexdigest()
+        if user.oidc_picture_hash == picture_hash:
+            return
+
+        try:
+            response = requests.get(picture, timeout=15, stream=True)
```
michael-genson: Can we disable redirects? Otherwise we could be making unsafe `GET` reqs.

michael-genson: Similarly, we aren't persisting the IP check from earlier.

mealie/core/security/providers/openid_provider.py
```diff
@@ -132,3 +149,89 @@ def required_claims(self):
...
+            response.raise_for_status()
+            content = self._read_capped(response)
+
+            with get_temporary_path() as temp_path:
+                temp_img = Path(temp_path).joinpath(str(uuid4()))
+                temp_img.write_bytes(content)
+
+                image = img.PillowMinifier.to_webp(temp_img)
+                dest = PrivateUser.get_directory(user_id) / "profile.webp"
+                dest.write_bytes(Path(image).read_bytes())
```
michael-genson: Can we use `shutil.move(image, dest)` instead?

michael-genson (changes requested):
Couple more things came up when I was looking at this. Also, when you rebase/merge into `mealie-next`, I think you need to re-order the alembic migration to put it at the end (you might want to just regenerate it).

mealie/schema/user/user.py
```diff
@@ -209,6 +209,7 @@ class PrivateUser(UserOut):
     password: str
     login_attemps: int = 0
     locked_at: datetime | None = None
+    oidc_picture_hash: str | None = None
```
michael-genson: Can we just use `external_avatar_hash` instead? So we don't tie the column to OIDC.

mealie/core/security/providers/openid_provider.py
doenke: Done

mealie/schema/user/user.py
doenke: Done

mealie/core/security/providers/openid_provider.py
doenke: All three done.

- Redirects: allow_redirects=False, plus an explicit 3xx reject — raise_for_status() treats redirects as success.
- IP check: same root cause. The check only vetted the claim's URL, so a redirect reached a host we never checked; with redirects off, the vetted host is the one we connect to.
- Migration: re-chained onto 69e942bab3aa and renamed to sort last — alembic heads now reports one head.

Heads-up: avatars served via a redirect (Gravatar) will now fail. Doesn't affect my own setup, but it might for others — happy to validate each hop instead if you think it's worth it.

michael-genson: In that case can we validate each redirect instead?

doenke: Done — redirects are followed again, but walked manually so every hop goes through the same validation before it's requested. Capped at 3 hops.

michael-genson (dismissed):
Thank you! I pushed a couple small fixes.

PR #7625, opened 2026-05-14
fix: enforce ownership check on recipe deletion (GHSA-x5v9-9jvh-7c7q)

hay-kot:

Fixes a BOLA (Broken Object Level Authorization) vulnerability where any authenticated same-household user could permanently delete recipes they don't own by sending a direct `DELETE /api/recipes/{slug}` request.

`can_delete` was delegating to `can_update`, which intentionally allows same-household users to edit unlocked recipes as a collaborative editing feature. Deletion is permanent and irreversible, so it now enforces ownership regardless of lock state. Admins retain full delete access.

Fixes GHSA-x5v9-9jvh-7c7q

## Testing

Added regression tests for the missing cases:
- `test_other_user_cant_delete_unlocked_recipe` — non-owner cannot single-delete an unlocked recipe
- `test_other_user_cant_bulk_delete_unlocked_recipes` — non-owner cannot bulk-delete unlocked recipes

All existing ownership tests continue to pass.

## AI / LLM Assistance

This PR was researched and implemented with Claude Code.

### DIFF
--- a/mealie/services/recipe/recipe_service.py
+++ b/mealie/services/recipe/recipe_service.py
@@ -69,8 +69,19 @@ class RecipeService(RecipeServiceBase):
     def can_delete(self, recipe_slugs: list[str]) -> bool:
         if self.user.admin:
             return True
-        else:
-            return self.can_update(recipe_slugs)
+
+        # Deletion requires ownership; collaborative editing rules (can_update) do not apply
+        model = self.group_recipes.model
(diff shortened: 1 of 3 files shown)


michael-genson (dismissed):
Thanks!

michael-genson:
Looks like a couple tests broke

michael-genson (approved):

PR #7629, opened 2026-05-14
fix: Protect sensitive data in query filter API (GHSA-8m57-7cv5-rjp8)

michael-genson:

_(REQUIRED)_

The query filter API allows filtering by arbitrary fields, including sensitive data (e.g. hashed passwords). The query filter API can also be used unauthenticated in public routes (e.g. recipes by user).

This PR explicitly sets which fields should be filterable using a custom `FilterableColumn` type. `FilterableColumn` is identical to `Mapped` but inserts `info={"filterable": True}` into the metadata. If the metadata is missing from a column in the query filter builder, then a `ValueError` is raised.

_(REQUIRED)_

Fixes GHSA-8m57-7cv5-rjp8

## Testing

Added comprehensive testing for the new private/restricted functionality.

## AI / LLM Assistance

_(REQUIRED)_

Copilot was used to help scope out the work and tinker with various different approaches. Code was thoroughly reviewed by a human.


hay-kot (dismissed):
Changes look good and I won't block this, but I have some concerns with an opt-out system. For example, should you be able to extract the `auth_method` field from a query? Probably not. I think a system where you opt-into querying instead of exposing all fields by default would be a more secure approach.

michael-genson:
Yeah this is fair, I was on the fence which way I go since the tradeoff is if you add a new table you kind of have to be familiar with the implementation. But I guess tests will catch that.

I'll go ahead and invert the logic!

mealie/db/models/recipe/labels.py
```diff
@@ -3,26 +3,26 @@
 from sqlalchemy import ForeignKey, String, UniqueConstraint, orm
```
michael-genson, line 3: This file was orphaned in the root models directory for some reason so I moved it to `/recipe`.

mealie/db/models/_filterable_column.py
```diff
@@ -0,0 +1,23 @@
...
+    def __class_getitem__(cls, item: type) -> type:
+        return Mapped[Annotated[item, mapped_column(info={"filterable": True})]]
+
+
+# SQLAlchemy doesn't play nice with mypy when overriding Mapped, so
+# we use this awkward workaround to make mypy happy
+if TYPE_CHECKING:
+    FilterableColumn = Mapped
+else:
+    FilterableColumn = _FilterableColumn
```
michael-genson, line 23: This is a silly workaround but I tried:
- Inheriting from Mapped (`class FilterableColumn(Mapped): ...`)
- Simplifying the assignment (`type FilterableColumn[T] = Mapped[Annotated[T, mapped_column(info={"filterable": True})]]`)

And they fail for one reason or another, and this works 🤷

hay-kot (approved):
LGTM 🚀 

Thanks for taking this one on!

PR #7650, opened 2026-05-22
feat: In-app AI Provider Configuration

michael-genson:

_(REQUIRED)_

Implements a new AI Provider system which allows users to configure one (or more) AI providers in-app. This completely replaces the old environment variable setup.

Users with the "can manage" permission (and admins) can manage AI providers in the group settings page:
image

An "AI provider" is an integration with any arbitrary AI service (OpenAI, or any OpenAI-compatible service). Unlike the previous environment variable implementation, users can mix and match completely unrelated providers (e.g. default to Ollama, but use OpenAI's whisper for audio transcriptions). Currently you can configure:
- default (used for most things)
- audio (used exclusively for importing videos and transcribing them)
- image (used for image OCR when creating recipes via image)

When adding or editing a provider you can set all the variables you previously set in env vars:
image

You can set the headers/params similarly in the advanced tab:
image

Providers are per-group, so admins can manage providers for other groups as well (on the admin page). I've also added this as an optional step in the setup wizard.

In addition to addressing community feedback with the difficulties/limitations of the env var configuration, this also opens the door for better configuration capabilities such as:
- Different providers for different functions (e.g. one for importing recipes, one for parsing)
- Multiple providers for error handling (e.g. if one provider fails, fallback to a backup, etc.)
- Configuring prompts in-app, replacing the current env var approach (currently the only remaining env var for OpenAI is the prompts directory)
- Allowing users to choose a model on the fly (e.g. maybe you have two different models you like for parsing, we can expose them both on the parsing page)

I also updated the admin debugger to let you choose which provider you want to debug:
image

I also added an announcement (which is slightly different for admins, and for users that can/can't manage groups):
image

_(REQUIRED)_

N/A, addresses a bunch of feedback around the current implementation

Considerations made in this PR:
- All the old env vars are deprecated (except the prompts directory). The alembic migration will read the env vars and populate one or two providers using them (since audio transcriptions use a different model, it necessitates a different provider, if enabled).
- The API key is never returned via the API, similar to how users and passwords work.
- The new implementation uses generic "AI Provider" language to decouple from OpenAI. In the future I want to more-or-less remove directly naming things with OpenAI to better communicate that you can use _any_ provider. Functionally this has always been true, but the naming makes it seem otherwise.
- I've deprecated the `OPENAI_WORKERS` and `OPENAI_SEND_DATABASE_DATA` env vars completely with no replacement. These were rather shortsighted anyway. I have some ideas to bring them back in a more meaningful way if the community wants them, but I suspect no one used these (or don't care if they're removed).
- I implemented a new "no autofill" plugin to disable autofill from password managers, etc. The provider form using the "password" field for the API key breaks password managers in multiple ways. We can theoretically use this elsewhere too, e.g. webhooks, but for now I've only implemented it on the new form.

## Testing

Lots of backend tests, manually walked through each AI-driven flow in the frontend.

## AI / LLM Assistance

_(REQUIRED)_

Copilot did a lot of the frontend work and helped debug some of the backend issues. Most of the backend code was written by hand, much of the frontend was written by Copilot. Everything was manually verified and tested by me (a human, probably).


BrysonC:
Excited for this feature! Unfortunate timing for me to be updating my AI configuration, I couldn't figure out why the docs didn't match my current instance.

primetime43:
@michael-genson I think for the model in here, it should by default give you a dropdown of options, then the option of custom text input. Its too easy to not know what to put or incorrectly configure it. For example, I want to use gpt5.5, how does the user know to do "gpt5.5" or it may be something else. Make this less error prone.

image

PR #7656, opened 2026-05-24
fix: return HTTP 200 for SPA fallback responses

### LINKED ISSUE #7512: [BUG] Backend SPA fallback returns HTTP `404` on every client-side route refresh, breaking deployments behind reverse proxies that intercept `4xx`
### First Check

Hitting any client-side route directly (e.g. `https://demo.mealie.io/g/home`, `/shopping-lists`, `/admin/...`) returns the Mealie SPA HTML but with HTTP status code **404**, not 200:

```shell
$ curl -I https://demo.mealie.io/g/home
HTTP/2 404
accept-ranges: bytes
cache-control: no-cache
content-type: text/html; charset=utf-8
date: Tue, 21 Apr 2026 18:45:53 GMT
etag: "c3a31524709229f82173bdbbbff8d546"
last-modified: Tue, 21 Apr 2026 03:45:40 GMT
server: uvicorn
content-length: 8838
```

Because `demo.mealie.io` apparently does not have a reverse proxy that intercepts upstream `4xx`, the browser renders the body anyway and the SPA hydrates, so the user never notices in this case.

_However_, anyone running Mealie behind a reverse proxy that intercepts `4xx` (e.g. [`tarampampam/error-pages`](https://github.com/tarampampam/error-pages), Traefik middleware, ...) sees a 404 error page on every refresh of any in-app route instead of the Mealie UI. 

The user has to manually navigate to the root (`https://recipes.example.com/`) every time to get a working UI; from there client-side navigation works, but any subsequent refresh on a another route returns a 404 page again.

### Steps to Reproduce

Verify the wrong status code on demo:

```shell
$ curl -I https://demo.mealie.io/g/home
HTTP/2 404
```

Also can be reproduced locally by deploying Mealie behind a reverse proxy (e.g. Traefik/Nginx/Caddy) that can intercept status codes.

1. Deploy v3.16.0 (`ghcr.io/mealie-recipes/mealie:v3.16.0`) behind any reverse proxy that intercepts upstream `4xx` status codes.
2. Log in and navigate to `https://recipes.example.com/g/home`.
3. Refresh the page.
4. The proxy's generic 404 page is shown instead of Mealie.
5. Manually navigate to `https://recipes.example.com/`; works (and URL changes to `/g/home`).
6 Refresh `/g/home`; fails again.

---

I have tried debugging this issue with Copilot (Opus 4.7), which believes the root cause to be the following

### Root cause

The SPA is mounted with Starlette's `html=True`:

```python
# mealie/routes/spa/__init__.py
app.mount("/", SPAStaticFiles(directory=__app_settings.STATIC_FILES, html=True), name="spa")
```

`StaticFiles(html=True).get_response()` has the following behavior on a missing path:

1. Look for a `404.html` in the static directory.
2. If found → return a `FileResponse` with that body and **`status_code=404`** (no exception raised).
3. Only if `404.html` is also missing → raise `HTTPException(404)`.

Nuxt with `ssr: false` generates a `404.html` (and `200.html`) in the build output whose contents **are the SPA shell** (they are effectively copies of `index.html` — see `globIgnores: ["404.html", "200.html", "index.html"]` in `frontend/nuxt.config.ts` where they are treated identically). So on a request for `/g/home`:

1. Starlette can't find `g/home`.
2. It serves the contents of `404.html` (= SPA shell) with `status_code=404`.
3. **No exception is raised.**
4. Control returns to `SPAStaticFiles.get_response`, the `except HTTPException` branch is skipped, `Cache-Control: no-cache` is attached because the response is HTML, and the 404 + SPA shell response goes back to the client.

This means:
- The behavior is the same in v3.15.0 and earlier — this is a long-standing bug, not a v3.16.0 regression. The 3.16.0 refactor just changed how the response is post-processed; it didn't change the status semantics.
- The `super().get_response("index.html", scope)` fallback inside the `except` block in `SPAStaticFiles.get_response` is **dead code** — it never runs.

### Proposed fix

Smallest change: after the `super().get_response(...)` call returns, detect the case where Starlette served `404.html` (the SPA shell) and rewrite the status to 200. This keeps the existing `Cache-Control` logic and the rarely-actually-hit `except` branch unchanged:

```diff
 class SPAStaticFiles(StaticFiles):
     async def get_response(self, path: str, scope):
         try:
             response = await super().get_response(path, scope)
         except HTTPException as ex:
             if ex.status_code == 404:
                 response = await super().get_response("index.html", scope)
             else:
                 raise ex

+        # StaticFiles(html=True) serves the Nuxt-generated 404.html (which IS
+        # the SPA shell — Nuxt creates 404.html and 200.html as copies of
+        # index.html when ssr: false) with status_code=404 for any unknown
+        # path. Rewrite the status to 200 so reverse proxies that intercept
+        # 4xx (tarampampam/error-pages, Traefik errors middleware, nginx
+        # proxy_intercept_errors, Caddy handle_errors, Cloudflare custom
+        # error pages, ...) don't replace the body with a generic error page.
+        if response.status_code == 404 and response.media_type == "text/html":
+            response.status_code = 200
+
         # Hashed assets (_nuxt/*) are safe to cache forever since new builds
         # produce new filenames. HTML must revalidate so browsers always fetch
         # the correct bundle references after a container rebuild ...
         if path.startswith("_nuxt/"):
             response.headers["Cache-Control"] = "public, max-age=31536000, immutable"
         elif path == "." or response.media_type == "text/html":
             response.headers["Cache-Control"] = "no-cache"

         return response
```

An alternative would be to mount `SPAStaticFiles(..., html=False)`, which would make the existing `try/except` fallback actually trigger — but that changes how `/` itself is resolved and is more invasive.

---

.

### Mealie Version

- Version: [v3.16.0](https://github.com/mealie-recipes/mealie/releases/tag/v3.16.0)
- Build: [5fc4851ef5261be3cee131ab17e5532fc4820f14](https://github.com/mealie-recipes/mealie/commit/5fc4851ef5261be3cee131ab17e5532fc4820f14)

### Deployment

Docker (Linux)

_No response_

roadhero:
### What this PR does / why we need it:

Mealie's `SPAStaticFiles` serves the correct HTML shell for all client-side routes (e.g. `/g/home`, `/shopping-lists`, `/admin/...`), but responds with HTTP status **404** instead of **200**. This is invisible during direct browser access because the SPA router handles it client-side, but it breaks any deployment behind a reverse proxy that intercepts upstream 4xx responses — Traefik `errors` middleware, nginx `proxy_intercept_errors`, Caddy `handle_errors`, Cloudflare custom error pages. In those setups, every page refresh shows a generic error page instead of Mealie.

**Root cause:** `StaticFiles(html=True)` serves `404.html` (which Nuxt generates as a copy of the SPA shell) with `status_code=404` *without* raising `HTTPException`. The existing `except HTTPException` fallback therefore never fires. The fix detects an HTML 404 response from the parent class and rewrites the status to 200.

**Changes to `mealie/routes/spa/__init__.py`:**

- Added a 4-line status rewrite (1 condition, 1 assignment) between the `try`/`except` block and the `Cache-Control` logic.

### Which issue(s) this PR fixes:

Fixes #7512

### Special notes for your reviewer:

- The existing `except HTTPException` block (serving `index.html` on 404) is dead code when `html=True` is set and `404.html` is present in the build output — Starlette never raises in that case. This PR intentionally leaves that block in place as a safety net for the unlikely event that `404.html` is absent from a custom build.
- The condition requires both `status_code == 404` and `media_type == "text/html"`. In the `StaticFiles(html=True)` context, this combination can only come from the `404.html` fallback — it cannot be triggered by missing JS/CSS/image assets (those have non-HTML media types).
- The `response_404()` helper (used for invalid recipe slugs by `serve_recipe_with_meta` etc.) creates its own `Response(...)` object directly. It does not go through `SPAStaticFiles.get_response` and is unaffected.

### Testing

Added `tests/unit_tests/test_spa_static_files.py` with regression tests against a temporary directory containing `index.html` and `404.html`:

- `test_unknown_path_returns_200_with_html_body` — the core regression test; verified to fail against pre-fix code (returns 404) and pass with the fix (returns 200).
- `test_known_path_returns_200` — direct `index.html` request still returns 200.

Existing `tests/integration_tests/test_spa.py` (15 tests): all pass; intentional `response_404()` calls in recipe metadata routes remain 404 as designed.

### AI/LLM Assistance

This PR was prepared with AI assistance from Anthropic Claude:

- **Bug investigation and root cause analysis** — used Claude to trace the StaticFiles 404 path through Starlette's source and confirm that `html=True` + a present `404.html` produces a non-raising 404 response.
- **Fix implementation** — the 4-line edit to `mealie/routes/spa/__init__.py` was drafted by Claude based on the analysis above, then I reviewed it line-by-line against Starlette's `StaticFiles.get_response` behaviour before committing.
- **Regression test** — the `test_spa_static_files.py` test was drafted by Claude; I verified it fails against pre-fix code and passes against the fix, and confirmed it does not interact with the existing `test_spa.py` fixtures.
- **PR body** — drafted with Claude assistance and edited by me.

I have read every line of the diff, reproduced the bug in a deployment behind Traefik with `errors` middleware enabled, and confirmed the fix resolves it. I take responsibility for the change.


michael-genson:
Please use the full PR template which includes AI/LLM assistance disclosure

roadhero:
Updated the PR description to include the AI/LLM Assistance section per the template — thanks for the catch.

ToshY:
@michael-genson Any updates on reviewing this? Would be nice to have.

michael-genson (approved):
Thanks for this!

knom:
Will this be merged into main?
I really would love it for my deployment!

michael-genson:
CI got stuck, this was supposed to be merged already. Just queued it up again

PR #7663, opened 2026-05-25
dev: switch to pnpm and update devcontainer

p0lycarpio:

_(REQUIRED)_

This pull request migrates the frontend and related tooling from Yarn v1 toward [pnpm](https://pnpm.io/) : a fast, up-to-date alternative. The project now uses Node.js 24 throughout (which was not the case in all workflows). The development container has been updated accordingly and now uses the same versions as the final image.

The goal is to improve dependency management, speed up installs, and modernize the dev environment. It also updates some dependencies and makes the development setup more consistent.

References to `yarn.lock` have been replaced, and `packageManager` is now set to `pnpm@11.18.0` in `package.json`

_(REQUIRED)_

N/A

Nuxt has been updated to the latest minor version to resolve a build issue encountered with pnpm. Warnings appear during the frontend build but can be addressed in a separate pull request.

Devcontainer has also been slimmed down (from 3.1 GB to 2.7 GB) by removing docker-in-docker.

## Testing

- Replacing Yarn with pnpm was tested locally by updating and rebuilding the Devcontainer several times without cache.
- The execution of various tasks (starting with “ui:” in the Taskfile) was also tested, as well as a full build.
- The workflows could ~~not~~ be tested ; except the releases ones.

## AI / LLM Assistance

_(REQUIRED)_

Not used for this PR.


frontend/package.json
```diff
@@ -8,8 +8,8 @@
     "start": "nuxt start",
     "generate": "nuxt generate --spa",
     "lint:js": "eslint .",
-    "lint": "yarn lint:js",
-    "lint:log": "yarn lint:js --debug",
+    "lint": "pnpm lint:js",
+    "lint:log": "lint:js --debug",
```
michael-genson: Should this be `pnpm lint:js --debug`?

.devcontainer/Dockerfile
```diff
@@ -28,8 +14,14 @@ RUN apt-get update \
     gnupg gnupg2 gnupg1
 
 # Install uv
-RUN pip install uv
+COPY --from=ghcr.io/astral-sh/uv:latest /uv /usr/local/bin/uv
 ENV UV_LINK_MODE=copy
 
+# Install Taskfile
+RUN curl -1sLf 'https://dl.cloudsmith.io/public/task/task/setup.deb.sh' | sudo -E bash \
+    && apt-get install task
```
michael-genson: Do we need `-y` here?

michael-genson (changes requested):
Looks great, thanks for doing this. A couple comments. Also, do we need this pnpm/action-setup workaround in the other workflows too, or is it e2e-specific?

.devcontainer/Dockerfile
p0lycarpio: Yes, thanks !

p0lycarpio:
Thanks for the feedback!
Since I wasn't able to test the other workflows, I think the workaround is also needed to get the correct version. The change has been made.

Xyolyp:
This one currently conflicts with `mealie-next`, and only in `frontend/yarn.lock` — so nothing substantive, just the lockfile. Given the PR replaces yarn with pnpm anyway, taking upstream's `yarn.lock` and letting your own changes regenerate should be enough.

(Not a maintainer, I just stumbled over it while going through the open PRs.)

michael-genson (approved):
Thank you!

PR #7718, opened 2026-05-31
chore: add 5-day dependency cooling period for supply-chain hardening

hay-kot:

Enforces a 5-day minimum release age before any dependency update is adopted, mitigating supply-chain attacks (compromised releases are usually caught and yanked within days).

- **Renovate** (`renovate.json`): top-level `minimumReleaseAge: "5 days"` covering all managers + `internalChecksFilter: "strict"`. Removes the now-redundant per-rule entries.
- **uv** (`pyproject.toml`): `exclude-newer = "5 days"` as defense-in-depth for manual `uv add`/`uv lock`.
- **CI / Taskfile / Dockerfile**: `UV_FROZEN=1` everywhere uv runs, so it installs from the lockfile instead of re-resolving (a rolling `exclude-newer` otherwise makes every uv command re-resolve and fail on in-window pins).

N/A — proactive supply-chain hardening.

The frontend uses Yarn Classic (no `minimum-release-age`); its npm deps are gated via Renovate's npm manager. Already-locked deps are unaffected — the window only governs adopting new versions.

## Testing

- Full PR CI is green (backend tests, package build, Trivy, docs) — these initially failed on the `exclude-newer` re-resolve and now pass with `UV_FROZEN`.
- Verified locally that a manual `uv lock` correctly refuses an in-window release.

## AI / LLM Assistance

Created with substantial assistance from Claude Code (Anthropic) — investigation, config, CI-failure diagnosis, and this description — all reviewed and directed by the author.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -27,7 +27,7 @@ jobs:
         uses: astral-sh/setup-uv@v4
 
       - name: Install dependencies
-        run: uv sync --only-group docs --no-install-project
+        run: uv sync --frozen --only-group docs --no-install-project
 
       - name: Build docs
         run: uv run --no-project mkdocs build -d site
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -20,6 +20,10 @@ concurrency:
 jobs:
   build:
     runs-on: ubuntu-latest
+    env:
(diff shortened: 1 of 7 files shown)


michael-genson (approved):
Nice, I knew renovate supported this, didn't know uv also does (of course it does)

PR #7719, opened 2026-05-31
fix: harden recipe content against stored XSS (chips, instructions, asset media)

hay-kot:

Closes three stored-XSS / HTML-injection vectors in recipe content and tightens how uploaded assets are served.

- **`frontend/app/components/Domain/Recipe/RecipeChips.vue`** — `truncateText()` measured length by assigning the tag/category name to `innerHTML`, which parsed (and executed) markup such as `` in the live document. It now delegates to a plain-text helper and never touches the DOM.
- **`frontend/app/lib/sanitize/text.ts`** (new) — small, testable `truncateText` helper that treats its input strictly as text.
- **`frontend/app/lib/sanitize/markdown.ts`** (new) — extracts and hardens the DOMPurify sanitizer previously inlined in `SafeMarkdown.vue`. Scriptable and form elements are stripped by the tag allowlist. `<iframe>` is only permitted when an admin has configured a host allowlist, and even then any iframe whose `src` is not an `https` URL on the allowlist is removed.
- **`frontend/app/components/global/SafeMarkdown.vue`** — now calls the shared sanitizer and passes the configured iframe host allowlist from `$appInfo`.
- **`mealie/core/settings/settings.py`** — adds `ALLOWED_IFRAME_HOSTS` (comma-separated) which extends a secure default list of trusted video providers (YouTube, Vimeo). Exposed to the client via `AppInfo` so the renderer can enforce it.
- **`mealie/routes/media/media_recipe.py`** — recipe asset responses now also send `X-Content-Type-Options: nosniff` alongside the existing `Content-Disposition: attachment`, so uploaded files cannot be served as active content in Mealie's origin.

Decision note: iframe embeds were previously unrestricted. Rather than dropping the feature, this makes it secure-by-default (trusted video hosts only) and lets operators opt into additional hosts via `ALLOWED_IFRAME_HOSTS`.

Addresses the following security advisories:

- GHSA-ffx9-4vvj-2rrq — `innerHTML` measurement sink in recipe chips
- GHSA-v89r-qfp9-95gv — scriptable/embed elements in rendered recipe instructions
- GHSA-gfwc-pjx4-mg9p — recipe asset content-type handling (adds `nosniff` on top of the existing attachment disposition)

The iframe host allowlist is the main design choice worth a look — defaults live in `DEFAULT_ALLOWED_IFRAME_HOSTS` and are extended (not replaced) by `ALLOWED_IFRAME_HOSTS`. Host matching is exact-or-subdomain over `https` only.

## Testing

- Frontend: new vitest coverage in `frontend/app/lib/sanitize/{markdown,text}.test.ts` — verifies scripts/handlers/form controls are stripped, iframe is removed unless its `https` src host is allowlisted (incl. subdomain and lookalike-host cases), and that truncation never parses HTML.
- Backend: `tests/unit_tests/test_config.py` covers the `allowed_iframe_hosts` default/extend/normalize behavior; `tests/integration_tests/user_recipe_tests/test_recipe_image_assets.py` asserts assets are served with `Content-Disposition: attachment` and `X-Content-Type-Options: nosniff`.
- All of the above pass locally; lint (eslint/ruff), format, and mypy are clean on the changed files.

## AI / LLM Assistance

This PR was produced with significant assistance from Claude Code (Anthropic). The agent performed the research, implementation, tests, and PR authoring; changes were reviewed and directed by a human maintainer.

### DIFF
--- a/docs/docs/documentation/getting-started/installation/backend-config.md
+++ b/docs/docs/documentation/getting-started/installation/backend-config.md
@@ -29,6 +29,7 @@
 | --------------------------- | :-----: | ----------------------------------------------------------------------------------- |
 | SECURITY_MAX_LOGIN_ATTEMPTS |    5    | Maximum times a user can provide an invalid password before their account is locked |
 | SECURITY_USER_LOCKOUT_TIME  |   24    | Time in hours for how long a users account is locked                                |
+| ALLOWED_IFRAME_HOSTS        |  `""`   | Comma-separated extra hostnames allowed as `<iframe>` sources in recipe content. Extends the built-in list of trusted video providers (YouTube, Vimeo). Subdomains are included automatically. Only `https` sources are permitted. Adding hosts here opts into rendering embeds from those origins to all viewers, including the public, so add only origins you trust. |
 
 ### Database
 
--- a/frontend/app/components/Domain/Announcement/Announcements/2026-05-31_1_iframe-embeds.vue
+++ b/frontend/app/components/Domain/Announcement/Announcements/2026-05-31_1_iframe-embeds.vue
@@ -1,0 +1,54 @@
+<template>
+  <div>
+    <p>
+      To harden Mealie against malicious content, <code>&lt;iframe&gt;</code> embeds in recipe
+      instructions, notes, and descriptions are now restricted to a trusted set of hosts.
+    </p>
+    <div class="mb-2">
+      By default, embeds are allowed only from well-known video providers:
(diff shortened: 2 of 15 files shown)


michael-genson (dismissed):
LGTM, is it worth adding an in-app announcement about the new `ALLOWED_IFRAME_HOSTS` env var?

michael-genson (approved):
LGTM

PR #7755, opened 2026-06-13
feat(frontend): add current week view toggle to meal planner

lainors7:

This PR introduces a new feature to the meal planner: a persistent **"Current Week"** view toggle. 

Currently, navigating the meal planner can cause dates to drift or require manual adjustments. With this feature enabled, the planner automatically shifts to display the current week view, aligning the 7-day display window precisely with the user-defined "start day of the week" from their household preferences. This provides a cleaner, more intuitive, and faster navigation experience for users who manage their meals on a strict weekly schedule.

### Summary of Changes

To implement this preference persistently across browser reloads, changes were made to both the global user preference state and the core planner component view logic:

1. **`frontend/app/composables/use-users/preferences.ts`**
   * Added the `useCurrentWeek` boolean property to the `UserMealPlanPreferences` interface.
   * Initialized it to `false` by default inside the local storage wrapper (`useLocalStorage`) so the setting persists natively.

2. **`frontend/app/pages/household/mealplan/planner.vue`**
   * Updated the calendar activator `<v-menu>` to conditionally display the current week text badge.
   * Injected a new `<v-switch>` tied to `currentWeekMode` inside the picker dropdown card.
   * Updated the initialization logic and added reactive `watch` bindings so that when "Current Week" is active, custom date/day inputs are disabled and locked within the calculated current week boundaries.

### UI Changes (Before / After)

* **Current Week Mode Activated:**
imagen

* **Custom Range Picker (Disabled automatically when Current Week is active):**
imagen

* **Context (User default start day is Monday, current day of testing is Saturday):**
imagen

---

This PR does not resolve an existing issue report. It implements a standalone quality-of-life feature request for persistent weekly navigation in the meal planner.

---

## Testing

I tested this new feature locally in a development environment:
1. Verified that toggling "Current Week" instantly snaps the calendar view range to the current week based on the household's "First Day of Week" preference.
2. Verified that manual date pickers and number inputs are properly disabled to avoid conflicting boundaries while the mode is active.
3. Reloaded the browser multiple times to ensure that the user's choice (`useCurrentWeek: true/false`) persists seamlessly via `LocalStorage` as expected.

---

## AI / LLM Assistance

An AI assistant (LLM) was utilized during the creation of this Pull Request to assist with formatting the code blocks into the repository's markdown templates and  proofreading the technical explanations in English. All core Vue/TypeScript implementation details, reactivity logic, and code adjustments were manually applied or extracted and integrated from the functional local workspace.


noxhirsch:
I really like this idea, and it's a feature I've been missing ever since I switched from another meal planning app.
One feature that I think would be very important in the weekly view - and beneficial in the normal view as well - is the ability to switch to the previous or next time period.
If you select a custom time period - e.g., 10 days - the new buttons would let you switch to the 10 days before or after that period.

I would just write "week" instead of "current week," so it would still make sense even when you switch to the next week.
To keep the time range button shorter for smaller screen, I'd remove the weekdays (since you know your week starts Monday/Sunday ...) and the brackets in the weekly view.

Quick mockup:
image

lainors7:
@noxhirsch Thanks so much for the feedback! I've implemented all of your suggestions:

- Previous/next week navigation (following your mockup): They shift by the exact length of the currently selected period, so if you pick a custom 10day range, the buttons jump 10 days back/forward.
- Changed the label so it still makes sense once you navigate away from the current week.
- Shorter date button — Removed the weekdays and the brackets from the range label; it now just shows the month + day

imagen

Xyolyp:
Two things I ran into while reading through the open PRs — I'm not a maintainer, just passing them on.

This no longer merges cleanly into `mealie-next`, though the conflict is confined to `frontend/app/pages/household/mealplan/planner.vue`, so a rebase should be quick.

The bigger one: alongside `en-US`, the diff also carries 41 other files under `frontend/app/lang/messages/`. Those are managed through Crowdin, and the contributor guidance is explicit that PRs touching non-English locales get rejected — so dropping them and keeping only `en-US.json` would likely save you a round trip.

michael-genson (changes requested):
Generally looks good, I think this makes a lot of sense. A few things to address:
- Default useCurrentWeek to true in preferences.ts rather than false -- likely the preferred behavior for most users out of the box.
- Rather than disabling the date picker and day-count inputs whenever "Current Week" is on, can we drop the disabling and just auto-turn-off "Current Week" when the user changes a custom range/date input? That removes the need to explicitly toggle it off first.
- Revert the "current-week": "Week" additions in every file under frontend/app/lang/messages/ except en-US.json -- that's the only one that should ever be hand-edited, the rest are Crowdin-generated.
- Drop the new dateRange format. Add a "compact" format instead ({ "month": "short", "day": "numeric" }, no weekday) to every locale file under frontend/app/lang/dateTimeFormats/, placed before "short" in each file, and use it for the week-range display. dateTimeFormats/ isn't Crowdin-covered, so this needs to be hand-added everywhere, not just en-US.json.
- Add test coverage for getCurrentWeekRange(startDay) across all firstDayOfWeek values and month/year boundaries, plus for navigate(direction)'s period-length calculation.

michael-genson:

lainors7:
@michael-genson Hey, sorry for the delay. I didn't have time to get to this earlier. Could you please reopen it? Thanks!

PR #7781, opened 2026-06-20
feat: rank exact and prefix matches above substring matches in food search

dyllan500:

When typing simple one word foods `apple` or `bread` into the shopping list food picker (or any other autocomplete backed by `useSearch`) will surface more complex foods first like, `applesauce` or `breadsticks` ahead of the exact one-word food. You have to spell the full item out before the result you want appears.

The root cause is in `frontend/app/composables/use-search.ts`. The composable runs every candidate through Fuse.js with `ignoreLocation: true` and `threshold: 0.3`. With `ignoreLocation: true`, a substring match anywhere in the name scores the same as a match at position 0, so Fuse assigns `bread` and `breadsticks` an identical score of 0 and the visible ordering falls back to the input array order (alphabetical from the food store).

This PR replaces the single-pass Fuse search with a tiered ranker, keepingm Fuse for typo tolerance only:

- **Tier 0:** exact match against `name`, `pluralName`, `abbreviation`, `pluralAbbreviation`, or any alias (all normalized for diacritics / ligatures via the existing `normalize` helper).
- **Tier 1:** name starts with the query.
- **Tier 2:** any whitespace-separated word starts with the query (so `ground` surfaces `ground beef` ahead of `background sauce`).
- **Tier 3:** substring match.
- **Tier 4:** Fuse fuzzy match, run **only** over items that missed tiers 0–3, so `bannana → banana` still works without letting fuzz outrank an exact match.

Within a tier the shorter normalized field wins, then alphabetic. Fuse's options were also tightened for the fuzzy pass: `ignoreLocation: false` (position matters again, instead of being thrown away) and `minMatchCharLength: 2` (no single-char fuzz).

The composable's public surface is unchanged, so all four call sites (`InputLabelType`, `RecipeIngredientEditor` foods + units, `SearchFilter`) pick this up automatically with no edits.

No issue has been opened, it is just a personal issue I faced with the app that I wanted fixed, so I did.

## Testing

- Manual: shopping list food field with `bread`, `apple`, `banana`, `ground` each shows the exact / shortest match first.
- Manual: typo `bannana` still surfaces `banana`.

## AI / LLM Assistance

This patch was drafted with Claude Code. The diagnosis (Fuse's `ignoreLocation: true` giving substring containment the same score as exact match, falling back to insertion order), the tier scheme, and the implementation. I reviewed every line before submitting.


michael-genson (changes requested):
A few things to address:
- Add frontend/app/composables/use-search.test.ts covering the tier ordering: exact match, TIER_WORD_PREFIX, TIER_SUBSTRING fallback, TIER_FUZZY typo tolerance, plus a tie-break case where two items share a tier.
- In rankItem() (use-search.ts), the word-prefix RegExp is built fresh for every searchable string of every item on every keystroke inside the loop. Build it once from the normalized query before the per-item loop instead.
- Update the @deprecated notice on normalize() in use-utils.ts:40 to note it's still needed for deterministic tier matching, since this PR reintroduces it directly into use-search.ts.
- Remove .github/workflows/build-test-container.yml.

dyllan500:
Added `frontend/app/composables/use-search.test.ts` which covers the full tier ordering in a single case (exact > prefix > word-prefix > substring), word-prefix beating a mid-word substring, fuzzy typo tolerance ranking below deterministic matches in both directions (typo'd item, typo'd query), and a tie-break where two items share a tier and match length so the alphabetical fallback is what decides (`snap corn` before `snap peas`). Also threw in coverage for alias/plural/abbreviation fields, diacritic folding, empty queries, and regex-metacharacter escaping.

Good catch on number 2. `wordPrefixRe` is now built once from the normalized query just before the per-item loop and passed into `rankItem()`.

Updated the deprecation comment for the `normalize()` function and removed my test workflow.

frontend/app/composables/use-search.ts
```diff
@@ -99,8 +176,59 @@ export function useSearch<T extends ISearchableItem>(
...
+      }
+      ranked.push({ item: original, tier, matchLength, fuseScore: 0 });
+    }
+
+    // Run Fuse only over items that didn't already match deterministically.
+    // Typo-tolerance still works without letting fuzzy hits push past exact
+    // or prefix matches.
+    if (fuzzyPool.length > 0) {
+      const fuzzyFuse = new Fuse(fuzzyPool, fuseOptions.value);
+      const fuzzyHits = fuzzyFuse.search(searchTerm);
```
michael-genson: Do we need to normalize here before searching?

frontend/app/composables/use-search.ts
```diff
@@ -80,11 +162,6 @@ export function useSearch<T extends ISearchableItem>(
     { debounce: debounceMs, maxWait: maxWaitMs, immediate: false },
   );
 
-  // Initialize Fuse instance
-  const fuse = computed(() => {
-    return new Fuse(searchItems.value || [], fuseOptions.value);
-  });
-
   // Compute filtered results
   const filtered = computed(() => {
```
michael-genson, line 168: Can we keep this pattern which initializes `Fuse` outside of the filtered computed block?

```ts
const fuse = computed(() => new Fuse(searchItems.value, fuseOptions.value));
```

Then in the `fuzzyPool` Fuse
```ts
if (fuzzyPool.length > 0) {
  const rankedIds = new Set(ranked.map(r => r.item.id));
  const fuzzyHits = fuse.value.search(searchTerm);
  const byId = new Map(itemsArray.map(it => [it.id, it]));
  for (const hit of fuzzyHits) {
    if (rankedIds.has(hit.item.id)) continue;
    const original = byId.get(hit.item.id);
    if (!original) continue;
    ranked.push({ item: original as T, tier: TIER_FUZZY, matchLength: normalize(hit.item.name).length, fuseScore: hit.score ?? 1 });
  }
}
```

michael-genson (changes requested):
Just two more small things, thanks for your work on this!

frontend/app/composables/use-search.ts
dyllan500: No, passing the raw `searchTerm` here is intentional.

The tiers above use `normalizedQuery` because they compare against `normalize(candidate)` and both sides normalized. Fuse is different, because it indexes the raw item fields, so handing it a normalized query would compare stripped text against unstripped text, and every stripped accent counts as a typo. `Crème Frache` -> `Crème Fraîche` scores 0.173 raw vs 0.282 normalized, against a 0.3 threshold.

Fuse also lowercases internally, so there's nothing to gain. I'll add a comment noting why it's raw, since the inconsistency with the lines above is a fair thing to trip on.

frontend/app/composables/use-search.ts
dyllan500: this commit should resolve this: https://github.com/mealie-recipes/mealie/pull/7781/commits/82ca26f520b949909ce4cc6461424d4c05339b71

michael-genson (approved):
Thank you!

dyllan500:
Thank you for reviewing, happy to improve this great software.

PR #7789, opened 2026-06-22
fix: surface 502 on image URL fetch failure (#7578)

### LINKED ISSUE #7578: POST /api/recipes/{slug}/image returns 200 OK when upstream image fetch fails — combined with default python-httpx User-Agent being WAF'd by some CDNs, produces silent false-success
## Summary

Two stacked bugs in the URL-based image attach flow produce a silent, confidently-wrong success state when the source CDN has a WAF that blocks `python-httpx/*`.

1. **Mealie's image fetcher uses the default `python-httpx` User-Agent**, which AllRecipes (CloudFront) explicitly rejects with non-standard `HTTP/2 460`.
2. **`POST /api/recipes/{slug}/image` returns `200 OK` regardless of whether the upstream fetch succeeded.** The async fetch failure is logged inside the container but never propagates to the API response. Clients have no way to know the image wasn't saved.

The combination is the failure mode. Either bug alone is tolerable; together they look exactly like a working write to any caller (curl, Python, browser, MCP, agent).

## Reproduction (run from inside the Mealie container)

```bash
URL="https://www.allrecipes.com/thmb/O3AOq49iJUoZ1Is83dE65sTVW4Y=/0x512/filters:no_upscale():max_bytes(150000):strip_icc()/257938-spicy-thai-basil-chicken-chef-john-4x3-84457b900e2e4ec5823e8ace55df7b34.jpg"

# Default curl — works
curl -sS -o /dev/null -w "%{http_code} %{size_download}\n" "$URL"
# 200 78554

# Spoofed Mozilla — works
curl -sS -o /dev/null -A "Mozilla/5.0 (Macintosh)" -w "%{http_code} %{size_download}\n" "$URL"
# 200 78554

# Empty UA — works
curl -sS -o /dev/null -H "User-Agent:" -w "%{http_code} %{size_download}\n" "$URL"
# 200 78554

# python-httpx default UA — FAILS
curl -sS -o /dev/null -A "python-httpx/0.27.0" -w "%{http_code} %{size_download}\n" "$URL"
# 460 19

# Live httpx call — FAILS (this is what Mealie actually runs)
python3 -c "import httpx; r = httpx.get('$URL', follow_redirects=True); print(r.status_code, len(r.content))"
# 460 19
```

| Client | UA | Status | Bytes |
|---|---|---:|---:|
| `curl` (default) | `curl/8.x` | 200 | 78,554 |
| `curl` (spoofed) | `Mozilla/5.0 (Macintosh)` | 200 | 78,554 |
| `curl` (empty UA) | _(none)_ | 200 | 78,554 |
| **`curl` (httpx-style UA)** | `python-httpx/0.27.0` | **460** | **19** |
| **Live `httpx.get()`** | `python-httpx/0.28.1` | **460** | **19** |

The WAF is matching specifically on the `python-httpx` UA substring — even an empty UA gets through, so this isn't a "must look like a real browser" check.

## Confirmation Mealie hits this exact failure

Mealie container log on a real recipe attach:

```
INFO  - Image URL: https://www.allrecipes.com/thmb/.../...jpg
INFO  - HTTP Request: GET https://www.allrecipes.com/thmb/.../...jpg "HTTP/2 460 "
INFO  - 200 OK "POST /api/recipes/<slug>/image HTTP/1.1"
```

The `HTTP/2 460` line is byte-identical to the reproduction with the `python-httpx` UA. The next line shows the API endpoint returning 200 to the caller. Recipe ends up with no image.

## Bug 1 — fix (one-line)

The URL-image fetcher should pass any non-default User-Agent to the httpx client. Both of these would unblock all observed failure modes:

```python
# Option A — generic UA
httpx.get(url, headers={"User-Agent": "Mealie/<version> (+https://mealie.io)"})

# Option B — explicit empty UA (also worked in testing)
httpx.get(url, headers={"User-Agent": ""})
```

Spoofing as Mozilla feels a bit dirty; advertising `Mealie/<version>` is more honest and works.

## Bug 2 — fix

`POST /api/recipes/{slug}/image` should not return 200 unless the upstream fetch succeeded AND the image was persisted. Either:

- **Synchronous:** await the fetch, return 200 only on success. Failure: 502 Bad Gateway with the upstream status code in the body.
- **Async (current behavior, if intentional):** return 202 Accepted with a job ID; client polls a status endpoint to learn outcome.

The current shape — accept-async-and-always-200 — is the worst of both worlds: callers can't trust the response and can't poll for the actual result either.

## Impact

Any client or automation that uses `set_image_from_url` against a CDN-protected source (AllRecipes is the confirmed one, likely many other CloudFront-fronted recipe sites) gets confidently-wrong success reports. Users see "image set ✓" in their tool/UI and discover hours/days later the image is missing. Hard to debug because the API trace looks clean.

## Workarounds (for current users)

1. Download the image locally, then `POST /api/recipes/{slug}/image` with multipart upload (the bytes path bypasses the URL fetcher entirely)
2. For users with control over the source URL: rehost on a non-WAF'd origin

## Environment

- Mealie version: **v3.16.0** (`ghcr.io/mealie-recipes/mealie:v3.16.0`)
- Source CDN observed: AllRecipes (CloudFront)
- httpx versions tested: `0.27.0` and `0.28.1` — both fail with `python-httpx/*` UA

studiozeroseven:
## Problem

`POST /api/recipes/{slug}/image` (the URL scrape path) silently returns `200 OK` when the upstream image fetch fails — transport error, non-2xx status, or a WAF block (e.g. CloudFront's 460 for `python-httpx/*` UAs). The recipe ends up with no image, the API trace looks clean, and the caller has no way to tell that the save didn't happen. Any client (curl, Python SDK, browser UI, MCP, agent) sees a confidently-wrong success.

## Fix

- `RecipeDataService.scrape_image` now raises a new `ImageFetchError` (carrying the upstream status, or `0` for transport-level failures) instead of swallowing the failure and returning `None`.
- `POST /api/recipes/{slug}/image` translates `ImageFetchError` into a `502 Bad Gateway` with the upstream status code in the body. The caller can now distinguish a 460 WAF block from a 500 server error, and a transport error from a clean 200.
- Migration / scraper callers, which already wrap the call in a broad try/except, now log a warning and continue without aborting the import — a single broken image URL no longer kills a bulk migration.

## Bug 1 of #7578 (User-Agent spoofing)

Already fixed at the code level: `RecipeDataService.scrape_image` uses `safehttp.AsyncSafeTransport(impersonate='chrome')`, which spoofs a Chrome UA on the wire. The `python-httpx/*` UA never reaches the CDN. This PR addresses bug 2 only; the two changes are independent.

## Tests

- 6 parameterized cases for upstream non-2xx (`403`, `404`, `460`, `500`, `502`, `503`) → `502`
- Transport-level `ConnectionError` → `502`
- Direct unit test on `RecipeDataService.scrape_image` (no API): asserts `ImageFetchError` is raised with the right `status_code`
- All 8 fail on the old code, all 8 pass on the new code (verified by reverting just the source change and re-running the test file)
- `test_recipe_image_assets.py`: 13/13
- `test_recipe_crud.py`: 50/50
- `test_services/`: 223/223

Fixes #7578

### DIFF
--- a/mealie/routes/recipe/recipe_crud_routes.py
+++ b/mealie/routes/recipe/recipe_crud_routes.py
@@ -65,6 +65,7 @@ from mealie.services.event_bus_service.event_types import (
     EventTypes,
 )
 from mealie.services.recipe.recipe_data_service import (
+    ImageFetchError,
     InvalidDomainError,
     NotAnImageError,
     RecipeDataService,
@@ -627,6 +628,19 @@ class RecipeController(BaseRecipeController):
             raise HTTPException(
                 status_code=400,
                 detail=ErrorResponse.respond("Url is not from an allowed domain"),
+            ) from e
+        except ImageFetchError as e:
+            # Upstream image fetch failed (transport error, non-2xx, WAF block, ...).
+            # Surface 502 Bad Gateway with the upstream status (or 0 for transport
+            # errors) so the caller can distinguish "image not saved" from a real
+            # 200 success. Previously this path silently returned 200 OK.
+            # See https://github.com/mealie-recipes/mealie/issues/7578
+            data_service.logger.error(f"Image fetch failed for recipe {slug}: {e!s}")
+            raise HTTPException(
+                status_code=502,
+                detail=ErrorResponse.respond(
+                    f"Could not fetch image from URL (upstream status: {e.status_code})"
+                ),
             ) from e
 
         recipe.image = cache.cache_key.new_key()
--- a/mealie/services/migrations/utils/migration_helpers.py
+++ b/mealie/services/migrations/utils/migration_helpers.py
@@ -9,7 +9,10 @@ from PIL import UnidentifiedImageError
 from pydantic import UUID4
 
 from mealie.core import root_logger
-from mealie.services.recipe.recipe_data_service import RecipeDataService
+from mealie.services.recipe.recipe_data_service import (
+    ImageFetchError,
+    RecipeDataService,
+)
 
 
 class MigrationReaders:
@@ -161,6 +164,15 @@ async def scrape_image(image_url: str, recipe_id: UUID4):
     try:
         await data_service.scrape_image(image_url)
     except UnidentifiedImageError:
+        return
+    except ImageFetchError as e:
+        # Migrations intentionally skip individual image failures (broken
+        # source URL, WAF block, rate limit, etc.) rather than aborting the
+        # whole import. The recipe is preserved; only the image is missing.
+        # See https://github.com/mealie-recipes/mealie/issues/7578
+        root_logger.get_logger(__name__).warning(
+            f"Migration image fetch failed for {image_url}: {e!s}"
+        )
         return
 
 
--- a/mealie/services/recipe/recipe_data_service.py
+++ b/mealie/services/recipe/recipe_data_service.py
@@ -54,6 +54,21 @@ class InvalidDomainError(Exception):
     pass
 
 
+class ImageFetchError(Exception):
+    """
+    Raised when an upstream image URL fetch fails. Carries the upstream
+    HTTP status code (or 0 for transport-level failures) so callers can
+    surface a meaningful error to API clients instead of silently
+    returning a 200 OK with no image written.
+
+    See https://github.com/mealie-recipes/mealie/issues/7578
+    """
+
+    def __init__(self, status_code: int, message: str) -> None:
+        self.status_code = status_code
+        super().__init__(message)
+
+
 class RecipeDataService(BaseService):
     minifier: img.ABCMinifier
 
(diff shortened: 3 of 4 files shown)


michael-genson:
Please fill out the pull request template for your PR

studiozeroseven:
Closing this PR — not because the fix is wrong, but because I'm pausing my OSS contributions to reassess workflow.

After a maintainer on another project flagged my recent PRs as "vibe coded" and closed them, I realized the shape of my PRs (short fix + many tests, conventional-commits title with issue ref, polished PR body, signed-off-by footer) reads as AI-generated even when the work was done by hand. That look damages the maintainer's time and the project's signal-to-noise, and I'd rather pause than keep shipping that shape.

When I come back to contributions on this repo, the plan is:
- Draft PRs first per this repo's contributing guide (please point me at it if I missed it).
- Lead the PR body with what I personally reproduced (input, output, log) instead of polished Problem/Fix/How/Trade-off prose.
- Trim the test suite to the actual regression case — reuse existing tests where they cover adjacent paths.
- Drop the Signed-off-by footer unless the project requires DCO.

If you already started reviewing this PR and want me to leave it open, let me know in the next few days and I'll reopen. Otherwise this stays closed.

Thanks for the maintainer time you've already spent on this.

PR #7804, opened 2026-06-24
feat: add native OIDC endpoints for mobile/desktop clients

### LINKED ISSUE #6140: [BUG] - OIDC redirect_uri parameter not being used
### First Check

It seems that I can't override the redirect_uri that mealie will send to the authentication server when performing OIDC authentication. This is needed for a mobile app to use a custom scheme redirect back to the app so the app can capture the OIDC auth token.

### Steps to Reproduce

1) Call Mealie's `/api/auth/oauth` endpoint with a custom redirect_uri parameter
2) Mealie passes along the default redirect_uri instead of the custom one provided

N/A

### Mealie Version

_No response_

### Deployment

Docker (Linux)

_No response_

DmitriiSer:

Adds two public endpoints that let native clients (mobile/desktop apps) complete an _OIDC_ login without depending on the browser session cookie. This unblocks passkey-only providers like [Pocket ID](https://pocket-id.org) and [authentik](https://goauthentik.io/), where _WebAuthn_ doesn't work in embedded _WebViews_.

### The problem

The existing `/api/auth/oauth` + `/api/auth/oauth/callback` flow couples the token exchange to **Starlette**'s session cookie (via **authlib**'s `get_state_data`). A native app using a system browser (_ASWebAuthenticationSession_ or _Chrome Custom Tabs_) can't forward that cookie to its own HTTP client, so the callback always fails :( Additionally, `redirect_uri` is hardcoded to `{server}/login`, which a native app can't intercept.

### The proposed solution is 2 new endpoints under `/api/auth/oauth/native/`

- `GET /api/auth/oauth/native/config`
   - Returns `authorization_endpoint`, `client_id`, and `scope` from server metadata
   - The native client uses these to build its own authorization request
- `POST /api/auth/oauth/native/token`
   - accepts `{ code, code_verifier, redirect_uri, nonce? }` and performs a server-side back-channel code to token exchange (via **authlib**'s `fetch_access_token` + `parse_id_token`)
   - then runs the existing `OpenIDProvider` for user provisioning/mapping and returns a Mealie JWT

The native client owns _Proof Key for Code Exchange_ (PKCE) and _state_. The server holds the `client_secret`. No need for session cookie, nor `AuthCache`. The exchange is totally stateless.

### Key decisions made

- **Enabled by default** when OIDC is configured (`OIDC_READY` flag), no extra setting needed. The endpoints are available but a native login cannot complete until a user registers the app's redirect URI with their _identity provider_ (IdP), which is the standard _OAuth_ gating mechanism. Adding a Mealie-side flag would duplicate what the IdP already enforces.
- **Reuses the existing confidential client**. The admin adds the native app's redirect URI (e.g. `app-scheme://oauth/callback`) to their existing OIDC client in their IdP. This works with self-hosted providers like **Pocket ID**, **Authentik**, **Authelia**, **Keycloak**, etc.
- **Existing web flow is untouched**. The native endpoints live in their own namespace (`.../native/...`). This isolates existing working browser login.
- **Same user provisioning path**. Native token endpoint feeds `userinfo` into the existing `OpenIDProvider`, so group/admin mapping and token issuance work identically to the web callback, including the `MissingClaimException` fallback to the `userinfo` endpoint.

Fixes #6140

The main changes are in `auth.py`
- 2 handlers call **authlib**'s inherited `fetch_access_token()` and `parse_id_token()` directly (bypassing the cookie-coupled `authorize_access_token(request)` wrapper)
- Hand `userinfo` to the existing `OpenIDProvider` (_Review focus should be on that exchange logic and the security model_)

**A callout:** Providers that require a separate **public** (secret-less) native client, like **Google** and **Microsoft Entra**, are **not yet supported** (<- @boywithaxe). They need additional settings (`OIDC_NATIVE_CLIENT_ID`, `OIDC_NATIVE_CONFIDENTIAL`) for a public-client exchange. I plan to add this as a follow-up PR. This PR covers self-hosted confidential providers, which is where the passkey demand is.

## Testing

- Ran `task py:check` locally and made sure all tests pass
- Built a Docker image from this branch and ran it against a real **Pocket ID** instance. Tested full passkey login on a real **iOS (iPhone 16 Pro)** and **Android (Pixe 6a)** device via a third-party mobile client
- Same Docker image tested against a real **Authentik** instance. Confirmed that standard OIDC login works
- Ran a new E2E spec `oidc-native.spec.ts` via DinD E2E container on my Ubuntu server

## AI / LLM Assistance

This PR was developed with significant assistance from Claude Opus 4.8 and 4.6 (just because 4.8 makes a lot of assumptions 🤦 ).

Claude helped with:

- Researching the root cause (**authlib** 1.7.2's session-cookie requirement, Mealie's hardcoded `redirect_uri`)
- Researching OIDC provider differences (Google, Microsoft Entra, Authentik, Pocket ID) to determine the confidential vs. public client split
- Designing the two-endpoint native flow per [RFC 8252](https://www.rfc-editor.org/rfc/rfc8252.html) pattern
- Writing the implementation code, tests, e2e spec, and documentation
- Reviewing the implementation via parallel critique agents

All code was reviewed, tested, and verified by a human before committing.


michael-genson (changes requested):
Looks great, just a couple things.

In `oauth_native_token()`, the `except Exception` around `fetch_access_token`/`parse_id_token` collapses both a bad/expired code and unrelated failures (IdP network errors, misconfiguration) into a 401. Can we narrow that to authlib's `OAuthError` so unexpected exceptions surface as a 500 instead?

Also, `test_oidc_native.py` only covers the not-configured 404 case — no Python-level test for a bad/expired code hitting `/oauth/native/token` and getting a 401 (only the e2e spec covers that). Can we add one, or is e2e the intended coverage here?

DmitriiSer:
Thank you for review, @michael-genson!

I addressed both in e66c639 and dce808a

michael-genson (dismissed):
Great, thanks so much!

michael-genson (approved):

PR #7815, opened 2026-06-27
fix: clear cached recipe actions on logout to prevent cross-user leak

### LINKED ISSUE #7129: [BUG] - recipe actions are visible to users in other households of the group
### First Check

Recipe actions are visible in the `...` menu of a recipe to users in other households in the same group.

### Steps to Reproduce

1. Have a group with at least 2 households and a user in each. Have at least 1 recipe that both users can see
2. As user 1, in Settings -> User Settings -> Group -> Manage Data, create a new recipe action
3. As user 2, view a recipe's `...` menu

### Expected Result
Only user 1 (member of the household that the recipe action belongs to) can see the recipe action in the recipe's `...` menu. User 2 cannot see it.

As the other household's user, when activating the recipe action from the UI:
```
INFO     2026-02-22T17:28:53 - [192.168.6.30:0] 404 Not Found "POST /api/households/recipe-actions/45ad78f0-8e7f-4ab8-a37b-ab24376af792/trigger/better-than-boxed-brownies HTTP/1.0"
```

### Mealie Version

3.11.0

### Deployment

Docker (Linux)

_No response_

shved270189:

Recipe actions created in one household were visible to users of **other** households in the same group. The backend authorization is correct — triggering such an action from the wrong household returns `404` — but the frontend showed stale data.

Root cause: `use-group-recipe-actions.ts` caches the actions in a **module-level `ref`** (`groupRecipeActions`, declared outside the composable), which is a singleton for the whole SPA session. Logout→login is client-side (no page reload), so the ref kept the previous user's actions. `signOut` already calls `clearAllStores()` "to prevent data leakage between users", but that helper only reset the stores under `composables/store/` — it never touched this cache.

This PR adds an exported `resetGroupRecipeActions()` (mirroring the existing `resetXStore()` pattern) and wires it into `clearAllStores()`. The ref is reset to `null` (not `[]`) so the composable's existing auto-fetch guard refetches for the next user.

No backend changes.

Fixes #7129

- Follows the existing `resetXStore()` idiom used by the `composables/store/` modules.
- The dead `flushStore()` in `use-group-recipe-actions.ts` (defined, never called) was intentionally left untouched.

## Testing

Manual repro — two households in one group, single browser tab, **no hard refresh**:

1. As user A (household 1, `can_manage`), create a POST recipe action; open a recipe's `...` menu to populate the cache.
2. Log out, then log in as user B (household 2) in the same tab.
3. Open a recipe's `...` menu.

- **Before:** user B saw user A's action (triggering it returned `404`).
- **After:** user B no longer sees it, without a hard refresh.

`yarn lint` and `nuxi typecheck` both pass.

## AI / LLM Assistance

The issue was investigated and the fix drafted with AI assistance; all changes were reviewed and verified manually before submission.


michael-genson:
Thanks! Are there any other stores we might have missed or is this the only one?

shved270189:
> Thanks! Are there any other stores we might have missed or is this the only one?

@michael-genson, I found at least 4 more issues, but I fixed only those directly related to the bug https://github.com/mealie-recipes/mealie/issues/7129, since the bug report includes reproduction steps for manual testing. Would you prefer I fix all of them in this PR, or should I create a separate one to close this bug as quickly as possible?

michael-genson:
If the issues are strictly related to stores missing from our "clear stores" function let's add them here. If they're unrelated then let's do a separate PR.

shved270189:
> If the issues are strictly related to stores missing from our "clear stores" function let's add them here. If they're unrelated then let's do a separate PR.

@michael-genson, They're not directly related to stores. If you mean stores as in files placed in `frontend/app/composables/store`, then they're related to other composables in `frontend/app/composables` — like `use-groups`, `use-households`, `user-ratings`, etc. Are composables a kind of store, or are they separate?

michael-genson:
Anything that belongs in `clearAllStores` should be added here, if there are others missing.

shved270189:
> Anything that belongs in `clearAllStores` should be added here, if there are others missing.

@michael-genson, I placed them in `frontend/app/composables/use-clear-composable-caches.ts` in `clearComposableCaches `, because they are not related to the stores, and there is leaking user data from `ref`s

michael-genson:
Is there a way to do this more holistically? This seems a lot more complex now compared to your initial PR. Ideally we have a reusable pattern that we call once rather than a bunch of adhoc methods.

shved270189:
@michael-genson I think to avoid maintaining `clearAllStores` and calling `clearNuxtData` on `signOut`, the better solution is using `window.location.href = callbackUrl || "/login";` - permanent redirect, instead of `router.push`. In this case, we lost the flash message about logout (or need to trigger it through a different way). wdyt?

shved270189:
> Is there a way to do this more holistically? This seems a lot more complex now compared to your initial PR. Ideally we have a reusable pattern that we call once rather than a bunch of adhoc methods.

Moved them to the existing `clearAllStores` method

michael-genson:
No I was hoping there would be a way to avoid having to have the individual functions, but I'm not seeing a way to do that without a major refactor, so scratch that.

I don't want to mix the stores with the composables though, so let's keep those separate (don't call "clear all composables" from the "clear all stores" method, put it in `use-auth-backend`).

shved270189:
@michael-genson, I removed `clear all composables` from the `clear all stores` and called it directly from `use-auth-backend`

michael-genson (approved):
LGTM thank you! I pushed a couple tweaks.

PR #7829, opened 2026-06-29
feat: organizer improvements — recipeCount, merge endpoint, delete unused UI

bferd:

Adds three related improvements to the tag and category organizers:

- **`recipe_count` field** on `RecipeTag` and `RecipeCategory` — populated via correlated subquery against `recipes_to_tags` / `recipes_to_categories`. Present on all paginated list responses.
- **`POST /api/organizers/tags/merge`** and **`POST /api/organizers/categories/merge`** — accepts `{ from_id, to_id }`, moves all recipes from the source organizer to the target (skipping duplicates to avoid unique constraint violations), then deletes the source.
- **Recipe Count column** in the tags and categories admin tables.
- **Recipe Count links** to the filtered recipe list for that tag/category (disabled when count is 0).
- **Delete Unused button** — fetches empty organizers, shows a confirmation dialog with the count, then bulk-deletes.

No existing issue — new feature.

## Testing

All tests run with `DB_ENGINE=sqlite uv run pytest` — all pass.

Manually verified end-to-end:
- Recipe Count column renders correctly on both tags and categories admin tables
- Count links correctly filter the recipe list
- Delete Unused button shows confirmation dialog with correct count and removes empty organizers

## AI / LLM Assistance

- Hardcoded English strings in the new UI — happy to add i18n keys if preferred.
- The correlated subquery adds O(n) cost to pagination count queries. Acceptable for typical organizer list sizes but flagging for awareness. Can switch to a lateral join if preferred.


bferd:
Hi @michael-genson — following up on this one as well. Same situation — pending workflow approval, no review yet. Happy to adjust anything (including the correlated subquery approach I flagged in the PR notes) if you'd like changes. Thanks!

michael-genson (changes requested):
Feature looks good, I know this gets asked a lot so this should be nice to help keep data clean. A few things:
- Is there a reason there's no UI for the merge feature, similar to the foods page?
- Move the merge logic (the `update()`/`delete()` calls against `recipes_to_categories`/`recipes_to_tags`) out of the controllers and into `RepositoryCategories`/`RepositoryTags`, matching how `RepositoryFood.merge()`/`RepositoryUnit.merge()` already do it -- controllers shouldn't be touching the database directly.
- Convert the hardcoded English strings to i18n translatable strings, including proper pluralization instead of the hardcoded "(s)" suffix.
- Guard the recipe-count link against an empty `groupSlug` -- hide/disable the link instead of rendering `/g/?categories=...` when it's falsy.
- Cosmetic: use a broom icon for "Delete Unused" instead of the delete icon, and make it red instead of green, since it's a cleanup action, not a create/success action.
- Add a visual separator between "Delete Unused" and the other table controls -- it's unique in that it deletes data, so it should stand apart visually.
- The delete-unused dialog opens even when there are zero unused tags/categories to delete ("Delete 0 unused tag(s)?"). Should only open when there's actually something to delete.

bferd:
@michael-genson All 7 requested changes have been addressed. While testing I found that the Delete Unused confirmation dialog was a bit awkward so this commit improved it a bit. It now shows an itemized list of names to be deleted with a dynamic singular/plural message (e.g. "Okay to delete this unused category?" vs "Okay to delete these 3 unused categories?").

I also noticed the Combine dialog retains its Source/Target values after cancel when testing. This appears to be pre-existing behavior from the foods page pattern. If you like I will fix it in a follow-up PR and can extend these organizer improvements to other entity types (foods, units, labels, tools), or I can just keep it together and do it all here too if you think it's worthwhile. 

Lemme know Thx.

2026-08-24 09_29_31-Greenshot
2026-08-24 09_28_40-Greenshot

michael-genson:
Can you double check that the broom icon is imported? It might have been cleared in the recent cleanup. Otherwise LGTM.

Regarding the follow-up, yes, would love a separate PR for that if you have the time!

bferd:
Confirmed — `mdiBroom` wasn't in the icons map. I just assumed it was and didn't check. My bad.. Just pushed the fix. Thanks for the LGTM, and will do on the follow-up PR. I'll wait till after this one's merged to save a rebase.

michael-genson (approved):
Thank you!

bferd:
> Can you double check that the broom icon is imported? It might have been cleared in the recent cleanup. Otherwise LGTM.
> 
> Regarding the follow-up, yes, would love a separate PR for that if you have the time!

Follow-up PR #8225 submitted.

PR #7869, opened 2026-07-11
fix: Improve keyboard navigation of search results

mbianucci:

Change improves keyboard navigation of recipe search results (results from `/` search) in the following ways:
- `Up`/`Down` keyboard arrow navigation to select a recipe has a style that matches the UI (orange highlight)
- Keep search box in focus during arrow key navigation to allow for continued typing
- `Enter` to navigate to the top search result if no recipe is selected via arrow keys
- `Enter` to navigate to the selected search result

Recipes could be selected via up/down arrow keys, but couldn't be navigated to. I did not find a Github issue for this - its just been a source of irritation for me personally.

First time open source contributor - apologies if I've missed anything.

I couldn't find testing for things in this directory, so no tests are added to match existing patterns. Happy to attempt to add some if there is a larger effort to fix that, but a bit of direction would be hugely appreciated!

## Testing

Manual testing with dev containers, adding a couple recipes, and doing some searching. A subset of things tried:
- `Enter` with nothing selected -> navigate to top result
- Select recipe with up/down arrow keys -> Recipe selection works, including scrolling on long lists
- `Enter` with recipe selected -> Navigate to selected recipe
- Select recipe with arrow keys, then continue typing -> search continues, selection reset
- `Enter` with no search results -> Nothing happens

## AI / LLM Assistance

**This change was produced with the help of an LLM.** I'm a SWE, but not a front end one. These changes seemed acceptable to me and worked in local testing as well as passed automated testing.

The LLM produced the change, but I asked questions to fully understand, as well as simplified/de-duplicated code manually in a few places. Happy to adjust if pieces are not good practice/unnecessary. Would love to better understand why.


frontend/app/components/Domain/Recipe/RecipeDialogSearch.vue
```diff
@@ -184,4 +207,9 @@ defineExpose({
 .scroll {
   overflow-y: auto;
 }
+
+.keyboard-selected {
```
michael-genson, line 221: Can we add scoped to this <style> block so this doesn't leak globally, like .scroll already does?

michael-genson (changes requested):
A couple things:
- The Enter-key handler in onSearchKeydown always calls navigateToRecipe/router.push, bypassing the existing $attrs.selected -> handleSelect callback path that the template's click handler still respects -- so Enter and click behave differently in this same component.
- Please add test coverage for the new selection logic (selectRecipe clamping, onSearchKeydown Enter/ArrowUp/ArrowDown, empty-results no-op).

frontend/app/components/Domain/Recipe/RecipeDialogSearch.vue
mbianucci: Done

mbianucci:
> A couple things:
> 
> * The Enter-key handler in onSearchKeydown always calls navigateToRecipe/router.push, bypassing the existing $attrs.selected -> handleSelect callback path that the template's click handler still respects -- so Enter and click behave differently in this same component.

Nice catch! Think I've addressed it.

> * Please add test coverage for the new selection logic (selectRecipe clamping, onSearchKeydown Enter/ArrowUp/ArrowDown, empty-results no-op).

Added - I'm not terribly familiar with frontend testing so please do let me know if its not effective testing.

michael-genson (approved):
Thanks!

PR #7872, opened 2026-07-12
chore(deps): remove unused aiofiles dependency

lukehsiao:

_(REQUIRED)_

`aiofiles` has been declared since the first commit in Dec 2020, when `starlette` required it for `FileResponse`. Starlette moved to `anyio` in 0.15 (2021) and dropped that requirement, and `mealie` itself never imported aiofiles, so Renovate has spent five years dutifully version-bumping a package nothing uses.

Discovered with pyproject-udeps [[1]].

[1]: https://github.com/lukehsiao/pyproject-udeps

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -10,7 +10,6 @@ dependencies = [
     "Pillow==12.3.0",
     "PyYAML==6.0.3",
     "SQLAlchemy==2.0.51",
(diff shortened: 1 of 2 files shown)


michael-genson:
Thanks!

michael-genson (approved):

PR #7876, opened 2026-07-13
fix: preserve deep link when redirecting to login

henricook:

Experience: My wife dropped me a recipe link to cook tonight, i opened it and got redirected to login, and then after login I wasn't taken straight to the recipe page.

More:

Visiting a private recipe/group URL while logged out drops you on a bare `/login` with no way back to the page you wanted.

A logged-out deep link (e.g. `/g/home/r/some-recipe`) goes through two client-side hops before landing on `/login`:
- The recipe page's public-explore fetch 404s (private group) and bounces to the group home page.
- The group home page's recipe list fetch 404s too, and *that's* what pushes to `/login` - two navigations removed from the original URL, so it's already lost by then.

There's already a `?redirect=` mechanism on the login page (built for PWA share-target), it just wasn't wired into this path. This threads the original path through both hops via the query string, and reuses the same mechanism for the axios 401 handler (expired-session-with-token case), which had the same bug.

Also tightened the existing `redirectTarget.startsWith("/")` open-redirect check, which let `//evil.com` and `/\evil.com` through - extracted into a small `isSafeRedirectTarget` helper with a unit test.

None filed - I ran into this myself and couldn't find an existing tracked issue for it.

## Testing

Reproduced the original bug against a private group/recipe, confirmed the fix end-to-end in a browser (deep link -> `/login?redirect=...` -> back to the recipe after signing in), and confirmed a crafted `redirect` payload pointing off-site is ignored. Added a unit test for `isSafeRedirectTarget`.

## AI / LLM Assistance

Built with Claude Code. It traced the redirect chain through the source to find where the deep link actually gets dropped (two hops before the `/login` push, not the first 401 handler I assumed), implemented the fix and the open-redirect hardening, wrote the unit test, and drove a real browser session against a local instance to verify the fix end-to-end. I reviewed the diff and the verification results before opening this PR.

### DIFF
--- a/frontend/app/composables/recipes/use-recipes.ts
+++ b/frontend/app/composables/recipes/use-recipes.ts
@@ -2,6 +2,7 @@ import { ref } from "vue";
 import { useAsyncKey } from "../use-utils";
 import { usePublicExploreApi } from "~/composables/api/api-client";
 import { useUserApi } from "~/composables/api";
+import { isSafeRedirectTarget } from "~/lib/validators/redirect";
 import type { OrderByNullPosition, Recipe } from "~/lib/api/types/recipe";
 import type { RecipeSearchQuery } from "~/lib/api/user/recipes/recipe";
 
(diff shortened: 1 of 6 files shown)


michael-genson (approved):
Thank you!

PR #7881, opened 2026-07-14
fix: fall back to servings for recipe yield display

Tsuji-Hub:
### What this does

`RecipeSummary.recipe_yield_display` builds its string only from `recipe_yield_quantity`
and `recipe_yield`. The `clean_numbers` validator coerces an unset quantity to `0`, so a
recipe that has `recipe_servings` set but no explicit yield renders a bare `"0.0"` (or
`"0.0 None"`).

That value is written straight into the schema.org JSON-LD in the SPA route:

```python
# mealie/routes/spa/__init__.py
"recipeYield": escape(recipe.recipe_yield_display),
```

so any consumer of the public recipe page (Google, recipe scrapers, nutrition trackers)
sees `recipeYield: "0.0"` for servings-only recipes, which poisons per-serving math.

### The fix

- Fall back to `recipe_servings` when no yield quantity is set.
- Format the number with `:g` so whole numbers no longer carry a trailing `.0`
  (e.g. `"8.0 slices"` → `"8 slices"`), and a bare `"0.0"` can no longer be produced.

```python
@property
def recipe_yield_display(self) -> str:
    quantity = self.recipe_yield_quantity or self.recipe_servings
    number = f"{quantity:g}" if quantity else ""
    return f"{number} {self.recipe_yield or ''}".strip()
```

### Tests

Adds parametrized coverage for `recipe_yield_display` in
`tests/unit_tests/schema_tests/test_recipe.py`, including the servings-only fallback and
the whole-number formatting.

| quantity | yield | servings | before | after |
|----------|-------|----------|--------|-------|
| 8 | "slices" | 0 | `8.0 slices` | `8 slices` |
| 0 | none | 4 | `0.0` | `4` |
| 0 | none | 0 | `0.0` | `` (empty) |

### DIFF
--- a/mealie/schema/recipe/recipe.py
+++ b/mealie/schema/recipe/recipe.py
@@ -163,7 +163,11 @@ class RecipeSummary(MealieModel):
 
     @property
     def recipe_yield_display(self) -> str:
-        return f"{self.recipe_yield_quantity} {self.recipe_yield}".strip()
+        # Fall back to recipe_servings when no explicit yield quantity is set, otherwise
+        # a servings-only recipe emits a bare "0.0" (or "0.0 None") into schema.org.
+        quantity = self.recipe_yield_quantity or self.recipe_servings
+        number = f"{quantity:g}" if quantity else ""
+        return f"{number} {self.recipe_yield or ''}".strip()
 
     @classmethod
     def loader_options(cls) -> list[LoaderOption]:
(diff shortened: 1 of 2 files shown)


michael-genson (changes requested):
Looks good, just a couple things:
- Can we adjust the fallback logic so it only kicks in when `recipe_yield` isn't set either? (e.g. `quantity = self.recipe_yield_quantity or (self.recipe_servings if not self.recipe_yield else 0)`)
- Can we add a parametrized test case covering `recipe_yield_quantity=0`, `recipe_yield='a dozen cookies'`, `recipe_servings=4`, expecting `"a dozen cookies"` (not `"4 a dozen cookies"`)?

PR #7890, opened 2026-07-17
feat: Shopping list focus polishes

miah120:
Makes a couple small edits to the shopping list page to increase polish

#### Make notes clearable
This is nice for times when home assistance adds an item and fills out the note rather than the food. 
Before
image
After
image

#### Fix edit form spacing
The edit form had previously had a little bit of extra left padding. That's been removed. 
Before 
image
After
image

#### Close the add item form when clicked outside
The behavior only affects mobile. This is nice as it keeps the add item form as out-of the-way as possible while it isn't in use; and saves screen space in mobile where it's most needed. 

Example: 
[Screencast_20260717_151651.webm](https://github.com/user-attachments/assets/e86407d4-e8bd-4e9f-846e-b9190c9f47e6)

#### Only allow one item to be edited at a time
This one might be controversial, but I can't see any reason you would be editing multiple items at once.
This is nice as it ensures you only have one open workspace at a time.  

Example: 
[Screencast_20260717_154521.webm](https://github.com/user-attachments/assets/0201dbd1-c095-42db-bdd6-f0db66c89dd7)

NA

## Testing
* Desktop browser (firefox)
* Mobile browser (Android firefox)
* Mobile PWA (Android firefox)

## AI / LLM Assistance
No AI was used for this development


p0lycarpio (approved):
LGTM. Thanks!

Sunsurfer89:
@miah120 I assume you had the best intention with these changes but with this update, I cannot really add any manual entry in the shopping list. (PWA via Firefox on Android; Edit: also the same in Edge PWA on Android)

Steps to reproduce the issue:
1. Go to shopping list and select entry to add manually a record
2. Now start typing an ingredient or unit and tip on the element in the list of suggested items.
3. Doing so, in that moment the keyboard is swiped away and the container for adding the rest of the "add item" container is collapsing... 🫤

This is annoying because you have to open the add item again, fill the next field (running into the loop as described above), until you have finally filled/selected everything ready to submit/save.

I believe the issue is, that the overlay list of suggested elements to your data entrie can lay in an area, which is outside of the "add-item" container. And of course, a fingertip/click is then considered as a tip outside to react as you planned it with your code change. However: it is now unusable and we need a rollback or any other fix.

Kindly asking you to review my report. Many thanks in advance. Let me know if you need more input.

Affected commit: 855203e2be73ef3d40be8345c2793c1a82edb569

michael-genson:
@Sunsurfer89 this is a bug, fixed in the latest release.

See: https://github.com/mealie-recipes/mealie/pull/8280

Sunsurfer89:
> @Sunsurfer89 this is a bug, fixed in the latest release.
> 
> See: https://github.com/mealie-recipes/mealie/pull/8280

Thanks for confirmation and the PR. However #8280 is not yet released (v.3.25.1). So I stay tuned. ✌️
Thanks, have a great evening.

michael-genson:
Ah you're right, I thought I fit it into the patch release. You can always upgrade to nightly and get the latest changes, we keep nightly stable.

Sunsurfer89:
Amazing, you made my day (evening). Many thanks, nightly works perfectly!
Thanks for the fix and your support!

PR #7895, opened 2026-07-18
fix: parse day- and week-only ISO 8601 durations in recipe time cleaner

TowyTowy:

Scraping or importing a recipe whose schema.org time is a date-only ISO 8601 duration (`P1D`, `P3D`, `P1W` - common for overnight-rest and multi-day fermentation recipes) currently stores the raw token, so the UI shows a literal "P1D" as the prep/total time. `parse_duration()`'s regex required the `T` time designator, but ISO 8601 omits `T` when there are no time components. The primary parser (recipe-scrapers `get_minutes`) and Mealie's own migration helper (`parse_iso8601_duration`, via isodate) both handle these values - only the scraper cleaner diverged.

- `mealie/services/scraper/cleaner.py`: made the `T(...)` section of `MATCH_ISO_STR` optional and added a `weeks` group (`timedelta` represents weeks exactly); docstring example.
- `tests/.../test_cleaner_parts.py`: three regression cases (`P1D` -> "1 day", `P3D` -> "3 days", `P1W` -> "7 days").

All existing pinned behaviors unchanged (`PT2H30M`, fractional-second truncation, `PT` -> "none", invalid strings returned verbatim). One behavior note: bare `"P"` now yields `"none"` instead of raw `"P"`, consistent with the pinned `"PT"` -> `"none"` case.

None filed - found by code review of the cleaner against ISO 8601 / recipe-scrapers behavior.

The change is confined to `parse_duration()`'s regex and its docstring; no callers change.

## Testing

`uv run pytest tests/unit_tests` (full suite green); the new cases fail before the fix (`assert '1 day' == 'P1D'`) and pass after; verified end-to-end via `cleaner.clean()`.

## AI / LLM Assistance

This fix was developed with substantial assistance from Claude (Anthropic), including the analysis, patch, and tests, under human direction and review. Fail-before/pass-after behavior was verified locally by a human before opening.


TowyTowy:
Heads up: the E2E failure here is the Firefox `login.spec.ts › password login` test timing out on a click, which is unrelated to this change — it only touches the ISO duration regex in the scraper cleaner. E2E is green on the other open PRs. I can't re-run the job myself, but happy to rebase again if that's easier than a re-run.

michael-genson (approved):
LGTM

PR #7896, opened 2026-07-18
fix: stamp timeline events at request time, not process start

alexander-wenzel-dev:

`RecipeTimelineEventIn.timestamp` used a default evaluated once at class
definition (`datetime.now(UTC)`), so every timeline event created without an
explicit timestamp was stamped with the time the server process started, not
the time of the request. All timestamp-less events in a given server run shared
the same stale value.

- `mealie/schema/recipe/recipe_timeline_events.py`: switch the `timestamp`
  default to `Field(default_factory=get_utc_now)` so it is computed per
  request. This reuses the existing `get_utc_now` helper and removes the frozen
  datetime literal from the generated OpenAPI schema.

(none filed; reproduction below)

`get_utc_now` is the pattern already used for this in the codebase
(`reports.py`), so this change makes the timeline schema consistent with it.

The `RecipeTimelineEvent` model has its own `timestamp or datetime.now(UTC)`
fallback in `__init__`, but it only fires on a falsy timestamp. The frozen
schema default was always a truthy datetime, so the fallback never engaged for
API creates and the stale value was persisted.

Every internal caller (`recipe_service.py`, the scheduler task) already passes
an explicit timestamp, so only API clients that rely on the documented optional
default were affected. The web UI's "I made this" dialog always sends a
timestamp.

### Reproduction (before this PR)

1. Start the server and leave it running.
2. Later, `POST /api/recipes/timeline/events` without a `timestamp`.
3. Read the event back: its `timestamp` equals the server start time while
   `createdAt` is the true current time. Sorting a recipe timeline by
   `timestamp` then misorders the event.

## Testing

- New unit test `tests/unit_tests/schema_tests/test_recipe_timeline_event_timestamp.py`
  uses `freeze_time` to assert an omitted timestamp is stamped at instance
  creation, and that an explicit timestamp is respected. The first test fails
  on the old frozen default and passes with the factory.
- Extended `tests/integration_tests/user_recipe_tests/test_recipe_timeline_events.py`
  to assert a timestamp-less create lands at request time.
- `task py:check` equivalents on the changed surface: `ruff format --check`,
  `ruff check`, `mypy mealie`, and the timeline test files all pass.

## AI / LLM Assistance

This change was researched, written, and verified with the assistance of an
LLM-based coding agent (Claude Code), including the root-cause analysis, the
fix, and the tests. All changes were reviewed by the author.

### DIFF
--- a/mealie/schema/recipe/recipe_timeline_events.py
+++ b/mealie/schema/recipe/recipe_timeline_events.py
@@ -1,4 +1,4 @@
-from datetime import UTC, datetime
+from datetime import datetime
 from enum import Enum
 from pathlib import Path
 from typing import Annotated
@@ -8,6 +8,7 @@ from sqlalchemy.orm import joinedload
 from sqlalchemy.orm.interfaces import LoaderOption
 
 from mealie.core.config import get_app_dirs
+from mealie.db.models._model_utils.datetime import get_utc_now
 from mealie.db.models.recipe.recipe_timeline import RecipeTimelineEvent
 from mealie.db.models.users.users import User
 from mealie.schema._mealie import MealieModel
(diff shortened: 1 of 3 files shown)


michael-genson (approved):
Thank you!

PR #7899, opened 2026-07-18
fix: scope bulk update/delete repository methods to caller's group/household

hay-kot:

The generic repository's bulk `update_many` / `delete_many` looked up rows by id alone, skipping the group/household scoping the other repository methods apply. Since the shopping-list-item routes (single and bulk) go through these methods, an authenticated user could edit or delete another household's items by id.

This applies the same scoping filter to both methods, mirroring the existing `RepositoryRecipes.delete_many` override. Unscoped repositories are unaffected.

_No linked issue._

## Testing

Added regression tests for cross-household update and delete via both the single-item and bulk routes; existing shopping-list suites pass.

## AI / LLM Assistance

Prepared with the assistance of an LLM (Claude), reviewed by a human before submission.


michael-genson (dismissed):
LGTM

PR #7905, opened 2026-07-19
fix: hide "No Label" heading when copying an unlabeled shopping list

brent20:

Copying a shopping list as text always grouped unlabeled items under Mealie's
internal "No Label" placeholder. When a list has no labels applied anywhere,
`itemsByLabel` ends up with just that one group, so the copied text always
started with a redundant `[No Label]` / `# No Label` heading. For my use-case, I frequently need to manually remove this when pasting the list.

This skips the heading only when it's the *sole* group in the list. If some
items are labeled and others aren't, "No Label" is still shown, since it's
useful there to separate the two.

No open issue currently tracks this. This is a small quality of life improvement. Happy to file one first if you'd prefer.

## Testing

Added two vitest cases: all-items-unlabeled (plain + markdown), and mixed
labeled/unlabeled. `yarn lint --max-warnings=0` and `yarn test:ci` pass locally.

## AI / LLM Assistance

I used Claude to help diagnose the cause and draft the fix and
tests. I reviewed and tested the change myself and I'm submitting it as my
own contribution. Thanks!


brent20:
@michael-genson Hi, just checking in on the status of this and wondering when this PR will be reviewed. Thank you!

michael-genson (approved):
Thank you!

brent20:
Thank you!

PR #7907, opened 2026-07-19
fix: remove aggregate recipe last made data

### LINKED ISSUE #6931: [Task] - Investigate/remove `last_made` field on recipe API

A while back we moved the "Last Made" property to be household-scoped, rather than how it used to be (global/group scoped). This is accessed via a different API. The old `last_made` field is still returned in the API. I'm not even sure if we keep this up to date anymore since the household change.

Either we need to:
- Keep it and document why we need/want it (and make sure we are updating it properly)
- Remove it from the API

zhangli091011:

- Removes the deprecated aggregate `lastMade` value from `Recipe` and `RecipeSummary` API schemas and generated frontend types.
- Keeps `recipes.last_made` as a deprecated database column, while removing runtime writes and the household-to-recipe synchronization event.
- Uses `households_to_recipes.last_made` as the authoritative value for frontend display, updates, sorting, and filtering.
- Makes household last-made updates monotonic with a conditional database update so older timestamps cannot overwrite newer ones.
- Preserves `PATCH /api/recipes/{slug}/last-made` for the existing "Made this" workflow.
- Rejects `lastMade` filters and ordering when recipe queries have no authenticated household context, including public cookbook and suggestion queries, instead of reading frozen aggregate data.
- Preserves Copy Me That `made_this` data by importing it into the current household association.
- Adds regression coverage for API schemas, cross-household isolation, public queries, imports, scheduler behavior, and monotonic updates.

Fixes #6931

The `recipes.last_made` database column is intentionally retained and marked deprecated. The existing last-made PATCH route is also retained because it writes household-scoped data and is used by the frontend.

## Testing

- Focused backend integration and unit suites: 186 passed; the final affected subset after review fixes: 91 passed.
- `uv run --frozen --no-sync ruff check ...`
- `uv run --frozen --no-sync ruff format --check ...`
- `uv run --frozen --no-sync mypy mealie`
- `yarn lint:js --max-warnings=0`
- `yarn build`
- `git diff --check`

The Windows environment cannot build `python-ldap` without MSVC Build Tools, so LDAP-independent backend tests used an external import-only LDAP stub that is not part of this PR. `yarn test:ci` is also blocked in this checkout before test collection because Vitest cannot find a tsconfig for `app/tests/setup.ts`; no frontend tests execute. Frontend lint and the production build pass.

## AI / LLM Assistance

AI assistance was used to investigate the issue and maintainer guidance, trace aggregate and household-scoped `lastMade` usages, implement the backend/frontend changes, add tests, and review the final diff. The changes were iteratively reviewed and verified locally with the checks listed above.


zhangli091011:
The branch is now updated to the latest `mealie-next` (`803d5e918f`), resolving the strict up-to-date gate. The new PR CI run is awaiting maintainer approval: https://github.com/mealie-recipes/mealie/actions/runs/30060174284. Could a maintainer please approve it so the required PostgreSQL, SQLite, and E2E checks can run?

zhangli091011:
The full PR CI suite, including PostgreSQL, SQLite, and E2E tests, is now passing. A review when convenient would be appreciated.

mealie/repos/repository_recipes.py
```diff
@@ -92,6 +93,22 @@ def _get_rating_col_alias(self) -> sa.ColumnElement | None:
...
         return sa.cast(effective_rating, sa.Float)
 
+    def _validate_last_made_query(self, query: PaginationQuery | RecipeSuggestionQuery) -> None:
+        if self.user_id:
+            return
+
+        if query.order_by and any(
+            decamelize(value.split(":", 1)[0].strip()) == "last_made" for value in query.order_by.split(",")
+        ):
+            raise HTTPException(400, "last_made is only available for authenticated household queries")
```
michael-genson, line 100: The frontend still offers "Last Made" sorting on public explore pages (`RecipeCardSection.vue` and `RecipeExplorerPageSearch.vue`), and falls back to the user's saved `orderBy` preference. Won't this make public group pages return an empty list for anyone who's used that sort? Could we hide the option (and ignore a saved `last_made` preference) when there's no logged-in household?

mealie/repos/repository_recipes.py
```diff
@@ -92,6 +93,22 @@ def _get_rating_col_alias(self) -> sa.ColumnElement | None:
...
+
+        if query.query_filter:
+            builder = QueryFilterBuilder(query.query_filter)
```
michael-genson: `QueryFilterBuilder` raises `ValueError` on malformed filters, and this runs before the existing try/except in both `add_pagination_to_query` and `find_suggested_recipes`, so a bad `queryFilter` on the public recipe and suggestion routes now returns a 500 instead of a 400.

mealie/services/household_services/household_service.py
```diff
@@ -87,13 +90,35 @@ def set_household_recipe(self, recipe_slug: str | UUID, data: HouseholdRecipeUpd
...
+            try:
+                household_recipe_out = self.repos.household_recipes.create(create_data)
+                return household_recipe_out.cast(HouseholdRecipeSummary)
+            except IntegrityError:
+                self.repos.session.rollback()
+                existing_household_recipe = self.repos.household_recipes.get_by_recipe(recipe.id)
+                if not existing_household_recipe:
+                    raise
+
+        if data.last_made:
```
michael-genson: Setting last made to an older date is a valid operation (e.g. correcting a mistake), and it's something we'd like to support in the UI eventually (not in this PR). Can we keep the backend update unconditional like it is today? For the child recipe check in `RecipeLastMade.vue`, could we fetch each child's household recipe (like we already do for the parent) and keep the newer-only comparison on the frontend?

michael-genson (changes requested):

zhangli091011:
Addressed the review feedback in 7c6f48c9 and merged the latest mealie-next:

- Public pages now hide last-made sorting and fall back when a saved last_made preference is present.
- Malformed queryFilter values still return 400 instead of 500.
- lastMade updates are unconditional again; the newer-only comparison for child recipes moved to the frontend.

Local checks passed: targeted backend tests, ruff, mypy, frontend ESLint, and Vitest (36 files / 391 tests). CI is awaiting maintainer approval to run.

michael-genson (approved):
Thanks for the updates, looks good!

PR #7910, opened 2026-07-19
fix: Repair shopping lists and recipes when a food is deleted. 

### LINKED ISSUE #7711: [BUG] - sqlite: Deleting a food that is on a shopping list leaves blank shopping list entry
### First Check

When deleting a food at /group/data/foods/ that is on a shopping list, the food is deleted and a blank entry (empty string) is left behind on the shopping list. Deletion should be prevented with a Conflict error, or the shopping list item converted to free text note.

### Steps to Reproduce

1) Run mealie with sqlite
2) Add a food ingredient to the shopping list
3) Delete the food from data management

n/a

### Mealie Version

Version: nightly
Build: 
[e24187fefb2c5a49764ff0e5d3998dfb31982356](https://github.com/mealie-recipes/mealie/commit/e24187fefb2c5a49764ff0e5d3998dfb31982356)

### Deployment

Docker (Linux)

_No response_

miah120:
When a food is deleted via data management, we will now search each recipe and each shopping list for items containing that food.
If any are found, we will prepend the deleted food to that items note. 

Fixes: #7711

## Testing
Tested in docker dev container. 

## AI / LLM Assistance
No AI used.

### DIFF
--- a/mealie/routes/unit_and_foods/foods.py
+++ b/mealie/routes/unit_and_foods/foods.py
@@ -1,8 +1,12 @@
 from functools import cached_property
 
+import sqlalchemy as sa
 from fastapi import APIRouter, Depends, HTTPException
 from pydantic import UUID4
 
+from mealie.db.models.household.shopping_list import ShoppingList, ShoppingListItem
+from mealie.db.models.recipe.ingredient import RecipeIngredientModel
+from mealie.db.models.recipe.recipe import RecipeModel
 from mealie.routes._base.base_controllers import BaseUserController
 from mealie.routes._base.controller import controller
 from mealie.routes._base.mixins import HttpRepo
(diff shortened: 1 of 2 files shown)


mealie/routes/unit_and_foods/foods.py
```diff
@@ -75,4 +81,35 @@ def update_one(self, item_id: UUID4, data: CreateIngredientFood):
     @router.delete("/{item_id}", response_model=IngredientFood)
     def delete_one(self, item_id: UUID4):
         self.checks.can_organize()
+        self._repair_recipes(item_id)
```
michael-genson, line 84: Going through `RecipeService.update_one` and `ShoppingListService` here runs into permission and scoping issues: `update_one` raises a 403 for locked recipes and other households' recipes, which fails the delete partway through, and `self.repos` is household-scoped, so other households' shopping list items are skipped. Could we move this into `RepositoryFood` instead, similar to how `merge` handles it? Something like overriding `delete` to load `RecipeIngredientModel` and `ShoppingListItem` rows by `food_id`, prepend the food name to each `note` (it's nullable, so the current f-string produces "Salt None"), then delete the food in the same commit. Setting `note` on the ORM rows keeps `note_normalized` in sync through the existing listener.

michael-genson (changes requested):

PR #7914, opened 2026-07-20
fix: harden server-initiated HTTP against SSRF and DNS rebinding

hay-kot:

Server-initiated HTTP (recipe scraping, webhooks, recipe actions, image fetching) could be pointed at internal/private addresses. The existing scraper guard also only validated the IP it looked up while curl re-resolved and connected separately, so a rebinding DNS answer slipped past it, and it missed some ranges (e.g. CGNAT) and IPv6 entirely. The webhook and recipe-action paths had no guard at all.

This routes everything through one SSRF-protected layer:

- Resolve each host once and validate **every** resolved address (private, loopback, link-local, CGNAT `100.64/10`, multicast, reserved, unspecified, IPv4-mapped IPv6); reject if any is unsafe.
- **Pin** the connection to the validated address (curl's RESOLVE option) so curl can't re-resolve to a rebound IP. Resolution uses `getaddrinfo`, so IPv6 is checked too. Redirects re-enter the transport and re-validate per hop.
- Add a sync `SafeTransport` + `safehttp.post` and move the webhook and recipe-action paths off raw `requests.post`.
- Add `HTTP_ALLOW_LIST` / `HTTP_DISALLOW_LIST` settings (precedence: deny > allow > default) so operators can reach a known internal server or block extra hosts.

Apprise notifications egress through the third-party library rather than our client, so they aren't covered here.

_No linked issue._

The safe transport pins per request through its curl session, so a single transport instance isn't safe for concurrent requests. All callers use a fresh transport per request (the concurrent image-size check now creates one per URL) and follow redirects sequentially.

## Testing

New `test_safehttp.py` covers the blocklist (incl. CGNAT/metadata/IPv6/IPv4-mapped), allow/deny precedence, reject-if-any-unsafe, and an end-to-end pin (a host that doesn't resolve to localhost only reaches it via the pin). Existing webhook/recipe-action tests updated to the new call path; scraper/image/webhook suites pass. `ruff` + `mypy` clean.

## AI / LLM Assistance

Prepared with the assistance of an LLM (Claude); design (pinning approach, allow/deny lists, scope) was reviewed with a human maintainer, who reviewed the changes before submission.


michael-genson (dismissed):
LGTM, is it worth an in-app announcement for breaking changes to notifiers/actions?

PR #7925, opened 2026-07-21
fix: Keep instruction sticky in ingredient linker

### LINKED ISSUE #6207: Constrain scrollable area in ingredient linker to ingredient list, not the recipe step text
When linking ingredients, we see the text of the recipe step being linked to, and the list of ingredients. Sometimes, the list is long enough to cause y-overflow scrolling:

Image

However, when scrolling down, we can't see the recipe step anymore! So I have to read, scroll down, toggle ingredient, scroll up, read, repeat.

Image

The FR/bug is to constrain the scrollable area to just the ingredient list, forcing the recipe step text to always be visible

miah120:
Previously, if a recipe had a long list of ingredients, you may have to scroll far enough through the recipe linker that the recipe instruction was completely gone. 
image

Now, the recipe instruction is sticky. 
image

Additionally, if the instruction itself is too long, it will be constrained to a separate scrollable area. 

Fixes #6207

## Testing
Tested on desktop (Firefox) and Mobile (Android Firefox)

## AI / LLM Assistance
No AI used.

### DIFF
--- a/frontend/app/components/Domain/Recipe/RecipePage/RecipePageParts/RecipePageInstructions.vue
+++ b/frontend/app/components/Domain/Recipe/RecipePage/RecipePageParts/RecipePageInstructions.vue
@@ -9,55 +9,65 @@
       max-width="600px"
       max-height="40%"
     >
-      <v-card-text class="pt-4">
-        <p>
-          {{ activeText }}
-        </p>
-        <v-divider class="my-4" />
-        <template v-if="Object.keys(groupedUnusedIngredients).length > 0">
-          <h4 class="ml-1">
-            {{ $t("recipe.unlinked") }}
-          </h4>
-          <template v-for="(ingredients, title) in groupedUnusedIngredients" :key="title">
-            <h4 v-if="title" class="py-3 ml-1 pl-4">
-              {{ title }}
-            </h4>
-            <v-checkbox-btn
-              v-for="ing in ingredients"
-              :key="ing.referenceId"
-              v-model="activeRefs"
-              :value="ing.referenceId"
-              class="ml-4"
-            >
-              <template #label>
-                <RecipeIngredientHtml :ingredient="ing" :scale="scale" />
+      <div class="grid">
+        <div class="sticky">
+          <v-card flat style="max-height: 40dvh; overflow-y: auto;">
+            <v-card-text>
+              <p>
+                {{ activeText }}
+              </p>
+            </v-card-text>
+          </v-card>
+          <v-divider />
+        </div>
+        <v-card flat>
+          <v-card-text>
+            <template v-if="Object.keys(groupedUnusedIngredients).length > 0">
+              <h4 class="ml-1">
+                {{ $t("recipe.unlinked") }}
+              </h4>
+              <template v-for="(ingredients, title) in groupedUnusedIngredients" :key="title">
+                <h4 v-if="title" class="py-3 ml-1 pl-4">
+                  {{ title }}
+                </h4>
+                <v-checkbox-btn
+                  v-for="ing in ingredients"
+                  :key="ing.referenceId"
+                  v-model="activeRefs"
+                  :value="ing.referenceId"
+                  class="ml-4"
+                >
+                  <template #label>
+                    <RecipeIngredientHtml :ingredient="ing" :scale="scale" />
+                  </template>
+                </v-checkbox-btn>
               </template>
-            </v-checkbox-btn>
-          </template>
-        </template>
+            </template>
 
-        <template v-if="Object.keys(groupedUsedIngredients).length > 0">
-          <h4 class="py-3 ml-1">
-            {{ $t("recipe.linked-to-other-step") }}
-          </h4>
-          <template v-for="(ingredients, title) in groupedUsedIngredients" :key="title">
-            <h4 v-if="title" class="py-3 ml-1 pl-4">
-              {{ title }}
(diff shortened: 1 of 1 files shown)


frontend/app/components/Domain/Recipe/RecipePage/RecipePageParts/RecipePageInstructions.vue
```diff
@@ -779,6 +779,14 @@ function openImageUpload(index: number) {
 </script>
 
 <style lang="css" scoped>
+.sticky {
+  position: sticky;
+  top: 0;
+  z-index: 2;
+  max-height: 40dvh;
+  overflow: scroll;
```
michael-genson: Can we use `overflow-y: auto` instead of `overflow: scroll`? `scroll` leaves an empty scrollbar gutter on Windows/Linux even when the instruction is one line.

frontend/app/components/Domain/Recipe/RecipePage/RecipePageParts/RecipePageInstructions.vue
```diff
@@ -9,11 +9,11 @@
       max-width="600px"
       max-height="40%"
     >
-      <v-card-text class="pt-4">
-        <p>
-          {{ activeText }}
-        </p>
+      <div class="sticky bg-surface pt-4 px-4">
+        {{ activeText }}
         <v-divider class="my-4" />
```
michael-genson: Can we move the divider out of the div so it stays pinned to the bottom edge of the header?

frontend/app/components/Domain/Recipe/RecipePage/RecipePageParts/RecipePageInstructions.vue
```diff
@@ -9,11 +9,11 @@
...
+      </div>
+      <v-card-text class="pt-4">
```
michael-genson: Do we need the extra `pt-4` here? We already have `pt-4` from the sticky div.

frontend/app/components/Domain/Recipe/RecipePage/RecipePageParts/RecipePageInstructions.vue
```diff
@@ -9,11 +9,11 @@
       max-width="600px"
       max-height="40%"
     >
-      <v-card-text class="pt-4">
-        <p>
-          {{ activeText }}
-        </p>
+      <div class="sticky bg-surface pt-4 px-4">
+        {{ activeText }}
```
michael-genson: Can we put `activeText` inside `<p>` tags? e.g. `<p class="mb-0">{{ activeText }}</p>`

frontend/app/components/Domain/Recipe/RecipePage/RecipePageParts/RecipePageInstructions.vue
```diff
@@ -9,11 +9,11 @@
       max-width="600px"
       max-height="40%"
     >
-      <v-card-text class="pt-4">
-        <p>
-          {{ activeText }}
-        </p>
+      <div class="sticky bg-surface pt-4 px-4">
```
michael-genson: Double check `bg-surface` works as expected and matches the `v-card` background color in dark mode, I don't think we use that anywhere else

michael-genson (changes requested):
Thanks for this! A couple tweaks:

miah120:
I ended up switching to a grid layout approach rather than a sticky one since I wasn't able to get things to work very well (particularly with the divider) with the sticky approach.

frontend/app/components/Domain/Recipe/RecipePage/RecipePageParts/RecipePageInstructions.vue
miah120: done

frontend/app/components/Domain/Recipe/RecipePage/RecipePageParts/RecipePageInstructions.vue
miah120: It had been important to the layout, but was a little messy. No longer necessary with the new grid approach

frontend/app/components/Domain/Recipe/RecipePage/RecipePageParts/RecipePageInstructions.vue
miah120: done

frontend/app/components/Domain/Recipe/RecipePage/RecipePageParts/RecipePageInstructions.vue
miah120: done

frontend/app/components/Domain/Recipe/RecipePage/RecipePageParts/RecipePageInstructions.vue
miah120: ftr, bg-surface does work in darkmode. But this was no longer necessary with the new grid approach so its not there anymore anyway

michael-genson:
Can you double check the branch? It doesn't seem to be sticky for me anymore.

miah120:
Ah. Forgot to also test in desktop 🤦‍♀️. Looks like the grid works for mobile and sticky works for desktop, so now everything should work in both modalities

michael-genson (approved):
Looks great now, thanks!

PR #7939, opened 2026-07-23
feat: Bottom sheets

miah120:

This PR adds bottom sheets to mealie in mobile in places where appropriate*. 
This has a couple benefits over popups: 
* Keeps the user in the original context: i.e., users can still see the page they're on which tends to read as less disruptive
* Keeps the actions at the bottom where users' thumbs tend to be
* Keeps actions close to the content. 

Desktop Before
image
Mobile Before
image

Desktop After
image
Mobile After
image

There are also a couple miscellaneous changes in here such as: 
* reround the shopping lists expansion panels 
* fix the shopping list check-all/unchecck-all icons
* moved the restore backup option into the card actions section
* color the delete recipe modal red/error instead of primary (from the admin page)

\*Best practices for bottom sheets is to only use them when there's minimal interactions. My litmus test for this was 3 or fewer interactions. E.g., confirm/cancel is 1 interaction, select + confirm is 2 interactions, etc. 

NA

There were several dialogs I didn't change with this PR. Most of them because they had too much content to fit inside bottom sheets, but also a couple I skipped over because I wasn't able to find them in the site. 

Dialogs with too much content:
  * Announcement Dialog
  * Cookbook editor/creator
  * AI Dialog
  * Group data create
  * Group data edit
  * Group data create
  * Shopping list ingredients
  * Recipe dialog bulk add
  * recipe print preferences
  * organizer edit
    * I think as-is it would be fine, but ik there is a plan to add colors to these organizers, which would probably be included here.
  * timeline event edit
  * Ingredient Link editor
  * Recipe page parse
  * User invite dialogue
  * cookbook query filter
  * Notifier create
  * Meal plan create
  * Shopping list label reorderer

Dialogs where idek what they are:
  * Recipe data alias manager
  * Recipe Assets
  * Recipe organizer dialogue

## Testing

## AI / LLM Assistance
No AI used


frontend/app/components/global/BaseDialog.vue
```diff
@@ -210,24 +223,5 @@ function deleteEvent() {
...
-
-function logDeprecatedProp(val: string) {
-  console.warn(
-    `[BaseDialog] The method '${val}' is deprecated. Please use v-model="value" to manage state instead.`,
-  );
 }
 </script>
-
-<style>
-.top-dialog {
```
michael-genson, line 229: Can we keep this rule, or move it to a shared stylesheet? `RecipeDialogSearch.vue:7` passes `content-class="top-dialog"` and the dialog branch below still sets `:content-class="top ? 'top-dialog' : undefined"`, so dropping it makes the `top` prop a no-op and global search loses its top pinning.

frontend/app/components/global/BaseDialogContent.vue
```diff
@@ -0,0 +1,140 @@
...
+    </v-toolbar>
+
+    <div style="flex: 1 1 auto; min-height: 0; overflow: auto">
+      <slot v-bind="{ submitEvent: () => emit('submit') }" />
+    </div>
+
+    <v-spacer />
+    <v-divider />
+
+    <v-card-actions class="pb-4">
```
michael-genson: Can we gate `pb-4` and the `flex-1-1-0` classes on `$vuetify.display.xs`, or keep the `<v-spacer/>` on larger displays? As written every dialog on desktop gets a stretched action row, including ones that were never converted to bottom sheets.

frontend/app/components/global/BaseDialogContent.vue
```diff
@@ -0,0 +1,140 @@
...
+        <v-btn
+          class="flex-1-1-0"
+          variant="text"
+          color="grey"
+          @click="emit('cancel')"
+        >
+          {{ cancelText }}
+        </v-btn>
+
+        <slot name="custom-card-action" />
```
michael-genson, line 39: The `custom-card-action` slots that weren't updated now sit beside a full-width cancel button — `RecipeDataAliasManagerDialog.vue:35`, `RecipePageParseDialog.vue:161`, `finder/index.vue:89`, and the `edit-dialog-custom-action` pass-through in `GroupDataPage.vue:47`. Can we bring those in line with the new layout?

frontend/app/components/global/BaseDialogContent.vue
```diff
@@ -0,0 +1,140 @@
...
+</template>
+
+<script setup lang="ts">
+import { useNuxtApp } from "#app";
+
+interface DialogProps {
+  color?: string;
+  title?: string;
+  icon?: string | null;
+  width?: number | string;
```
michael-genson: Are `width`, `maxWidth`, `top`, `keepOpen`, and `disableSubmitOnEnter` used in here? They look declared and defaulted but never read — can we drop them if BaseDialog is handling all of that?

frontend/app/components/global/BaseDialogContent.vue
```diff
@@ -0,0 +1,140 @@
...
+
+  // actions
+  canDelete?: boolean;
+  canConfirm?: boolean;
+  canSubmit?: boolean;
+  disableSubmitOnEnter?: boolean;
+}
+
+interface DialogEmits {
+  (e: "update:modelValue", value: boolean): void;
```
michael-genson: Same question for `update:modelValue` and `close` — I only see submit, cancel, confirm, and delete emitted.

frontend/app/components/global/BaseDialog.vue
```diff
@@ -4,7 +4,60 @@
...
+      :content-props="{
+        style: 'overflow: hidden',
+      }"
+      :width="width"
+      :max-width="maxWidth ?? undefined"
+      @keydown.enter="submitOnEnter"
+      @click:outside="emit('cancel')"
+      @keydown.esc="emit('cancel')"
+    >
+      <BaseDialogContent
```
michael-genson: Both branches list the same ~17 props and 3 slots by hand. Would a computed bindings object with `v-bind` be worth it here, so the two can't drift?

michael-genson (changes requested):
Thanks for this! A few things before it goes in:

frontend/app/components/global/BaseDialog.vue
miah120: done

frontend/app/components/global/BaseDialogContent.vue
miah120: fixed

frontend/app/components/global/BaseDialogContent.vue
miah120: I ended up fixing this in a bit of a round-about way. I switched the card actions to use a grid layout (with auto columns) rather than the standard flex layout. This lets consumers use custom actions without needing to add flex-1-1-0 in order for things to render properly.

frontend/app/components/global/BaseDialogContent.vue
miah120: fixed; also removed them from the shared bindings.

frontend/app/components/global/BaseDialogContent.vue
miah120: fixed

michael-genson (approved):
LGTM, thanks! I pushed a couple small things

PR #7940, opened 2026-07-23
fix: Add bottom padding to prevent fab overlapping content

miah120:
Previously, when on the recipe list page, the "return to top" fab could overlap with the content of the page and either make it difficult or impossible to interact with the context menu of the last recipe. 
Before:
image

Now, there is padding added at the bottom to ensure there's room for the fab
image

NA

## Testing

## AI / LLM Assistance
No AI used.


p0lycarpio:
Thanks!

p0lycarpio (approved):

PR #7941, opened 2026-07-23
feat: skip expiring share link for fully-public groups/households

bferd:

- Adds a check on the recipe Share button: when the recipe's group **and** household are
  both fully public (`private_group === false && private_household === false`), Share now
  triggers an instant native share (mobile) or clipboard copy (desktop) of the plain recipe
  URL, instead of opening the expiring-link dialog.
- For every other combination of group/household privacy, the existing expiring-token
  share dialog is untouched and behaves exactly as before.
- Rationale: when an instance (or a specific group/household within it) is already
  configured to be publicly viewable without login, the plain recipe URL is already fully
  functional and permanent — generating a time-limited token for it is redundant friction
  for the user. The existing token flow remains necessary for non-public groups/households,
  where it's the only way to grant temporary outside access.
- Change is entirely in `frontend/app/components/Domain/Recipe/RecipeContextMenu/RecipeContextMenuContent.vue`:
  added `useGroupSelf`, an `isFullyPublic` computed property, and a `sharePlainLink()`
  function (using `useShare`/`useClipboard` from `@vueuse/core`, mirroring the existing
  pattern in `RecipeDialogShare.vue`). The share menu item's click handler now branches on
  `isFullyPublic` before deciding whether to open the dialog.
- `RecipeDialogShare.vue` is unchanged — no new share tokens are created or stored for the
  new public-link path, by design.
- No backend changes. No new endpoints, no schema changes.

N/A

- The public/private determination deliberately uses `=== false` rather than `!value`,
  since `useGroupSelf()`/`useHouseholdSelf()` load asynchronously and may be `undefined`
  on first render. This ensures an unloaded/unknown privacy state safely falls through to
  the existing (safe) dialog behavior rather than being mistakenly treated as public.
- This mirrors the server-side gating exactly: `get_public_group` (in
  `mealie/core/dependencies/dependencies.py`) 404s the entire explore namespace for a
  private group before any household-level check runs, and `controller_public_recipes.py`
  separately checks `household.preferences.private_household` per recipe. The frontend's
  `&&` of both flags matches this two-gate backend behavior, so the button's decision
  never diverges from what's actually reachable anonymously.
- `navigator.share` / `navigator.clipboard` both require a secure context (HTTPS or
  `localhost`) per browser spec — this was initially mistaken for a bug while testing over
  plain HTTP on a LAN IP (`Clipboard not supported` fired on both mobile and desktop
  identically), until testing was moved to an HTTPS-exposed dev subdomain, which resolved
  it. No code change was needed; noting it here in case a future reviewer testing over
  plain HTTP hits the same false negative.

## Testing

Manually tested against 4 seeded accounts covering all group/household privacy
combinations, over HTTPS (to get real `navigator.share`/`navigator.clipboard` behavior
rather than a secure-context false negative):

| Account  | Group   | Household | Result |
|----------|---------|-----------|--------|
| pubpub   | Public  | Public    | ✅ Instant share — native share sheet (mobile) / clipboard + toast (desktop), no dialog |
| pubpriv  | Public  | Private   | ✅ Existing expiring-link dialog, unchanged |
| privpub  | Private | Public    | ✅ Existing expiring-link dialog, unchanged |
| privpriv | Private | Private   | ✅ Existing expiring-link dialog, unchanged |

Confirmed via direct API calls that each seeded account's actual `privateGroup`/
`privateHousehold` values matched the intended matrix before running the UI pass, so the
above results reflect the real decision logic rather than a data-seeding mismatch.

`pubpub` on desktop (Windows share flyout via `navigator.share`):

Desktop

`pubpub` on mobile (Android native share sheet via `navigator.share`):

Mobile

Both confirm the plain recipe URL (`https://mealietest.schroth.ca/g/public/r/leek-and-mushroom-soup`)
is populated with no token or expiration query params.

ESLint clean. `DB_ENGINE=sqlite uv run pytest` run as a sanity check (no backend code
changed by this PR, so no new backend tests were needed). One pre-existing failure
(`test_pg_connection_args`) observed, unrelated to this change — caused by local
`.env` using custom Postgres credentials/DB name rather than the test's hardcoded
`mealie`/`mealie`/`mealie` defaults; confirmed not present in this PR's diff and expected
to be absent in CI, which sets its own clean env vars.

## AI / LLM Assistance

Claude Code was used to generate the initial implementation following the existing image
import pattern. All generated code was reviewed, tested manually, and verified against the
codebase conventions before submission.

### DIFF
--- a/frontend/app/components/Domain/Recipe/RecipeContextMenu/RecipeContextMenuContent.vue
+++ b/frontend/app/components/Domain/Recipe/RecipeContextMenu/RecipeContextMenuContent.vue
@@ -111,6 +111,7 @@ import { useGroupSelf } from "~/composables/use-groups";
 import { useHouseholdSelf } from "~/composables/use-households";
 import { alert } from "~/composables/use-toast";
 import { usePlanTypeOptions } from "~/composables/use-group-mealplan";
+import { isRecipeFullyPublic } from "~/lib/recipe/recipe-visibility";
 import type { Recipe } from "~/lib/api/types/recipe";
 import type { GroupRecipeActionOut, ShoppingListSummary } from "~/lib/api/types/household";
 import type { PlanEntryType } from "~/lib/api/types/meal-plan";
@@ -215,11 +216,6 @@ const groupSlug = computed(() => route.params.groupSlug as string || auth.user.v
 const firstDayOfWeek = computed(() => {
   return household.value?.preferences?.firstDayOfWeek || 0;
 });
-
-const isFullyPublic = computed(() =>
-  group.value?.preferences?.privateGroup === false
-  && household.value?.preferences?.privateHousehold === false,
-);
 
 const { share, isSupported: shareIsSupported } = useShare();
 const { copy, copied, isSupported: clipboardIsSupported } = useClipboard();
@@ -326,6 +322,9 @@ const shoppingLists = ref<ShoppingListSummary[]>();
 const recipeRef = ref<Recipe | undefined>(props.recipe);
 const recipeRefWithScale = computed(() =>
   recipeRef.value ? { scale: props.recipeScale, ...recipeRef.value } : undefined,
+);
+const isFullyPublic = computed(() =>
+  isRecipeFullyPublic(recipeRef.value, group.value, household.value),
 );
 const isAdminAndNotOwner = computed(() => {
   return (
--- a/frontend/app/components/Domain/Recipe/RecipeContextMenu/RecipeContextMenuContent.vue
+++ b/frontend/app/components/Domain/Recipe/RecipeContextMenu/RecipeContextMenuContent.vue
@@ -100,12 +100,14 @@
 </template>
 
 <script setup lang="ts">
+import { useClipboard, useShare } from "@vueuse/core";
 import RecipeDialogAddToShoppingList from "~/components/Domain/Recipe/RecipeDialogAddToShoppingList.vue";
 import RecipeDialogPrintPreferences from "~/components/Domain/Recipe/RecipeDialogPrintPreferences.vue";
 import RecipeDialogShare from "~/components/Domain/Recipe/RecipeDialogShare.vue";
 import { useLoggedInState } from "~/composables/use-logged-in-state";
 import { useUserApi } from "~/composables/api";
 import { useGroupRecipeActions } from "~/composables/use-group-recipe-actions";
+import { useGroupSelf } from "~/composables/use-groups";
 import { useHouseholdSelf } from "~/composables/use-households";
 import { alert } from "~/composables/use-toast";
 import { usePlanTypeOptions } from "~/composables/use-group-mealplan";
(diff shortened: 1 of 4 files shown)


frontend/app/components/Domain/Recipe/RecipeContextMenu/RecipeContextMenuContent.vue
```diff
@@ -204,15 +206,47 @@ const i18n = useI18n();
...
+const groupSlug = computed(() => route.params.groupSlug as string || auth.user.value?.groupSlug || "");
 
 const firstDayOfWeek = computed(() => {
   return household.value?.preferences?.firstDayOfWeek || 0;
 });
 
+const isFullyPublic = computed(() =>
+  group.value?.preferences?.privateGroup === false
+  && household.value?.preferences?.privateHousehold === false,
+);
```
michael-genson: Do we have an existing composable for this? If possible I'd like to see this somewhere reusable, e.g. in `useShare`

frontend/app/components/Domain/Recipe/RecipeContextMenu/RecipeContextMenuContent.vue
```diff
@@ -425,7 +459,12 @@ const eventHandlers: { [key: string]: () => void | Promise<any> } = {
     });
   },
   share: () => {
-    shareDialog.value = true;
+    if (isFullyPublic.value) {
+      sharePlainLink();
+    }
+    else {
+      shareDialog.value = true;
+    }
```
michael-genson, line 466: Similarly this logic might work better in `useShare` but I haven't looked at it in a while so if I'm mistaken this is fine as-is

frontend/app/components/Domain/Recipe/RecipeContextMenu/RecipeContextMenuContent.vue
bferd: I'll take a look at this and below coment tomorrow and report. Thanks.

frontend/app/components/Domain/Recipe/RecipeContextMenu/RecipeContextMenuContent.vue
bferd: Confirmed and agree — pulling this into a shared useShareLink() composable now. Also found a real bug while looking at this: RecipeDialogShare.vue's shareRecipe() checks if (shareIsSupported) instead of if (shareIsSupported.value), so on browsers without the Web Share API (confirmed on Firefox desktop), clicking the share row does nothing at all — it silently skips the clipboard fallback. Since the fix lives naturally in the same composable I'm extracting here, want me to fix it as part of this PR, or would you rather I split it into a separate PR? Standing by...

@michael-genson

frontend/app/components/Domain/Recipe/RecipeContextMenu/RecipeContextMenuContent.vue
bferd: No existing composable for this — checked, `privateGroup/privateHousehold` aren't combined anywhere else in the codebase. I'd lean toward keeping it separate from `useShare` though, rather than folding it in — it's a different concern (deriving visibility from group/household state vs. the actual share/clipboard mechanics), and `RecipeDialogShare.vue` would end up pulling in visibility logic it never uses if they're combined. That said, there's no second use case for it anywhere else in the codebase or on the roadmap right now, so I'm not sure it earns its own composable yet — happy to extract a small `useGroupHouseholdVisibility()` if you'd still like it reusable, otherwise I'll leave it inline for now... lemme know.

michael-genson: I'm fine with keeping it inline in `RecipeDialogShare`, I didn't realize `useShare` was Vue and not us.

frontend/app/components/Domain/Recipe/RecipeContextMenu/RecipeContextMenuContent.vue
michael-genson: Apologies I thought `useShare` was something we owned but it's from Vue. I'm fine with leaving it in this component instead.

Regarding the bug, let's fix that in this PR.

frontend/app/components/Domain/Recipe/RecipeContextMenu/RecipeContextMenuContent.vue
michael-genson: This seems to use the current user's group/household settings but ultimately it's the recipe's group/household settings that matter for sharing.

michael-genson: Let's also add a test for this.

bferd: Confirmed and fixed — you're right, `useGroupSelf()`/`useHouseholdSelf()` were reading the viewer's own group/household, not the recipe's. `isFullyPublic` now requires `recipe.groupId`/`recipe.householdId` to match the viewer's own before trusting those preferences, plus checks `recipe.settings.public` — a separate per-recipe flag I'd missed, since the backend's actual gate in `controller_public_recipes.py` is a three-part condition, not just group+household privacy. If the recipe belongs to a different household than the viewer's own, it now correctly falls through to the existing dialog instead of guessing.

Tested end-to-end: an account with a fully-public group/household viewing a recipe owned by a _private_ household in the same group now correctly gets the dialog, not instant share.

Also extracted the logic into a standalone `isRecipeFullyPublic()` function `(frontend/app/lib/recipe/recipe-visibility.ts)` with 8 unit tests covering each condition — per your comment, this made it actually testable without mounting the whole component.

frontend/app/components/Domain/Recipe/RecipeContextMenu/RecipeContextMenuContent.vue
bferd: Fixed the `shareIsSupported` bug from my earlier reply — confirmed via the same test pass as isFullyPublic thread below, and manually re-verified on Firefox desktop that the clipboard fallback now actually fires.

michael-genson (approved):
Thank you! This one has bugged me forever

PR #7946, opened 2026-07-25
feat: Meal Planner UI Improvements

miah120:
Goals: 
* quick scrolling between weeks
* standardize buttons
* improve visual hierarchy
  * increase meal plan entry size
* allow more actions from the view page
  * while these could be left behind the edit page, allowing for some quick changes to be done with lower friction means that the edit page can be saved for larger scale changes (e.g., rearranging multiple recipes, filling out multiple days worth of meals, etc.) while the view page is good for both viewing and tweaking. 
* separate out common components 
* separate out common functionality into composables
* animate meal plan entries in and out

Files changed: 
* frontend/app/components/Domain/Household/GroupMealPlanDayContextMenu.vue
  * No longer necessary. Looks like this one may have been made before base button group took on the majority of the generic functionality this had. 
* frontend/app/pages/household/mealplan/planner.vue
  * use new useAddToShoppingListDialog composable
  * add new quick week scrolling buttons
  * remove unused css
  * standardize buttons into menu/button group
* frontend/app/pages/household/mealplan/planner/view.vue
  * use shared meal plan components
  * add animations
  * allow for meal plan creation/editing/removing  
* frontend/app/pages/household/mealplan/planner/edit.vue
  * use shared meal plan components
  * add animations
* frontend/app/lib/icons/icons.ts
  * icons 
* frontend/app/lang/messages/en-US.json
  * "remove from plan" string 
* frontend/app/composables/use-group-mealplan.ts
  * move shared logic here 
* frontend/app/composables/shopping-list-page/use-add-to-shopping-list-dialog.ts
  * move shared logic here 
* frontend/app/components/global/BaseButtonGroup.vue
  * use colors/icons/disabled properties
  * add loading property 
* frontend/app/components/global/Transition/SpinTransition.vue
  * animation
* frontend/app/components/Domain/Recipe/RecipeCardMobile.vue
  * allow a new context menu slot; this is so note-only meals can be edited/removed
  * pass through leading/append items for editing/removal
* frontend/app/components/Domain/Recipe/RecipeContextMenu/RecipeContextMenuContent.vue
  * use new useAddToShoppingListDialog composable
* frontend/app/components/Domain/Recipe/RecipeContextMenu/RecipeContextMenu.vue
  * use shared type
  * pass through mealplan events (I couldn't figure out a way to do this automatically)
* frontend/app/components/Domain/Mealplan/MealPlanNoteMenu.vue
* frontend/app/components/Domain/Mealplan/MealPlanLayout.vue
* frontend/app/components/Domain/Mealplan/MealPlanDayHeader.vue
* frontend/app/components/Domain/Mealplan/MealPlanDay.vue
* frontend/app/components/Domain/Mealplan/MealPlanCreateDialog.vue
  * custom shared components 

Before and after: 
***
Overall:  
| Before | After |
|--------|--------|
| image | image | 

Standardized buttons + Quick scrolling: 
| Before | After |
|--------|--------|
| image | imageimage |

Better day headers:
| Before | After |
|--------|--------|
| image | image |

New actions now accessible from the view mode: 
image
image

Animations: 
[Screencast_20260725_152906.webm](https://github.com/user-attachments/assets/31c6afaf-b036-4f07-89da-5b397c311517)

NA

Sorry for such a big pr 😅 

I did my best to keep everything clean here, but there was one issue I wasn't able to work around: 
In RecipeContextMenu.vue, I had to resort to manually forwarding the mealplan events. 
I'm sure there must be a way to not need that, but from what I can find online the only way to do this is binding to $attrs, which raises issues w this component. 

I know you tend to be more vue savvy than I am, so if there's something I'm missing I'd love to get that cleaned up. 

## Testing

## AI / LLM Assistance
No AI used.


michael-genson:
> Sorry for such a big pr

https://github.com/mealie-recipes/mealie/pull/5184

> +24,006 -19,604

michael-genson:
FYI https://github.com/mealie-recipes/mealie/pull/8152 redoes a lot of the editor page so there will probably be merge conflicts. It doesn't replace your PR, I like a lot of the changes here, but it does replace a part of it.

frontend/app/pages/household/mealplan/planner/view.vue
```diff
@@ -1,108 +1,97 @@
...
+    />
+    <MealPlanLayout :mealplans="mealplans">
+      <template #default="{ day }">
+        <MealPlanDay :day="day.date" :actions="actions" :recipes="day.recipes">
+          <SpinTransition>
+            <v-card v-if="day.sections.length" variant="flat" class="pl-4 pr-2">
+              <SpinTransition>
+                <div v-for="section in day.sections" :key="section.title">
+                  <div class="py-2 d-flex flex-column">
+                    <div class="primary" style="width: 50px; height: 2.5px" />
```
michael-genson: `class="primary"` is Vuetify 2 syntax — under the pinned Vuetify 4 it applies no background, so this bar renders invisible. Can we make it `bg-primary`?

frontend/app/pages/household/mealplan/planner.vue
```diff
@@ -167,16 +185,20 @@ const state = ref({
...
-const shoppingLists = ref<ShoppingListSummary[]>();
-
 const firstDayOfWeek = computed(() => {
   return household.value?.preferences?.firstDayOfWeek || 0;
 });
 
+function changeWeek(step: number) {
+  const { start, end } = weekRange.value;
+  state.value.range = [
+    addDays(start, step * numberOfDays.value),
```
michael-genson: Should this step by the displayed span (`weekRange.end - weekRange.start + 1`) rather than `numberOfDays`? With `numberOfDaysPast > 0` the chevrons re-show days that are already on screen, and a custom picker range skips days.

frontend/app/components/Domain/Mealplan/MealPlanDay.vue
```diff
@@ -0,0 +1,176 @@
...
+          children: [
+            {
+              text: $t('meal-plan.add-day-to-list'),
+              icon: $globals.icons.cartCheck,
+              event: 'shopping-list',
+              loading: addAllLoading,
+              disabled: !props.recipes.length,
+            },
+            ...commonButtons,
+            ...randomButtons,
```
michael-genson: Flattening both lists here means "Side" appears twice, and the dice entries read as plain "Breakfast"/"Lunch"/"Dessert" with nothing marking them as random. Can we nest them under a "Random Meal" parent the way `inlineButtons` does below?

frontend/app/components/Domain/Recipe/RecipeCardMobile.vue
```diff
@@ -156,6 +164,7 @@ const props = withDefaults(defineProps<Props>(), {
 });
 
 defineEmits<{
+  [key: string]: [] | [slug: string];
```
michael-genson: Can we declare the events explicitly instead of an index signature? As written this turns off emit-name checking for the whole card — a typo'd event compiles fine and silently does nothing — and it's used in a lot of places.

frontend/app/components/Domain/Recipe/RecipeContextMenu/RecipeContextMenuContent.vue
```diff
@@ -1,6 +1,6 @@
 <template>
   <RecipeDialogShare v-model="shareDialog" :recipe-id="recipeId" :name="name" />
-  <RecipeDialogPrintPreferences v-model="printPreferencesDialog" :recipe="recipeRef" />
+  <RecipeDialogPrintPreferences v-model="printPreferencesDialog" :recipe="recipeRef as NoUndefinedField<Recipe>" />
```
michael-genson: This was `:recipe="recipeRef"` before, so the cast is silencing a genuinely possible `undefined`. Can we guard with `v-if="recipeRef"`, or widen the dialog's prop type, instead of asserting?

frontend/app/components/Domain/Recipe/RecipeCardMobile.vue
```diff
@@ -96,25 +96,30 @@
...
-                  print: false,
-                  printPreferences: false,
-                  share: true,
-                }"
-                @deleted="$emit('delete', slug)"
-              />
+              <slot name="context-menu">
+                <RecipeContextMenu
+                  v-if="isOwnGroup && showRecipeContent"
+                  v-bind="$attrs"
```
michael-genson: On your question about this one: `inheritAttrs` is still default-true, so `class="mb-2"` from `view.vue` lands on the root card *and* gets re-applied here. Setting `inheritAttrs: false` alongside the forwarding would keep it to one target.

michael-genson (changes requested):
Thanks for this! Nice to see the planner broken up. A few things:

frontend/app/pages/household/mealplan/planner/view.vue
miah120: Huh, I actually don't remember adding something like that... If its the same to you I'll just remove it

frontend/app/pages/household/mealplan/planner.vue
miah120: Yep, fixed

frontend/app/components/Domain/Recipe/RecipeContextMenu/RecipeContextMenuContent.vue
miah120: I mostly added this for dev purposes so I could know when the linter was yelling at me vs a preexisting problem. Since this is using an `NoUndefinedField` rather than just requiring `recipeRef` itself to not be undefined, the v-if method wouldn't help here. We could make a `recipeHasNoUndefinedFields(): is NoUndefinedField<Recipe>` type of function, but that would probably be a pretty big lift, and a pain to maintain.

Alternatively, as you mentioned we could make the dialog just take a recipe instead of a `NoUndefinedField<Recipe>`, the place it uses that prop already has a related linter error so maybe it's just fine.

But I think that kinda starts creeping away from the real goals of this PR. Honestly it might make sense to have a dedicated PR come though for cleaning up all the linter errors like this...  

Basically if you want me to fix it I can, but otherwise I'll just remove the assertion.

frontend/app/components/Domain/Recipe/RecipeCardMobile.vue
miah120: fixed

frontend/app/components/Domain/Recipe/RecipeCardMobile.vue
miah120: I think if we want the emits to be declared explicitly, then it's probably easier to just emit them explicitly too :/ 

Looks like there is a way for vue components to make use of [generic types](https://stackoverflow.com/questions/78584096/is-it-possible-to-pass-a-generic-type-argument-to-vue-props), which would allow us to specify that this component only emits the components included in the buttons added by the user... but it would mean changing things to stop using the SFC pattern which would be a pretty big lift here, especially considering we'd need to transition over a few components. 

If it's alright with you, I've just removed the v-bind and opted to emit the events manually.

frontend/app/components/Domain/Mealplan/MealPlanDay.vue
miah120: Didn't do this originally because BaseButtonGroup didn't support submenus. I made BaseMenu to support that, so now the random options show up behind a submenu.

frontend/app/components/Domain/Recipe/RecipeContextMenu/RecipeContextMenuContent.vue
michael-genson: I'm good with just removing the assert for now, I agree it's a bit of a rabbit hole

frontend/app/pages/household/mealplan/planner/view.vue
michael-genson: Yeah that should be fine, very possible it came from somewhere else during the refactor

frontend/app/components/Domain/Recipe/RecipeCardMobile.vue
michael-genson: Yeah that's fine with me

michael-genson (approved):
Looks great, thanks for your work on this!

PR #7949, opened 2026-07-26
dev: Update front end unit test dependencies

miah120:
Updates front end unit test dependencies to work with the new nuxt version 

note: the --api.host=0.0.0.0 allows the unit test UI to be accessible via localhost. Otherwise it gets trapped in the container. The "yarn run" prefix also seems to be load bearing though I'm not really sure why... 

NA

## Testing
Tested in devcontainer

## AI / LLM Assistance
No AI used.


michael-genson (approved):
Thank you!

PR #7950, opened 2026-07-26
fix: Clear InputLabelType item-id as null instead of empty string

### LINKED ISSUE #7947: [BUG] - Cannot remove unit from shopping list item (fails silently)
### First Check

When I try to remove (clear) the unit from a shopping list item that has one, the change doesn't get saved but no error is shown in the UI; after the next refresh (manual or background refresh), the unit re-appears.

Screen recording:

https://github.com/user-attachments/assets/e908d95a-0b04-4ea2-8503-850e027e4946

### Steps to Reproduce

1. Add a shopping list item with a quantity and unit (e.g. go [here to the demo sites](https://demo.mealie.io/shopping-lists/be863cbc-4e7b-47cb-8a5f-fb482e43ddac), and add "1 kilogram of apples")
2. Edit the item that has a quantity and unit, to remove the unit (click the "x" button in the unit field), and save the item
3. The unit disappears (as expected). After a few seconds, the unit re-appears (bug)

Browser console shows this error:

```
Failed to load resource: the server responded with a status of 422 ()
CSHLN-dT.js:3 AxiosError: Request failed with status code 422
    at ls (5XpxOu-z.js:4:10504)
    at XMLHttpRequest.g (5XpxOu-z.js:4:16708)
    at Q.request (5XpxOu-z.js:6:2125)
    at async Object.safe (CSHLN-dT.js:3:18880)
    at async Object.put (CSHLN-dT.js:3:19258)
    at async wn.updateMany (CSHLN-dT.js:3:16292)
    at async b (BOaSGyGO2.js:1:22134)
    at async Object.x [as process] (BOaSGyGO2.js:1:22427)
    at async s (BOaSGyGO2.js:1:22815)
    at async c (BOaSGyGO2.js:1:23029)
```

### Mealie Version

3.21.0 (but has been around for a while) / reproduced on the current demo site

### Deployment

Docker (Linux)

_No response_

lehnerpat:
Empty string IDs fail UUID validation on shopping list item updates (422), so clearing a unit/food/label could not persist.

_(REQUIRED)_

This PR fixes a bug in the frontend `InputLabelType` component, which defaulted the `itemId.value` to empty string when a falsy input was given (e.g. when clearing the field with the "x" button). The backend rejects the empty string when UUIDs are expected; absence must be indicated by `null` in JSON, not empty string.

This fixes issues with clearing units and foods from shopping list items, which has been failing silently and then reverting on next data refresh.

### Screen recordings:

Before:

https://github.com/user-attachments/assets/e908d95a-0b04-4ea2-8503-850e027e4946

After:

https://github.com/user-attachments/assets/3891ab56-fd6a-4680-bba0-18290c930681

_(REQUIRED)_

Fixed #7947

This only fixes the specific bug that broke the clearing of food/unit; the bug was exacerbated by the shopping list view not handling API errors correctly (silently swallowing them). I think that should be fixed too, but I'm not familiar enough with the shopping list logic, so I didn't want to include that in this PR.

Even though `InputLabelType` is a "global" component, from what I can tell it's only used for shopping list items, so this code change shouldn't affect anything else; but I'd appreciate a second opinion on that.

## Testing

Manually, running the dev site locally.

## AI / LLM Assistance

_(REQUIRED)_

I used AI to diagnose the issue based on the bug symptoms I observed (see #7947), and to make the fix. I've manually reviewed the change.

### DIFF
--- a/frontend/app/components/global/InputLabelType.vue
+++ b/frontend/app/components/global/InputLabelType.vue
@@ -42,7 +42,7 @@ import { useSearch } from "~/composables/use-search";
 const modelValue = defineModel<MultiPurposeLabelSummary | IngredientFood | IngredientUnit | null>({ default: () => null });
 
 // support v-model:item-id binding
-const itemId = defineModel<string | undefined>("item-id", { default: undefined });
+const itemId = defineModel<string | null | undefined>("item-id", { default: undefined });
 
 const props = defineProps({
   items: {
@@ -81,7 +81,7 @@ const itemVal = computed({
     return modelValue.value;
   },
   set: (val) => {
-    itemId.value = val?.id || "";
+    itemId.value = val?.id ?? null;
     modelValue.value = val;
   },
 });


michael-genson (approved):
Thanks!

PR #7954, opened 2026-07-26
feat: add option to group checked shopping list items by label

chasestech:

_(REQUIRED)_

This feature makes it a bit easier to reuse shopping lists by allowing the user to toggle grouping of checked-off list items by their labels:

- **Sorting Logic**: Added `updateCheckedItemsByLabel` in `use-shopping-list-sorting.ts` to group checked items by label, sort the label categories following the existing label order from `labelSettings`, and preserve the descending check-off order (most recent to least recent) within each category.
- **UI Toggle**: Added a tag toggle button in the checked items section header on `[id].vue` to easily switch between flat and grouped views.
- **User Preference**: Added `groupCheckedByLabel` option to `UserShoppingListPreferences` so the toggle state persists in browser local storage across sessions.
- **Event Propagation**: Added `@click.stop` to the action buttons container in the checked items section header to prevent clicking action buttons from collapsing/expanding the section.

_(REQUIRED)_

N/A

- Category grouping for checked items is optional and defaults to `false` (flat list).

## Testing

- Added unit tests in `use-shopping-list-sorting.test.ts` verifying category grouping, handling of uncategorized items, category ordering from `labelSettings`, and internal item order preservation.
- Ran backend and frontend dev environments locally and manually verified UI grouping, toggling, and button click propagation.

### Before
mealie-current

### After
grouped-list
ungrouped-list

## AI / LLM Assistance

_(REQUIRED)_

Developed with Gemini 3.6 Flash and manually reviewed/tested.

### DIFF
--- a/frontend/app/composables/shopping-list-page/sub-composables/__tests__/use-shopping-list-sorting.test.ts
+++ b/frontend/app/composables/shopping-list-page/sub-composables/__tests__/use-shopping-list-sorting.test.ts
@@ -172,4 +172,32 @@ describe("use-shopping-list-sorting", () => {
       expect(sortedList.listItems).toEqual(undefined);
     });
   });
+
+  describe("updateCheckedItemsByLabel", () => {
+    const { updateCheckedItemsByLabel, t } = wrapper();
+    test("groups checked items by label and preserves internal order", () => {
+      const sortedList = {
+        ...MOCK_SHOPPING_LIST,
+      };
+      const checkedItems = [
+        { ...MOCK_ITEM, id: "c1", checked: true, label: MOCK_LABEL2.label, labelId: "2", updatedAt: "2026-01-02" },
+        { ...MOCK_ITEM, id: "c2", checked: true, label: MOCK_LABEL.label, labelId: "1", updatedAt: "2026-01-03" },
+        { ...MOCK_ITEM, id: "c3", checked: true, label: MOCK_LABEL.label, labelId: "1", updatedAt: "2026-01-01" },
+        { ...MOCK_ITEM, id: "c4", checked: true, updatedAt: "2026-01-04" },
+      ];
+      const result = updateCheckedItemsByLabel(sortedList, checkedItems);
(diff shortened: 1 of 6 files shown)


michael-genson:
Code generally looks okay, but I'm not convinced this is something we want to implement. Can you go more into detail on the motivation behind this? Generally checked items are just there for reference, I don't think I've seen anything like this on other similar shopping list apps. I'm also concerned with the burden of having multiple different ways of rendering the shopping list (we used to do something similar for unchecked items, but have since consolidated into the current iteration).

chasestech:
Hey @michael-genson, thanks for taking the time to review. Perhaps the way I'm using shopping lists isn't the intended method, but I'll share a bit of my reasoning.

I use shopping lists as persistent, reusable lists for each of my preferred stores. The contents are mostly staple items that don't change often. When it's time to shop, I refer back to my lists to see what I’ve previously checked off that needs restocking.

Currently, checked items are organized only by completion date, which makes finding specific items difficult in lists with dozens or hundreds of entries. Categorizing checked items would also allow me to look in my food storage by section (e.g., checking the fridge specifically for Dairy ingredients that I may need to replenish).

I'm open to suggestions on how this feature could be improved, or if I can optimize my workflow with Mealie's current capabilities.

michael-genson:
I see the usecase, though right now shopping lists are set up to clear items as they get checked. They get pruned from the database when there are too many (> 100, I think?) so there would have to be additional changes to make shopping lists viable as reusable.

The shopping list page is already quite complex so I worry about adding conditional rendering.

chasestech:
I see that in the codebase now, you're correct about the 100 item limit for checked items. I'm OK with closing this PR if you think that's best, or I can continue to work on this feature to see how to make the lists more reusable and address the current pruning behavior.

michael-genson:
Yeah for now I don't think we'll get this in. I'm definitely open to accommodating your usecase of reusing shopping lists better, but I'd like to see a more targeted solution instead of just adding labels to the checked items.

PR #7964, opened 2026-07-28
Draft PR. Fix all the warning in pytest.

bestxmg:
fix: eliminate pytest warnings and fix pre-existing test failures across migrations, backups, and timezone handling

While setting up the dev environment and running the test suite, I noticed the full `pytest` run was producing a large number of warnings, and turned up several pre-existing test failures along the way. This PR cleans all of that up so the suite runs clean (0 failures, 0 warnings), split into 10 focused commits — one per problem:

- **`mealie/services/migrations/utils/database_helpers.py`** — `_get_or_set_generic()` was calling `.model_dump()` on category/tag models before returning them, so `Recipe.recipe_category`/`Recipe.tags` ended up holding plain dicts instead of model instances for every recipe imported via a migrator. Fixed by removing the `.model_dump()` call.
- **`mealie/routes/handlers.py`** — replaced the deprecated `status.HTTP_422_UNPROCESSABLE_ENTITY` with `HTTP_422_UNPROCESSABLE_CONTENT`.
- **`mealie/services/backups_v2/alchemy_exporter.py`** (2 commits):
  - `dump()` sorted tables topologically via `self.meta.sorted_tables`, which can't resolve the intentional circular FK between `ai_providers`/`ai_provider_settings` and warns on every backup. Verified table order is actually inconsequential here — `restore()` disables FK enforcement for the whole operation via `ForeignKeyDisabler` — so switched to plain `self.meta.tables.values()`.
  - `drop_all()` used the deprecated `Inspector.from_engine()`; replaced with `sqlalchemy.inspect()`.
- **`tests/unit_tests/test_config.py`** — `monkeypatch.setenv()` was being passed raw `bool` values in the LDAP/OIDC settings tests; wrapped in `str()`.
- **`tests/fixtures/fixture_shopping_lists.py`** — teardown deleted a `ShoppingList` whose ORM-cached `list_items` collection was stale relative to the DB (a test had already deleted one item directly), causing a row-count-mismatch warning. Added `session.expire_all()` before the teardown delete.
- **`pyproject.toml` / `uv.lock`** (3 commits):
  - Added `httpx2` as a dev dependency — starlette's `TestClient` prefers it and warns when falling back to `httpx`, which is what nearly every integration test was doing.
  - Added a narrowly-scoped `filterwarnings` entry suppressing pyRdfa's deprecated `datetime.utcnow()` call (unmaintained third-party dependency, already flagged elsewhere in `mealie/app.py`).
  - Added a second `filterwarnings` entry suppressing sqlite's deprecated datetime-adapter warning, caused by `GroupDataExportsModel.expires` being declared a `String` column while always storing a real `datetime`. **This is a real, pre-existing bug I did not fix** — see Special notes below.
- **`tests/unit_tests/services_tests/scheduler/tasks/test_create_timeline_events.py`** and **`tests/integration_tests/user_household_tests/test_group_mealplan.py`** — 4 pre-existing test failures, all one root cause: these tests build "today's" meal-plan date with `datetime.now(UTC).date()`, but the app's "today" logic (both the timeline-events scheduler task and the `/households/mealplans/today` endpoint) intentionally resolves "today" using the household's *local* timezone. On a machine where local date and UTC date currently differ (reproduced on `Pacific/Auckland`, UTC+12), plans land on the wrong calendar day as far as the app is concerned. Fixed by building/comparing dates with `tzlocal()` instead, matching what the app actually does.

Each commit message has the full reason / consequence / fix for its specific change.

N/A — these weren't filed as GitHub issues; they were discovered incidentally while setting up the dev environment and running the full test suite locally.

One issue was intentionally **not** fixed, only flagged and suppressed: `GroupDataExportsModel.expires` (`mealie/db/models/group/exports.py`) is declared as a `String` column but is always populated with a real `datetime` (see `Exporter.run()` and the `GroupDataExport` Pydantic schema, which types it `datetime`). The correct fix is changing the column to `NaiveDateTime`, but that needs an Alembic migration and verification against Postgres — unlike sqlite, Postgres won't silently coerce a `datetime` bound against a live `VARCHAR` column, so this needs more careful review than I was comfortable doing unsupervised. I left a comment on the suppression explaining this; happy to follow up with the migration in a separate PR if maintainers want it.

The `ai_provider_settings`/`ai_providers` FK-cycle fix (dropping the topological sort in `dump()`) touches backup/restore code — I verified it with the full backup/restore round-trip test suite and traced through `ForeignKeyDisabler` to confirm table order was never actually load-bearing there, but flagging it for extra reviewer attention given the blast radius if I'm wrong.

## Testing

Ran the full test suite (`uv run pytest`) after each individual commit and once more at the end after the full split — 0 failures, 0 warnings, only the pre-existing intentional skips (LDAP tests requiring an external service, long-running scraper/parser tests) remain. Also specifically re-ran `test_alchemy_exporter.py` and `test_backup_v2.py` (full backup/restore round-trip) after the FK-cycle fix, and confirmed the 4 timezone-related failures are reproducible on a clean `mealie-next` checkout with none of these changes applied, to rule out a regression from this PR itself.

## AI / LLM Assistance

Claude (Anthropic, Sonnet 5) was used extensively throughout this PR: diagnosing the root cause of each warning/failure (including tracing through SQLAlchemy/starlette internals and reproducing the timezone bug), writing the fixes, splitting the working changes into these 10 atomic commits, and drafting the commit messages and this PR description. I reviewed and directed the work, decided what was safe to auto-fix vs. flag for review (e.g. deliberately not touching the `GroupDataExportsModel.expires` schema issue), and ran/verified the test suite results myself before approving the commits.


bestxmg:
Don't merge please. It's still working in progress.

bestxmg:
Don't merge, please. It's our coursework. It's still working in progress.

PR #7966, opened 2026-07-28
feat: add rating as a query filter field

dominikrein:

Adds "Rating" to the field list of the query filter builder, in all three places the builder is used:

- `frontend/app/components/Domain/Household/GroupMealPlanRuleForm.vue` — meal plan rules
- `frontend/app/components/Domain/Cookbook/CookbookEditor.vue` — cookbook editor
- `frontend/app/pages/g/[groupSlug]/recipes/finder/index.vue` — recipe finder

`RecipeModel.rating` is already declared as a `FilterableColumn`, so filtering by rating works today when the filter string is set through the API — it is just not offered in the UI. This makes rules like "pick a random dinner rated 4 or higher" or a cookbook of "everything rated 4+" possible without hand-writing a query filter string.

The change is purely additive: one entry per `fieldDefs` list, 5 lines each. There are no backend changes, no new translation keys (`general.rating` already exists, so every locale is covered), and no new field type — the builder's `"number"` type is already fully implemented (numeric input, parsing, serialization) but was not used by any field list until now.

The entry is placed after `user_id` so it groups with the recipe's own properties rather than sitting among the organizers.

image

None — I did not find an existing issue for this.

Two behaviours are worth being aware of. I chose to leave both as they are rather than expand the scope of this PR, but I'm happy to revisit if you'd rather handle them here:

- `rating` is the average across all users' ratings, not the requesting user's own rating. For single-user instances these are identical. A separate "my rating" field would need work on the filter path and a design decision, so it felt out of scope here.
- Unrated recipes have a `NULL` rating and are therefore excluded from every comparison, including `rating < 3`. This is standard SQL behaviour and matches how the existing `last_made` field already behaves.

I also considered constraining the input to 0–5, but that would require extending the `FieldDefinition` interface and would affect every field type, which turns an additive change into a refactor.

## Testing

Verified manually against a local instance, for all three usages:

- The field appears in the dropdown at the intended position and renders a numeric input.
- The operator list for the field is the six comparison operators.
- A rule saved as `rating >= 4` is persisted correctly and parsed back into a numeric field after a page reload. This path mattered most to me, since `type: "number"` had no production usage before this change.
- `POST /api/households/mealplans/random` respects the threshold: across repeated calls it only ever returned the rated recipe, never the unrated one.
- `yarn lint --max-warnings=0` clean, `yarn test` 151/151 passing.

## AI / LLM Assistance

Claude Code was used to locate the relevant code, write the change, and verify it against a local instance. I reviewed the full diff before submitting.


michael-genson (approved):
Thank you!

PR #7969, opened 2026-07-28
dev: Shopping list unit test

miah120:
Shopping list composables are now mostly all tested. 

Also removed some unused code from the shopping list. 

NA

## Testing
Yep, that's the idea

Tested in browser to make sure the code that was removed didn't break anything. 

## AI / LLM Assistance
No AI used.


michael-genson (approved):
Awesome, thank you! Thanks for also cleaning up some dead code.

PR #7984, opened 2026-07-31
feat: Improve recipe parsing ui

miah120:
Goals: 
* Improve the mobile ingredient parsing ui
* Generally improve the ui by...
  * streamlining presented information
  * establishing a visual hierarchy
  * adding some animations
* Clean up the code by splitting the components into smaller parts
* Improve unit test coverage 

Files changed:
* frontend/app/components/Domain/Recipe/RecipePage/RecipePage.vue
  * Don't exit edit mode when done parsing ingredients 
* frontend/app/components/Domain/Recipe/RecipeIngredientEditor.vue
* frontend/app/components/Domain/Recipe/RecipeIngredientEditorLayout.vue
* frontend/app/components/Domain/Recipe/RecipePage/RecipePageParts/RecipePageIngredientEditor.vue
  * fix layout for mobile  
* frontend/app/components/Domain/Recipe/RecipePage/RecipePageParts/RecipePageParseDialog.vue
* frontend/app/components/Domain/Recipe/RecipePage/RecipePageParts/RecipeParseDialog/ParseDialogChangeParser.vue
* frontend/app/components/Domain/Recipe/RecipePage/RecipePageParts/RecipeParseDialog/ParseDialogInfo.vue
* frontend/app/components/Domain/Recipe/RecipePage/RecipePageParts/RecipeParseDialog/ParseDialogParse.vue
* frontend/app/components/Domain/Recipe/RecipePage/RecipePageParts/RecipeParseDialog/ParseDialogReview.vue
* frontend/app/components/Domain/Recipe/RecipePage/RecipePageParts/RecipeParseDialog/RecipePageParseDialog.vue
  * break the dialog out into smaller components 
* frontend/app/components/global/Transition/SwipeTransition.vue
  * Add some nice animations 
* frontend/app/composables/recipes/__tests__/use-parse-ingredients-dialog.test.ts
  * Unit tests 
* frontend/app/composables/recipes/use-parse-ingredients-dialog.ts
  * Created to centralize the dialogs functionality 
* frontend/app/composables/use-users/preferences.ts
  * track "don't show this again" in local storage 
* frontend/app/lang/messages/en-US.json
  * strings 

***
Before/After:
|  | Before | After |
|--------|--------|--------|
| banner desktop | image | image |
| banner mobile | image | image |
| info step desktop | NA | image |
| info step mobile | NA | image |
| parse step desktop | image | image | 
| parse step mobile | image | image | 
| review step desktop | image | image | 
| review step mobile | image | image | 

***
Animations: 
[Screencast_20260730_201700.webm](https://github.com/user-attachments/assets/e3c95224-40f1-4744-a369-b68300e8ed6d)

NA

## Testing
Tested in light and dark mode, on desktop and mobile (android) firefox. 
New composable for the parse dialog is fully unit tested 🎉

## AI / LLM Assistance
No AI used.

### DIFF
--- a/frontend/app/components/Domain/Recipe/RecipeIngredientEditor.vue
+++ b/frontend/app/components/Domain/Recipe/RecipeIngredientEditor.vue
@@ -1,227 +1,182 @@
 <template>
-  <div>
-    <v-text-field
-      v-if="model.title || showTitle"
-      v-model="model.title"
-      density="compact"
-      variant="underlined"
-      hide-details
-      class="mx-1 mt-3 mb-4"
-      :placeholder="$t('recipe.section-title')"
-      style="max-width: 500px"
-      @click="$emit('clickIngredientField', 'title')"
-    />
-    <v-row
-      :no-gutters="mdAndUp"
-      density="comfortable"
-      class="d-flex flex-wrap my-1"
-    >
-      <v-col
-        sm="12"
-        md="2"
-        cols="12"
-        class="flex-grow-0 flex-shrink-0"
+  <v-card class="my-2">
+    <div v-if="enableDragHandle || enableContextMenu || model.title || showTitle" class="d-flex pt-2 align-center">
+      <template v-if="enableDragHandle">
+        <v-icon class="ma-2 handle" size="large">
+          {{ $globals.icons.arrowUpDown }}
+        </v-icon>
+      </template>
+
+      <v-text-field
+        v-if="model.title || showTitle"
+        v-model="model.title"
+        density="compact"
+        variant="underlined"
+        hide-details
+        class="mx-1 mt-3 mb-4"
+        :placeholder="$t('recipe.section-title')"
+        style="max-width: 500px"
+        @click="$emit('clickIngredientField', 'title')"
+      />
+      <BaseButtonGroup
+        v-if="enableContextMenu"
+        hover
+        :large="false"
+        class="ml-auto"
+        :buttons="btns"
+        @toggle-section="toggleTitle"
+        @toggle-subrecipe="toggleIsRecipe"
+        @insert-above="$emit('insert-above')"
+        @insert-below="$emit('insert-below')"
+        @delete="$emit('delete')"
+      />
+    </div>
+    <div class="d-flex ga-2 pa-2" :class="$vuetify.display.smAndDown ? 'flex-column' : ''">
+      <v-number-input
+        v-model="model.quantity"
+        variant="filled"
+        :precision="null"
+        :min="0"
+        hide-details
+        inset
+        control-variant="stacked"
+        density="compact"
+        :placeholder="$t('recipe.quantity')"
+        @keypress="quantityFilter"
+      />
+      <v-autocomplete
+        ref="unitAutocomplete"
+        v-model="model.unit"
+        v-model:search="unitSearch"
+        auto-select-first
+        hide-details
+        density="compact"
+        variant="filled"
+        return-object
+        :items="filteredUnits"
+        :custom-filter="() => true"
+        item-title="name"
+        :placeholder="$t('recipe.choose-unit')"
+        clearable
+        :menu-props="{ attach: props.menuAttachTarget, maxHeight: '250px' }"
+        @keyup.enter="handleUnitEnter"
       >
-        <v-number-input
-          v-model="model.quantity"
-          variant="solo"
-          :precision="null"
-          :min="0"
-          hide-details
-          control-variant="stacked"
-          inset
-          density="compact"
-          :placeholder="$t('recipe.quantity')"
-          @keypress="quantityFilter"
-        >
-          <template v-if="enableDragHandle" #prepend>
-            <v-icon
-              class="mr-n1 handle"
-            >
-              {{ $globals.icons.arrowUpDown }}
-            </v-icon>
-          </template>
-        </v-number-input>
-      </v-col>
-      <v-col
-        sm="12"
-        md="2"
-        cols="12"
-      >
-        <v-autocomplete
-          ref="unitAutocomplete"
-          v-model="model.unit"
-          v-model:search="unitSearch"
-          auto-select-first
-          hide-details
-          density="compact"
-          variant="solo"
-          return-object
-          :items="filteredUnits"
-          :custom-filter="() => true"
-          item-title="name"
-          class="mx-1"
-          :placeholder="$t('recipe.choose-unit')"
-          clearable
-          :menu-props="{ attach: props.menuAttachTarget, maxHeight: '250px' }"
-          @keyup.enter="handleUnitEnter"
-        >
-          <template #prepend>
-            <v-tooltip v-if="unitError" location="bottom">
-              <template #activator="{ props: unitTooltipProps }">
-                <v-icon
-                  v-bind="unitTooltipProps"
-                  class="ml-2 mr-n3 opacity-100"
-                  color="primary"
-                >
-                  {{ $globals.icons.alert }}
-                </v-icon>
-              </template>
-              <span v-if="unitErrorTooltip">
-                {{ unitErrorTooltip }}
-              </span>
-            </v-tooltip>
-          </template>
-          <template #no-data>
-            <div class="caption text-center pb-2">
-              {{ $t("recipe.press-enter-to-create") }}
-            </div>
-          </template>
-          <template #append-item>
-            <div v-if="showCreateUnit" class="px-2">
-              <BaseButton
-                block
-                size="small"
-                @click="createAssignUnit()"
-              />
-            </div>
-          </template>
-        </v-autocomplete>
-      </v-col>
+        <template v-if="unitError" #prepend-inner>
+          <v-tooltip location="bottom">
+            <template #activator="{ props: unitTooltipProps }">
+              <v-icon
+                v-bind="unitTooltipProps"
+                class="opacity-100"
+                color="primary"
+              >
+                {{ $globals.icons.alert }}
+              </v-icon>
+            </template>
+            <span v-if="unitErrorTooltip">
+              {{ unitErrorTooltip }}
+            </span>
+          </v-tooltip>
+        </template>
+        <template #no-data>
+          <div class="caption text-center pb-2">
+            {{ $t("recipe.press-enter-to-create") }}
+          </div>
+        </template>
+        <template #append-item>
+          <div v-if="showCreateUnit" class="px-2">
+            <BaseButton
+              block
+              size="small"
+              @click="createAssignUnit()"
+            />
+          </div>
+        </template>
+      </v-autocomplete>
 
       <!-- Foods Input -->
-      <v-col
+      <v-autocomplete
         v-if="!state.isRecipe"
-        m="12"
-        md="4"
-        cols="12"
-        class=""
+        ref="foodAutocomplete"
+        v-model="model.food"
+        v-model:search="foodSearch"
+        auto-select-first
+        hide-details
+        density="compact"
+        variant="filled"
+        return-object
+        :items="filteredFoods"
+        :custom-filter="() => true"
+        item-title="name"
+        :placeholder="$t('recipe.choose-food')"
+        clearable
+        :menu-props="{ attach: props.menuAttachTarget, maxHeight: '250px' }"
+        @keyup.enter="handleFoodEnter"
       >
-        <v-autocomplete
-          ref="foodAutocomplete"
-          v-model="model.food"
-          v-model:search="foodSearch"
-          auto-select-first
-          hide-details
-          density="compact"
-          variant="solo"
-          return-object
-          :items="filteredFoods"
-          :custom-filter="() => true"
-          item-title="name"
-          class="mx-1 py-0"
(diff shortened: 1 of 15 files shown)


p0lycarpio:
The new information banners look really good, especially on mobile. Is there a specific reason for placing the text fields inside a card? It takes up more space and creates a "stacking" effect of multiple elevation layers.

miah120:
> The new information banners look really good, especially on mobile. 

Thanks! 


>Is there a specific reason for placing the text fields inside a card?

Largely because it just felt natural to me as a way to group the various icons in the mobile view; and I didn't feel like trying out a bunch of different options if no one was gonna notice. I can put forward some alternate methods, cause yeah the stacking effect does bug me a lil

miah120:
Yeah, I think this looks better.


image


image

frontend/app/components/Domain/Recipe/RecipePage/RecipePageParts/RecipeParseDialog/ParseDialogReview.vue
```diff
@@ -0,0 +1,87 @@
...
+      v-model="state.parser"
+      :available-parsers="availableParsers"
+      @update:model-value="$emit('changeParser', $event)"
+      @parse="$emit('parse')"
+    />
+    <div>
+      <v-card-title class="text-center pt-0 pb-8">
+        {{ $t("recipe.parser.review-parsed-ingredients") }}
+      </v-card-title>
+      <v-card-text style="max-height: 60vh; overflow-y: auto;">
```
p0lycarpio: This style makes only the ingredients scrollable and doesn't take up the full height of the card on mobile. Removing it seems more responsive to me.

```suggestion
      <v-card-text>
```

frontend/app/components/Domain/Recipe/RecipeIngredientEditor.vue
```diff
@@ -11,229 +11,191 @@
...
+          @insert-below="$emit('insert-below')"
+          @delete="$emit('delete')"
+        />
+      </template>
+      <template #form>
+        <div class="flex-grow-1">
+          <div class="d-flex ga-2 py-2" :class="$vuetify.display.mdAndDown ? 'flex-column' : ''">
+            <v-number-input
+              v-model="model.quantity"
+              variant="filled"
```
p0lycarpio, line 38: Is there a particular reason why the field variant was changed from "solo" to "filled"?

p0lycarpio:
Overall, it looks better; however, I’m confused by the new parser information step that displays the green checkmark and the text. Seeing that screen made me think all my ingredients had been parsed and there was nothing left to do which isn't the case when some ingredients don't exist yet.

I wonder if the previous version wasn't more concise and didn't save a click or a step. In fact, perhaps it would be better to simply shorten the message and ask the user to "adjust/create unmatched ingredients"?

What do you think?

frontend/app/components/Domain/Recipe/RecipePage/RecipePageParts/RecipeParseDialog/ParseDialogReview.vue
miah120: Ah, yep. Looks like that got copied over from the original implementation. I agree it doesn't look the best. My guess is it was left over from when they made all popups take up the full height of the page on mobile. 

Took it out :)

frontend/app/components/Domain/Recipe/RecipeIngredientEditor.vue
miah120: In general, elevating content is a way to draw additional attention to it. When you have a big table of cells, you can't really give your attention to every cell at once, so it just feels a little off. 

That's why the term "solo" is used; its a good style for when you have a solo input field (e.g., a search field).

miah120:
> I wonder if the previous version wasn't more concise and didn't save a click or a step. In fact, perhaps it would be better to simply shorten the message and ask the user to "adjust/create unmatched ingredients"?
> 
> What do you think?

I'm down to workshop the verbiage here -- I agree it's a little clunky -- I'm just not the best writer so I figured I wouldn't try myself or I'd end up bungling it worse.

As for the extra page; yes, it is an extra click (or two with the new checkbox), but I still think it's worth keeping around. In particular, it serves as a quick explanation for new users as to what they should expect. Is everyone gonna read it? Probably not. But its a decent way to not leave people entirely confused on what to expect.  

However, it's also entirely useless to anyone who's already familiar with this workflow since it's static content; thus the "don't show this again" option. This clears up all the subsequent screens which allows for a more focused, less noisy workflow. 

> Overall, it looks better; however, I’m confused by the new parser information step that displays the green checkmark and the text. Seeing that screen made me think all my ingredients had been parsed and there was nothing left to do which isn't the case when some ingredients don't exist yet.

Yeah that's fair, I hadn't really considered that. I think a different icon there might make it more clear that action is still necessary... probably not a warning icon since that implies something went wrong, but maybe like the info icon or robot icon or something. I'll browse the mdi list and see if anything jumps out. 

Thanks for the feedback! I'll workshop a few things and see if something else works better.

frontend/app/components/Domain/Recipe/RecipeIngredientEditor.vue
p0lycarpio: All right, I see. That makes sense

miah120:
Mdi gives us the progress check and progress tag icons, here's how they look w what I think is the most appropriate colors. 

I think progress check w the success color seems most fitting to me. 

| success | info |
|--------|--------|
| image | image |
| image | image |

I also added a bit more structure to the page which should hopefully help.
image

What do you think?

p0lycarpio:
Thank you for these suggestions and improvements! I agree that the progress check icon is indeed more appropriate for this step, and the last suggestion (with the revised structure and headings) is also essential!

Come to think of it, I wonder if the title and description for this step could be dynamic based on the remaining items. We could display contextual title and text depending on the analysis result (whether everything has been automatically analyzed or manual verification is needed), while also adjusting the icon accordingly (partial or full).

The current state already contains everything needed (`parsedIngs` logic + `shouldReview`), so exposing derived counters (e.g., `autoParsedCount`, `needsReviewCount`, `allAutoParsed`) from the dialog’s existing state would help maintain a single source of truth and avoid duplicating logic across components.

Of course, it's not a deal-breaker; the PR already brings a lot of improvements.

What do you think?

miah120:
> Thank you for these suggestions and improvements! I agree that the progress check icon is indeed more appropriate for this step, and the last suggestion (with the revised structure and headings) is also essential!

Glad you like it :) I'll commit those changes then

> Come to think of it, I wonder if the title and description for this step could be dynamic based on the remaining items. We could display contextual title and text depending on the analysis result (whether everything has been automatically analyzed or manual verification is needed), while also adjusting the icon accordingly (partial or full).
> ... 
> What do you think?

Hmm, so the way I'd think about it at least is to group people into new users and experienced users. 

For experienced users, they (ideally) won't see the info screen at all, however some sort of visual hint to show progress in the workflow or how many items need action could be nice although admittedly I don't really care about having that personally. Something like this: 
[Screencast_20260827_093611.webm](https://github.com/user-attachments/assets/a9763942-ef28-46a4-ac35-efc353f71eec)

However, I'd argue that any additional information would be redundant or add unnecessary noise. We don't need to tell an experienced user that we auto parsed everything, we can just take them to the review page. 

For newer users, I think for the most part having static content is best since the idea is this screen only gets shown the first time the user goes through the workflow, so we want to make sure they get the relevant information. But having a little readout like this could make it feel a little more dynamic: 
| Some to review | All auto parsed  |
|--------|--------|
| image | image | 

I think we can leave the icon as-is given that there will always be the review step after this page. 

What do you think?

miah120:
Also I realized when I actually implemented that (they were just mockups for the screenshot), you cant actually get to the "all ingredients parsed" page because we just take them immediately to the review page. Which... we could opt to always show the info page, but imo it works either way so it's probably fine to keep as is.

miah120:
The unit tests are currently failing due to some expected numerical changes. I'm working on fixing them, but there was also a drop in coverage after some of the changes here that doesn't really make sense to me, but I'm looking into it.

frontend/app/components/Domain/Recipe/RecipePage/RecipePageParts/RecipeParseDialog/ParseDialogInfo.vue
```diff
@@ -0,0 +1,38 @@
+<template>
+  <SpinTransition>
```
p0lycarpio: `SpinTransition` doesn't exists and throw an error in console. Did you mean `SwipeTransition`?

frontend/app/components/Domain/Recipe/RecipePage/RecipePageParts/RecipeParseDialog/RecipePageParseDialog.vue
```diff
@@ -0,0 +1,117 @@
...
+            :available-parsers="availableParsers"
+            :parser="parser"
+            @parse="parseIngredients"
+            @change-parser="(newParser) => parser = newParser"
+          />
+        </div>
+      </SwipeTransition>
+    </v-container>
+    <template v-if="state.step !== ParseStep.LOADING" #custom-card-action>
+      <SpinTransition>
```
p0lycarpio: Same comment as for SpinTransition

frontend/app/components/Domain/Recipe/RecipePage/RecipePageParts/RecipeParseDialog/RecipePageParseDialog.vue
```diff
@@ -0,0 +1,117 @@
+<template>
+  <BaseDialog
+    :model-value="modelValue"
+    :title="$t('recipe.parse-ingredients')"
+    :icon="$globals.icons.fileSign"
+    disable-submit-on-enter
+    @update:model-value="emit('update:modelValue', $event)"
+  >
+    <v-container fluid class="pa-2 ma-0" style="background-color: rgb(var(--v-theme-background));">
```
p0lycarpio: I suggest adding the `overflow-x-hidden` class, which prevents the scrollbar from appearing during the animation.
It would make more sense to put this in the `SwipeTransition` component, but I wasn't able to get it to work.

I also removed the `style` tag, which doesn't seem necessary.

```suggestion
    <v-container fluid class="pa-2 ma-0 overflow-x-hidden">
```

frontend/app/components/Domain/Recipe/RecipePage/RecipePageParts/RecipeParseDialog/ParseDialogChangeParser.vue
```diff
@@ -0,0 +1,58 @@
+<template>
+  <v-card
+    variant="outlined"
+    color="info"
+    class="d-flex justify-space-between align-center"
+    @click="$emit('parse')"
```
p0lycarpio, line 6: Clicking on the current parser restarts the analysis, but `SwipeTransition` is replayed, and a user who hasn't hidden the first step is redirected back to it.
Is this intentional?

p0lycarpio (changes requested):
Thanks for all the work you've put into this!
I really like the first page with the dynamic display of parsed and remaining items; it seems clear to me.
In hindsight, I feel that displaying progress via the stepper clutters the various step pages without adding much value. Perhaps it could be placed at the top of the dialog, like a loading bar? Personally, I would just remove it.

I’ve added a few comments on the code. Thanks again!

frontend/app/components/Domain/Recipe/RecipePage/RecipePageParts/RecipeParseDialog/ParseDialogInfo.vue
miah120: Removed

frontend/app/components/Domain/Recipe/RecipePage/RecipePageParts/RecipeParseDialog/RecipePageParseDialog.vue
miah120: Replaced with swipe transition

frontend/app/components/Domain/Recipe/RecipePage/RecipePageParts/RecipeParseDialog/RecipePageParseDialog.vue
miah120: Added to the swipe transition

frontend/app/components/Domain/Recipe/RecipePage/RecipePageParts/RecipeParseDialog/ParseDialogChangeParser.vue
miah120: It's not really intentional... but I'm not sure its worth it to fix. It's about ~15-20 lines of code to hide, and involves adding extra props/event handlers to like 7-8 different functions/components. If you think it's important, then I can make the change, but otherwise I think it's fine to leave it as is.

p0lycarpio (approved):
I've made few fixes. Thanks!

michael-genson (approved):
LGTM

PR #7990, opened 2026-08-01
[codex] Use Codex for Mealie import ingredient matching

tbhanson96:
## What changed

- Route Mealie's AI-backed web recipe import through the Codex structured recipe extraction path.
- Add known Mealie foods and units to the Codex prompt so Codex can return `foodId` and `unitId` when there is a clear catalog match.
- Preserve structured ingredients through scraper cleanup instead of flattening them back to display strings.
- Keep local food/unit matching as a fallback when Codex omits an ID or returns no clear match.

## Why

The previous web and social import flows let AI extract ingredient names but matched Mealie ingredient identifiers only afterward. Supplying known catalog entries to Codex lets imports choose existing food and unit identifiers up front, improving ingredient reuse for recipe scaling and shopping list aggregation.

## Validation

- `UV_FROZEN=1 /home/tim/.local/bin/uv run --no-sync pytest tests/integration_tests/user_recipe_tests/test_recipe_create_from_social.py tests/integration_tests/user_recipe_tests/test_recipe_create_from_openai.py`
- `UV_FROZEN=1 /home/tim/.local/bin/uv run --no-sync pytest`

Note: local test setup used `uv sync --no-install-package python-ldap` with a virtualenv-only LDAP import shim because this environment cannot build `python-ldap` without `aarch64-linux-gnu-gcc`, and `sudo` requires a password.


tbhanson96:
Closing because this PR was opened against the wrong repository target. Re-opening against the fork instead.

PR #8001, opened 2026-08-03
feat: Instruction time

miah120:
Adds a way to track the time each step in a recipe takes. This is useful for tracking at a glance how long each component of the recipe will take. 

In the future, I'd like to work towards using this to create a timeline for the recipe that can take into account which steps need to happen before others, and which steps can be done in parallel in order to ensure each component of each recipe finishes at the same time; but I figure this is helpful enough as is. 

Edit mode:
image

View/Cook mode:
image

NA

## AI / LLM Assistance
No AI used.


mealie/schema/recipe/recipe_step.py
```diff
@@ -21,3 +21,5 @@ class RecipeStep(MealieModel):
     text: str
     ingredient_references: list[IngredientReferences] = []
     model_config = ConfigDict(from_attributes=True)
+
+    cook_time: str | None = None
```
michael-genson, line 25: Since this is a new API field, could we store it as an integer number of minutes instead of an `"H:MM"` string? A formatted string isn't validated server-side and will be harder to work with if we build on this later (e.g. the timeline idea). If so, could we include the unit in the name (e.g. `duration_minutes`, both here and on the DB column)? That also avoids confusion with the recipe-level `cook_time`, which is a free-text string.

frontend/app/components/Domain/Recipe/RecipePage/RecipePageParts/CookTime.vue
```diff
@@ -0,0 +1,81 @@
...
+        @click="time = undefined"
+      />
+    </div>
+  </div>
+  <div v-else-if="isEditMode">
+    <BaseButton
+      secondary
+      color="info"
+      class="rounded-pill"
+      @click="time = '0:00'"
```
michael-genson, line 45: Do we need the "add, then set" flow? This saves a `0:00` time as soon as the button is clicked. Could we instead always show the input in edit mode (a single minutes field would keep steps from getting cluttered), and treat empty or 0 as `null`? That would also remove the need for the add and clear buttons.

frontend/app/components/Domain/Recipe/RecipePage/RecipePageParts/CookTime.vue
```diff
@@ -0,0 +1,81 @@
...
+    >
+      {{ $t('recipe.perform-time') }}
+    </BaseButton>
+  </div>
+  <div v-else-if="time" style="min-width: fit-content;">
+    <v-chip color="info" size="large" :prepend-icon="$globals.icons.clockOutline">
+      {{ time }}
```
michael-genson, line 52: Can we format this with units (e.g. `1h 30m`)? `1:30` on its own is ambiguous.

mealie/alembic/versions/2026-09-04-16.14.42_afdc42353aa4_add_cook_time_to_recipe_steps.py
```diff
@@ -0,0 +1,33 @@
...
+
+"""
+
+import sqlalchemy as sa
+from alembic import op
+
+
+# revision identifiers, used by Alembic.
+revision = "afdc42353aa4"
+down_revision: str | None = "b3f1c9a27d84"
```
michael-genson, line 15: `mealie-next` has added migrations after `b3f1c9a27d84`, so this will create a second Alembic head. Can you rebase and point this at the current head?

michael-genson (changes requested):
Love this feature, it's definitely something I've thought about quite a bit

granto:
Thanks for working on this—I'm developing TimerVana, a step-based timer app, and would like to support the format Mealie adopts rather than open a competing instruction-time PR.

We have a local, AI-assisted implementation of optional instruction durations, with tests for persistence, clearing values, export/import, and migration preservation. We have also exercised Mealie-to-TimerVana transfer and duration preservation in our web development environment and installed iOS TestFlight app on a Mac. I'd be happy to contribute relevant tests or downstream compatibility feedback here.

One question on the proposed integer `duration_minutes`: would integer seconds with an explicit unit in the field name be worth considering? That would preserve short steps such as 30 seconds or 1 minute 30 seconds without requiring a more complicated editor. I understand the preference for a simple minutes field; if whole minutes is the preferred contract, we can adapt the consumer accordingly.

Is there a useful way to help with this PR's remaining work? No dependency on TimerVana or a separate sharing feature is needed for instruction durations.

Disclosure: our local implementation and test work used AI assistance, and this comment was drafted with AI assistance.

granto:
Following up with a concrete contribution, if useful: I've prepared a patch addressing the four review requests using **integer minutes**, setting aside the seconds question so it doesn't hold this work up.

Branch: https://github.com/granto/mealie/tree/concretehelp/pr8001-minutes

Your two original commits are retained with your authorship, rebased onto `mealie-next` at `eb20622970856fd114e45e5139a2485857ff101d`. The separately attributed follow-up commits are:
- https://github.com/granto/mealie/commit/34b14e7be5b230cfc311adc8a2c963a966d2535d
- https://github.com/granto/mealie/commit/47f8c06cb175ef6ec2575079963308dd8f5c3165

Changes:
- Nullable, validated `duration_minutes` / `durationMinutes`, with zero normalized to null.
- One always-visible minutes field in edit mode; blank clears it.
- Unit-labeled duration display, rather than ambiguous `H:MM`.
- Migration attached to the upstream head at that base, plus regenerated API types without the unrelated type changes.
- Save-path validation so invalid input cannot silently save the prior duration and leave edit mode.

Verification: 45 focused tests passed against SQLite and 45 against PostgreSQL (including persistence, clearing, export/import, and populated migration preservation); the final full frontend suite passed 484 tests, and the Nuxt production build passed. Focused independent AI-assisted review/testing also covered the save guard.

Limitations: full backend suite, a clean whole-project typecheck, full browser acceptance, and upstream CI have not been completed for this branch. Local frontend dependencies were isolated copies rather than a fresh lockfile install. This is an offered patch, not a claim that it is merge-ready; migration ancestry should be rechecked against the base when adopted.

Would you prefer a PR against your branch, or to take/adapt these commits yourself? Because the original commits were rebased, these are not promised to cherry-pick conflict-free onto the old head. I haven't opened a replacement upstream PR, and there is no TimerVana dependency or sharing feature bundled here.

**AI assistance:** the follow-up implementation, tests, review, and this comment were produced with AI assistance. Your original commits remain separately attributed; if adopting this work, the PR's current AI disclosure would need updating accordingly.

PR #8016, opened 2026-08-05
fix: keep the whole image visible after rotating it in the image cropper

### LINKED ISSUE #5602: [BUG] - Image cropper should mount/react with the max(width, height) of the input image
### First Check

When creating a recipe with an image, you can crop/rotate the image.
However, for a rectangular image (where width is significantly different from height, or vice versa), rotating the image doesn't cause the Cropper component to change the view window. This means some of the now rotated image is no longer visible.

### Steps to Reproduce

1. Go to `/g/home/r/create/image`
2. Upload a rectangular image

Image

3. Rotate the image

Image

Some of the image is now not shown in the Cropper.

My expectation is that either the Cropper component should be mounted with the max(width, height) as its dimensions (ie. as a square). Or it should react to what the new dimensions of the image are after rotation.

N/A

### Mealie Version

mealie:nighlty as at 2025-06-29

### Deployment

Docker (Linux)

_No response_

derekdiliu:

Rotating a rectangular image in the cropper cut part of it off. A quarter turn swaps the image's width and height, but the cropper kept the box it had computed for the old orientation, so the rotated image no longer fit inside it. You had to zoom out by hand to get the rest back.

After a rotation the cropper's aspect ratio now follows the image's current orientation, and the cropper re-measures itself so the visible area is recomputed. Sizing is left alone until the first rotation, so nothing changes on mount.

That makes `cropperHeight` redundant, so it's dropped along with its two callers. In `RecipeLastMade.vue` this also means the cropper is no longer pinned to `20vh`, which used to crop the image before you'd touched anything.

### Before / after

Rotating a 1600x600 image on `/g/<group>/r/create/image`, corners colour-coded so cropping is obvious.

Before, only a narrow band survives the rotation:

Before: rotating leaves only a narrow band of the image visible

After:

After: the whole rotated image stays visible

Fixes #5602

In case you wonder why not just call the library's own methods: `refresh()` only re-reads the container size, so it works where the container has no definite height and does nothing where it does. `reset()` recomputes the visible area but resets `appliedImageTransforms` first, which undoes the rotation.

Since the box follows the image, a rotated 3000x200 panorama gives a 405x6079 cropper. Correct, but very tall. That isn't new, a 200x3000 image already does this today, so I left it rather than add a height cap. Happy to add one if you'd prefer.

## Testing

Manual, with test images whose four corners are colour-coded so any cropping is immediately visible.

Rotated both call sites (`create/image` and the "I Made This" dialog) repeatedly in both directions at 1600x600, 4000x3000, 600x1600, 3000x200 and 120x90. The whole image stays visible in every case, the mounted size is unchanged from before this PR, four rotations return to the starting state with no drift, and cropping and saving still produce the expected image. `task ui:check` passes.

No automated test: the frontend suite has no component-rendering setup, and this is layout geometry the cropper library computes.

## AI / LLM Assistance


michael-genson (dismissed):
Thanks! This was definitely annoying behavior

derekdiliu:
Merged mealie-next in to resolve the conflict from #8043. Since create/image.vue is now a redirect stub, the cropper-height removal moved to its new caller in RecipeImportImages.vue. The fix itself is unchanged.

michael-genson (approved):

PR #8018, opened 2026-08-05
fix: admin unable to edit recipes in other households (#7810)

### LINKED ISSUE #7810: [BUG] - Admin account can't modify other housholds recipes
### First Check

Administrator account can't modify recipes if they belong to a different household than the Administrator's

### Steps to Reproduce

1) Create 2 households in the same group
2) Create a user in the SECOND household with which you create/import a recipe
3) Make sure the Administrator account belongs to the FIRST household (or any other household that is not the SECOND one)
4) Using the Administrator account, browse the SECOND household recipes, open one and see the modify button is not present (thou, it is possible to delete the recipe, using the "..." menu)

No relevant log to provide

### Mealie Version

3.20.0
commit: dbe84cc91b9e1dfc88b783a1e31d8a8495800337

### Deployment

Docker (Linux)

_No response_

bferd:

**Core fix — admin recipe-edit bypass (#7810):**
- `mealie/services/recipe/recipe_service.py` — `can_update()` now returns `True` immediately for site admins, mirroring the existing (and already-tested) bypass in `can_delete()`. Previously, admins could view and delete recipes in other households, but not edit them — an inconsistency, not an intentional restriction, confirmed against the codebase's existing precedent for what "admin" means elsewhere (unscoped by group/household everywhere else in the app).
- `frontend/app/composables/recipes/use-recipe-permissions.ts` — matching admin bypass added to `canEditRecipe`, so the Edit button actually renders for admins in this scenario (the backend fix alone wouldn't have surfaced the option in the UI).

**Permissions UI consistency pass** (surfaced while working through #7810 — the admin user-management screen didn't accurately reflect how permissions actually work):
- `frontend/app/pages/admin/manage/users/[id].vue` and `create.vue` — the Administrator checkbox now greys out (and live-updates) the `canManageHousehold`/`canManage` fields it already force-overrides server-side (`_set_permissions()`), and `canOrganize`/`canInvite` are greyed out unless `canManage` or admin is set, matching the actual backend enforcement at `PUT /households/permissions`.
- `frontend/app/types/auto-forms.ts` and `frontend/app/components/global/AutoForm.vue` — added an optional `tooltip` field to the shared form-field type, rendered as a small hover info icon (matching the existing tooltip pattern used in `GroupAIProviderSettingsEditor.vue`). Kept the existing `hint` mechanism fully intact and unchanged — `common-settings-form.ts` (the first-run setup wizard) still depends on it.
- `frontend/app/composables/use-users/user-form.ts` — added accurate tooltip explanations to all six Permissions checkboxes, each verified against actual backend enforcement rather than assumed from the label text.
- `frontend/app/pages/household/members.vue` — added the same tooltip explanations to the household members table's column headers, which previously had no per-column explanation at all for "Manage Household."
- `frontend/app/components/Domain/Household/HouseholdPreferencesEditor.vue` — added a tooltip on "Lock recipe edits from other households" clarifying that admins always bypass this setting (directly reflecting the #7810 fix).
- `frontend/app/components/Domain/Recipe/RecipeSettingsSwitches.vue` — added a tooltip on the recipe "Public" switch clarifying it also depends on group/household privacy settings.
- `frontend/app/lang/messages/en-US.json` — all associated copy (only this locale file touched; other languages are managed via Crowdin per the project's translation workflow).
- `frontend/app/pages/admin/manage/users/[id].vue` and `create.vue` — follow-up fix: the original disabled-fields logic only greyed out `canManageHousehold`/`canManage` when Administrator was checked, but missed that `_set_permissions()` also force-sets `canOrganize`, `canInvite`, and `advanced` true for admins. All five fields are now
  correctly greyed out under Administrator. Also fixed a related gap where toggling a field's enabling condition back off (e.g. unchecking Manage Group, or unchecking Administrator) left dependent fields checked-but-now-editable instead of resetting them — greyed-out now always means "off and locked," never "off-looking but still true underneath."

Fixes #7810

- The admin bypass in `can_update()` is deliberately the first check, matching `can_delete()`'s existing structure exactly — no new pattern introduced.
- `AutoForm.vue` now supports both `hint` (persistent, always-visible text) and `tooltip` (hover-only) as independent, coexisting optional fields — this was a deliberate choice after confirming another existing form (`common-settings-form.ts`) still relies on `hint` for the setup wizard, where always-visible explanatory text is more appropriate than hover-only.
- The UI scope here grew from the original bug investigation once we noticed the Permissions panel didn't accurately represent backend behavior (e.g., "Administrator" implicitly forces other checkboxes true server-side, but the UI let you toggle them independently). Happy to split any of this into a separate PR if you'd prefer a narrower diff for #7810 alone.

## Testing
- Backend: `DB_ENGINE=sqlite uv run pytest` — 2,550 passed, 16 skipped (expected: LDAP tests requiring CI service, long-running scraper/parser tests), no failures.
- Frontend: `yarn vitest run` — 152 passed across 18 test files.
- New backend test: `test_admin_update_recipes_in_other_households`, parametrized over `is_private_household` and `household_lock_recipe_edits`, mirroring the existing `test_admin_delete_recipes_from_other_households`.
- New frontend test case in `use-recipe-permissions.test.ts` covering the admin cross-household scenario.
- Manually verified end-to-end across multiple groups and households: admin edit access now works as expected in the previously-blocked scenario, and all six tooltips render and read correctly.
- Manually verified the full toggle-on/toggle-off round trip for Administrator and Manage Group, confirming dependent fields both force-set correctly and reset correctly in both directions (this surfaced the gap described above).

## AI / LLM Assistance


frontend/app/composables/recipes/use-recipe-permissions.ts
```diff
@@ -13,6 +13,9 @@ export function useRecipePermissions(
     if (!user?.id) {
       return false;
     }
+    if (user.admin) {
```
michael-genson: Could we move this below the `groupId` check? Admins can view public recipes from other groups, but the backend is group-scoped, so the Edit button would show and then 404 on save.

michael-genson (changes requested):
Thanks for the fix! Could we split the permissions UI changes (tooltips, admin user form logic, members table headers) into a separate PR, as you offered? That keeps this one to the #7810 fix and its tests.

One note for that follow-up: `canOrganize` and `canInvite` don't depend on `canManage` on the backend (`_set_permissions()` and `PUT /households/permissions` set them independently), so they shouldn't be greyed out or reset when `canManage` is off.

frontend/app/composables/recipes/use-recipe-permissions.ts
bferd: Done. moved below the groupId check

bferd:
Addressed both points:

- Moved the admin bypass below the `groupId` check in `use-recipe-permissions.ts` - added a test covering the cross-group case you flagged (admin in a different group can no longer get a false-positive Edit button).
- Split the permissions UI changes out into #8429, and while doing that, confirmed `canOrganize`/`canInvite` are independent of `canManage` on the backend (`_set_permissions()` and `PUT /households/permissions` set them separately) - removed the incorrect coupling there.

This PR is now scoped to just the #7810 fix and its tests.

michael-genson:
Looks good thanks! Can you rebase? Then I'll approve

bferd:
Squashed and rebased, gtg

michael-genson (approved):
Thanks!

PR #8021, opened 2026-08-06
fix: recipe image lost after using "Get" (fetch by URL) then saving

### LINKED ISSUE #8007: [BUG] Edit → Image → Get removes existing recipe image instead of replacing it
### First Check

Summary

Using Edit → Image → Get with a valid image URL downloads and converts the image successfully, but the recipe loses its image after saving instead of replacing it.

### Steps to Reproduce

Open any recipe that already has a recipe image.
Click Edit.
Open the Image section.
Click Get.
Enter a valid public image URL.
Wait for the image preview/update to complete.
Save the recipe.

The existing recipe image should be replaced with the downloaded image.

Actual Behavior

After saving, the recipe no longer displays an image.

The original image is removed even though the replacement image was successfully downloaded and processed.

Additional Information

The backend successfully:

downloads the remote image
creates:
original.webp
min-original.webp
tiny-original.webp
serves the generated image successfully through the media endpoint

The generated original.webp is valid and can be viewed directly.

If I download that exact original.webp from Mealie and immediately upload it using Edit → Image → Upload, the image displays correctly.

This is reproducible:

on my self-hosted instance
on demo.mealie.io
in both Edge and Chrome
using multiple public image URLs

The issue appears to be specific to Edit → Image → Get.

INFO  Image URL: https://midwesternhomelife.com/wp-content/uploads/2022/08/chicken-gumbo-sloppy-joe-recipe-4.jpg

INFO  original.webp created
INFO  min-original.webp created
INFO  tiny-original.webp created

INFO  POST /api/recipes/<recipe>/image 200 OK

INFO  GET /api/media/recipes/<recipe>/images/original.webp?rnd=6&version=

INFO  PUT /api/recipes/<recipe> 200 OK

Then clicking Upload and uplading from local download:
INFO  original.webp created
INFO  min-original.webp created
INFO  tiny-original.webp created

INFO  PUT /api/recipes/<recipe>/image 200 OK

INFO  GET /api/media/recipes/<recipe>/images/original.webp?rnd=7&version=246

INFO  PUT /api/recipes/<recipe> 200 OK

### Mealie Version

Version
[v3.22.0](https://github.com/mealie-recipes/mealie/releases/tag/v3.22.0)
Build
[1eea0254a35e82bf80c255909890ba49c8cb660b](https://github.com/mealie-recipes/mealie/commit/1eea0254a35e82bf80c255909890ba49c8cb660b)

### Deployment

Docker (Linux)

_No response_

bferd:
- Fixes a bug where using **Edit → Image → Get** (fetch recipe image by URL) causes the recipe to lose its image after saving, even though the image is correctly downloaded and processed.
- **Root cause:** `RecipeImageUploadBtn.vue`'s `getImageFromURL()` emitted the same event as the delete-image button on success instead of a refresh event, wiping `recipe.image` in the frontend's local state immediately. If the user then saved, that empty value was persisted, overwriting the correct value the backend had already written.
- `mealie/routes/recipe/recipe_crud_routes.py`: `scrape_image_url` (`POST /{slug}/image`) now returns `UpdateImageResponse(image=recipe.image)`, matching the existing pattern used by the upload endpoint (`PUT /{slug}/image`).
- `RecipeImageUploadBtn.vue`: `getImageFromURL()` now reads the new image version from the response and emits a `refresh` event with that value, instead of wrongly emitting `delete`.
- `RecipePageEditorToolbar.vue`: added a `refreshImage(image)` handler that applies the new image value to `recipe.value.image` before bumping `imageKey`, replacing the previous no-op `imageKey++`.
- The existing `deleteImage()` flow (actual image deletion) was left untouched.

Fixes #8007

Root cause and reproduction details are documented in my comment on #8007. Reproduced on both v3.22.0 (prod) and `mealie-next` (dev) before diagnosing.

## Testing
- `DB_ENGINE=sqlite uv run pytest` — full suite passes
- `ruff format . --check` and `ruff check` — clean, no new issues
- Manually reproduced the bug before the fix (confirmed via API: `recipe.image` field wiped to `""` after Get + Save, despite the correct image file being written to disk)
- Manually verified the fix: `recipe.image` now updates to a valid value immediately after Get, persists correctly through Save, and the image displays as expected
- Regression-tested Upload and Delete flows on other recipes — both still work correctly

## AI / LLM Assistance

### DIFF
--- a/frontend/app/components/Domain/Recipe/RecipeImageUploadBtn.vue
+++ b/frontend/app/components/Domain/Recipe/RecipeImageUploadBtn.vue
@@ -85,11 +85,12 @@ import { useUserApi } from "~/composables/api";
 
 const UPLOAD_EVENT = "upload";
 const DELETE_EVENT = "delete";
+const REFRESH_EVENT = "refresh";
 
 const props = defineProps<{ slug: string }>();
 
 const emit = defineEmits<{
-  refresh: [];
+  refresh: [image: string];
   upload: [fileObject: File];
   delete: [];
 }>();
@@ -125,8 +126,9 @@ async function deleteImage() {
 
 async function getImageFromURL() {
   loading.value = true;
-  if (await api.recipes.updateImagebyURL(props.slug, url.value)) {
-    emit(DELETE_EVENT);
+  const { data } = await api.recipes.updateImagebyURL(props.slug, url.value);
+  if (data?.image) {
+    emit(REFRESH_EVENT, data.image);
   }
   loading.value = false;
   menu.value = false;
--- a/frontend/app/components/Domain/Recipe/RecipePage/RecipePageParts/RecipePageEditorToolbar.vue
+++ b/frontend/app/components/Domain/Recipe/RecipePage/RecipePageParts/RecipePageEditorToolbar.vue
@@ -4,7 +4,7 @@
       class="my-2"
       :slug="recipe.slug"
       @upload="uploadImage"
-      @refresh="imageKey++"
+      @refresh="refreshImage"
       @delete="deleteImage"
     />
     <RecipeSettingsMenu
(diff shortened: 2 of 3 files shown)


michael-genson (approved):
Thank you!

PR #8028, opened 2026-08-07
fix: prevent recipe tags from hiding card actions

### LINKED ISSUE #8027: [BUG] - Long recipe tags can push the action menu outside recipe cards
### First Check

On the desktop Recipes page, recipe tags are displayed in the same row as the favourite button, rating and action menu.

When a recipe has long tags, the tags can overflow the available space and push the action menu outside the recipe card. The menu then becomes unavailable. Cards with two tags can also have a different height from cards without tags.

I expect the action menu to remain visible and the recipe cards to keep a consistent layout regardless of the number or length of tags.

A possible solution is to follow the layout used by `RecipeCardMobile.vue`: display up to two tags on a separate row above the action buttons, truncate long tag names with an ellipsis, and keep the footer height consistent. When there are no tags, the action buttons can be vertically centred in the footer.

This would be a frontend-only change to `RecipeCard.vue`. I have prepared a draft pull request and can adjust the solution based on maintainer feedback.

### Current behaviour

Image

### RecipeCardMobile reference

Image

### Proposed result

Image

### Steps to Reproduce

1. Open the Recipes page using the desktop card layout.
2. Create or edit a recipe.
3. Add two tags to the recipe. Use long tag names to make the problem easier to see.
4. Return to the Recipes page.
5. Observe that the tags can push the action menu outside the card and affect the card layout.

No relevant logs are produced. This is a frontend layout issue and no console or API errors were observed.

### Mealie Version

Development build from the `mealie-next` branch.
Commit: af15ba1f45940ce6e9e5c3406cb833730804918f
Docker image: Local VS Code Dev Container

### Deployment

Docker (Linux)

VS Code Dev Container running in Docker Desktop on macOS.
Browser: Google Chrome.

feilongcheng:

Long recipe tags can push the action menu outside desktop recipe cards. Cards with tags can also have an inconsistent layout compared with cards without tags.

This PR:

- Uses the existing tag layout in `RecipeCardMobile.vue` as a reference.
- Moves recipe tags to a separate row above the card actions.
- Continues to display a maximum of two tags.
- Truncates long tag names with an ellipsis.
- Keeps the recipe card footer height consistent.
- Vertically centres the action buttons when no tags are present.
- Removes the vertical padding from the action row.

### Before

Recipe card layout before the fix

### After

Recipe card layout after the fix

- Fixes #8027
- Also addresses #8052

## Testing

- `task ui:check`
  - Frontend lint passed
  - 18 test files passed
  - 151 tests passed
- Manually tested with:
  - No tags
  - One tag
  - Two tags
  - More than two tags
  - Long English and Chinese tag names
  - Two-column and three-column desktop layouts
  - Favorite, rating and action-menu controls

## AI / LLM Assistance

I used ChatGPT to discuss the UI approach and help draft parts of the implementation. I reviewed and understood the final changes, tested the affected layouts manually, and made the final implementation decisions.

### DIFF
--- a/frontend/app/components/Domain/Recipe/RecipeCard.vue
+++ b/frontend/app/components/Domain/Recipe/RecipeCard.vue
@@ -176,7 +176,7 @@ const cursor = computed(() => showRecipeContent.value ? "pointer" : "auto");
   display: flex;
   flex-direction: column;
   justify-content: flex-end;
-  height: 84px;
+  min-height: 84px;
 }
 .recipe-card-footer--no-tags {
   justify-content: center;
@@ -190,12 +190,7 @@ const cursor = computed(() => showRecipeContent.value ? "pointer" : "auto");
 }
 .recipe-card-tags .v-chip {
   flex: 0 1 auto;
-  min-width: 0;
-  max-width: calc(50% - 2px);
   margin: 0 !important;
-}
-.recipe-card-tags .v-chip:only-child {
-  max-width: 100%;
 }
 .recipe-card-tags .v-chip__content {
   display: block;
@@ -205,6 +200,5 @@ const cursor = computed(() => showRecipeContent.value ? "pointer" : "auto");
 }
 .recipe-card-actions {
   width: 100%;
-  min-height: 40px;
 }
 </style>
--- a/frontend/app/components/Domain/Recipe/RecipeCard.vue
+++ b/frontend/app/components/Domain/Recipe/RecipeCard.vue
@@ -40,55 +40,62 @@
           {{ name }}
         </v-card-title>
 
-        <slot name="actions">
-          <v-card-actions
-            v-if="showRecipeContent"
-            class="px-1"
-          >
-            <RecipeFavoriteBadge
-              v-if="isOwnGroup"
-              :recipe-id="recipeId"
-              show-always
-            />
-            <div v-else class="px-1" /> <!-- Empty div to keep the layout consistent -->
+        <div
+          class="recipe-card-footer"
+          :class="{ 'recipe-card-footer--no-tags': tags.length === 0 }"
+        >
+          <RecipeChips
+            v-if="tags.length > 0"
+            class="recipe-card-tags px-4"
+            :truncate="false"
+            :items="tags"
+            :title="false"
+            :limit="2"
+            small
+            url-prefix="tags"
+            v-bind="$attrs"
+          />
 
-            <RecipeCardRating
-              :model-value="rating"
-              :recipe-id="recipeId"
-            />
-            <v-spacer />
-            <RecipeChips
-              :truncate="true"
-              :items="tags"
-              :title="false"
-              :limit="2"
-              small
-              url-prefix="tags"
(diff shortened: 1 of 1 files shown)


frontend/app/components/Domain/Recipe/RecipeCard.vue
```diff
@@ -165,4 +172,39 @@ const cursor = computed(() => showRecipeContent.value ? "pointer" : "auto");
   line-clamp: 8;
   overflow: hidden;
 }
+.recipe-card-footer {
+  display: flex;
+  flex-direction: column;
+  justify-content: flex-end;
+  height: 84px;
+}
```
michael-genson, line 180: Can we switch this to `min-height` instead of `height`?

frontend/app/components/Domain/Recipe/RecipeCard.vue
```diff
@@ -165,4 +172,39 @@ const cursor = computed(() => showRecipeContent.value ? "pointer" : "auto");
...
+.recipe-card-tags .v-chip {
+  flex: 0 1 auto;
+  min-width: 0;
+  max-width: calc(50% - 2px);
+  margin: 0 !important;
+}
+.recipe-card-tags .v-chip:only-child {
+  max-width: 100%;
+}
```
michael-genson: We can drop `max-width`, Vuetify should handle this automatically, `.v-chip` already sets `min-width: 0; max-width: 100%`. That makes the `:only-child` rule unnecessary too.

frontend/app/components/Domain/Recipe/RecipeCard.vue
```diff
@@ -165,4 +172,39 @@ const cursor = computed(() => showRecipeContent.value ? "pointer" : "auto");
...
+.recipe-card-tags .v-chip__content {
+  display: block;
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap;
+}
+.recipe-card-actions {
+  width: 100%;
+  min-height: 40px;
```
michael-genson: Do we need this `min-height`? It ties Vuetify's `.v-card-actions { min-height: 52px }` on specificity, so it only wins by stylesheet order, and either value fits in the footer anyway.

michael-genson (changes requested):
Generally looks good, a few comments.

frontend/app/components/Domain/Recipe/RecipeCard.vue
feilongcheng: I updated it to use min-height.

frontend/app/components/Domain/Recipe/RecipeCard.vue
feilongcheng: I removed the redundant width rules and the :only-child rule.

frontend/app/components/Domain/Recipe/RecipeCard.vue
feilongcheng: I removed the custom min-height.

michael-genson (approved):
Thank you!

PR #8033, opened 2026-08-08
dev: fix all the warning in pytest

### LINKED ISSUE #7965: [Task] - Fix all the warnings in pytest

fix: eliminate pytest warnings and fix pre-existing test failures across migrations, backups, and timezone handling

While setting up the dev environment and running the test suite, I noticed the full `pytest` run was producing a large number of warnings, and turned up several pre-existing test failures along the way. This PR cleans all of that up so the suite runs clean (0 failures, 0 warnings), split into 10 focused commits — one per problem:

- **`mealie/services/migrations/utils/database_helpers.py`** — `_get_or_set_generic()` was calling `.model_dump()` on category/tag models before returning them, so `Recipe.recipe_category`/`Recipe.tags` ended up holding plain dicts instead of model instances for every recipe imported via a migrator. Fixed by removing the `.model_dump()` call.
- **`mealie/routes/handlers.py`** — replaced the deprecated `status.HTTP_422_UNPROCESSABLE_ENTITY` with `HTTP_422_UNPROCESSABLE_CONTENT`.
- **`mealie/services/backups_v2/alchemy_exporter.py`** (2 commits):
  - `dump()` sorted tables topologically via `self.meta.sorted_tables`, which can't resolve the intentional circular FK between `ai_providers`/`ai_provider_settings` and warns on every backup. Verified table order is actually inconsequential here — `restore()` disables FK enforcement for the whole operation via `ForeignKeyDisabler` — so switched to plain `self.meta.tables.values()`.
  - `drop_all()` used the deprecated `Inspector.from_engine()`; replaced with `sqlalchemy.inspect()`.
- **`tests/unit_tests/test_config.py`** — `monkeypatch.setenv()` was being passed raw `bool` values in the LDAP/OIDC settings tests; wrapped in `str()`.
- **`tests/fixtures/fixture_shopping_lists.py`** — teardown deleted a `ShoppingList` whose ORM-cached `list_items` collection was stale relative to the DB (a test had already deleted one item directly), causing a row-count-mismatch warning. Added `session.expire_all()` before the teardown delete.
- **`pyproject.toml` / `uv.lock`** (3 commits):
  - Added `httpx2` as a dev dependency — starlette's `TestClient` prefers it and warns when falling back to `httpx`, which is what nearly every integration test was doing.
  - Added a narrowly-scoped `filterwarnings` entry suppressing pyRdfa's deprecated `datetime.utcnow()` call (unmaintained third-party dependency, already flagged elsewhere in `mealie/app.py`).
  - Added a second `filterwarnings` entry suppressing sqlite's deprecated datetime-adapter warning, caused by `GroupDataExportsModel.expires` being declared a `String` column while always storing a real `datetime`. **This is a real, pre-existing bug I did not fix** — see Special notes below.
- **`tests/unit_tests/services_tests/scheduler/tasks/test_create_timeline_events.py`** and **`tests/integration_tests/user_household_tests/test_group_mealplan.py`** — 4 pre-existing test failures, all one root cause: these tests build "today's" meal-plan date with `datetime.now(UTC).date()`, but the app's "today" logic (both the timeline-events scheduler task and the `/households/mealplans/today` endpoint) intentionally resolves "today" using the household's *local* timezone. On a machine where local date and UTC date currently differ (reproduced on `Pacific/Auckland`, UTC+12), plans land on the wrong calendar day as far as the app is concerned. Fixed by building/comparing dates with `tzlocal()` instead, matching what the app actually does.

Each commit message has the full reason / consequence / fix for its specific change.
This is the link of the test log before our draft PR and the test log after our draft PR, just for your information:
https://docs.google.com/document/d/11POS6tgElf94wpqiMG1RWixGImYeQAgjyShYeJZnW3M/edit?usp=sharing

bestxmg:
fix: eliminate all pytest warnings across the test suite

Running the full test suite (`task py:check`) at the base of this branch produces 156 pytest warnings and 4 test failures. This PR resolves all of them, in 10 focused commits — one per root cause. Verified by running `task py:check` at the baseline and after every individual commit: zero new warnings or failures were introduced anywhere in the chain, and the count only ever goes down.

- **`mealie/services/migrations/utils/database_helpers.py`** — `_get_or_set_generic()` called `.model_dump()` on each `CategoryOut`/`TagOut` before returning it, so `Recipe.recipe_category`/`Recipe.tags` held plain dicts instead of model instances for every recipe imported via a migrator. Removed the `.model_dump()` call. Fixes `UserWarning: Pydantic serializer warnings` (37 occurrences, across all recipe migration tests).
- **`mealie/routes/handlers.py`** — replaced the deprecated `status.HTTP_422_UNPROCESSABLE_ENTITY` with `HTTP_422_UNPROCESSABLE_CONTENT`. Fixes `StarletteDeprecationWarning` (25 occurrences, one per 422 response across many tests).
- **`tests/integration_tests/user_household_tests/test_group_mealplan.py`**, **`tests/unit_tests/services_tests/scheduler/tasks/test_create_timeline_events.py`** — these tests built meal-plan dates with `datetime.now(UTC).date()`, but the app's "today" logic (both the timeline-events scheduler task and the `/households/mealplans/today` endpoint) intentionally resolves "today" using the household's *local* timezone. On a machine where local date and UTC date differ, plans land on the wrong calendar day as far as the app is concerned. Switched date construction to `tzlocal()`. Fixes 4 pre-existing test failures (unrelated to warnings, discovered incidentally).
- **`pyproject.toml`** (2 commits) — added narrowly-scoped `filterwarnings` entries for two warnings we can't fix directly:
  - pyRdfa3 (unmaintained transitive dependency) calls deprecated `datetime.datetime.utcnow()` — 65 occurrences in `test_cleaner.py` alone.
  - `GroupDataExportsModel.expires` is declared a `String` column but always holds a real `datetime`, so the value reaches sqlite3's deprecated implicit adapter — 2 occurrences. **This is a real, pre-existing type-declaration bug, not fixed here** — see Special notes below.
- **`mealie/services/backups_v2/alchemy_exporter.py`** (2 commits):
  - `dump()` sorted tables via `self.meta.sorted_tables`, which can't resolve the intentional circular FK between `ai_providers`/`ai_provider_settings` and warns on every backup (4 occurrences). Verified table order is inconsequential here — `restore()` disables FK enforcement for the whole operation — so switched to `self.meta.tables.values()`.
  - `drop_all()` used the deprecated `Inspector.from_engine()`; replaced with `sqlalchemy.inspect()` (11 occurrences).
- **`tests/fixtures/fixture_shopping_lists.py`** — `list_with_items`'s teardown cascades a delete across `list_items` using a collection cached when the fixture seeded it; a test that deletes one item directly leaves that cached count stale by teardown time. Fetches the real `ShoppingList` ORM instance via `session.get()` and expires just that object rather than the whole session.
- **`tests/unit_tests/test_config.py`** — LDAP/OIDC settings tests passed raw `bool` values to `monkeypatch.setenv()`, which requires strings. Wrapped in `str()`. Fixes 4 `PytestWarning` variants.
- **`pyproject.toml` / `uv.lock`** — added `httpx2` as a dev dependency; Starlette's `TestClient` prefers it and warns when falling back to `httpx`, which nearly every integration test was doing (10 occurrences, the last warning remaining before this commit).

Fixes #7965

## Testing

Ran `task py:check` (format, lint, type-check, full test suite) at the base commit and again after each of the 10 commits individually — 11 full runs in total. Result: 0 new warnings or failures introduced at any point in the chain; every commit either eliminated exactly the warning category it targeted, or (in one case) showed no observable change in that specific run because the targeted warning is order/data-dependent within the full suite and didn't happen to fire in either of that pair's runs — verified separately via an isolated, reliable reproduction of that specific test file before/after. Final state: 0 failed, 0 warnings, 2546 passed, 16 skipped.

## AI / LLM Assistance

Claude code

### DIFF
--- a/mealie/services/backups_v2/alchemy_exporter.py
+++ b/mealie/services/backups_v2/alchemy_exporter.py
@@ -181,7 +181,7 @@ class AlchemyExporter(BaseService):
 
             result = {
                 table.name: [dict(row) for row in connection.execute(table.select()).mappings()]
-                for table in self.meta.sorted_tables
+                for table in self.meta.tables.values()
             }
 
         return jsonable_encoder(result)
--- a/mealie/services/backups_v2/alchemy_exporter.py
+++ b/mealie/services/backups_v2/alchemy_exporter.py
@@ -248,11 +248,11 @@ class AlchemyExporter(BaseService):
 
     def drop_all(self) -> None:
         """Drops all data from the database"""
-        from sqlalchemy.engine.reflection import Inspector
+        from sqlalchemy import inspect
         from sqlalchemy.schema import DropConstraint, DropTable, MetaData, Table
 
         with self.engine.begin() as connection:
-            inspector = Inspector.from_engine(self.engine)
+            inspector = inspect(self.engine)
 
             # We need to re-create a minimal metadata with only the required things to
             # successfully emit drop constraints and tables commands for postgres (based
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -103,7 +103,9 @@ python_files = 'test_*'
 python_functions = 'test_*'
 testpaths = ["tests"]
 filterwarnings = [
+    # Suppress unmaintained third-party pyRdfa utcnow() deprecation warning
     "ignore::DeprecationWarning:pyRdfa.*",
+    # Suppress Python 3.12 default SQLite datetime-adapter deprecation warning
     "ignore:The default datetime adapter is deprecated as of Python 3.12:DeprecationWarning",
 ]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -78,6 +78,7 @@ dev = [
     "types-urllib3==1.26.25.14",
     "pydantic-to-typescript2==1.0.6",
     "freezegun==1.5.5",
+    "httpx2==2.7.0",
 ]
 
 [build-system]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -103,6 +103,7 @@ python_functions = 'test_*'
 testpaths = ["tests"]
 filterwarnings = [
     "ignore::DeprecationWarning:pyRdfa.*",
+    "ignore:The default datetime adapter is deprecated as of Python 3.12:DeprecationWarning",
 ]
 
 [tool.coverage.report]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -101,6 +101,9 @@ python_classes = '*Tests'
 python_files = 'test_*'
 python_functions = 'test_*'
 testpaths = ["tests"]
+filterwarnings = [
+    "ignore::DeprecationWarning:pyRdfa.*",
+]
 
 [tool.coverage.report]
 skip_empty = true
--- a/mealie/routes/handlers.py
+++ b/mealie/routes/handlers.py
@@ -25,7 +25,7 @@ def register_debug_handler(app: FastAPI):
     async def validation_exception_handler(request: Request, exc: RequestValidationError):
         exc_str = f"{exc}".replace("\n", " ").replace("   ", " ")
         log_wrapper(request, exc)
-        content = {"status_code": status.HTTP_422_UNPROCESSABLE_ENTITY, "message": exc_str, "data": None}
-        return JSONResponse(content=content, status_code=status.HTTP_422_UNPROCESSABLE_ENTITY)
+        content = {"status_code": status.HTTP_422_UNPROCESSABLE_CONTENT, "message": exc_str, "data": None}
+        return JSONResponse(content=content, status_code=status.HTTP_422_UNPROCESSABLE_CONTENT)
 
     return validation_exception_handler
--- a/mealie/services/migrations/utils/database_helpers.py
+++ b/mealie/services/migrations/utils/database_helpers.py
@@ -45,7 +45,7 @@ class DatabaseMigrationHelpers:
                     )
                 )
 
-            items_out.append(item_model.model_dump())
+            items_out.append(item_model)
         return items_out
 
     def get_or_set_category(self, categories: Iterable[str]) -> list[RecipeCategory]:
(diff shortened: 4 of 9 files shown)


bestxmg:
@michael-genson 
Our PR for **fixing all the warnings** in pytest. is ready. Please take a look. I have rebased our PR to the latest code (8th August) and run  **task py:check** on my machine. This is the summary:
image

bestxmg:
It’s worth noting that the four test failures in commit `2914d001` (local time not UTC) were already present in the baseline and were not introduced by us. However, we have also fixed these failures.

mealie/services/backups_v2/alchemy_exporter.py
```diff
@@ -181,7 +181,7 @@ def dump(self) -> dict[str, list[dict]]:
 
             result = {
                 table.name: [dict(row) for row in connection.execute(table.select()).mappings()]
-                for table in self.meta.sorted_tables
+                for table in self.meta.tables.values()
             }
```
michael-genson, line 185: My understanding is that `tables.values()` does not guarantee foreign key ordering, which can cause issues during restore. Is that not the case here?

pyproject.toml
```diff
@@ -101,6 +102,10 @@ python_classes = '*Tests'
 python_files = 'test_*'
 python_functions = 'test_*'
 testpaths = ["tests"]
+filterwarnings = [
+    "ignore::DeprecationWarning:pyRdfa.*",
+    "ignore:The default datetime adapter is deprecated as of Python 3.12:DeprecationWarning",
+]
```
michael-genson, line 110: Can we add a comment here why this is needed? I know you explained this in the PR but I also want it here

michael-genson (changes requested):
Thanks for this! Just one note

mealie/services/backups_v2/alchemy_exporter.py
michael-genson: Oops nevermind you explained it in your PR body, thanks!

pyproject.toml
bestxmg: Yep, our group has added this. Please take a look.

michael-genson (approved):
Great, thanks so much!

PR #7825, opened 2026-06-29
feat: add plain-text recipe import via OpenAI

bferd:
## What this PR does / why we need it:

Adds a plain-text recipe import feature using OpenAI. Users can paste any unstructured recipe text — typed, transcribed from a cookbook, or copied from a source that doesn't support URL scraping — and have it automatically parsed into a structured Mealie recipe.

Changes:
- `mealie/services/openai/prompts/recipes/parse-recipe-text.txt` — new prompt instructing the model to extract recipe data from plain text faithfully without inventing content
- `mealie/services/recipe/recipe_service.py` — added `build_recipe_from_text()` to `OpenAIRecipeService` and `create_from_text()` to `RecipeService`; uses `ai_enabled` (not `image_provider_enabled`) since no vision capability is needed
- `mealie/routes/recipe/recipe_crud_routes.py` — new `POST /api/recipes/create/text` endpoint accepting `{"text": "..."}` with optional `translateLanguage` query param
- `mealie/schema/recipe/recipe_scraper.py` — added `CreateRecipeFromText` schema
- `frontend/app/pages/g/[groupSlug]/r/create/plaintext.vue` — new page with textarea, disabled-when-empty submit button, and optional translation selector; hidden when AI is not enabled
- `frontend/app/lib/api/user/recipes/recipe.ts` — added `createOneFromText()` API method
- `frontend/app/pages/g/[groupSlug]/r/create.vue` — added "Create from Text" entry to the create subpages list
- `frontend/app/lang/messages/en-US.json` — added i18n keys
- `tests/integration_tests/user_recipe_tests/test_recipe_create_from_text.py` — integration test with mocked OpenAI response

## Which issue(s) this PR fixes:

Closes #4561 (supersedes stalled PR with clean implementation and tests)

## Special notes for your reviewer:

This is modeled directly on the existing image import feature (PR #3974). The key difference is that text import only requires `ai_enabled`, not `image_provider_enabled`, since no vision model is needed.

## Testing

Integration test passes with SQLite:
`DB_ENGINE=sqlite uv run pytest tests/integration_tests/user_recipe_tests/test_recipe_create_from_text.py -v`

Manually tested end-to-end with unstructured recipe text via the UI with a real OpenAI-compatible provider.

## AI / LLM Assistance

Claude Code was used to generate the initial implementation following the existing image import pattern. All generated code was reviewed, tested manually, and verified against the codebase conventions before submission.

PR #7618, opened 2026-05-13
feat: add Force OpenAI Scraper option to URL and bulk import

zdenek-stursa:
## What this PR does / why we need it:

Adds a **"Force OpenAI Scraper"** checkbox to the URL import page (`/g/{household}/r/create/url`) and the bulk import page (`/g/{household}/r/create/bulk`).

When checked, the request bypasses the default HTML scraper (`RecipeScraperPackage`) and routes directly to the OpenAI LLM scraper (`RecipeScraperOpenAI`). This is useful when the standard scraper fails on sites that use JavaScript rendering or non-standard markup, but the page content is still parseable by an LLM.

The checkbox is only shown when OpenAI is configured on the server (`$appInfo.enableOpenai`).

**File-by-file summary:**

- `mealie/schema/recipe/recipe_scraper.py` — added `use_openai: bool = Field(False, alias="useOpenAI")` to `ScrapeRecipe`
- `mealie/schema/recipe/recipe.py` — added same field to `CreateRecipeByUrlBulk` with `populate_by_name=True`
- `mealie/services/scraper/scraper.py` — added `use_openai` parameter to `create_from_html()`; when `True`, initialises `RecipeScraper` with `[RecipeScraperOpenAITranscription, RecipeScraperOpenAI]` instead of the default strategy list
- `mealie/services/scraper/recipe_bulk_scraper.py` — passes `use_openai` to each `create_from_html()` call
- `mealie/routes/recipe/recipe_crud_routes.py` — extracts `use_openai` from the request (via `isinstance` check) and passes it to `create_from_html()`
- `mealie/services/openai/prompts/recipes/scrape-recipe.txt` — prompt now explicitly requires `@context` and `@type` fields; fixes a systematic issue where some models omit them, causing `recipe_scrapers` to raise `NoSchemaFoundInWildMode`
- `mealie/services/scraper/scraper_strategies.py` — added `setdefault` fallback for `@context`/`@type` in `RecipeScraperOpenAI.get_html()` as a safety net
- `frontend/app/lang/messages/en-US.json` — added `force-openai-scraper` and `force-openai-scraper-description` i18n keys
- `frontend/app/composables/use-users/preferences.ts` — added `forceOpenAI: boolean` to `UserRecipeCreatePreferences` (persisted in localStorage)
- `frontend/app/composables/use-new-recipe-options.ts` — added `enableForceOpenAI` prop and `forceOpenAI` computed
[...]

PR #8043, opened 2026-08-10
feat: Unified AI recipe page

michael-genson:

_(REQUIRED)_

Today we have tons of different ways of making recipes, some using AI and some not. This PR unifies all AI methods under one "Import with AI" page. From here you can:
- Pass a URL, scrape the content and send it to AI
- Pass a video URL, download it, transcribe it, etc.
- Upload images to be transcribed by AI
- Pass arbitrary text (HTML, JSON, or literally just typing freeform) to AI

Users can also opt to create new organizers based on the recipe content, as well as translate the recipe. Previously these options were selectively available, but now they're generally available for all AI methods.

Screenshot 2026-08-09 at 8 48 51 PM

On the backend the AI import process is rearchitected into a workflow and steps:
1. First, ingest the content and compile it into a master markdown file (internal only)
2. Read the markdown file and produce a recipe
3. Add organizers (tags, categories, tools)

Each step is its own prompt, so we don't overload the AI with one giant request. This new architecture also allows us to extend the workflow in the future, and even potentially allow users to inject their own pre/post processing.

A future win will be to build in a parsing step, but I wanted to keep this PR as limited as I can.

---

The image-specific page is deprecated, as well as a ton of the old prompts (since architecture has changed to be multi-step). Some endpoints are also deprecated as a result.

We still have the old non-AI import methods (URL, HTML/JSON) and those still fall back to AI as they used to. I didn't think it made sense to deprecate the AI fallback here since it effectively makes them worse. However, unlike those pages, the new AI page _exclusively_ uses AI and skips our recipe scraper pipeline.

_(REQUIRED)_

Replaces https://github.com/mealie-recipes/mealie/pull/7618
Replaces https://github.com/mealie-recipes/mealie/pull/7825

Been wanting to work on this one for a while, finally had a few days.

## Testing

Pytest for backend, threw a lot of recipes at GPT for the frontend.

## AI / LLM Assistance

_(REQUIRED)_

Written by Claude. Planned, architected, and tested by a human (me!).


bferd:
Hey @michael-genson, the unified AI import architecture is a really nice consolidation, nice work pulling all those paths together into one workflow.

I pulled mealie-next and put the new "Import with AI" page through a full test matrix on my dev instance (text, URL, image, video, JSON, translate, organizer creation, and combined-source imports). Found a couple of issues I'll file as separate issues so they're each easy to track and reproduce:

- URL + text notes combined silently drops the URL content: when both a URL and text/notes are provided, only the text survives into the compiled source; the URL is never fetched. Traced to StructuredDataCompiler.can_compile() not checking for a URL before claiming the input. For comparison, image + text notes works correctly: I verified the transcribed image content and my note's corrections (an ingredient substitution and a serving-size change) both landed in the merged output, so this looks isolated to the URL path specifically.
- Not sure which side is the bug here - AI-generated recipes (from a directive like "create a recipe for X") populate ingredients/instructions when "Translate the recipe into my language" is checked, which looks like the translate instruction is unintentionally overriding the base prompt's transcribe-only constraint ("do not create, infer, or make up any information"). Before I write this up, it might be worth deciding whether fully-imagined recipes (no real source, just a request) are in scope for this page at all. If not, the fix is probably making that constraint hold consistently regardless of translate state; if so, it's more of a "make generation actually reliable" feature. I personally have used AI to create recipes based on my ideas so if this could be a user choice that is where my vote would lie. Note: if the translate button is not checked the AI imagined recipe is not created with that note entered.
- Couldn't fully test video URL end to end since my provider doesn't offer an audio transcription API, but along the way I noticed a couple of related toast issues I can help with: no clear message when ffmpeg/ffprobe aren't installed, and some backend errors (like OpenAIServiceError) surface to the user as raw exception class names instead of a friendly message. Happy to take a pass at improving those.

Nothing urgent, just wanted to give you a heads-up before the individual issues land and want to discuss thoughts on the 2nd bullet direction further.  **I will await your response on that one.**

I also like the parsing step you mentioned as a future direction. It's something I currently enforce as a rule in a Claude project I use for recipe import (with API access), always finishing with the parse step. Another idea from that same setup: for text-based recipe inputs with no source image, it scrapes a copyright-free image as a cover photo (after my approval), which could be a nice way to close the gap I hit with plain-text imports (recipe created fine, but no image). Again though, not sure if that is part of scope.

bferd:
Quick follow-up on the first bullet above (URL + text notes) — I want to walk back the "bug" framing a bit.

I noticed the helper text under the Recipe URL field actually says *"other pages are read only if you don't paste content below"*, which describes exactly this behavior as intentional rather than a bug in the traditional sense.

That said, I still think it's worth a second look, because it's inconsistent with the **image + notes** case, which does merge both sources rather than one replacing the other. The **notes** field's own label/placeholder ("Recipe text, HTML, JSON, or notes") and helper copy ("this is used as notes instead, so you can say what to name the recipe or correct a detail") both read as "notes annotate/correct the primary source," not "notes fully replace it." So a note like "substitute pepper jack for mozzarella" alongside a URL reads to a user like a correction to apply on top of the fetched recipe — not a signal to skip the URL entirely.

So the real question is less "is this a bug" and more: is skipping the URL fetch when notes are present the intended design, or should URL + notes merge the way image + notes does? I didn't want to lock in a "bug" framing on something that might be working as designed. Thoughts?

michael-genson:
@bferd thanks for the feedback! I've addressed this in https://github.com/mealie-recipes/mealie/pull/8067.

PR #8075, opened 2026-08-12
fix: skip image request when a recipe has no image

### LINKED ISSUE #4804: [BUG] - Do not log 404 error on recipe with no image
### First Check

Recipes that genuinely have no cover image are resulting in a 404 error in the logs. This results in a false positive in some of my security tools, banning the users ip address from communication with my server.

Expected:
- If the recipe contains no image, there should be no request to retrieve the "min-original.webp" image. 
OR
- the API should recognize that the recipe has no image, and return an http 200 code with an empty body.

### Steps to Reproduce

1. Open the chrome dev tools so you can see the full network log when the error is thrown.
2. Create a recipe manually, and do not attach an image.
3. Go to the recipes page. You will see all of your recipes including those without images
4. a call to the api is made "/api/media/recipes/4d647a74-bad9-476e-8a3d-ad132a13406e/images/min-original.webp?rnd=1&version=" and a 404 is returned and logged.

Log in Chrome browser from demo site
`GET https://demo.mealie.io/api/media/recipes/631d6769-9e8b-4199-83c0-ddf7eaafa086/images/min-original.webp?rnd=1&version=jL92 404 (Not Found)`

NGINX logs in proxy-host-*access.log
`[31/Dec/2024:03:47:58 +0000] - 404 404 - GET https xxxx.xxxx.com "/api/media/recipes/4d647a74-bad9-476e-8a3d-ad132a13406e/images/min-original.webp?rnd=1&version=" [Client xxx.xx.xxx.xxx] [Length 22] [Gzip -] [Sent-to xxx.xx.x.x] "Mozilla/5.0 (Linux; Android 10; K) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/131.0.0.0 Mobile Safari/537.36" "https://xxxx.xxxxx.com/g/xxxxx"
`

### Mealie Version

Version
v2.4.1

Build
[2922782e5ea44e1c5d2690c9aea84134a044a7cc](https://github.com/mealie-recipes/mealie/commit/2922782e5ea44e1c5d2690c9aea84134a044a7cc)

### Deployment

Docker (Linux)

_No response_

henry1113nz:

Recipes without an image still request the image file. The server has nothing to
serve, so it returns 404. The UI hides this well — `RecipeCardImage` uses the
failed request itself as its signal and swaps in the placeholder icon — but the
request is made every time, and self-hosted instances end up with a log full of
404s. #4804 reports a security tool reading those repeated 404s as an attack and
banning the reporter's own users.

The information needed to skip the request is already available at every render
site: a recipe's `image` field holds the cache key when an image exists and is
empty when it does not. `RecipeList.vue` already guards on it (`v-if="recipe.image"`).
This PR applies the same guard to the places that were missing it.

Per file:

- **`RecipeCardImage.vue`** — render the image only when `imageVersion` is set,
  instead of rendering it and waiting for the error. `imageVersion` is also added
  to the existing `watch` so an upload to a previously imageless recipe shows up
  without a reload.
- **`RecipeCard.vue`**, **`RecipeCardMobile.vue`** — the `image` prop defaulted to
  `"abc123"`, a truthy placeholder that makes "no image" indistinguishable from
  "has an image". Defaults to `undefined` now.
- **`planner/view.vue`** — never passed `image`, so it inherited the default above
  and requested an image for every entry, including note-only entries with no
  recipe at all.
- **`planner/edit.vue`** — never passed `image-version`. Without this the guard in
  `RecipeCardImage` would hide images that do exist.
- **`RecipePageInfoCard.vue`** — the landscape branch was missing the `recipe.image`
  check the portrait branch already had.
- **`RecipePrintView.vue`** — an imageless recipe rendered a broken image in the
  print view. The column is skipped when there is no image.

Backend is untouched. Returning 404 for a file that does not exist is correct
behaviour; the bug is asking for it in the first place.

Fixes #4804

Two things worth a look:

- `planner/edit.vue` did not pass `image-version` before this PR. It has to be
  added together with the guard, otherwise meal plan entries that do have images
  would stop showing them.
- The `"abc123"` prop defaults looked deliberate but appear to be leftovers — they
  are the reason `planner/view.vue` was requesting images for note-only entries.
  Happy to split that into its own commit if you would rather keep it separate.

## Testing

Local dev instance, Chromium, counting requests to `/api/media/recipes/*` on a
fresh load of each page. Same recipes and same method before and after.

| Page | Before | After |
| --- | --- | --- |
| Recipe list (6 recipes, 1 with an image) | 1× 200, 5× 404 | 1× 200 |
| Recipe page, no image | 1× 404 | none |
| Recipe page, with image | 1× 200 | 1× 200 |
| Meal plan, view | 1× 200, 1× 404 | 1× 200 |
| Meal plan, edit | 1× 200, 1× 404 | 1× 200 |

All five pages render the same before and after. Also checked by hand:

- uploading an image to a recipe that had none shows the image without a reload
- print view of an imageless recipe no longer shows a broken image
- `task ui:check` passes

image

image

## AI / LLM Assistance

An LLM was used to help trace which components render recipe images and to draft
the change. The investigation posted on #4804 is my own, and I reviewed the diff
and verified the behaviour by hand in a local instance before opening this PR.


michael-genson:
Thanks!

PR #8078, opened 2026-08-13
feat: support customizing app icon, favicon, banner title, and HTML title

jspeyside:

Adds operator-configurable branding for self-hosted instances: the app logo and the top-bar/tab title can now be overridden, falling back to the current Mealie defaults when unset. This follows the same pattern as the existing `THEME_*` env vars.

New env vars (all optional):
- `BRANDING_NAME` – top app-bar title, browser tab title, PWA manifest name (default `Mealie`)
- `BRANDING_LOGO_PATH` – path to a custom logo file (svg/png/ico/webp/jpg/jpeg), used in the header, on the login/register/setup pages, and as the favicon

An invalid or missing logo path is silently treated as unset and falls back to the built-in assets — no crash, no broken image.

**Backend**
- `mealie/core/settings/branding.py` (new): `Branding` settings class, mirrors `themes.py`'s `env_prefix` pattern. `name: str` and `logo_path: str | None`; the `logo_file` property resolves the configured path only if it exists, is a file, and has an allowed image suffix.
- `mealie/core/settings/settings.py`: wires `branding: Branding` into `AppSettings`.
- `mealie/schema/admin/about.py`: `branding_name`/`branding_logo_url` are now fields directly on the existing `AppInfo` schema — no separate `AppBranding` model.
- `mealie/routes/app/app_about.py`: `GET /about` now populates the branding fields; a single new `GET /about/branding-logo` route serves the logo file (mirrors the existing `/about/theme` route, same `Cache-Control` policy).
- `mealie/routes/spa/manifest.py`: PWA manifest `name`/`short_name` use `settings.branding.name`.
- `mealie/routes/spa/__init__.py`: **unchanged** — the earlier `index.html` rewrite / `SPAStaticFiles` override has been removed entirely; branding is applied client-side only (see below).

**Frontend**
- `frontend/app/plugins/app-info.client.ts`: already fetched `/api/app/about` and provided `$appInfo`; now also calls `useHead()` once that data loads, to set the branded `<title>` and favicon `<link>` tags.
  - This is necessary because the frontend runs with `ssr: false`: the client re-applies `nuxt.config.ts`'s static defaults on every boot, which would otherwise clobber the branded title/favicon moments after page load.
- `AppHeader.vue` / `AppLogo.vue`: render the custom logo image when configured (`$appInfo.brandingLogoUrl`), else fall back to the existing MDI glyph / inline SVG. Header banner title now reads `$appInfo.brandingName`. Other in-app uses of the MDI icon glyph (recipe placeholders, stats cards, etc.) are intentionally left as-is — those are semantic icons, not branding.

image

**Docs**
- `backend-config.md`: "Branding" section updated for the two remaining env vars, with an explicit note that this only covers the app name, tab title, and logo — not every other place "Mealie" appears (translations, emails, links, etc.).

N/A — not tied to a tracked issue.

This revision addresses review feedback:
- Dropped the `index.html`/`SPAStaticFiles` rewrite entirely. Title/favicon are now set client-side only via `useHead`, so there's no more risk from rewriting how every HTML response is served, and the caching regression (conditional `/` requests coming back as a 304 with a body, dropped `ETag`/`Last-Modified`) is gone — `mealie/routes/spa/__init__.py` is back to its pre-branding state.
- Branding values now come from the existing `/api/app/about` response instead of a new endpoint/schema/plugin — `AppBranding` and `branding.client.ts` are both gone.
- Settings trimmed to `BRANDING_NAME` + a single `BRANDING_LOGO_PATH`, with one file route (`/about/branding-logo`) instead of two.
- Docs now explicitly scope branding to app name/tab title/logo only.
- `frontend/app/lib/api/types/admin.ts` was regenerated via the codegen pipeline rather than hand-edited.

Other notes:
- `logo_path` takes no request-derived input — it's an operator-set env var pointing at a fixed on-disk path — so there's no path-traversal surface; the suffix allowlist is purely to avoid serving arbitrary files with a wrong `Content-Type`.
- Because branding is now applied entirely client-side through `/api/app/about`, it behaves identically in the Nuxt dev server and the production static build — there's no longer a dev/prod distinction to call out here.

## Testing

- Rewrote the `Branding` settings unit tests for the single-`logo_path` shape (missing file, directory, disallowed suffix, env var override).
- Rewrote the `/about` integration tests to cover the branding fields on `AppInfo` and the new `/about/branding-logo` route (defaults, custom logo, missing-file fallback); removed the now-obsolete `apply_branding()` HTML-rewrite tests from `test_spa.py`.
- `ruff check`/`ruff format`/`mypy` clean on all changed files; `eslint` clean on all changed frontend files; full branding/about/spa pytest suite (30 tests) passes.
- Manually ran the FastAPI backend and Nuxt dev server together with `BRANDING_NAME`/`BRANDING_LOGO_PATH` set and confirmed via the API that `/api/app/about` returns the overridden name/logo URL (through the dev proxy as well as directly) and that `/about/branding-logo` serves the file with the correct content type. Ran a full visual test in a browser.
## AI / LLM Assistance

This PR was implemented end-to-end by Claude (Anthropic's AI coding assistant, via Claude Code) — planning, backend/frontend code, new tests, and docs — based on requirements I provided in conversation. I reviewed the plan before implementation, then manually tested the running app myself and reported two real bugs back to the assistant (a client-side hydration issue that reset the branded title/favicon after page load, and a mislabeled test favicon file), both of which were diagnosed and fixed in the same session. I have reviewed the resulting diff.


jspeyside:
rebased

michael-genson (changes requested):
Thanks for the PR! I like the idea of letting self-hosters customize branding, but I'm not convinced on the implementation. I'd like to keep this as small as possible to limit the maintenance burden.

A few thoughts on reshaping it:

- Since the frontend isn't server-rendered, could we set the title and favicon only on the client (via `useHead`) and drop the `index.html` rewrite and the `SPAStaticFiles` change? Rewriting how every HTML response is served is where most of the risk is. As written, it also breaks caching: a conditional request for `/` comes back as a 304 with a body, and the `ETag`/`Last-Modified` headers are dropped, so `index.html` is re-downloaded on every load.
- Could the branding values come from the existing `/api/app/about` app info (which the frontend already loads on startup) instead of a new endpoint, schema, and plugin?
- Could we trim the settings down to `BRANDING_NAME` (header, tab title, PWA name) and a single `BRANDING_LOGO_PATH` used for both the header icon and the favicon? That would mean one file route instead of two.

I'd also like the docs to be explicit that this only covers the app name, tab title, and logo, and not every place "Mealie" appears (translations, emails, links, etc.).

Also, `frontend/app/lib/api/types/admin.ts` is generated, so could you regenerate it with `task dev:generate` rather than editing it by hand?

jspeyside:
Updated to address review feedback

mealie/core/settings/branding.py
```diff
@@ -0,0 +1,25 @@
...
+ALLOWED_IMAGE_SUFFIXES = {".svg", ".png", ".ico", ".webp", ".jpg", ".jpeg"}
+
+
+class Branding(BaseSettings):
+    name: str = "Mealie"
+    logo_path: str | None = None
+    model_config = SettingsConfigDict(env_prefix="branding_", extra="allow")
+
+    @property
+    def logo_file(self) -> Path | None:
```
michael-genson: Can we log a warning when `BRANDING_LOGO_PATH` is set but we fall back to the default logo (missing file, directory, or unsupported extension)? Otherwise a typo'd path is pretty hard to debug. Probably best to do it once at startup rather than here, since this runs on every `/api/app/about` call.

michael-genson (commented):
Looks good, just one small change and we're good to go!

mealie/core/settings/branding.py
```diff
@@ -0,0 +1,40 @@
...
+        self._logo_file = None
+        if not self.logo_path:
+            return self
+
+        file_path = Path(self.logo_path)
+        if file_path.is_file() and file_path.suffix.lower() in ALLOWED_IMAGE_SUFFIXES:
+            self._logo_file = file_path
+        else:
+            # Deferred import to avoid a circular import: root_logger -> config -> settings -> branding
+            from mealie.core.root_logger import get_logger
```
michael-genson: I think this crashes on startup when the path is bad: `Branding()` is built while `mealie.core.config` is still importing, so this import fails with `ImportError: ... partially initialized module 'mealie.core.config'`. Can we move the warning into `lifespan_fn` in `mealie/app.py` with the other startup logs, and leave the validator to just resolve the path?

michael-genson (commented):
Thanks for the update! One more:

mealie/core/settings/branding.py
jspeyside: Good catch! I moved the warning log alongside the other startup logs.

PR #8083, opened 2026-08-14
fix: Exporting using ZIP file results in empty fields

### LINKED ISSUE #8045: [BUG] - Exporting using ZIP file results in empty fields
### First Check

Sharing/Importing a ZIP recipe results in many ingredients being blank.
This is running 3.22.0 on source and destination servers.

Image

Using the attached file I was able to replicate the same problem on the demo side. 

[korean-bbq-steak.zip](https://github.com/user-attachments/files/30883612/korean-bbq-steak.zip)

### Steps to Reproduce

1) On Server A, download the ZIP version of any recipe (recipe.zip)
2) Import recipe.zip on Server B

n/a

### Mealie Version

_No response_

### Deployment

Docker (Linux)

_No response_

xiaofang2016:

* Fixes recipe import when ingredient food records are referenced by the imported recipe.
* Checks whether the referenced ingredient food already exists before creating a new record.
* Reuses an existing ingredient food in the current group when the normalized name, plural name, or alias matches.
* Creates the ingredient food when no matching record exists and assigns the newly created ID to the imported data.
* This addresses the empty ingredient fields described in #8045. The underlying issue occurs during the import process rather than during ZIP file generation.
* After the data migration is completed, refresh the food list in data management so the user can see the latest added food immediately.

Fixes #8045

The main change is in the ingredient food handling during recipe import.

The implementation first attempts to resolve an existing food record and then falls back to matching by normalized name, plural name, or alias before creating a new food record. This is intended to avoid creating duplicate ingredient foods during imports.

## Testing

* Tested importing a recipe from a ZIP file.
* Verified that ingredient food records are created when they do not already exist.
* Verified that existing ingredient foods are reused when a matching normalized name, plural name, or alias is found.
* Verified that the imported ingredient data receives the corresponding food ID.
* Search some new added food in  Food Data page without refreshing any page or restarting the server.

## AI / LLM Assistance

An LLM was used to assist with understanding the issue, reviewing the relevant code, and discussing the implementation approach. I reviewed the suggested changes, implemented the solution, and verified the resulting code and behaviour myself.


xiaofang2016:
hi, can anyone help review my PR? What should I do to make it faster for an approval?

mealie/services/recipe/recipe_service.py
```diff
@@ -280,12 +325,13 @@ def _process_recipe_data(self, key: str, data: list | dict | Any):
...
         data["household_id"] = str(self.user.household_id)
 
-        # make sure categories and tags are valid
+        # make sure categories and tags and food are valid
         if key == "recipe_category":
             return self._transform_category_or_tag(data, self.repos.categories)
         elif key == "tags":
             return self._transform_category_or_tag(data, self.repos.tags)
-
+        elif key == "food":
```
michael-genson, line 387: Units go through the same `auto_init` lookup, so a unit from another server is still dropped to `None` here. Could we handle `unit` too? `DataMatcher` (`mealie/services/parser_services/_base.py`) already does the name/plural/alias lookup for both foods and units (including unit abbreviations), so reusing it would cover both without the hand-written query. Note that it fuzzy-matches by default, so you may want to pass a threshold of `100` for exact matches only.

michael-genson (changes requested):

mealie/services/recipe/recipe_service.py
xiaofang2016: Hi, @michael-genson thanks for the review. I have updated the unit process in the same way as food, and I removed my query function so I can reuse the _get_data_matcher function in _base.py. That really inspires me a lot on how to make contribution in Mealie.Thanks again for your advise and I will keep on working . Please review it again and I hope this time can satisfy the requirements.

michael-genson (approved):
Thanks for the updates, looks good!

PR #8094, opened 2026-08-15
chore: treat all warnings as errors in pytest unless explicitly ignored.

### LINKED ISSUE #8095: Warnings pass silently in tests/CI
Python/library warnings raised during tests are currently just printed and ignored — a test can emit a DeprecationWarning or similar and still pass. That means a regression introducing a new warning gets merged without anyone noticing, since CI shows green either way.

Proposed/Possible Solution(s)?
Add "error" to pyproject.toml's pytest filterwarnings, so any warning not explicitly excused fails the test instead of passing silently. Known, unfixable third-party warnings (e.g. from unmaintained dependencies) can stay explicitly ignored by name.

bestxmg:

Adds `"error"` to `pyproject.toml`'s `filterwarnings`, so tests now fail on any warning instead of letting it pass silently. Previously a regression could introduce a new warning and still pass CI cleanly.

Fixes #8095

## Testing

- Full suite passes: `2637 passed, 0 failed, 16 skipped`.
- Verified the check actually works: temporarily reintroduced a known warning elsewhere in the code, confirmed it correctly failed 4 tests, then reverted that test change.

## AI / LLM Assistance

Claude Code


bestxmg:
Hi @michael-genson,

Thanks again for your help merging our first PR #8033 last time!

Could you please take a look at this one when you get a chance?

We're a study group from the University of Waikato. In a previous PR we fixed all the existing warnings surfaced by pytest, but there was nothing stopping a new warning from being introduced unnoticed in the future. This PR closes that gap: the test suite now treats any warning as a failure unless it's explicitly ignored, so both `task py:check` locally and GitHub CI will catch it going forward instead of letting it pass silently.

Thanks!

michael-genson (approved):
LGTM, thanks again!

PR #8103, opened 2026-08-16
fix: show registration API error messages

### LINKED ISSUE #5834: [BUG] - 'Something went wrong' when trying to register new user with invalid token - no logs or errors presented to user on why
### First Check

I am unable to give a new user a group ID from a group that already exists to join on registration. When the new user tries to create the account, the 'Something went wrong' message pops up (would also like to request a more descriptive error message is used) and nothing happens.

This can be circumvented by creating an invite link, but I would rather not create individual invite links for each of the people that I want to invite since it seems like the invite links are one per person according to my testing.

### Steps to Reproduce

1) Go to the Registration page for Mealie
2) Create a new user (you can even make a new group to test it with
3) Get the Group Id from the group you just created through registering or an already existing group Id
4) Register by joining that group

I don't have access to the server hosting my Mealie instance until later today, but the front end log returns this:
```
{
  "message": "Request failed with status code 400",
  "name": "AxiosError",
  "stack": "Em@https://demo.mealie.io/_nuxt/D38o0MPC.js:44:1034\ng@https://demo.mealie.io/_nuxt/D38o0MPC.js:44:5669\n",
  "config": {
    "transitional": {
      "silentJSONParsing": true,
      "forcedJSONParsing": true,
      "clarifyTimeoutError": false
    },
    "adapter": [
      "xhr",
      "http",
      "fetch"
    ],
    "transformRequest": [
      null
    ],
    "transformResponse": [
      null
    ],
    "timeout": 10000,
    "xsrfCookieName": "XSRF-TOKEN",
    "xsrfHeaderName": "X-XSRF-TOKEN",
    "maxContentLength": -1,
    "maxBodyLength": -1,
    "env": {},
    "headers": {
      "Accept": "application/json, text/plain, */*",
      "Content-Type": "application/json",
      "Accept-Language": "en-US",
      "Authorization": "Bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIwZGZjZDRmOC0yY2VjLTQwMTUtOTM3NC03MDU0ZWE2OGI0NTIiLCJleHAiOjE3NTM5NzE2MTMsImlzcyI6Im1lYWxpZSJ9.4kVpmHipTyuA7tkg-mK0eWlzVtuTwEokdLc861kJAaY"
    },
    "baseURL": "/",
    "withCredentials": true,
    "method": "post",
    "url": "/api/users/register",
    "data": "{\"email\":\"yikifin718@devdigs.com\",\"username\":\"Testame\",\"fullName\":\"Testa Me\",\"password\":\"123456789\",\"passwordConfirm\":\"123456789\",\"locale\":\"en-US\",\"advanced\":false,\"groupToken\":\"b56c1d09-1dcb-4c0f-a67a-f4b20fe1191b\"}",
    "allowAbsoluteUrls": true
  },
  "code": "ERR_BAD_REQUEST",
  "status": 400
}
```

The Response data states that it is
`"Request failed with status code 400", name: "AxiosError", code: "ERR_BAD_REQUEST", message: "Invalid group token"`
even though I copied the group token from the home group in the demo page 

### Mealie Version

_No response_

### Deployment

Docker (Linux)

_No response_

YibingZhang325:

Registration failures can show a specific backend error, such as "Invalid group token", followed by the generic "Something Went Wrong!" message.

The registration page now skips its generic fallback when the error contains `detail.message`, which the existing Axios interceptor already displays. Errors without that message still use the generic fallback. Successful registration still resets the form and redirects to login.

The change is limited to the registration page and its tests. Global Axios error handling is unchanged.

Fixes #5834

## Testing

- Added six registration-page tests covering handled backend messages, missing errors, string details, empty messages, validation arrays, and successful registration.
- Verified the duplicate-message test fails against the original page and passes with the fix.
- `task ui:check`: ESLint and all 397 tests across 37 files passed.
- `task py:check`: formatting, Ruff, and mypy passed; pytest reported 2879 passed, 18 skipped, and 3 failures caused by sandbox restrictions on binding localhost sockets. Re-running the full affected `tests/unit_tests/pkgs/safehttp/test_transport.py` file outside the sandbox passed all 32 tests.

## Release notes

Avoid showing an extra generic error after a specific registration error.

## AI / LLM Assistance

OpenAI Codex assisted with implementing the review changes and running the checks.


frontend/app/plugins/axios.ts
```diff
@@ -95,9 +96,10 @@ export default defineNuxtPlugin((nuxtApp) => {
         }
       }
 
-      if (error?.response?.data?.detail?.message) {
-        alert.error(error.response.data.detail.message as string);
-      };
+      const errorMessage = getApiErrorMessage(error);
```
michael-genson: This makes the global interceptor toast string `detail` values too, which the backend uses in a lot of places (e.g. the 401 `"Could not validate credentials"`, and various 404s), so those would start showing untranslated toasts across the app. The registration error in #5834 already uses `detail.message`, which this handled before. Could we revert the interceptor change and keep the fix to the registration page?

michael-genson (changes requested):

frontend/app/plugins/axios.ts
YibingZhang325: Thanks for the feedback! I reverted the interceptor changes and kept the fix on the registration page. I also updated the tests and PR description. Could you take another look when you have time?

michael-genson (approved):
Thank you!

PR #8111, opened 2026-08-17
fix: prevent duplicate iOS camera photos from overwriting recipe step assets

### LINKED ISSUE #5623: [BUG] - Taking multiple photos for Steps only keeps first photo
### First Check

Hello. First, thank-you for such a great project!

When I go to add a step under Instructions for a recipe, and I tap the dots and select Upload image, Take photo. I can then take a photo using my device's camera, and after a few seconds I see the link embedded in the step. Great, so far so good. However, if I take more than one picture it only ever keeps the first picture. I can see the embedded links are identical (ie. of the first picture). One time, I went through an entire recipe, taking pictures along the way, only to find that at the end, I had several pictures of just the first picture, repeated over and over.

Couple things to note:
- iOS (iPad)
- This only happens when taking a picture within the app. If I use the upload from existing photos this does not happen. So, as long as I take the photos with just the regular camera app, then upload the picture in Mealie, it works just fine.
- Looking at the attached log below, I can see when I take the first picture. However, I don't see a similar output to the logs when I take the second picture.

Thank-you.

### Steps to Reproduce

1) Upload PIcture > Take Photo
2) Photo is uploaded and link is embedded in the Step's description
3) Repeat step 1). Second photo is disregarded and the link to the first photo is embedded again

INFO     2025-07-03T09:42:54 - [192.168.2.2:45832] 200 OK "GET /api/households/self/recipes/photo-test HTTP/1.1"
INFO     2025-07-03T09:44:28 - [192.168.2.2:31340] 200 OK "GET /api/organizers/tools?page=1&perPage=-1&orderBy=name&orderDirection=asc HTTP/1.1"
INFO     2025-07-03T09:43:15 - [127.0.0.1:51722] 200 OK "GET /api/app/about HTTP/1.1"
INFO     2025-07-03T09:43:22 - [192.168.2.2:2388] 200 OK "POST /api/recipes/photo-test/assets HTTP/1.1"
INFO     2025-07-03T09:43:44 - [192.168.2.2:35194] 200 OK "GET /api/organizers/tools?page=1&perPage=-1&orderBy=name&orderDirection=asc HTTP/1.1"
INFO     2025-07-03T09:43:22 - [192.168.2.2:2388] 200 OK "GET /api/media/recipes/90858b97-b8a0-404c-af12-87932fc931e0/assets/image-jpg.jpg HTTP/1.1"
INFO     2025-07-03T09:43:44 - [192.168.2.2:7094] 200 OK "GET /api/organizers/categories?page=1&perPage=-1&orderBy=name&orderDirection=asc HTTP/1.1"
INFO     2025-07-03T09:43:48 - [192.168.2.2:22532] 200 OK "POST /api/recipes/photo-test/assets HTTP/1.1"
INFO     2025-07-03T09:43:46 - [127.0.0.1:55498] 200 OK "GET /api/app/about HTTP/1.1"
INFO     2025-07-03T09:43:50 - [192.168.2.2:22532] 200 OK "GET /api/organizers/categories?page=1&perPage=-1&orderBy=name&orderDirection=asc HTTP/1.1"
INFO     2025-07-03T09:43:50 - [192.168.2.2:36010] 200 OK "GET /api/organizers/tools?page=1&perPage=-1&orderBy=name&orderDirection=asc HTTP/1.1"
INFO     2025-07-03T09:44:16 - [127.0.0.1:46070] 200 OK "GET /api/app/about HTTP/1.1"
INFO     2025-07-03T09:44:28 - [192.168.2.2:31340] 200 OK "PUT /api/recipes/photo-test HTTP/1.1"
INFO     2025-07-03T09:44:47 - [127.0.0.1:47576] 200 OK "GET /api/app/about HTTP/1.1"

### Mealie Version

mealie:v2.7.1
d24a518bac7c89df27118286c8d9588f2edbbc44

### Deployment

Docker (Linux)

Docker with Portainer, running on Proxmox LXC. Installed using PostgreSQL database.

RisonG:

When taking multiple photos with the in-app camera (e.g. on iOS) to insert into a recipe step,
only the first photo was ever kept — every subsequent photo's embedded image link pointed to the
same file as the first one.

Root cause: iOS Safari always names camera-captured files `image.jpg`, regardless of how many
photos are taken in a session (photos picked from the library instead keep a unique filename,
which is why that path worked fine). The `/api/recipes/{slug}/assets` upload endpoint derives the
on-disk filename deterministically from the client-supplied `name` via `slugify()`, with no
uniqueness check, and opens the destination file with `"wb"` — so a second upload with the same
name silently overwrote the first file on disk while also inserting a duplicate `RecipeAsset`
record pointing at the same URL.

- `mealie/routes/recipe/recipe_crud_routes.py`: in `upload_recipe_asset`, check whether the
  destination file already exists before writing; if so, append a short random suffix (uuid4 hex)
  to the filename so the new asset gets a unique path instead of overwriting the previous one.
- `tests/integration_tests/user_recipe_tests/test_recipe_image_assets.py`: added
  `test_recipe_assets_create_duplicate_name`, a regression test that uploads two assets with the
  same `name` (mirroring the iOS `image.jpg` scenario) and asserts both files are kept, with
  distinct filenames, and both appear on the recipe.

This is a server-side fix, so it protects against filename collisions from any client, not just
the iOS camera capture flow.

Fixes #5623

The fix only touches the collision-handling branch inside `upload_recipe_asset`; the happy path
(unique names) is unchanged, so the existing `test_recipe_assets_create` test still passes as-is.

## Testing

Added a new integration test (`test_recipe_assets_create_duplicate_name`) that uploads two assets
with an identical `name` and verifies:
- both HTTP requests succeed
- the two assets get different `fileName`s
- both files exist on disk
- the recipe ends up with 2 asset records (not 1 overwritten one)

Ran the full asset test file locally: `uv run pytest tests/integration_tests/user_recipe_tests/test_recipe_image_assets.py -v` — all 6 tests pass.

## AI / LLM Assistance

Claude Code (Anthropic) was used to investigate the root cause (tracing the upload flow from the
iOS camera capture in the frontend through to the backend asset-upload endpoint), implement the
fix, and write the regression test. The diagnosis and fix were reviewed and verified by me before
committing.

### DIFF
--- a/mealie/routes/recipe/recipe_crud_routes.py
+++ b/mealie/routes/recipe/recipe_crud_routes.py
@@ -3,7 +3,7 @@ from collections import defaultdict
 from collections.abc import AsyncIterable, Awaitable, Callable
 from shutil import copyfileobj
 from typing import Annotated
-from uuid import UUID
+from uuid import UUID, uuid4
 
 import orjson
 import sqlalchemy
@@ -817,7 +817,6 @@ class RecipeController(BaseRecipeController):
             raise HTTPException(status_code=400, detail="Missing required fields")
 
         file_name = f"{file_slug}.{extension}"
-        asset_in = RecipeAsset(name=name, icon=icon, file_name=file_name)
 
         recipe = self.service.get_one(slug)
 
@@ -829,6 +828,14 @@ class RecipeController(BaseRecipeController):
                 status_code=400,
                 detail=f"File name {file_name} or extension {extension} not valid",
             )
+
+        # Client-supplied names aren't guaranteed to be unique (e.g. iOS camera captures are all
+        # named "image.jpg"), so avoid silently overwriting an existing asset with the same name.
+        if dest.is_file():
+            file_name = f"{file_slug}_{uuid4().hex[:8]}.{extension}"
+            dest = recipe.asset_dir / file_name
+
+        asset_in = RecipeAsset(name=name, icon=icon, file_name=file_name)
 
         with dest.open("wb") as buffer:
             copyfileobj(file.file, buffer)
--- a/tests/integration_tests/user_recipe_tests/test_recipe_image_assets.py
+++ b/tests/integration_tests/user_recipe_tests/test_recipe_image_assets.py
@@ -42,6 +42,49 @@ def test_recipe_assets_create(api_client: TestClient, unique_user: TestUser, rec
     assert recipe_respons["assets"][0]["name"] == payload["name"]
 
 
+def test_recipe_assets_create_duplicate_name(
+    api_client: TestClient, unique_user: TestUser, recipe_ingredient_only: Recipe
+):
+    """
+    Regression test for https://github.com/mealie-recipes/mealie/issues/5623
+
+    iOS Safari names every camera-captured photo "image.jpg", so uploading multiple assets in a
+    row with an identical client-supplied name must not silently overwrite the previous asset.
+    """
+    recipe = recipe_ingredient_only
+    payload = {
+        "name": "image",
+        "icon": random_string(10),
+        "extension": "jpg",
+    }
+
+    first_response = api_client.post(
+        f"/api/recipes/{recipe.slug}/assets",
+        data=payload,
+        files={"file": data.images_test_image_1.read_bytes()},
+        headers=unique_user.token,
+    )
+    assert first_response.status_code == 200
+    first_file_name = first_response.json()["fileName"]
+
+    second_response = api_client.post(
+        f"/api/recipes/{recipe.slug}/assets",
+        data=payload,
+        files={"file": data.images_test_image_1.read_bytes()},
+        headers=unique_user.token,
+    )
+    assert second_response.status_code == 200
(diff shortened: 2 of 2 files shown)


mealie/routes/recipe/recipe_crud_routes.py
```diff
@@ -830,6 +829,14 @@ def upload_recipe_asset(
...
             )
 
+        # Client-supplied names aren't guaranteed to be unique (e.g. iOS camera captures are all
+        # named "image.jpg"), so avoid silently overwriting an existing asset with the same name.
+        if dest.is_file():
+            file_name = f"{file_slug}_{uuid4().hex[:8]}.{extension}"
+            dest = recipe.asset_dir / file_name
+
+        asset_in = RecipeAsset(name=name, icon=icon, file_name=file_name)
+
```
michael-genson, line 839: Can we place this above the path containment check right above this?

michael-genson (changes requested):
One small change, thanks for this!

PR #8123, opened 2026-08-17
feat: add connectivity test for AI providers

elliot-ultra:

Adds a "Test Connection" feature for AI providers, so users can verify their configuration works before relying on it — right now the only way to find out a provider is misconfigured is to hit a confusing failure somewhere else in the app (recipe parsing, image import, etc.).

OpenAI doesn't have a dedicated connectivity-test endpoint, but the `/models` endpoint (listing available models) is free/unmetered on most OpenAI-compatible providers, so this uses that instead of a real chat completion. The configured model name is cross-checked against the returned model list to catch typos.

- **Test button** on the create/edit provider dialog — tests the unsaved form values, or (when editing) the saved provider merged with whatever's currently in the form, so editing + testing before saving actually reflects your in-progress changes
- **Auto Test toggle + per-row retry** on the providers list, with color-coded latency and a details dialog for warnings/failures
- Three result states: success (green), connected-but-model-not-confirmed (yellow, non-blocking), and connection failed (red)
- Error messages are sanitized — long non-JSON error bodies (e.g. a typo'd base_url landing on a Cloudflare block page) get truncated instead of dumping raw HTML into the UI, and 401s never pass through provider error text since some providers echo back a masked API key fragment

Not tied to a bug report — implements the feature proposed and discussed in #8051, which includes a full writeup of the design rationale and real-provider testing notes.

The model-list cross-check is deliberately a soft warning, not a hard failure, and this was validated against a real edge case: Moonshot AI (Kimi) omits models the account hasn't paid for from `/models` — a model missing from the list doesn't necessarily mean the name is wrong. Treating it as a hard block would have actively prevented users from using legitimately-correct-but-gated models.

This feature only validates connectivity (key/base_url/model name) — it cannot validate provider capabilities. DeepSeek passes this test (valid key, valid model) but doesn't support multimodal input, so it would still fail later if selected as the Image Provider. There's no standardized way to query capability support across OpenAI-compatible providers, so this is an accepted, out-of-scope limitation, not an oversight.

Scope was kept intentionally tight — this doesn't touch the existing `admin_debug.py` OpenAI debug endpoint (site-admin-only, image-upload-capable functional test), which serves a different purpose and audience than this group-manager-facing connectivity check.

## Testing

- Unit tests for `OpenAIService.test_connection()`: success/model-match, model-not-found warning, empty-model-list (no false warning), generic failure, timeout override, long-error-message truncation, 401-doesn't-leak-key-fragment, HTTP-status-in-message
- Integration tests for both new routes: unsaved test, saved test, override-merge behavior when editing, permission checks (`can_manage` required), never persists on test, 404 for unknown provider
- Manually tested against live accounts: OpenAI, DeepSeek, Moonshot/Kimi (both the international and domestic endpoints), Qwen (DashScope), and Gemini's OpenAI-compat layer — covering valid configs, wrong API keys, and wrong base_urls (including one that hit an unrelated Cloudflare-protected domain)
- Self-reviewed the diff before opening this PR and fixed what it found: a race condition where a slow in-flight test could overwrite the result shown for a different provider, the edit dialog testing stale saved config instead of current form values, and an unnecessary DB query pattern in `OpenAIService.__init__`

## AI / LLM Assistance

Used Claude Code extensively throughout this PR — from initial design discussion (informed by real testing against OpenAI, DeepSeek, Moonshot/Kimi, Qwen, and Gemini) through implementation, test writing, and a self-review pass that caught and fixed several bugs before submitting (a race condition on rapid provider switching, stale-config testing in edit mode, and an unnecessary DB query pattern). I directed the design decisions, ran the actual manual testing against live provider accounts, and reviewed/verified the generated code and test results myself before pushing.


michael-genson:
Thanks for the detailed writeup, and for doing the real provider testing, that context is genuinely useful.

That said, I'm not convinced on the added value over the admin test endpoint we already have. `POST /api/admin/debug/openai/{provider_id}` has been there a while, with a UI at `/admin/debug/openai` linked from the admin sidebar, and it's a stronger check than this one since it runs a real completion and confirms the model actually responds, rather than only confirming the key and base URL are valid. Your own notes make that case, DeepSeek passes a `/models` check but still can't complete an image import.

The gaps I do see are that the existing endpoint is admin only so a group manager can't reach it, it requires a saved provider so you can't test before saving, and it sits on a separate debug page instead of next to the config. Those seem worth closing. Could that be done with a test button on the provider dialog and a group scoped route, reusing what's already in `admin_debug.py` rather than adding a second implementation?

The part I'd want to leave out is the layer around it, the auto test on page load, the per row badges, the persisted preference, and the latency color coding. That's a lot of new surface for something most people click once when setting a provider up, and the 1000ms/1200ms thresholds are a policy we'd then have to defend.

elliot-ultra:
Yes, the existing debug functionality is already quite powerful. Initially, the goal was to minimize changes and avoid issues arising from unfamiliarity with the project, so the focus was primarily on adding new code. Adding repetitive functionality is not a good solution from a project perspective, and this has now been adjusted, reusing methods from admin_debug.py.

Regarding the auto-test feature you mentioned, its value lies in allowing users to quickly check connectivity when configuring multiple AI providers, rather than having to manually test each one in the editing page. Of course, this is based on theoretical scenarios. Considering the project's characteristics, it's unlikely that users will configure many AI providers to parse recipes in practice, so this feature has been removed.

This commit adds recipe-image.jpg and recipe.json for the image check: recipe-image.jpg serves as the test image for verifying an AI provider's image recognition capability. recipe.json lists the keywords expected to appear in the recipe shown in that image—if the AI's response contains one of them, the image check passes.

| Success | Doesn't Support Structured Outputs | Run Out of Balance | Unsupported Model |
| --- | --- | --- | --- |
| image |  image |  image | image |

mealie/services/openai/openai.py
```diff
@@ -157,6 +205,74 @@ def get_client(self, provider: AIProviderOut) -> AsyncOpenAI:
...
+        """
+        try:
+            response = await self.ping(provider, "Hello, checking to see if I can reach you.")
+        except Exception as e:
+            return AIProviderTestResult(success=False, message=_truncate_error_message(_describe_ping_error(e)))
+
+        if not response:
+            return AIProviderTestResult(success=False, message="No response received from the provider.")
+
+        image_test_passed, image_test_message = await self._test_image_recognition(provider)
```
michael-genson, line 236: Can we drop the image recognition check (and `testing/recipe-image.jpg`, `recipe.json`, and the image check UI in the dialog)? The text ping already covers what I was after. The keyword match is another rule we'd have to defend. It also runs on every provider whatever its role, so a correctly set up text-only default provider gets a red "Image check" X.

mealie/services/openai/openai.py
```diff
@@ -118,34 +153,47 @@ def build_message(self) -> dict:
...
-        self.default_provider = (
+        settings = get_app_settings()
+        self.custom_prompt_dir = settings.OPENAI_CUSTOM_PROMPT_DIR
+
+        super().__init__()
+
+    # provider_settings/default_provider/audio_provider/image_provider are lazy: most callers
+    # (e.g. test_connection) only need get_client() for one already-known provider and never
+    # touch these, so eagerly loading them in __init__ would be up to 4 wasted DB lookups.
+    @cached_property
```
michael-genson, line 169: Do we need the lazy-loading `cached_property` change here? It isn't related to the feature, and a few lookups on a button people click once aren't worth changing how the service loads its providers.

mealie/services/openai/openai.py
```diff
@@ -28,6 +30,39 @@
...
+
+
+def _truncate_error_message(text: str, limit: int = _MAX_TEST_ERROR_MESSAGE_LENGTH) -> str:
+    text = " ".join(text.split())
+    if len(text) <= limit:
+        return text
+    return f"{text[:limit].rstrip()}…"
+
+
+def _describe_ping_error(e: Exception) -> str:
```
michael-genson, line 47: Can we drop these error helpers? Also, this route is now open to group managers, not just admins. Can we avoid sending the provider's response body back? Otherwise a manager could point `base_url` at an internal host and read its error pages. The status code or exception type alone might be enough.

frontend/app/components/Domain/Group/GroupAIProviderDialog.vue
```diff
@@ -132,15 +166,59 @@ const defaultForm = () => ({
 
 const formData = reactive(defaultForm());
 
+const testing = ref(false);
+const testResult = ref<AIProviderTestResult | null>(null);
+
+// Bumped every time the dialog starts testing a (possibly different) provider, so a slow
+// response that's still in flight when the user switches providers gets ignored instead of
+// silently overwriting the result now being shown for someone else's config.
+let testRequestId = 0;
```
michael-genson, line 175: Do we need the request-id tracking? The dialog is modal, so disabling the Test button while a test is running should cover it.

frontend/app/components/global/BaseButtonGroup.vue
```diff
@@ -25,6 +25,7 @@
             :color="btn.color"
             :large="large"
             :disabled="btn.disabled"
+            :loading="btn.loading"
```
michael-genson, line 28: Double check this, I don't think anything uses `loading` now that the per-row retry is gone.

frontend/app/components/Domain/Group/GroupAIProviderSettingsEditor.vue
```diff
@@ -96,14 +96,14 @@
       variant="tonal"
       class="pa-0 mb-4"
     >
-      <v-row no-gutters>
-        <v-col :cols="10">
+      <v-row no-gutters align="center">
```
michael-genson, line 99: Can we revert this layout change? It looks like it was for the per-row badges that were removed.

michael-genson (changes requested):

mealie/services/openai/openai.py
elliot-ultra: Thanks for the review, I have already changed points 2-6 according to the suggestions. I still think the image capability check is an important part of the test. I tried qwen3-max as an AI model, which is set up correctly and passes the text check, but doesn't support images. The message in the dialog let me know this provider can't be used for image recognition. In order to make the testing process easier to understand I changed the message text and UI, using more neutral text to describe AI capability, rather than "[X] Image check", but if you'd still prefer to drop it, I'll remove it. Also, I shortened the error message, the response body is no longer returned at all, the result only carries the exception type and status code as you suggested. The full error is logged server-side instead.
The screenshots show the connection test:
| Success |  Doesn't Support Structured Outputs | Run Out of Balance | Unsupported Model | Text-only Model |
| --- | --- | --- | --- | --- |
| image | image | image | image | image |

michael-genson (approved):
Looks good, thanks for your work on this!

PR #8126, opened 2026-08-17
feat: filter meal plan rules by food label

Xyolyp:

Meal plan rules can already filter on individual **foods**, but not on the group a food belongs to. Setting up something as ordinary as a fish day means selecting every fish food by hand — and the rule then quietly rots, because every newly imported fish food is missing from that hand-maintained list. The rule keeps working and just stops being correct.

Mealie already stores the grouping this needs: a food's **Food Label** (`MultiPurposeLabel`). This PR makes it available as a rule criterion.

Discussed beforehand in #8125.

**Changes**

- `frontend/app/composables/use-query-filter-builder.ts` — adds a `foodLabel` field type and an `isMultiSelectType()` predicate. Organizers and food labels behave identically wherever a field's value is *a list of ids picked from a store* rather than free input, so those call sites now use the new predicate while organizer-specific handling keeps using `isOrganizerType()`. `OrganizerBase.slug` becomes optional, since labels have no slug and nothing reads it.
- `frontend/app/components/Domain/QueryFilterBuilder.vue` — adds a picker for the new field type, fed by the existing label store, plus a `storeForType()` helper so value hydration works for both organizers and labels.
- `frontend/app/components/Domain/Household/GroupMealPlanRuleForm.vue` — exposes the field as `recipe_ingredient.food.label_id`.
- `frontend/app/lang/messages/en-US.json` — one new string (`general.food-label`). No other locales touched.
- `docs/` — the planner rules paragraph listed only tags and categories; it now mentions ingredients and their food labels.
- tests, see below.

**No backend change.** `QueryFilterBuilder` already traverses relationship chains and applies the joins itself, and `IngredientFoodModel.label_id` is a `FilterableColumn`, so the attribute is reachable as-is. No schema change, no API change, no migration, and existing rules are untouched.

A deliberate design decision worth a reviewer's opinion: food labels stay **out** of the `Organizer` enum. They belong to foods rather than recipes, and there is no label organizer page to link to, so declaring them a recipe organizer felt wrong. The cost is a little duplicated markup in `QueryFilterBuilder.vue`. The alternative — adding `Label = "labels"` to the enum — is mechanical and I'm happy to switch if you prefer it.

None — this came out of discussion #8125.

Two things I chose *not* to do, both open to being overruled:

- The same field would fit the **cookbook editor**, which already offers the individual-food field. Left out to keep this PR to one surface.
- The recipe → ingredient join is one-to-many and the random draw applies its SQL limit before deduplication, so a recipe with three matching ingredients carries roughly three times the weight of a recipe with one. This is pre-existing behaviour of the current food filter rather than something this PR introduces, but a label filter makes it far more likely to be noticed. I'd rather fix that separately than fold it in here.

Two limitations users will run into, which may deserve documentation beyond what this PR adds:

- A food carries exactly **one** label, so labels used as shopping aisles ("Frozen", "Produce") can't double as ingredient kinds without relabelling.
- Only ingredients parsed into a food with a label can match; recipes with free-text ingredients never will.

## Testing

Unit tests (`tests/unit_tests/repository_tests/test_query_filter_builder.py`):

- the attribute chain resolves to `IngredientFoodModel.label_id`
- the compiled SQL joins `recipes` → `recipes_ingredients` → `ingredient_foods`
- `NOT IN` compiles to a subquery, so excluding a label excludes the whole recipe rather than matching its other ingredients

Integration tests (`tests/integration_tests/user_household_tests/test_group_mealplan.py`), alongside the existing tag/category rule tests:

- a food label rule only draws recipes containing an ingredient with that label
- excluding a label excludes a recipe even when it also contains an allowed ingredient

Frontend tests (`frontend/app/composables/__tests__/use-query-filter-builder.test.ts`): operator choices, query filter string construction for `IN`/`NOT IN`, and that `foodLabel` counts as multi-select but explicitly *not* as an organizer.

Manually verified against a dev instance as well: created a "Fish" label, a labelled food and recipes, and confirmed the random meal draw honours both the inclusive and the exclusive rule, and that a saved rule loads back into the editor correctly.

`yarn lint --max-warnings=0`, `yarn test:ci`, `ruff`, and `mypy` are clean locally.

## AI / LLM Assistance

Substantial. The codebase exploration, the implementation, the tests and the first draft of this description were produced with an LLM (Claude), working against a running dev instance. I reviewed the result and the behaviour before opening this PR. Stating it plainly so reviewers can calibrate: the design decisions above are ones I'd like scrutinised rather than taken on trust, particularly the choice to reuse food labels instead of introducing a separate ingredient grouping concept.

## Release Notes

Meal plan rules can now filter by a food's label, so a rule like "Friday is fish day" no longer requires listing every individual fish. Labels can be included or excluded, and excluding a label excludes any recipe containing an ingredient with it.

### DIFF
--- a/frontend/app/composables/use-query-filter-builder.ts
+++ b/frontend/app/composables/use-query-filter-builder.ts
@@ -18,7 +18,7 @@ export interface FieldPlaceholderKeyword {
 
 export interface OrganizerBase {
   id: string;
-  slug: string;
+  slug?: string;
   name: string;
 }
 
@@ -28,6 +28,7 @@ export type FieldType
     | "boolean"
     | "date"
     | "relativeDate"
+    | "foodLabel"
     | RecipeOrganizer;
 
 export type FieldValue
@@ -208,6 +209,14 @@ export function useQueryFilterBuilder() {
     );
   };
 
+  /**
+   * Field types whose value is a list of ids picked from a store, rather than free input.
+   * Food labels are not recipe organizers, but behave identically from the filter's point of view.
+   */
+  function isMultiSelectType(type: FieldType): boolean {
+    return isOrganizerType(type) || type === "foodLabel";
+  };
+
   function getFieldFromFieldDef(field: Field | FieldDefinition, resetValue = false): Field {
     const updatedField = {
       logicalOperator: logOps.value.AND,
@@ -215,7 +224,7 @@ export function useQueryFilterBuilder() {
     } as Field;
 
     let operatorChoices: FieldRelationalOperator[];
-    if (updatedField.fieldChoices?.length || isOrganizerType(updatedField.type)) {
+    if (updatedField.fieldChoices?.length || isMultiSelectType(updatedField.type)) {
       operatorChoices = [
         relOps.value["IN"],
         relOps.value["NOT IN"],
@@ -318,10 +327,10 @@ export function useQueryFilterBuilder() {
         isValid = false;
       }
 
-      if (field.fieldChoices?.length || isOrganizerType(field.type)) {
+      if (field.fieldChoices?.length || isMultiSelectType(field.type)) {
         if (field.values?.length) {
           let val: string;
-          if (field.type === "string" || field.type === "date" || isOrganizerType(field.type)) {
+          if (field.type === "string" || field.type === "date" || isMultiSelectType(field.type)) {
             val = field.values.map(value => `"${value.toString()}"`).join(",");
           }
           else {
@@ -368,5 +377,6 @@ export function useQueryFilterBuilder() {
     buildQueryFilterString,
     getFieldFromFieldDef,
     isOrganizerType,
+    isMultiSelectType,
   };
 }
--- a/frontend/app/composables/use-query-filter-builder.ts
+++ b/frontend/app/composables/use-query-filter-builder.ts
@@ -18,7 +18,6 @@ export interface FieldPlaceholderKeyword {
 
 export interface OrganizerBase {
   id: string;
-  slug?: string;
   name: string;
 }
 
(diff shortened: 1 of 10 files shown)


frontend/app/composables/use-query-filter-builder.ts
```diff
@@ -18,7 +18,7 @@ export interface FieldPlaceholderKeyword {
 
 export interface OrganizerBase {
   id: string;
-  slug: string;
+  slug?: string;
```
michael-genson, line 21: Nothing reads `slug` anywhere in here or in QueryFilterBuilder, so could we just drop it from `OrganizerBase` rather than making it optional?

And let's just add labels to the Organizer enum, we already have foods in there which aren't really organizers either. Then `RecipeOrganizerSelector` can handle the picker and we can drop the new autocomplete, `storeForType()` and `isMultiSelectType()`.

frontend/app/lang/messages/en-US.json
```diff
@@ -178,6 +178,7 @@
     "wednesday": "Wednesday",
     "yes": "Yes",
     "foods": "Foods",
+    "food-label": "Food Label",
```
michael-genson, line 181: `data-pages.foods.food-label` is already "Food Label", can we reuse that instead of sending a duplicate string to Crowdin?

michael-genson (changes requested):
Thanks, this is really cleanly put together, and the tests are appreciated. Couple of thoughts:

Xyolyp:
@michael-genson I've addressed your findings. Please let me know if there are any further concerns.

michael-genson (dismissed):
Looks good, thanks!

michael-genson:
Can you take a look at the test failure? I think if you merge in `mealie-next` and update the signature on `test_deep_traversal_to_food_label_works` you should be good.

michael-genson (approved):

PR #8133, opened 2026-08-18
fix: rescue markdown-fenced responses when a provider ignores strict schema output

lcs-crr:

Some OpenAI-compatible providers accept Mealie's `response_format=json_schema` (strict structured outputs) request without error, but don't actually enforce it — and the one confirmed against Ollama Cloud additionally wraps its JSON answers in a `‎``json ... `‎`` markdown code fence regardless of the requested response_format. That combination failed `client.chat.completions.parse()`: the SDK's own client-side validation rejected the fenced (otherwise valid) JSON *before* Mealie ever saw the content, so the fence handling could not rescue it, and any rescue path that stayed inside `parse()` would have cost a second API call per request.

`_get_raw_response()` in `mealie/services/openai/openai.py` now sends the request via `client.chat.completions.with_streaming_response.parse(..., response_format=response_schema)` and reads the raw body with `json.loads(await response.text())` inside the `async with`. This keeps the wire request identical to what `parse()` would send (it builds the same `response_format` payload itself, and `openai` is pinned to `==3.0.0`), while skipping the SDK's client-side validation gate entirely: `parse()`'s JSON validation runs as a `post_parser` that only executes inside the parsed-response path, which the raw-response path never triggers. Validation happens instead in `OpenAIBase.parse_openai_response()`, which strips a wrapping markdown fence before JSON validation (added to `_preprocess_response()` in `mealie/schema/openai/_base.py`, alongside its existing null-byte stripping). No private SDK imports are used. A provider that replies in prose (not fence-wrapped JSON) now fails with a clear error instead of being silently retried — that retry path is intentionally gone, per review discussion.

Parity with `parse()`'s error behavior is preserved: the body is validated into `ChatCompletion` and `LengthFinishReasonError`/`ContentFilterFinishReasonError` are raised when a provider truncates a response mid-schema (`finish_reason: "length"`) or filters it, so these surface as clear errors rather than confusing JSON-validation failures.

_(none filed — happy to open one if you'd like a record, otherwise this can be left blank/removed)_

## Testing

Unit tests covering the JSON parsing scenarios end to end:

- `test_openai_service.py`: strict `json_schema` request sent via `parse()` with plain JSON parsing unchanged; markdown-fenced JSON parsing successfully on the first call (no second API call); empty `choices` returning `None`; non-JSON response raising a clear error; `finish_reason: "length"` and `finish_reason: "content_filter"` raising the SDK's errors.
- `test_openai_base.py`: plain JSON; `‎``json` fences; bare `‎`` fences; fences with surrounding whitespace; literal and escaped null bytes stripped (pre-existing behavior, previously untested); prose raising `ValidationError`; fence-wrapped prose raising; empty/`None` response raising.
- Updated one existing test in `test_openai_parser.py` whose mock was coupled to the internal contract this PR changes.

All unit and integration tests pass (2860 passed, 16 skipped — the standard skips), `mypy` and `ruff` clean.

The wire request is byte-identical to `parse()`'s (verified against the pinned SDK's source, which posts the body with the same `response_format` and adds client-side parsing only afterwards), so compliant providers see no difference server side. The fence-stripping fix itself was previously verified end-to-end against a real Ollama Cloud account and a running Mealie instance (the fence case only surfaced under a real recipe-parsing prompt); this rework is covered by the unit tests above and the payload-equivalence verification, not re-verified live.

## AI / LLM Assistance

Diagnosed via Claude Code, which root-caused the original failures through direct API testing against Ollama Cloud and by reading the actual backend logs from a real running instance. The rework to send via `parse()`'s raw-response path was suggested by michael-genson in review and confirmed by reading the installed SDK's implementation of `parse()` and its streaming wrapper; the `finish_reason` handling restores the parity michael-genson flagged in review. I reviewed and committed the final diff, following TDD for the test updates.


michael-genson (changes requested):
Thanks for the thorough writeup and the tests, and for tracking this down against a real provider.

I'm hesitant on the fallback itself. Providers should respect the OpenAI spec, and the failure you hit looks specific to Ollama Cloud, so I'm not sure we want to carry a second code path to work around a provider bug.

The bigger issue is the cost. A provider that always wraps its output in a fence now pays two API calls on every request, since the fence stripping in `_preprocess_response` can't rescue the first attempt, the SDK validates internally before we ever see the content. That seems avoidable though, `parse()` is just `create()` plus a client side `model_parse_json`, and the `response_format` sent to the server is the same either way. Since we already validate in `parse_openai_response()`, calling `create()` with the same schema and parsing ourselves would behave identically server side in one call, handle the fence case on the first try, and drop the `except ValidationError` path entirely.

lcs-crr:
Reworked according to your suggestions. 
Just saw #8357 but on that `parse()` path, the fence-stripping in `_preprocess_response` is dead code for the case it's meant to rescue.
The SDK's client-side validation raises on a whole-response-wrapped-in-fences before Mealie ever sees it but the same global regex still runs on any response that's structurally valid JSON, so a field value that merely contains ``` or ** gets mangled anyway.

mealie/services/openai/openai.py
```diff
@@ -276,9 +275,11 @@ def get_prompt(self, name: str, data_injections: list[OpenAIDataInjection] | Non
 
     async def _get_raw_response(
         self, prompt: str, content: list[dict], response_schema: type[T], provider: AIProviderOut
-    ) -> ChatCompletion:
+    ) -> T | None:
+        from openai.lib._parsing._completions import type_to_response_format_param
```
michael-genson: I'm hesitant to rely on a private import here — Renovate bumps `openai` regularly and there's no stability guarantee on `openai.lib._parsing._completions`.

Use `client.chat.completions.with_streaming_response.parse(...)` instead, passing `response_format=response_schema` and reading `json.loads(await raw.text())` inside the `async with`. Same request, and `parse()`'s validation is a deferred `post_parser` the raw-response path never runs, so parsing still lands in `parse_openai_response()`. Use `with_streaming_response` rather than `with_raw_response` — the latter returns `LegacyAPIResponse`, where `.text` becomes async in the next major.

mealie/services/openai/openai.py
```diff
@@ -290,9 +291,18 @@ async def _get_raw_response(
...
             model=provider.model,
-            response_format=response_schema,
+            # The same response_format payload parse() would send, built with the SDK's own
+            # strict-schema helper. Calling create() directly skips the SDK's client-side
+            # validation, so responses we can parse ourselves (e.g. markdown-fenced JSON)
+            # are validated by parse_openai_response() below instead of being discarded.
+            response_format=type_to_response_format_param(response_schema),
         )
 
+        if not response.choices:
```
michael-genson: Should we check `finish_reason` here? `parse()` raised `LengthFinishReasonError` / `ContentFilterFinishReasonError` and we lose that on this path, so a truncated response fails as a generic JSON parse error instead of "hit the token limit".

michael-genson (commented):
Thanks for the rework — single call is what I wanted. Couple more:

Please also retitle; it still says "fall back to json_object mode", which is gone now, and the title lands as the squash commit.

michael-genson (approved):
Thanks for the rework, looks good!

Issue #8100, opened 2026-08-16
[BUG] - Registration flow allows proceeding to next step without valid group name


Luna-81:
### First Check

- [x] This is not a feature request.
- [x] I added a very descriptive title to this issue (title field is above this).
- [x] I used the GitHub search to find a similar issue and didn't find it.
- [x] I searched the Mealie documentation, with the integrated search.
- [x] I already read the docs and didn't find an answer.
- [x] This issue can be replicated on the demo site (https://demo.mealie.io/).

### What is the issue you are experiencing?

On the registration page, the "Next" button does not properly validate the group name field. Users can proceed to the next step even when the group name is empty or invalid, despite error messages being displayed.

### Steps to Reproduce

1. Access the Mealie registration page
2. **Scenario A (Empty field):** Leave the "Group" field completely empty and click the "Next" button directly. The user is allowed to proceed to the account details step without any error.
3. **Scenario B (Field touched):** Click into the "Group" field, then click out to trigger the `required` validation error. The error message "required" appears, but the "Next" button remains clickable and allows progression.

### Please provide relevant logs

INFO     2026-08-16T02:12:42 - [172.18.0.1:50320] 200 OK "GET /api/validators/user/name?name=user8 HTTP/1.1"

INFO     2026-08-16T02:12:43 - [127.0.0.1:47370] 200 OK "GET /api/app/about HTTP/1.1"

INFO     2026-08-16T02:12:50 - [172.18.0.1:55120] 200 OK "GET /api/validators/user/email?email=user8@gmail.com HTTP/1.1"

INFO     2026-08-16T02:13:14 - [127.0.0.1:44916] 200 OK "GET /api/app/about HTTP/1.1"

INFO     2026-08-16T02:13:15 - [172.18.0.1:48194] 200 OK "GET /api/validators/group?name= HTTP/1.1"

INFO     2026-08-16T02:13:17 - [172.18.0.1:48194] 200 OK "GET /api/validators/user/name?name=user8 HTTP/1.1"

INFO     2026-08-16T02:13:18 - [172.18.0.1:48194] 200 OK "GET /api/validators/user/name?name=user8 HTTP/1.1"

INFO     2026-08-16T02:13:25 - [172.18.0.1:40864] 422 Unprocessable Entity "POST /api/users/register HTTP/1.1"

### Mealie Version

- Mealie Version: `mealie:dev` (commit: 23603cc2, branch: mealie-next)
- Deployment: Docker

### Deployment

[...]

PR #8140, opened 2026-08-19
fix: prevent proceeding when group token is empty

### LINKED ISSUE #8139: [BUG] - Group token field allows empty input to proceed to next step
### First Check

On the "Join a Group" registration page, the Group Token field is marked as required but does NOT block navigation to the next step when left empty. Users can click "Next" and proceed to the account details page without entering a token.

### Steps to Reproduce

1. Go to /register
2. Select "Join a Group"
3. Leave the Group Token field empty
4. Click "Next"
5. Observe that the page advances to the account details step

N/A - Frontend validation issue, no backend logs produced.

### Mealie Version

mealie:dev (commit: 48e551d, branch: mealie-next)
Deployment: Docker (Windows)
Database: SQLite

### Deployment

Docker (Windows)

Related to #8100 (group name validation) but this is a different field - group token validation.

This should be prevented.
Image

This is correct.

Image

Luna-81:

This PR fixes two issues in the "Join a Group" registration flow:

1. **Validation Fix**: Previously, users could click "Next" even when the Group Token field was empty because `safeValidate()` returned a Promise instead of a boolean.

2. **UX Improvement**: The "Next" button now visually reflects the form state — it turns grey (disabled) when the token field is empty, and green (enabled) when the field has content.

**What changed:**
- Made `safeValidate()` async to properly await validation results
- Fixed `validateToken()` to check for empty values
- Made `provideToken.next()` async to block progression when token is empty
- Added `:disabled="!isTokenValid"` to the "Next" button for color feedback

**Testing:**
1. Go to /register
2. Select "Join a Group"
3. Leave Group Token empty → "Next" button is grey and cannot proceed
4. Enter any text in Group Token → "Next" button turns green and can proceed

Testing Result:
image
image

Similar to #8100 (group name validation) and #8233 (registration button colors), but this is for the group token flow.

Fixes #8139

## AI / LLM Assistance

N/A


frontend/app/pages/register/index.vue
```diff
@@ -371,10 +373,11 @@ const initial = {
...
 const domTokenForm = ref<VForm | null>(null);
 function validateToken() {
-  return true;
+  return Boolean(token.value && token.value.trim());
 }
+const isTokenValid = computed(() => validateToken());
 const provideToken = {
-  next: () => {
+  next: async () => {
     if (!safeValidate(domTokenForm as Ref<VForm>)) {
```
michael-genson: `safeValidate` is async now, so this needs an `await`. Without it, the check is on a Promise (always truthy) and never blocks.

frontend/app/pages/register/index.vue
```diff
@@ -327,9 +328,10 @@ const inputAttrs = {
 const i18n = useI18n();
 const isDark = useDark();
 
-function safeValidate(form: Ref<VForm | null>) {
+async function safeValidate(form: Ref<VForm | null>) {
```
michael-genson: There's already an async `safeValidate` in `frontend/app/composables/use-users/user-registration-form.ts` that awaits `validate()` and returns `result.valid`. Could we reuse that instead of keeping a second copy here?

michael-genson (changes requested):

frontend/app/pages/register/index.vue
Luna-81: I've moved all the changes to a new PR: #8455

This PR (#8140) can be closed in favor of #8455, which addresses the review feedback:
- Reused the existing async `safeValidate()` from `useUserRegistrationForm`
- Added `await` before `safeValidate()` in `provideToken.next()`

PR #8145, opened 2026-08-19
fix: strip trailing footnote markers from parsed ingredient names

### LINKED ISSUE #5422: [BUG] - Recipie Parser can't handle prices
### First Check

Some of the websites I use append price data in their ingredients lists and the recipe parsers can't handle that, when it really should be parsed to notes.

> 2 Tbsp cooking oil* ($0.10) 
> 1/4 tsp cayenne pepper ($0.02)

Side note, they also can't handle the asterisk some sites add when they have a note on an ingredient. It should be possible to filter that out of the ingredient name.  

### Steps to Reproduce

Import recipe with priced ingredients

N/A

### Mealie Version

v2.8.0 

### Deployment

Other (please specify below)

Nix

henry1113nz:

Some recipe sites append an asterisk to an ingredient to point at a footnote
elsewhere on the page (`2 Tbsp cooking oil*`). The brute parser keeps that
asterisk in the food name, because it joins the remaining tokens together with
no cleanup step. The NLP parser already handles it.

- `parser_utils/string_utils.py` — adds `remove_footnote_markers()`, which strips
  trailing asterisks. It is anchored to the end of the string, so asterisks
  elsewhere (`A*B`, `5* star anise`) are left alone.
- `brute/process.py` — applies it to the food name at both return points. The
  early return is the single-token case (`salt*`), which would otherwise be missed.

Before / after, same inputs through the brute parser:

| input | food before | food after |
| --- | --- | --- |
| `2 Tbsp cooking oil*` | `cooking oil*` | `cooking oil` |
| `2 Tbsp cooking oil* ($0.10)` | `cooking oil*` | `cooking oil` |
| `1 cup all-purpose flour**` | `all-purpose flour**` | `all-purpose flour` |
| `3 cloves garlic *` | `garlic *` | `garlic` |
| `1 lb chicken breast*, diced` | `chicken breast*` | `chicken breast` |
| `1 cup milk` | `milk` | `milk` |
| `1 tsp salt (optional)` | `salt` | `salt` |

Brute parser on the ingredient debug page, before and after:
image
image

Fixes #5422

The issue reports two problems. The price handling described in it no longer
reproduces — both parsers already put `($0.10)` into the note. Only the asterisk
was still outstanding, so that is all this PR changes. I posted the verification
for both halves on the issue.

On impact, in case it looks larger than it is: this does not cause duplicate
foods. `IngredientFoodModel.normalize()` strips the asterisk, so `cooking oil*`
and `cooking oil` normalise to the same string and match at 100. The problem is
that when the food does not exist yet, the new record is created from the
un-normalised name, leaving the user with a food permanently called
`cooking oil*`.

Only the brute parser is touched. The NLP parser handles these inputs correctly
already, so changing it would add risk for no benefit.

## Testing

Three cases added to the existing parametrised list in `test_brute_parser.py`
(trailing marker, marker alongside a price, and a double marker). They were added
first and confirmed failing against the old behaviour, then confirmed passing
after the change. The other 43 cases in that file are unaffected.

Also verified by hand in a local instance on the ingredient parser debug page,
with the NLP parser as a control.

## AI / LLM Assistance

I picked this issue, read through the existing discussion, and reproduced it in a
local instance before starting. Two earlier commenters had reported that they
could not reproduce it, so I tested both halves separately across both parsers
and posted the results and screenshots on #5422. Working out that the price half
was already fixed, that only the asterisk remained, and that it was specific to
the brute parser was my own conclusion, and it is what set the scope of this PR.

An LLM was then used to help trace the parser code path, to draft the change and
the test cases, and to work out the normalisation and fuzzy-matching behaviour
described above. I reviewed the diff, ran the tests locally, and confirmed the
before/after behaviour in the running instance myself.

### DIFF
--- a/mealie/services/parser_services/brute/process.py
+++ b/mealie/services/parser_services/brute/process.py
@@ -3,7 +3,7 @@ import unicodedata
 
 from pydantic import BaseModel, ConfigDict
 
-from ..parser_utils import check_char, move_parens_to_end
+from ..parser_utils import check_char, move_parens_to_end, remove_footnote_markers
 
 
 class BruteParsedIngredient(BaseModel):
@@ -145,7 +145,7 @@ def parse(ing_str, parser) -> BruteParsedIngredient:
     if len(tokens) == 1:
         ingredient = tokens[0]
         # TODO Refactor to expect BFP to be returned instead of Tuple
-        return BruteParsedIngredient(food=ingredient, note=note, amount=amount, unit=unit)
+        return BruteParsedIngredient(food=remove_footnote_markers(ingredient), note=note, amount=amount, unit=unit)
 
     try:
         # try to parse first argument as amount
@@ -208,4 +208,4 @@ def parse(ing_str, parser) -> BruteParsedIngredient:
     if unit_note not in note:
         note += " " + unit_note
 
-    return BruteParsedIngredient(food=ingredient, note=note, amount=amount, unit=unit)
+    return BruteParsedIngredient(food=remove_footnote_markers(ingredient), note=note, amount=amount, unit=unit)
--- a/mealie/services/parser_services/parser_utils/string_utils.py
+++ b/mealie/services/parser_services/parser_utils/string_utils.py
@@ -3,6 +3,7 @@ from fractions import Fraction
 
 compiled_match = re.compile(r"(.){1,6}\s\((.[^\(\)])+\)\s")
 compiled_search = re.compile(r"\((.[^\(])+\)")
+compiled_footnote = re.compile(r"\s*\*+\s*$")
 
 
 def move_parens_to_end(ing_str) -> str:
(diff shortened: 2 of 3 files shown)


Xyolyp:
Tested this locally. Everything in the description checks out, and the parity argument holds up:

- `tests/unit_tests/services_tests/ingredient_parser/test_brute_parser.py` passes (46 cases) with the three new ones.
- I ran `remove_footnote_markers()` against a set of inputs beyond the added tests. Trailing markers go, everything else stays: `'oil *'` → `'oil'`, `'oil* '` → `'oil'`, while `'A*B'`, `'5* star anise'` and `'salt*!'` are untouched, and `'5 * 5 flour*'` → `'5 * 5 flour'` — so only the anchored end is affected, as documented.
- I verified the claim that the NLP path already handles this rather than taking it on trust: `parse_ingredient("2 Tbsp cooking oil*")` returns the name `cooking oil`, and `"1 cup all-purpose flour**"` returns `all-purpose flour`. So the brute parser really was the odd one out, and this brings the two in line.

One edge case, more a question than an objection: an ingredient consisting solely of markers (`"*"` or `"**"`) now yields an empty food name instead of `"*"`. I'd guess empty is the better of the two, and it's hardly a realistic input — but if an empty food name is handled differently downstream, it might be worth a guard.

Not a maintainer, just gave it a proper run since it had been sitting without a response.

michael-genson (approved):
Thank you!

Issue #4936, opened 2025-01-23
[BUG] - Cannot merge ingredient units while in a shopping list


robobenklein:
### First Check

- [x] This is not a feature request.
- [x] I added a very descriptive title to this issue (title field is above this).
- [x] I used the GitHub search to find a similar issue and didn't find it.
- [x] I searched the Mealie documentation, with the integrated search.
- [x] I already read the docs and didn't find an answer.
- [ ] This issue can be replicated on the demo site (https://demo.mealie.io/).

### What is the issue you are experiencing?

I attempted to merge two units "teaspoon" and "teaspoons" while they were part of a shopping list:

### Steps to Reproduce

- Create recipes with details ingredients & units across households
- Add multiples of the recipes to a shopping list
- Attempt to merge two units being used

### Please provide relevant logs

```
mealie-mealie-1    | ERROR    2025-01-22T23:29:58 - (psycopg2.errors.ForeignKeyViolation) update or delete on table "ingredient_units" violates foreign key constraint "shopping_list_items_unit_id_fkey" on table "shopping_list_items"
mealie-mealie-1    | DETAIL:  Key (id)=(5eb35573-66e2-421e-8b2e-55368c12e0b8) is still referenced from table "shopping_list_items".
mealie-mealie-1    | 
mealie-mealie-1    | 
mealie-mealie-1    | [SQL: DELETE FROM ingredient_units WHERE ingredient_units.id = %(id)s::UUID]
mealie-mealie-1    | [parameters: {'id': '5eb35573-66e2-421e-8b2e-55368c12e0b8'}]
mealie-mealie-1    | (Background on this error at: https://sqlalche.me/e/20/gkpj)
mealie-mealie-1    | ERROR    2025-01-22T23:29:58 - (psycopg2.errors.ForeignKeyViolation) update or delete on table "ingredient_units" violates foreign key constraint "shopping_list_items_unit_id_fkey" on table "shopping_list_items"
mealie-mealie-1    | DETAIL:  Key (id)=(5eb35573-66e2-421e-8b2e-55368c12e0b8) is still referenced from table "shopping_list_items".
mealie-mealie-1    | 
mealie-mealie-1    | 
mealie-mealie-1    | [SQL: DELETE FROM ingredient_units WHERE ingredient_units.id = %(id)s::UUID]
mealie-mealie-1    | [parameters: {'id': '5eb35573-66e2-421e-8b2e-55368c12e0b8'}]
[...]

Issue #3624, opened 2024-05-20
[BUG] - Merging ingredient doesnt work if referenced by a shopping list


salco:
### First Check

- [X] This is not a feature request.
- [X] I added a very descriptive title to this issue (title field is above this).
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Mealie documentation, with the integrated search.
- [X] I already read the docs and didn't find an answer.
- [X] This issue can be replicated on the demo site (https://demo.mealie.io/).

### What is the issue you are experiencing?

If you want to merge an ingredient that is already used inside a shopping list nothing is show to the user and it doesnt work.  
In reality we get a http response: 500 when we look at the network log.  
Also if we look at the database (for me it's postgres) you see that the error is:
```
[21915] ERROR:  update or delete on table "ingredient_foods" violates foreign key constraint "shopping_list_items_food_id_fkey" on table "shopping_list_items"
```

To confirm:
I've deleted the shopping list in question and retry the merge and all went well.

### Steps to Reproduce

1. create `tomayto` and `tomato`
2. create 2 recipes, one using `tomayto` the other using `tomato`
3. create a new shopping list: I give you the freedom to find a funny name 😉 
4. use the 2 recipes to add all ingredient to the shopping list
5. then go to the the option or with the api call direcly `api/foods/merge` and merge `tomayto` into `tomato`

Voila you should get the response 500 in the browser logs.

### Please provide relevant logs

```
2024-05-19 21:46:29.354 EDT [21915] ERROR:  update or delete on table "ingredient_foods" violates foreign key constraint "shopping_list_items_food_id_fkey" on table "shopping_list_items"
2024-05-19 21:46:29.354 EDT [21915] DETAIL:  Key (id)=(※※※※※※※※-※※※※-※※※※-※※※※-※※※※※※※※※※※※) is still referenced from table "shopping_list_items".
2024-05-19 21:46:29.354 EDT [21915] STATEMENT:  DELETE FROM ingredient_foods WHERE ingredient_foods.id = '※※※※※※※※-※※※※-※※※※-※※※※-※※※※※※※※※※※※'::UUID
```

### Mealie Version

1.6.0

### Deployment

Docker (Linux)

### Additional Deployment Details

**Details**
Version: v1.6.0
Build: 418a8ec72bead3f625827846aafd644c851a4fcc
Mode de l'application: Production
Mode démo: Non démo
Port de l'API: 9000
Documentation de l'API: Activé
[...]

PR #8146, opened 2026-08-19
fix: repoint shopping list items when merging a food or unit

henry1113nz:

Merging a food that a shopping list item points at fails. `RepositoryFood.merge()`
moves the recipe ingredients from the old food to the new one and then deletes the
old food, but shopping list items reference the food through their own `food_id`
column rather than through that relationship, so nothing repoints them first.

The symptom depends on the database:

- **PostgreSQL** (as reported in the issue): the foreign key is enforced, the
  delete raises `shopping_list_items_food_id_fkey`, and the route returns HTTP 500
  with no feedback in the UI.
- **SQLite**: foreign keys are not enforced by default, so the delete succeeds and
  the shopping list item is left pointing at a food that no longer exists. The item
  silently shows no food, which is harder to notice than the 500.

Changes:

- `repos/repository_foods.py` — repoints shopping list items from the old food to
  the new one before the delete.
- `repos/repository_units.py` — the same fix for units. `RepositoryUnit.merge()`
  contains identical code and shopping list items also carry a `unit_id`, so this
  is the case reported in #4936, which was marked as a duplicate of this issue.

This is what the reporter asked for in the thread — repointing the reference rather
than blocking the merge.

Addresses #3624 (and #4936, marked as a duplicate of it).

I have deliberately not used a closing keyword, because the thread also covers
deleting a food, which this PR does not change. See the note below.

**The deletion case is not covered here, and I would like your call on it before
adding it.** @ChrisLane raised deletion in the thread, and your reply was:

> Ideally when deleting foods, existing shopping list items are either deleted, or
> their food is converted to a note before deletion. Ingredients should probably do
> this too as long as we're adding custom food deletion logic.

I confirmed plain deletion has the same defect — on SQLite the shopping list item
survives with a `food_id` pointing at a row that no longer exists. But the two
options you listed lead to different user-visible behaviour, and picking one is a
product decision rather than a bug fix, so I did not want to guess. If you tell me
which you prefer I am happy to add it to this PR.

One implementation note: I used an explicit `UPDATE` rather than adding a
`shopping_list_items` relationship to the food and unit models, to keep the change
local to the merge methods.

## Testing

A regression test was added next to each existing merge test:
`test_food_merger_with_shopping_list_reference` and
`test_unit_merger_with_shopping_list_reference`. Both create a shopping list item
that references the food/unit being merged away, then assert the item points at the
surviving record. Both were confirmed failing against the old behaviour before the
fix was applied.

The existing `test_food_merger` and `test_unit_merger` only exercise recipe
ingredient references, which is why this went unnoticed.

Full `tests/unit_tests/repository_tests/` (156 tests) and the shopping list
integration suites (50 tests) pass.

## AI / LLM Assistance

I picked this issue and reproduced it locally before starting, including confirming
that the deletion path described in the comments has the same defect. An LLM was
used to help trace which tables reference `ingredient_foods`, to draft the fix and
the regression tests, and to check whether units were affected in the same way. I
reviewed the diff and ran the tests locally, including confirming both new tests
failed before the change and passed after it.

### DIFF
--- a/mealie/repos/repository_foods.py
+++ b/mealie/repos/repository_foods.py
@@ -1,6 +1,7 @@
 from pydantic import UUID4
-from sqlalchemy import select
+from sqlalchemy import select, update
 
+from mealie.db.models.household.shopping_list import ShoppingListItem
 from mealie.db.models.recipe.ingredient import IngredientFoodModel
 from mealie.schema.recipe.recipe_ingredient import IngredientFood
 
@@ -17,6 +18,13 @@ class RepositoryFood(GroupRepositoryGeneric[IngredientFood, IngredientFoodModel]
         to_model = self._get_food(to_food)
 
         to_model.ingredients += from_model.ingredients
+
+        # Shopping list items reference the food directly rather than through the ingredients
+        # relationship, so they have to be repointed explicitly. Without this the delete below
+        # either violates a foreign key constraint or leaves the item pointing at a missing food.
+        self.session.execute(
+            update(ShoppingListItem).where(ShoppingListItem.food_id == from_food).values(food_id=to_food)
+        )
 
         try:
             self.session.delete(from_model)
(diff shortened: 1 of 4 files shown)


michael-genson (approved):
Thank you!

PR #8150, opened 2026-08-19
fix: use oe instead of œ in French food names

seb49:

- Updates `mealie/repos/seed/resources/foods/locales/fr-FR.json`.
- Replaces the `œ` ligature with `oe` in French food names.
- This makes the seeded French food names easier to search and enter in everyday use.

For example:

- `œuf` → `oeuf`
- `œufs` → `oeufs`

While `œ` is linguistically correct in French, `oe` is commonly used in practice when entering food names, especially when users are typing on a standard keyboard without an easy way to enter the ligature.

This change is limited to the French food seed data and does not modify the seeding logic.

No issue currently exists for this change.

This is intentionally a small and targeted change.

The goal is not to change the French spelling, but to improve consistency with how users commonly type and search for these food names.

## Testing

- Verified that the modified file remains valid JSON.
- Reviewed the resulting diff to ensure that only the intended French food names were changed.
- Tested the French food seed data in Mealie.

## AI / LLM Assistance

An LLM was used to assist with preparing this pull request description and reviewing the proposed change.

The code/data change itself was manually made in `fr-FR.json`.


michael-genson:
Translations are handled through [Crowdin](https://crowdin.com/project/mealie), so you'll want to make updates there. They will then sync back here!

Issue #5680, opened 2025-07-11
[BUG] - Food Labels not translating


Matthijsy:
### First Check

- [x] This is not a feature request.
- [x] I added a very descriptive title to this issue (title field is above this).
- [x] I used the GitHub search to find a similar issue and didn't find it.
- [x] I searched the Mealie documentation, with the integrated search.
- [x] I already read the docs and didn't find an answer.
- [x] This issue can be replicated on the demo site (https://demo.mealie.io/).

### What is the issue you are experiencing?

In the nightly version / v3.0 the labels are now determined based on the keys of the food seed files. However, these keys are always in English, and as far as I can see in Crowdin also not part of the translation process. This results in the labels always being created in English. 

### Steps to Reproduce

1. Set the Language to something different than English (Dutch for example)
2. Go to Data Management
3. Go to Labels
4. Click Seed

I would expect to have the labels in my local language, but they are always in English. 

### Please provide relevant logs

No Logs

### Deployment

Docker (Linux)

Fjodor42:
If someone could point me to the relevant parts of the code, I would be happy to attempt to fix this, as I was about the report the issue myself, before finding this bug report.

To that end, I imagine some sort of solution where labels are still generated from the keys in the localisation file for foods, but that generation being done by something (probably at build) that goes through said file and checks if the resulting labels are present in the localisation file for, well, labels, adding them if not. That should present them in Crowdin for translation.

Then, obviously, the UI should pick up localisation from that file, instead of what I presume is currently happening, namely pulling (some of) the labels from the keys in the food localisation file.

Does that sound reasonable?

Goldi95:
Maybe it helps: startet today with mealie with my first installation and also noticed the bug.
I am running the latest version and noticed that while **32** labels are imported into the database during the seed process, the Crowdin project currently only lists **20** labels for translation:

<img width="1224" height="687" alt="Image" src="https://github.com/user-attachments/assets/f9857b17-07d5-432f-b332-613fb8f17b05" />

Crowdin reference: https://crowdin.com/editor/mealie/92/en-de

Hopefully, this helps in syncing the seed data keys with the Crowdin source files!

gurkburk76:
Also, the swedish labels seem to be translated but they still show up as english, so a free bump on this i guess.
https://crowdin.com/editor/mealie/92/en-sv?view=side-by-side&filter=basic&value=0

notDavid:
[...]

PR #7429, opened 2026-04-08
fix: seed labels before foods in setup wizard to prevent race condition

Xenov-X:
## What this PR does / why we need it:
                                                                                                                                                          
During the first-time setup wizard, when "Use Seed Data" is enabled, many foods are created without their expected label associations. For example, `canned potato` gets labelled `Canned Food`, but `potato` and `new potato` get no label at all — despite all being defined with labels in the seed data file. This happens regardless of locale.                                                                  
                                                                  
In `frontend/app/pages/admin/setup.vue`, `seedData()` fires all three seeders concurrently via `Promise.all`. The food seeder (`IngredientFoodsSeeder.load_data` in `mealie/repos/seed/seeders.py`) looks up each label by name in the database:                                 

```                                                                                                                                                                                                         
  for label, values in self.load_file(file).items():              
      label_out = self.get_label(label)  # returns None if label doesn't exist yet                                                                                                                       
```                                                                                                                                                                     

Since seedFoods() and seedLabels() run in parallel, whichever labels happen to have been committed by the time the food seeder reaches a given category will be found; the rest return None. The result is non-deterministic: which foods get labels depends entirely on timing.                                                                                                                              

### Manual workaround
Skip the wizard's seed data option. Manually seed via Admin > Site Settings > Data Management — seed Labels first, then Foods. When done sequentially, all foods receive their correct labels.

### Fix
`seedLabels()` now completes before `seedFoods()` begins. `seedUnits()` still runs in parallel with `seedFoods()` since they are independent.       

## Which issue(s) this PR fixes:

Fixes https://github.com/mealie-recipes/mealie/issues/5680

## Testing
  - Fresh Docker deploy, run through setup wizard with "Use Seed Data" enabled                                                                                                                           
  - Verified all seeded foods have their expected labels in Data Management > Foods
  - Verified seeding still works when done manually via Admin > Data Management   

## AI / LLM Assistance

- Used to assist typing up PR
- Used to track down issue root

###  Potential improvements
JS is not my wheelhouse - there might be a better way to do this?

PR #8170, opened 2026-08-20
fix: registration with seed data returns a 500

justadityaraj:

Signing up with **Seed Data** ticked returns a 500, and the new account ends up with no seeded foods or units.

Two separate defects on that path:

**1. The seeder gets unscoped repositories.** Registration is unauthenticated, so `mealie/routes/users/registration.py:32` deliberately builds repos with `group_id=None`. Those were passed straight into `SeederService`, and `IngredientFoodsSeeder.load_data` reads `self.repos.group_id` into `SaveIngredientFood.group_id`, which is a required `UUID4`:

```
mealie/repos/seed/seeders.py:116: ValidationError: 1 validation error for SaveIngredientFood
group_id
  UUID input should be a string, bytes or UUID object [type=uuid_type, input_value=None, input_type=NoneType]
```

It escapes as a 500 rather than being swallowed, because the `try/except` in `seed()` wraps the `create()` call, while the failure happens inside the `load_data` generator as the loop advances it.

Fixed by scoping the repos to the group that was just created. That is the same thing `_fetch_or_register_new_household` already does 50 lines above (`registration_service.py:63`), so this call site was simply the one that was missed.

**2. Foods were seeded before labels.** Foods resolve their label by name as they are built, so with no labels present yet, every seeded food was written with `label_id = None`. Reordering to labels-first fixes it.

That is the same ordering `frontend/app/pages/admin/setup.vue` already uses after #7429 (which fixed issue #5680). That PR only touched the setup wizard, so the registration path kept the old order.

`git blame` puts the first defect at eb170cc7 (#3970, Households), which changed `SeederService(self.repos, user, group)` to `SeederService(self.repos)`, moving the group onto `repos.group_id` without scoping the repos here. It went unnoticed because `seed_data` had no test coverage at all.

## Testing

Reproduced the reporter's exact traceback locally, then confirmed the fix:

- Before: `POST /api/users/register` with `seedData: true` → 500, `ValidationError ... input_value=None` at `seeders.py:116`, preceded by `Seeding Ingredient Foods`.
- After: 201, with 2687 foods, 24 units, and every food carrying a label.

I also checked that the second half is doing real work: with only the scoping fix applied, registration succeeds but `assert all(food.label_id is not None for food in foods)` fails for all 2687 foods.

Added `test_user_registration_with_seed_data` to `tests/integration_tests/user_group_tests/test_group_registration.py`. It fails on `mealie-next` with the reporter's ValidationError and passes with this change, and the label assertion locks in the ordering half. The counts reuse the constants already asserted in `test_group_seeder.py`.

`tests/integration_tests/user_group_tests` + `tests/integration_tests/user_tests`: **72 passed, 9 skipped** (the skips are the LDAP tests that need the CI service). `ruff check`, `ruff format --check` and `mypy` are clean on both changed files.

The labels-before-foods reorder is technically a second change. I kept it in because a user who signs up with seed data still gets an unusable food list without it, and #7429 already established the ordering. Happy to split it out if you would rather have them separate — it is one line.

Scoped to `group_id` only, matching `_fetch_or_register_new_household`, since foods, units and labels are all group-level.

## AI / LLM Assistance

Claude Code, used throughout: investigating the traceback, locating the root cause, writing the fix and the test, and drafting this description. I reproduced the bug and ran the tests, lint and type checks locally myself, and verified each claim above against the source rather than taking the tooling's word for it.

### DIFF
--- a/tests/integration_tests/user_group_tests/test_group_registration.py
+++ b/tests/integration_tests/user_group_tests/test_group_registration.py
@@ -2,6 +2,7 @@ from fastapi.testclient import TestClient
 
 from mealie.repos.all_repositories import get_repositories
 from mealie.repos.repository_factory import AllRepositories
+from mealie.repos.seed.seeders import IngredientFoodsSeeder, IngredientUnitsSeeder
 from mealie.schema.response.pagination import PaginationQuery
 from tests.utils import api_routes
 from tests.utils.factories import user_registration_factory
@@ -52,9 +53,6 @@ def test_new_user_group_permissions(api_client: TestClient):
 
 
 def test_user_registration_with_seed_data(api_client: TestClient, unfiltered_database: AllRepositories):
-    CREATED_FOODS = 2687
-    CREATED_UNITS = 24
-
     registration = user_registration_factory()
     registration.seed_data = True
 
@@ -67,6 +65,16 @@ def test_user_registration_with_seed_data(api_client: TestClient, unfiltered_dat
     foods = group_repos.ingredient_foods.page_all(PaginationQuery(page=1, per_page=-1)).items
     units = group_repos.ingredient_units.page_all(PaginationQuery(page=1, per_page=-1)).items
 
-    assert len(foods) == CREATED_FOODS
-    assert len(units) == CREATED_UNITS
+    seeded_food_names = {
+        attributes["name"]
+        for label in IngredientFoodsSeeder.load_file(IngredientFoodsSeeder.get_file(registration.locale)).values()
+        for attributes in label["foods"].values()
+    }
+    seeded_unit_names = {
+        unit["name"]
+        for unit in IngredientUnitsSeeder.load_file(IngredientUnitsSeeder.get_file(registration.locale)).values()
+    }
+
+    assert {food.name for food in foods} == seeded_food_names
+    assert {unit.name for unit in units} == seeded_unit_names
     assert all(food.label_id is not None for food in foods)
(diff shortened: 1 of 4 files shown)


tests/integration_tests/user_group_tests/test_group_registration.py
```diff
@@ -46,3 +49,24 @@ def test_new_user_group_permissions(api_client: TestClient):
     assert user.get("canInvite") is True
     assert user.get("canManage") is True
     assert user.get("canOrganize") is True
+
+
+def test_user_registration_with_seed_data(api_client: TestClient, unfiltered_database: AllRepositories):
+    CREATED_FOODS = 2687
+    CREATED_UNITS = 24
```
michael-genson: Can we calculate this, or otherwise not rely on it being a constant? We used to have something like this but removed it because it was a pain to keep updating.

justadityaraj: Done, the same locale files the seeders read (`IngredientFoodsSeeder.get_file` / `load_file`) and compares name sets rather than counts, so it tracks the resources on its own and tells you which food went missing when it does break.

michael-genson: Ah I thought we removed those. If you wouldn't mind doing that in this PR I'd appreciate it!

justadityaraj: Done, below are the results.

```
The reading moved to `tests/utils/seed_data.py` as `seeded_food_names` / `seeded_unit_names` / `seeded_label_names`. Labels resolve through `MultiPurposeLabelSeeder.get_file`, so they come out of the foods file the same way the seeder itself gets them.

Side benefit of comparing name sets instead of counts: I injected a regression to check the assertions still bite, and the failure now reads

E         Extra items in the right set:
 E         'pound'

instead of `23 != 24`. Verified the helpers return exactly the old numbers first, and that an unknown locale still falls back to en-US. Seven tests green across both files.
```

michael-genson (approved):
Thanks!

PR #8172, opened 2026-08-21
fix: show progress on ingredient parser review buttons

Xyolyp:

The four buttons in the parser review step (create the missing unit or food, add the parsed text
as an alias) each do an API round trip with no feedback at all. On a slow connection nothing
appears to happen and you click again.

Uses the `state.loading` object the dialog already has. Unit and food are tracked separately,
since one ingredient can be missing both and only the button you pressed should react.

None that I could find. Noticed while using the parser on a German recipe library.

## AI / LLM Assistance

The change was written with Claude Code against a running dev instance. I reviewed it and checked
the button states before opening.

## Release Notes

The buttons for creating a missing unit or food, and for adding a parsed name as an alias, now
show a progress indicator while the change is saved.

### DIFF
--- a/frontend/app/components/Domain/Recipe/RecipePage/RecipePageParts/RecipeParseDialog/ParseDialogParse.vue
+++ b/frontend/app/components/Domain/Recipe/RecipePage/RecipePageParts/RecipeParseDialog/ParseDialogParse.vue
@@ -46,6 +46,7 @@
         :icon="$globals.icons.units"
         color="warning"
         size="small"
+        :loading="state.loading.unit"
         @click="createMissingUnit"
       >
         {{ $t("recipe.parser.missing-unit", { unit: currentMissingUnit }) }}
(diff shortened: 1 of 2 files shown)


miah120:
Ooh I like the extra feedback here. Imo you could probably drop the disabled prop since vuetify buttons already get disabled while loading, and the buttons get really hard to see when disabled.

Xyolyp:
Good catch, dropped in 8bb1def — thanks.

Checked it against the pinned Vuetify (4.1.2) to be sure the guard wasn't doing anything: `isDisabled` is only `group?.disabled.value || props.disabled`, so `loading` really doesn't feed into it. It does everything else though — `.v-btn--loading { pointer-events: none }`, `tabindex="-1"`, `aria-busy="true"` — so the button can't be activated by mouse or keyboard while the request is in flight, and `aria-busy` is arguably the better semantic here than `aria-disabled` anyway. `:disabled` was only buying the dimmed styling you didn't want.

`:loading` bindings stay as they were. Lint clean, 267 frontend tests pass.

michael-genson (approved):
LGTM

PR #8173, opened 2026-08-21
fix: don't default to the English-only parser in other languages

Xyolyp:

The Natural Language Processor only has an English model (`SUPPORTED_LANGUAGES == ["en"]`). On a
recipe in another language it still gets the amount right, because digits are language
independent, but the unit never gets tagged and ends up inside the food name.

I ran 16 German ingredient lines from two recipes I had been experimenting with through
`/api/parser/ingredient`, with German units seeded:

| input | Natural Language Processor | Brute Parser |
| --- | --- | --- |
| `1/2 Dose gehackte Tomaten` | 0.5 · — · `Dosen gehackte Tomaten` | 0.5 · Dose · `gehackte Tomaten` |
| `2 EL Olivenöl` | 2 · — · `EL Olivenöl` | 2 · Esslöffel · `Olivenöl` |
| `1 TL Salz` | 1 · — · `TL Salz` | 1 · Teelöffel · `Salz` |
| `1 Päckchen Backpulver` | 1 · — · `Päckchen Backpulver` | 1 · Packung · `Backpulver` |
| `1 Bund Petersilie` | 1 · — · `Bund Petersilie` | 1 · Bund · `Petersilie` |

4 of 16 units recognised by the NLP parser (only g, ml, kg, which also occur in English recipes)
against 14 of 16 for the brute parser, which resolves abbreviations through the database.

Everyone still starts on the NLP parser, and its name makes it sound like the better of the two.

`useParsingPreferences()` now takes a default that only applies when nothing is stored, and the
dialog passes `brute` outside English locales. A stored preference is never overridden.

The dialog also shows a hint when the NLP parser is selected in a non-English locale. That part
is what reaches existing users: `mergeDefaults: true` persists the default on first use, so
anyone who has opened the dialog already has `{"parser":"nlp"}` stored and would never see a
changed default.

One new `en-US` string, no other locales touched.

None directly. Related: #7108, where a maintainer confirms the English-only limitation and
recommends the brute or OpenAI parser. This makes the product say that instead of relying on
users finding the issue.

Locale detection is `locale.startsWith("en")`.

I left the backend default alone, since the server has no reliable notion of the caller's
language. Happy to move it if you'd rather decide this server-side.

## Testing

| locale | stored preference | preselected | hint |
| --- | --- | --- | --- |
| `de-DE` | none | Brute Parser | no |
| `en-US` | none | Natural Language Processor | no |
| `de-DE` | `nlp` | Natural Language Processor | yes |

`pnpm lint` and `pnpm vitest run` pass.

## AI / LLM Assistance

The measurements above, the change and the tests were produced with Claude Code against a running
dev instance. The parser table is reproducible via `/api/parser/ingredient`.

## Release Notes

Outside English, the ingredient parser no longer defaults to the Natural Language Processor,
which only understands English recipes. If it is selected anyway, the dialog explains why that
may not work.

### DIFF
--- a/frontend/app/components/Domain/Recipe/RecipePage/RecipePageParts/RecipeParseDialog/ParseDialogChangeParser.vue
+++ b/frontend/app/components/Domain/Recipe/RecipePage/RecipePageParts/RecipeParseDialog/ParseDialogChangeParser.vue
@@ -14,6 +14,15 @@
       </BaseButton>
     </v-card-actions>
   </v-card>
+  <v-alert
+    v-if="showNlpLanguageHint"
+    type="info"
+    variant="tonal"
+    density="compact"
+    class="mt-3 text-body-2"
+  >
+    {{ $t("recipe.parser.natural-language-processor-english-only") }}
+  </v-alert>
   <BaseDialog
     v-model="open"
     bottom-sheet
@@ -41,7 +50,7 @@
 import type { MenuItem } from "~/components/global/BaseOverflowButton.vue";
 import type { Parser } from "~/lib/api/user/recipes/recipe";
 
-defineProps<{ availableParsers: MenuItem[] }>();
+defineProps<{ availableParsers: MenuItem[]; showNlpLanguageHint: boolean }>();
 const emit = defineEmits<{ parse: [] }>();
 const currentParser = defineModel<Parser>({ default: "nlp" });
 
--- a/frontend/app/composables/recipes/use-parse-ingredients-dialog.ts
+++ b/frontend/app/composables/recipes/use-parse-ingredients-dialog.ts
@@ -29,8 +29,12 @@ export function useParseIngredientsDialog(
   const foodStore = useFoodStore();
   const foodData = useFoodData();
 
-  const parserPreferences = useParsingPreferences();
+  // The natural language parser is trained on English recipes, so it isn't a sensible default for
+  // other languages: it recognises the quantity but leaves the unit in the food name.
+  const isEnglishLocale = computed(() => (i18n.locale.value || "").toLowerCase().startsWith("en"));
+  const parserPreferences = useParsingPreferences(isEnglishLocale.value ? "nlp" : "brute");
   const parser = ref<Parser>(parserPreferences.value.parser || "nlp");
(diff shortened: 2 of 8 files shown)


frontend/app/components/Domain/Recipe/RecipePage/RecipePageParts/RecipeParseDialog/ParseDialogChangeParser.vue
```diff
@@ -40,7 +49,7 @@
 import type { MenuItem } from "~/components/global/BaseOverflowButton.vue";
 import type { Parser } from "~/lib/api/user/recipes/recipe";
 
-defineProps<{ availableParsers: MenuItem[] }>();
+defineProps<{ availableParsers: MenuItem[]; showNlpLanguageHint: boolean }>();
```
michael-genson, line 52: `ParseDialogParse.vue:2` renders this component too and doesn't pass `showNlpLanguageHint`. We should pass it in there.

frontend/app/composables/recipes/use-parse-ingredients-dialog.ts
```diff
@@ -29,8 +29,12 @@ export function useParseIngredientsDialog(
   const foodStore = useFoodStore();
   const foodData = useFoodData();
 
-  const parserPreferences = useParsingPreferences();
+  // The natural language parser is trained on English recipes, so it isn't a sensible default for
+  // other languages: it recognises the quantity but leaves the unit in the food name.
+  const isEnglishLocale = computed(() => (i18n.locale.value || "").toLowerCase().startsWith("en"));
+  const parserPreferences = useParsingPreferences(isEnglishLocale.value ? "nlp" : "brute");
```
michael-genson, line 35: Can we add tests to `use-parse-ingredients-dialog.test.ts`? Its `useParsingPreferences` mock drops the argument and the setup locale is `en-US`, so nothing exercises the new default or the hint.

michael-genson (changes requested):
Feature looks good, two things:

Xyolyp:
@michael-genson I've adressed your remarks, let me know if I took a valid approach.

michael-genson (approved):
Thank you!

PR #8176, opened 2026-08-22
feat: add pantry status to recipes

CharlesDay:

  Adds a pantry-status card to recipe pages so users can quickly see whether the recipe’s structured
  ingredients are marked as on hand for the current household.

  - Adds `RecipePantryStatus`, which lists on-hand and missing structured foods and explains when ingredients have not been structured yet.
  - Reuses existing “Add to Shopping List” dialog to add only missing structured ingredients.
  - Supports referenced/sub-recipes and avoids recursive reference loops.
  - Adds a small opt-in filter to the existing shopping-list dialog; its normal behavior is unchanged.
  - Adds `en-US` translation keys only; other locales remain managed by Crowdin.
  - Does not use AI or require a database migration.

  This checks whether a food is marked on hand, not whether the household has a sufficient quantity.

  No existing issue, this is a new feature proposal.

  The main behavior to review is the distinction between structured foods and unstructured ingredient
  text. Only structured foods can be reliably compared with the household’s on-hand list. (using the ing parser is the best friend here)

  ## Testing

  - Added unit tests for pantry-status calculation, including referenced recipes, duplicate foods,
  and unstructured ingredients.
  - Ran ESLint successfully.
  - Ran the full Vitest suite successfully: 28 files, 269 tests.
  - Hosted this on my self-hosted with the below screenshots of the feature

Pantry Check ui element/look (ignore the version history, that is another feature I am using locally)
Screenshot 2026-08-22 at 1 42 22 AM

Opened "Add missing to shopping list" modal
Screenshot 2026-08-22 at 1 42 40 AM

Resulting additions to the grocery list
Screenshot 2026-08-22 at 1 41 33 AM

  ## AI / LLM Assistance
  
  This PR was developed with OpenAI Codex assistance. Codex helped implement the feature, add tests
  and translations, and run local validation. The feature behavior, scope, and final changes were
  directed and reviewed by the contributor by deploying to local self hosted using the same build workflows as upstream.


miah120:
Oh this looks neat! 

If I could suggest a few UI tweaks tho, here are my thoughts: 
1. Use the shopping cart icon instead of a warning icon and switch from "missing" to "needed". While this is important for users to see (particularly when meal planning), mealie doesn't actually know what's in your pantry or not. So we don't know that the items are missing (and therefore needing a warning) just that they're needed for the recipe beyond what you already keep on hand. 
3. Remove the logic around structured foods vs unstructured foods. If there's an unstructured food, I think it's best to assume the user needs it. If they don't, they can always uncheck it before adding it to the shopping list, and if that's too much work they can add it as a food and mark it on hand. 
4. hide the card in cook mode. If you've started cooking this recipe and haven't bought all the ingredients you need, there's nothing mealie can do to save you 😔
5. it might be good to make it dismissable. I could see this being a bit of a nuisance if the user does actually have the ingredients ready. 
6. It could also be nice to hide the on-hand chip if no items are on hand.

As a bit of an aside, I considered suggesting making the card flat to match the ingredients/organizers sections, but given that this is actionable where the others aren't I think the extra emphasis is warranted. 

But I'm not a maintainer or anything, I'm just big on frontend dev lol 

Also thanks for writing up tests for this!

CharlesDay:
> Oh this looks neat!
> 
> If I could suggest a few UI tweaks tho, here are my thoughts:
> 
> 1. Use the shopping cart icon instead of a warning icon and switch from "missing" to "needed". While this is important for users to see (particularly when meal planning), mealie doesn't actually know what's in your pantry or not. So we don't know that the items are missing (and therefore needing a warning) just that they're needed for the recipe beyond what you already keep on hand.
> 2. Remove the logic around structured foods vs unstructured foods. If there's an unstructured food, I think it's best to assume the user needs it. If they don't, they can always uncheck it before adding it to the shopping list, and if that's too much work they can add it as a food and mark it on hand.
> 3. hide the card in cook mode. If you've started cooking this recipe and haven't bought all the ingredients you need, there's nothing mealie can do to save you 😔
> 4. it might be good to make it dismissable. I could see this being a bit of a nuisance if the user does actually have the ingredients ready.
> 5. It could also be nice to hide the on-hand chip if no items are on hand.
> 
> As a bit of an aside, I considered suggesting making the card flat to match the ingredients/organizers sections, but given that this is actionable where the others aren't I think the extra emphasis is warranted.
> 
> But I'm not a maintainer or anything, I'm just big on frontend dev lol
> 
> Also thanks for writing up tests for this!

Nice suggestions, I have applied the suggestions. See below screenshots. It now shows needed instead of missing, it is dismissible, disappears during cook mode, and removed the unstructured. If all ingredients are unstructured though it will not know what is on hand/needed and thus all are marked as needed. The items must be structures for the on hand to show correctly. Also if none are on hand then the chip does not show. Also shows the shopping icon rather than the warning

Screenshot 2026-08-29 at 1 44 25 AM


Screenshot 2026-08-29 at 1 43 15 AM

noxhirsch:
Our "workflow" is that we create a meal plan once a week for the entire week, add the ingredients from the meal plan view (where I can already see what I have and what I need) to the shopping list, and then buy everything. By the time we open the recipes, we already have everything we need, so the pantry check is useless to us because it warns us that we're missing ingredients we've already bought exactly for this recipe.

For us, this feature would only make sense if we tracked every ingredient in the pantry, including fresh ingredients. That kind of detailed inventory tracking feels more like a Grocy-style workflow, though. We only mark things like spices or some canned foods as "on hand". In our case, there are hardly any recipes that we can cook without shopping first, so there's no need to warn me with a big and colorful box above the ingredients list that I still need to buy things. I also don't want to dismiss it every time. 

So I think it would be helpful to be able to disable it completely, or at least make it much less prominent - for example, by placing it below the ingredients list and/or turning it into a relatively neutral, compact button that opens the "Add to Shopping List" dialog (e.g. a shopping cart icon with a missing-ingredient counter next to the copy icon).

Quick mockup (if disabling isn't an option):
image
I'm just a regular user expressing my personal opinion - not a maintainer ;)

michael-genson:
Hey there, thanks for the PR! I'm going to go ahead and close this one. The "Add to Shopping List" dialog already unchecks ingredients that are marked on hand, so most of this is already covered there, and as noted above a card on every recipe page is pretty noisy for households that only track a few staples as on hand.

PR #8183, opened 2026-08-23
feat: reverse proxy auth

danepowell:

Adds support for auto-login via reverse proxy auth

Fixes https://github.com/mealie-recipes/mealie/discussions/801

Thanks a lot for any and all feedback! This is my first contribution to Mealie. Let me know if you need help with anything. :heart: 

## Testing

As the documentation says, set the new `REVERSE_PROXY_AUTH_*` headers to match your reverse proxy and authentication provider (e.g., Traefik and Authelia). Simply click the Login button and you should immediately be logged in.

I've tested this locally and it works great.

## AI / LLM Assistance

I provided a high-level architecture to Copilot as discussed in https://github.com/mealie-recipes/mealie/discussions/801, Copilot handled the implementation, I reviewed it for accuracy. I think this is a good, clean, minimal implementation that aligns well with existing auth methods.


mealie/alembic/versions/2026-09-01-00.00.00_f3a8b6c1d2e4_add_reverse_proxy_auth_method.py
```diff
@@ -0,0 +1,28 @@
...
+Revises: 69e942bab3aa
+Create Date: 2026-09-01 00:00:00.000000
+
+"""
+
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "f3a8b6c1d2e4"
+down_revision = "69e942bab3aa"
```
michael-genson, line 13: `mealie-next` already has a migration on top of `69e942bab3aa`, so this will create a second Alembic head. Can you rebase and point this at the current head?

mealie/routes/auth/auth.py
```diff
@@ -298,6 +299,30 @@ async def oauth_native_token(
...
+async def reverse_proxy_login(
+    request: Request,
+    response: Response,
+    session: Session = Depends(generate_session),
+):
+    """Authenticate a user using a username forwarded by a trusted reverse proxy header"""
+    if not settings.REVERSE_PROXY_AUTH_READY:
+        raise HTTPException(status_code=status.HTTP_404_NOT_FOUND)
+
+    username = request.headers.get(settings.REVERSE_PROXY_AUTH_HEADER)
```
michael-genson, line 312: Right now anyone who can reach Mealie directly can set this header. Could we add an optional setting (e.g. `REVERSE_PROXY_AUTH_TRUSTED_IPS`) that, when set, rejects requests whose `request.client.host` isn't in the list? It can default to unset so it doesn't complicate simple setups.

mealie/routes/admin/admin_about.py
```diff
@@ -61,4 +62,6 @@ def check_app_config(self):
             is_up_to_date=APP_VERSION == "develop" or APP_VERSION == "nightly" or get_latest_version() == APP_VERSION,
             oidc_ready=settings.OIDC_READY,
             oidc_disabled=not settings.OIDC_AUTH_ENABLED,
+            reverse_proxy_auth_ready=settings.REVERSE_PROXY_AUTH_READY,
```
michael-genson, line 65: These aren't shown anywhere in `site-settings.vue`. Could we add a status card like the ones for LDAP/OIDC, or drop these fields?

michael-genson (changes requested):

mealie/routes/auth/auth.py
peterdey: One thing to note: this will fail if HOST_IP is set.  Uvicorn will rewrite `request.client.host` to `X-Forwarded-For` header (i.e. the client IP), and you lose the gateway IP.

mealie/core/settings/settings.py
```diff
@@ -431,6 +431,29 @@ def OIDC_READY(self) -> bool:
         """Validates OIDC settings are all set"""
         return self.OIDC_FEATURE.enabled
 
+    # ===============================================
+    # Reverse Proxy Auth Configuration
+    REVERSE_PROXY_AUTH_ENABLED: bool = False
+    REVERSE_PROXY_AUTH_HEADER: str = "X-Forwarded-User"
+    REVERSE_PROXY_AUTH_SIGNUP_ENABLED: bool = True
```
peterdey, line 438: `False` might be a safer default?

danepowell:
Thanks for the feedback, I've manually tested these changes and everything is still working.

mealie/routes/auth/auth.py
```diff
@@ -297,6 +299,46 @@ async def oauth_native_token(
...
+):
+    """Authenticate a user using a username forwarded by a trusted reverse proxy header"""
+    if not settings.REVERSE_PROXY_AUTH_READY:
+        raise HTTPException(status_code=status.HTTP_404_NOT_FOUND)
+
+    trusted_ips = settings.reverse_proxy_auth_trusted_ips
+    trusted_hosts, trusted_networks = trusted_ips
+    if trusted_hosts or trusted_networks:
+        # Without a trusted proxy list, anyone who can reach Mealie directly could set this header themselves.
+        client_ip = request.client.host if request.client else None
```
michael-genson: Can we check this against something other than `request.client.host`? With the default `HOST_IP="*"`, uvicorn replaces it with the first `X-Forwarded-For` entry, so anyone can send `X-Forwarded-For: <trusted ip>` and get past this. If `HOST_IP` is set to the proxy, it becomes the end user's IP instead, so the list would have to hold every user's IP. Would a shared secret header that the proxy sets (e.g. `REVERSE_PROXY_AUTH_SECRET`) be simpler?

frontend/app/pages/login.vue
```diff
@@ -315,7 +324,18 @@ whenever(
 onBeforeMount(async () => {
   if (isCallback()) {
     await oidcAuthenticate(true);
+    return;
```
michael-genson: Can we set `reverseProxyLoggingIn` to false before this return too? With OIDC and reverse proxy both enabled, a failed OIDC callback moves to `?direct=1` on the same page and the spinner never clears.

michael-genson (commented):
Thanks for the updates! Couple more:

PR #8202, opened 2026-08-24
feat: show recipes in a preferred unit system

Xyolyp:

Lets a reader view a recipe's quantities in the unit system they actually cook in — metric, imperial (UK) or US customary — and rewrite oven temperatures between °C and °F. **Nothing is written back**: the recipe keeps exactly what its author typed, this only changes what is rendered.

This picks up @agorton's work in #7573, which he has kindly offered to hand over. The picker rules, the preference model and the toggle placement are his; the main change is where the conversion happens. #7573 can be closed in favour of this.

Closes #3881 (discussion, 106 upvotes, open since 2024).

units-en

units-menu

units-metric2

### Why the conversion lives in the client

Since #7121 every `IngredientUnit` carries a `standardQuantity` / `standardUnit` pair, and that pair already ships with every ingredient in the recipe payload. So the browser has everything it needs, which buys three things:

- **No round trip.** No `GET /recipes/{slug}/conversions` endpoint, no `RecipeConversionResponse` schema, no shared Pint registry, no request per toggle. Cook mode, printing and the ingredient list inherit the conversion because they read the same `displayedRecipe` computed.
- **The scale is available.** #7573 noted a limitation: the display unit was picked from the *unscaled* magnitude, so 10× a 100 g ingredient read "1000 g". Here the unit is chosen from `quantity × standardQuantity × scale` while the returned quantity stays unscaled, so existing `useParsedIngredientText(ingredient, scale)` call sites keep working untouched and the same ingredient reads "1 kg".
- **It is arithmetic.** Eight canonical units across two dimensions. Pint is not earning its keep on this path.

### Files

- `frontend/app/composables/recipes/use-unit-conversion.ts` — the whole conversion. Pure functions that take a name resolver as an argument (so they test without an i18n instance), plus a thin composable that binds translations.
- `frontend/app/composables/recipes/use-unit-system.ts` — resolves user preference → household default → "as written", and builds the option lists.
- `frontend/app/components/Domain/Recipe/RecipeUnitSystemToggle.vue` — mirrors `RecipeScaleEditButton` and sits next to it in `RecipePageScale`. Hidden in edit mode, and hidden entirely on recipes where nothing is convertible.
- `frontend/app/components/Domain/Recipe/RecipePage/RecipePage.vue` — a `displayedRecipe` computed as the single substitution point.
- Backend: `UnitSystem` / `TemperatureUnit` enums, `users.preferred_*` (nullable) and `household_preferences.default_*`, one migration. No conversion logic server-side.

### Deliberately out of scope

- **Volume never becomes weight.** Cups to grams needs a density per food, which Mealie does not store. This is the single most common misunderstanding in #3881 and it deserves its own feature and data model.
- **Ingredients without standardization data are passed through untouched**, never guessed at. Seeded units have it; hand-made units may need `Standard Quantity` / `Standard Unit` filling in under Data Management → Units.
- **Imperial does not invent a cup.** Pint's `imperial_cup` is 284 ml while UK and AU recipes generally mean 250 ml, and there is no agreed imperial teaspoon or tablespoon. Imperial therefore renders volumes below an imperial pint in millilitres, which is how UK recipes are written anyway. @agorton reached the same conclusion in #7573.
- Shopping lists. Their `display` string is built server-side and items are aggregated there.

### Two things worth flagging

**Temperature is a separate preference** (`system | celsius | fahrenheit`, defaulting to "match the unit system"). Measuring in grams while owning a Fahrenheit oven is common, and it is the request that keeps recurring in #3881 — @Spinkman's "I like my temps in F and measurements in metric".

**Metric units render as decimals, customary ones as fractions.** The seeded units do not set `fraction`, so it defaults to `true`; without overriding this on the converted unit the parser renders 236.59 ml as "236 ⅝".

Closes #3881

- The picker thresholds in `pickDisplayUnit` are the judgement calls most worth a second opinion. US volumes run on cups all the way up ("8 ½ cups" rather than "4 pints") with tbsp/tsp below a quarter cup; #7573 switched to pints above 473 ml. Neither is obviously right.
- One bug carried over from #7573 that is worth knowing about if any of that code is reused: its temperature regex accepted a bare lowercase `c`, and `c` is the seeded abbreviation for *cup*. `"Add 2 c water"` became `"Add 35.6°F water"`, and `"1/2 c sugar"` matched on the `2`. Here the unit letter must be uppercase unless a degree sign is present; `350F`, `350 F`, `180°C` and `4°c` all still match. There are three regression tests for it.
- `test_admin_household_actions` needed a two-line change: it looks every response preference key up in a hand-written payload, so *any* new preference field breaks it with a `KeyError`. Not specific to this PR, but it will bite the next person too.
- The generated `mealie/schema/*/__init__.py` files come out of `task dev:generate` with a different `__all__` ordering than what is committed, across eight unrelated files. I left those alone and added only the two new exports by hand, to keep this diff reviewable — but it means the checked-in files are drifting from the generator.
- `useUnitSystem()` calls `useHouseholdSelf()`, which is a module-level singleton and only fetches once. Logged-out viewers never trigger it.

## Testing

`pnpm lint` clean, `pnpm vitest run` 325 passing (51 new), `uv run mypy mealie` clean, full `uv run pytest tests` green. The 10 remaining `ruff` findings are pre-existing on `mealie-next` in files this PR does not touch.

The new frontend tests cover every canonical-unit × system pair, the threshold edges (999 g vs 1000 g, oz↔lb, ml↔l), scale awareness in both directions, passthrough for the five ways an ingredient can be unconvertible, non-mutation of the input, and the temperature rewriter — including the `2 c` / `1/2 c` / `vitamin C` / `350Free` false positives and °F↔°C round trips with ranges.

Verified by hand against a dev instance: `2 cup` → `473 ml`, `8 ounce` → `227 g`, `4 ounce` → `113 g`; under US customary `350 ml` → `1 ½ c`; `"Bake at 350-375°F"` → `"Bake at 177-191°C"`. Toggling back to "as written" restores the authored values, and the toggle is absent in edit mode.

One gap in my manual testing: the scale interaction is covered by unit tests but I did not reproduce it in the browser — the servings stepper would not respond under automation and I stopped chasing it rather than report it as verified.

## AI / LLM Assistance

Substantial. The implementation, the tests and the first draft of this description were written with Claude Code (Opus 5), working from a design I reviewed and against a running dev instance. I read every line, ran all suites locally, and verified the Vuetify and Pint behaviour claims above against the pinned sources rather than taking the model's word for them.

## Release Notes

Recipes can now be viewed in metric, imperial or US customary units, with oven temperatures converted between °C and °F. Set a default in your profile or for your whole household, and switch any individual recipe with the button next to the serving scaler. Conversion is display-only — recipes keep the units they were written in.


frontend/app/lib/api/types/household.ts
```diff
@@ -345,6 +353,7 @@ export interface IngredientUnit {
 }
 export interface IngredientUnitAlias {
   name: string;
+  [k: string]: unknown;
```
deanmalan: What is `k`?

frontend/app/composables/recipes/use-unit-conversion.ts
```diff
@@ -0,0 +1,282 @@
...
+}
+
+/**
+ * Match a temperature: a number or a range, then an optional degree sign, then F or C.
+ *
+ * The unit letter must be uppercase unless a degree sign is present. Without that guard
+ * "1/2 c sugar" and "2 c water" read as 2 degrees Celsius — "c" is the seeded abbreviation
+ * for cup. The trailing `(?![A-Za-z])` keeps "vitamin C" and "350Free" out.
+ */
+const TEMPERATURE_PATTERN = /(\d+(?:\.\d+)?)(?:\s*[-–—]\s*(\d+(?:\.\d+)?))?(?:\s*°\s*([FCfc])|\s*([FC]))(?![A-Za-z])/;
```
deanmalan: Some websites say `425 degrees F` instead of using the `°`.

frontend/app/lib/api/types/household.ts
Xyolyp: Good catch, and thank you — that line should not have been in the diff at all.

`[k: string]: unknown` is a TypeScript index signature: it says "any other string key maps to an unknown value". `k` is just the parameter name pydantic2ts emits; it carries no meaning. It appears wherever a Pydantic model allows extra fields, and `mealie-next` already has 15 of them in this file.

The problem is that my `task dev:generate` run added 18 more, plus 3 in `meal-plan.ts`, none of which have anything to do with this PR. The committed generated files have drifted from what the current generator produces, so regenerating sweeps that drift into whatever PR happens to touch a schema. I have reset both files and re-added only the 20 lines this feature actually needs, so the generated-type diff is now `UnitSystem`/`TemperatureUnit` and the six preference fields, and nothing else. `meal-plan.ts` drops out of the PR entirely.

Worth flagging separately: the same thing happens to `mealie/schema/*/__init__.py`, where the generator reorders `__all__` across eight unrelated files. I left those alone for the same reason.

frontend/app/composables/recipes/use-unit-conversion.ts
Xyolyp: You are right, and it was a real gap — `425 degrees F` did not match. Fixed.

The pattern now accepts `350°F`, `350 F`, `350F`, `425 degrees F`, `425 degrees Fahrenheit`, `425 deg F`, `425 deg. F` and the Celsius/centigrade spellings, in either case for the spelled-out forms.

One case I deliberately left alone: a bare `425 degrees` with no unit. US recipes mean Fahrenheit by it and European ones Celsius, and there is nothing in the text to tell them apart — guessing would silently corrupt half of them. There is a test asserting `Turn the dough 90 degrees` stays put, too.

The reason the pattern is not simply case-insensitive: `c` is the seeded abbreviation for *cup*, so an `i` flag makes `Add 2 c water` read as 2°C. A bare letter therefore still has to be uppercase; only a degree sign or the word "degrees" licenses a lowercase one. 13 new tests cover the spellings and the false positives.

michael-genson:
Hey there, thanks for the PR. I wasn't happy with the implementation here since it hardcodes a ton of unit conversion ratios so I opened https://github.com/mealie-recipes/mealie/pull/8292 to leverage the existing conversion library `pint` to pre-generate conversion tables.

Going to close this one out in favor of https://github.com/mealie-recipes/mealie/pull/8292, happy to continue discussion there!

PR #8210, opened 2026-08-25
fix: allow SMTP encryption without authentication

### LINKED ISSUE #6212: [BUG] - SMTP transport encryption is coupled to authentication
### First Check

It is impossible to use TLS/SSL without setting User/Password. This requires to unnecessarily create a mailbox on the mailserver, where one could just use unauthenticated SMTP (i.e. mynetworks in Postfix)

```
Reason: Missing required values for ['SMTP_USER', 'SMTP_PASSWORD'] because SMTP_AUTH_STRATEGY is not None
```

Authentication and transport encryption are not directly related. While sending credentials over plaintext, is generally a bad idea (not considering local delivery), this check hardly makes sense to me, if any, you could argue it should be inverse.  

### Steps to Reproduce

1) Set `SMTP_HOST` and `SMTP_FROM_NAME` and `SMTP_FROM_EMAIL``
2) Start the application

```
Reason: Missing required values for ['SMTP_USER', 'SMTP_PASSWORD'] because SMTP_AUTH_STRATEGY is not None
```

### Mealie Version

v3.2.1

### Deployment

Docker (Linux)

_No response_

justadityaraj:
### Problem

SMTP validation currently treats `TLS` and `SSL` as authentication modes and requires `SMTP_USER` and `SMTP_PASSWORD` for either. Transport encryption and SMTP authentication are independent, and the sender already skips `login()` when credentials are absent.

### Changes

- Allow TLS and SSL connections without SMTP credentials.
- Continue to reject configurations that provide only one of username or password.
- Add regression cases for authenticated, unauthenticated, and partially configured TLS/SSL settings.
- Document SMTP credentials as an optional pair.

### User impact

Mealie can send through trusted relays that provide encrypted SMTP without requiring users to create an unnecessary mailbox.

Fixes #6212

## Testing

- `pytest tests/unit_tests/test_config.py -k smtp`: 9 passed.
- Broader configuration suite excluding an existing Windows-only SQLite path assertion: 58 passed.
- Ruff check and format check: passed.
- Mypy on the changed settings and test modules: passed.
- MkDocs build: passed (with the repository's existing `demo_url` warning).

The locked Python environment was installed on Windows with `python-ldap` omitted because that package requires native OpenLDAP/MSVC build dependencies; the SMTP tests do not depend on LDAP.

## AI / LLM Assistance

LLM assistance was used to analyze the validation path, implement the focused change, and add regression tests. I reviewed the final diff and all reported verification results before submission.

### DIFF
--- a/docs/docs/documentation/getting-started/installation/backend-config.md
+++ b/docs/docs/documentation/getting-started/installation/backend-config.md
@@ -53,8 +53,8 @@
 | SMTP_FROM_NAME                                  | Mealie  | Required For email                                |
 | SMTP_AUTH_STRATEGY                              |   TLS   | Required For email, Options: 'TLS', 'SSL', 'NONE' |
 | SMTP_FROM_EMAIL                                 |  None   | Required For email                                |
-| SMTP_USER<super>[&dagger;][secrets]</super>     |  None   | Required if SMTP_AUTH_STRATEGY is 'TLS' or 'SSL'  |
-| SMTP_PASSWORD<super>[&dagger;][secrets]</super> |  None   | Required if SMTP_AUTH_STRATEGY is 'TLS' or 'SSL'  |
+| SMTP_USER<super>[&dagger;][secrets]</super>     |  None   | Optional; must be set together with SMTP_PASSWORD |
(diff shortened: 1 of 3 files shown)


michael-genson (approved):
Thanks!

Issue #8224, opened 2026-08-27
[BUG] - Organizer data pages (Tags/Categories/Foods/Tools) show stale data until hard refresh


bferd:
### First Check

- [x] This is not a feature request.
- [x] I added a very descriptive title to this issue (title field is above this).
- [x] I used the GitHub search to find a similar issue and didn't find it.
- [x] I searched the Mealie documentation, with the integrated search.
- [x] I already read the docs and didn't find an answer.
- [ ] This issue can be replicated on the demo site (https://demo.mealie.io/).

### What is the issue you are experiencing?

The Tags, Categories, Foods, and Tools data management pages cache their table data in module-scope singleton stores (use-tag-store.ts, use-category-store.ts, use-food-store.ts, use-tool-store.ts) that only fetch once per SPA session. An initialized guard means revisiting the page never triggers a refetch, and nothing outside each store's own create/update/delete actions ever invalidates it.

As a result, if a recipe's tags/categories/foods/tools are changed from a different page (e.g. the recipe editor), the organizer data pages keep showing stale values — most visibly the recipeCount column — until a hard page reload resets the module state.

Confirmed to affect all four pages identically via the same shared store factory (use-store-factory.ts / use-actions-factory.ts).

Expected: the recipeCount (and other displayed fields) should reflect current data without requiring a hard refresh, either by refetching on page visit or by invalidating the store when underlying data changes elsewhere.

Additional context: the module-scope store caching architecture predates recipeCount, but recipeCount specifically (added in #7829 - mine, didn't notice pre-existing bug at the time unfortunately) is the field where this staleness became visible, since it's the one value on this page most likely to change from actions taken elsewhere in the app (e.g. editing a recipe's tags). Found while working on organizer improvements for foods/tools in a separate PR — not fixed there since it's a pre-existing, unrelated architectural issue. May look into a fix separately if time allows.

### Steps to Reproduce

1. Open the Tags (or Categories/Foods/Tools) data page, note a recipeCount
2. In a different tab/page, add that tag to another recipe via the recipe editor
3. Return to the Tags data page (soft navigation, no hard refresh)
4. recipeCount is unchanged; clicking the count still correctly shows the live/updated set    since that navigates to a fresh API-backed search

### Please provide relevant logs

none — UI/state staleness, not a runtime error.

### Mealie Version

Source build, mealie-next branch @ 8d602777 (post v3.24.0)

### Deployment

Docker (Linux)

Issue #8222, opened 2026-08-27
[BUG] - dev/code-generation types generator produces non-deterministic output depending on model processing order


bferd:
### First Check

- [x] This is not a feature request.
- [x] I added a very descriptive title to this issue (title field is above this).
- [x] I used the GitHub search to find a similar issue and didn't find it.
- [x] I searched the Mealie documentation, with the integrated search.
- [x] I already read the docs and didn't find an answer.
- [ ] This issue can be replicated on the demo site (https://demo.mealie.io/).

### What is the issue you are experiencing?

The type generator (dev/code-generation/main.py types, wrapped by task dev:generate) can produce incorrect output for shared model classes depending on which module processes them first in a given run.

Root cause: pydantic2ts temporarily flips each model's `extra` config to "forbid" before generating its JSON schema (to suppress an index signature), then reverts it — but the revert logic (`if x is not None: revert`) skips models whose original extra value was unset/None. Combined with pydantic v2 caching a model's compiled JSON schema the first time it's built, whichever module processes a shared model class first in a run gets a clean schema; any other module reusing that same class afterward reuses the cached forbid-mode schema instead, and gets stray `[k: string]: unknown` index signatures added to interfaces that shouldn't have them.

This makes output non-deterministic across runs (depends on module processing order) and can silently corrupt files unrelated to whatever a contributor is actually working on.

### Steps to Reproduce

1. Run task dev:generate (or python dev/code-generation/main.py types)
2. Check frontend/app/lib/api/types/household.ts and meal-plan.ts for `[k: string]: unknown` appearing on interfaces like RecipeTag, RecipeCategory, RecipeTool, Recipe, Nutrition, etc. that don't have it in the currently committed files
3. recipe.ts and labels.ts may come out clean in the same run since they can be processed earlier - output depends on module order, so results may vary run to run

### Please provide relevant logs

No runtime error - this is a code-generation correctness bug, found by running the actual pipeline and diffing generated output against the committed type files.

### Mealie Version

Source build, mealie-next branch @ 8d602777 (post v3.24.0)

### Deployment

Docker (Linux)

Issue #8221, opened 2026-08-27
[BUG] - MultiPurposeLabel.foods / shopping_list_items mis-declared as scalar, silently drop all but one related row


bferd:
### First Check

- [x] This is not a feature request.
- [x] I added a very descriptive title to this issue (title field is above this).
- [x] I used the GitHub search to find a similar issue and didn't find it.
- [x] I searched the Mealie documentation, with the integrated search.
- [x] I already read the docs and didn't find an answer.
- [ ] This issue can be replicated on the demo site (https://demo.mealie.io/).

### What is the issue you are experiencing?

MultiPurposeLabel.foods and MultiPurposeLabel.shopping_list_items (mealie/db/models/labels.py, lines 28-29) are declared as scalar relationships instead of lists, even though many foods or shopping-list items can share the same label.

As a result, label.foods and label.shopping_list_items only return one row instead of all of them whenever a label is used more than once.

Expected: both should be Mapped[list["IngredientFoodModel"]] and Mapped[list["ShoppingListItem"]], matching the relationship right below them which is declared correctly the same way.

Found while building a follow-up organizer improvements PR to add Delete-Unused support for labels, which needs the full list of foods/items using a label to work correctly.

### Steps to Reproduce

1. Create a MultiPurposeLabel
2. Attach it to two or more IngredientFoodModel rows (or two or more ShoppingListItem rows)
3. In backend code, access label.foods (or label.shopping_list_items) — only one row is returned instead of the full set

### Please provide relevant logs

No runtime error or log output — this is a static code-correctness finding in the SQLAlchemy model declaration, confirmed by reading the ORM mapping (mealie/db/models/labels.py:28-29) and verifying the FK/constraint shape, not by reproducing a crash.

### Mealie Version

Source build, mealie-next branch @ 8d602777 (post v3.24.0)

### Deployment

Docker (Linux)

Issue #8220, opened 2026-08-27
[BUG] - GET /api/groups/labels/{id} returns 500 instead of 404 for missing label


bferd:
### First Check

- [x] This is not a feature request.
- [x] I added a very descriptive title to this issue (title field is above this).
- [x] I used the GitHub search to find a similar issue and didn't find it.
- [x] I searched the Mealie documentation, with the integrated search.
- [x] I already read the docs and didn't find an answer.
- [ ] This issue can be replicated on the demo site (https://demo.mealie.io/).

### What is the issue you are experiencing?

GET /api/groups/labels/{id} returns a 500 Internal Server Error instead of a 404 Not Found when the provided id does not correspond to an existing label.

Root cause: get_one() in mealie/routes/groups/controller_labels.py (line 91) calls self.repo.get_one(item_id), which returns None when the id doesn't exist. The route declares response_model=MultiPurposeLabelOut, so FastAPI tries to serialize None into that model and fails with a ResponseValidationError, which surfaces as a 500 rather than the controller explicitly checking for None and raising a 404 — the pattern used elsewhere (e.g. tags, 
categories get_one).

Found while building a followup organizer improvements that touches recipe_count/merge/Delete-Unused support for labels in a separate PR. 
Left out of scope there since it's unrelated to that change.

### Steps to Reproduce

1. Call GET /api/groups/labels/{random-or-deleted-uuid}
2. Observe 500 response instead of 404

### Please provide relevant logs

fastapi.exceptions.ResponseValidationError: 1 validation error:
  {'type': 'model_attributes_type', 'loc': ('response',), 'msg': 'Input should be a valid dictionary or object to extract fields from', 'input': None}
  File "/opt/mealie-dev/mealie/routes/groups/controller_labels.py", line 91, in get_one
    GET /api/groups/labels/{item_id}

Reproduced via:
curl http://<mealie-instance>/api/groups/labels/<valid-but-nonexistent-uuid> -H "Authorization: Bearer <token>"
-> HTTP 500

### Mealie Version

Source build, mealie-next branch @ 8d602777 (post v3.24.0)

### Deployment

Docker (Linux)

PR #8225, opened 2026-08-27
feat: organizer improvements for foods, tools, and labels

bferd:

Follow-up to #7829, extending organizer improvements (`recipe_count`, merge, Delete Unused) to Foods, Tools, and Labels, plus a new Edit-dialog recipe-list feature and a fix for a pre-existing Combine dialog bug.

**Backend**
- `recipe_count` added to Foods and Tools (correlated subquery, mirroring the existing Tag/Category pattern). Labels intentionally excluded — they have no direct recipe relationship (only via foods/shopping-list items).
- Merge endpoints added for Tools (`POST /organizers/tools/merge`) and Labels (`POST /groups/labels/merge`). Foods' merge already existed via a different mechanism (direct FK reassignment rather than an association table) and was left as-is.
- Delete Unused (`get_empty()` + `GET .../empty`) added for Foods, Tools, and Labels.
- New batched tag-removal endpoint: `POST /organizers/tags/{id}/remove-from-recipes`, removing a tag from a list of recipes in one transaction — powers the new removable-chip Edit dialog for Tags.
- Fixed `GET /organizers/tags/{id}` silently returning `recipes: []` — the route declared `response_model=RecipeTagResponse` but called `self.mixins.get_one()` without `override_schema`, so it defaulted to `TagOut` (no `recipes` field). Fixed to match the existing `get_one_by_slug` pattern.
- `RepositoryMultiPurposeLabels.get_empty()` uses `.any()` on `MultiPurposeLabel.foods`/`.shopping_list_items` — this originally used `.has()` to work around #8221 (those relationships were incorrectly declared as scalar), but #8221 has since been fixed in #8226 (merged into `mealie-next`), and this branch has been synced to that, so `.any()` is now correct.

**Frontend**
- Foods/Tools/Labels data pages: `recipe_count` column (Foods/Tools only), merge dialog (Tools/Labels), and Delete Unused button/dialog, matching the existing Tags/Categories pattern.
- New Edit-dialog "Associated Recipes" feature:
  - **Tags**: recipes shown as removable chips (styled like the tag chips on recipe cards). Clicking × stages a removal locally; nothing is sent to the server until Confirm, and closing the dialog via Cancel, clicking outside, or Esc discards all staged changes. Helper text clarifies this behavior.
  - **Categories/Foods/Tools**: recipes shown as a plain, non-removable, clickable list that navigates to the recipe.
  - **Labels**: not included — no direct recipe relationship.
  - All lists truncate to 10 items with a "+N more" indicator (reusing the existing Delete Unused i18n pattern).
  - Only built on the group-scoped data pages (`/g/{slug}/organizers/...`); not extended to the public/explore Organizers pages — see notes below.
- Fixed a pre-existing bug: the Combine (merge) dialog's Source/Target selections weren't reset on Cancel, click-outside, or Esc — only on a successful merge — across all five merge dialogs (Tags, Categories, Foods, Tools, Labels).
- Added missing `recipeCount`/`CategoryMerge`/`TagMerge`/`TagRecipesRemove` TypeScript interfaces to `recipe.ts`, which had gone stale relative to the backend since #7829. Hand-added rather than regenerated, since the codegen tool has a separate pre-existing ordering bug (#8222, not fixed here) that corrupts unrelated files when run — verified these hand-added interfaces are byte-identical to what the real codegen pipeline produces for this file.
- Units intentionally untouched throughout.

N/A — follow-up to #7829, no tracked issue for this specific scope.

While building this, found and filed several unrelated pre-existing issues, left untouched here as out of scope: #8220, #8221, #8222, #8224.

- `RecipeToolMerge` and `MultiPurposeLabelMerge` were added as separate schema classes, mirroring the existing `TagMerge`/`CategoryMerge` pattern, rather than a single generic `OrganizerMerge` schema. If you would prefer to consolidate let me know - wasn't sure if the per-entity classes are intentional convention or just organic growth.
- There's no shared merge-dialog component - each of the five data pages (Tags/Categories/Foods/Tools/Labels) copy-pastes its own `fromX`/`toX` refs and `mergeX()` handler around `BaseDialog`. That's why the Combine dialog reset bug had to be fixed in five places rather than one, and why Labels' brand-new merge dialog (built this session) inherited the same bug via copy-paste before the fix. Might be worth a future refactor into a shared `MergeDialog` component to prevent this recurring.
- The new recipe-list feature is only on the group-scoped data pages, not public/explore. The removable Tags chips are the one genuinely new capability (no other way to bulk-remove a tag from selective recipes); the display-only lists on Categories/Foods/Tools are largely redundant with clicking the item's card, which already surfaces its recipes, so I didn't see a strong case for extending to public/explore there but flagging in case you see it differently.

## Testing:

- `DB_ENGINE=sqlite uv run pytest` — full suite, 2989 passed, 16 skipped (expected: long-running scraper/parser tests, LDAP tests requiring an external service). Numbers reflect a sync with current `mealie-next`, which added substantial new test coverage of its own.
- New integration tests added for Foods/Tools/Labels merge, Foods/Tools/Labels Delete Unused, and the Tags remove-from-recipes endpoint.
- Manually verified in the browser: recipe counts accurate after merge/edit, merge dialogs reassign correctly for all five entity types, Delete Unused lists genuinely-unused items, Tags chips stage/discard/commit correctly across Cancel/Confirm, Categories/Foods/Tools recipe links navigate correctly, truncation ("+N more") verified on Tags and Categories with 11+ associated recipes, Combine dialog reset verified via Cancel/click-outside/Esc on all five merge dialogs.

## AI / LLM Assistance:

Claude Code was used to generate the implementation across several passes (backend, frontend, and bug fixes). All code was reviewed, and all functionality was manually tested.


bferd:
Screenshots:

Screenshot_20260909_110358

Screenshot_20260909_110258

Screenshot_20260909_110534

Screenshot_20260909_110618

mealie/repos/repository_factory.py
```diff
@@ -167,6 +167,101 @@ def merge(self, from_tag: UUID4, to_tag: UUID4) -> TagOut | None:
...
+        q = super()._query(override_schema=override_schema, with_options=with_options)
+        count_sq = (
+            select(func.count(recipes_to_tools.c.recipe_id))
+            .where(recipes_to_tools.c.tool_id == Tool.id)
+            .correlate(Tool)
+            .scalar_subquery()
+        )
+        return q.options(with_expression(Tool.recipe_count, count_sq))
+
+    def get_empty(self) -> Sequence[Tool]:
```
michael-genson: These new `get_empty()` queries aren't scoped to the group, so `/empty` returns unused items from every group on the instance and "Delete Unused" shows them. Can we filter on `group_id` here and in the label and food versions? The existing tag/category ones have the same gap, so worth fixing all five while you're in here. Can we also add tests covering this — an unused food/tool/label created in another group shouldn't show up in the first group's `/empty` response?

frontend/app/pages/group/data/foods.vue
```diff
@@ -425,9 +490,27 @@ async function handleEdit() {
...
+const editRecipes = ref<RecipeSummary[]>([]);
+const editRecipesPreview = computed(() => editRecipes.value.slice(0, EDIT_RECIPES_PREVIEW_LIMIT));
+const editRecipesRemaining = computed(() => Math.max(editRecipes.value.length - EDIT_RECIPES_PREVIEW_LIMIT, 0));
+
+async function onEditDialogOpen(item: IngredientFoodWithOnHand) {
+  editRecipes.value = [];
+  if (!item?.id) {
+    return;
+  }
+  const { data } = await userApi.recipes.search({ foods: [item.id], perPage: -1 });
```
michael-genson: `perPage: -1` pulls every matching recipe just to show 10. Can we request a single page and use the response's `total` for the "+N more" count instead?

michael-genson (changes requested):

frontend/app/pages/group/data/foods.vue
bferd: Trade off would be capping at 10 removals per visit. The likelihood of users having more than 10 tags per is low but I think it would feel broken if so. Better to deal with that now. Going to add 10 per page and refresh on next page.

mealie/repos/repository_factory.py
bferd: Confirmed. This was an oversight. I have remedied and tested across groups and will be in the ~next commit

frontend/app/pages/group/data/foods.vue
bferd: Fixed - this and the same pattern in `tools.vue` and `categories.vue` no longer pull everything with `perPage: -1`. All three (plus `tags.vue`) now fetch one page of 10 at a time, sorted alphabetically, with Previous/Next buttons replacing the old "+N more" indicator. Tags additionally refills one recipe at a time as chips are removed, since it's the only one of the four with removable items. Also added handling for the case where that list gets long enough to exceed a URL length limit. Will push shortly for re-review.

PR #8226, opened 2026-08-27
fix: correct MultiPurposeLabel.foods/shopping_list_items relationship cardinality

### LINKED ISSUE #8221: [BUG] - MultiPurposeLabel.foods / shopping_list_items mis-declared as scalar, silently drop all but one related row

bferd:
- `mealie/db/models/recipe/labels.py`: fixes `MultiPurposeLabel.foods` and `MultiPurposeLabel.shopping_list_items`, which were annotated as scalar (`Mapped["IngredientFoodModel"]` / `Mapped["ShoppingListItem"]`) instead of `Mapped[list[...]]`, despite being the "one" side of a one-to-many relationship. 
- SQLAlchemy infers `uselist` from the annotation shape, so the static typing was wrong even though runtime behavior was masked by `uselist=False` on the reverse `.label` relationships. Now matches the correctly-typed sibling relationship `shopping_lists_label_settings` on the same model.
- No other files changed — confirmed via codebase search that `label.foods` and `label.shopping_list_items` aren't accessed anywhere outside the ORM declaration itself (no schema, service, or controller code depends on them), so this is a pure type-correctness fix with no behavior change.

Fixes #8221

Type-annotation-only fix, no migration needed (no columns/FKs changed). Verified mypy passes clean on the changed file.
`RepositoryMultiPurposeLabels.get_empty()` in #8225 (still open, review pending) uses `.has()` on `MultiPurposeLabel.foods`/`.shopping_list_items` as a workaround for the scalar-relationship issue fixed here. `.has()` will keep working fine after this merges, but `.any()` would be the more idiomatic choice once these are properly `list[...]`-typed. Not blocking either PR just flagging so it doesn't get lost; I will follow up with a small commit on #8225 once this lands.

## Testing
Full test suite passes locally (`DB_ENGINE=sqlite uv run pytest`) — 2701 passed, 
16 skipped, 0 failures — including `tests/integration_tests/user_household_tests/test_shopping_list_labels.py` (9 tests). 

## AI / LLM Assistance

### DIFF
--- a/mealie/db/models/recipe/labels.py
+++ b/mealie/db/models/recipe/labels.py
@@ -25,8 +25,10 @@ class MultiPurposeLabel(SqlAlchemyBase, BaseMixins):
     group_id: FilterableColumn[GUID] = mapped_column(GUID, ForeignKey("groups.id"), nullable=False, index=True)
     group: Mapped["Group"] = orm.relationship("Group", back_populates="labels")
 
-    shopping_list_items: Mapped["ShoppingListItem"] = orm.relationship("ShoppingListItem", back_populates="label")
-    foods: Mapped["IngredientFoodModel"] = orm.relationship("IngredientFoodModel", back_populates="label")
+    shopping_list_items: Mapped[list["ShoppingListItem"]] = orm.relationship(
+        "ShoppingListItem", back_populates="label"
+    )
+    foods: Mapped[list["IngredientFoodModel"]] = orm.relationship("IngredientFoodModel", back_populates="label")
     shopping_lists_label_settings: Mapped[list["ShoppingListMultiPurposeLabel"]] = orm.relationship(
         "ShoppingListMultiPurposeLabel", back_populates="label", cascade="all, delete, delete-orphan"
     )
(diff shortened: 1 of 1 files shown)


michael-genson (approved):
Thanks!

PR #8227, opened 2026-08-27
fix: labels get-by-id returns 404 instead of 500 for missing id

### LINKED ISSUE #8220: [BUG] - GET /api/groups/labels/{id} returns 500 instead of 404 for missing label

bferd:
- Fixes GET /api/groups/labels/{id} returning 500 instead of 404 when the label doesn't exist
- `controller_labels.py`'s get-by-id endpoint called `self.repo.get_one()` directly instead of `self.mixins.get_one()`; `HttpRepo.get_one()` is what raises the 404 on a missing item, so bypassing it let `None` fall through to response_model serialization and throw a 500
- One-line fix: swap to `self.mixins.get_one(item_id)`, matching Tags/Categories/Foods/Tools/Units, all of which already use the mixin correctly
- Added `test_label_get_one_404` regression test

Fixes #8220

Audited all six organizer get-by-id endpoints before fixing — Labels was the only one bypassing `self.mixins.get_one()`. No changes needed elsewhere.

## Testing
- Full suite: `DB_ENGINE=sqlite uv run pytest` — all passed
- `ruff check` and `ruff format --check` clean on both changed files
- New test `test_label_get_one_404` passes, confirmed alongside existing `test_label_create_duplicate`

## AI / LLM Assistance

### DIFF
--- a/mealie/routes/groups/controller_labels.py
+++ b/mealie/routes/groups/controller_labels.py
@@ -70,7 +70,7 @@ class MultiPurposeLabelsController(BaseCrudController):
 
     @router.get("/{item_id}", response_model=MultiPurposeLabelOut)
     def get_one(self, item_id: UUID4):
-        return self.repo.get_one(item_id)
+        return self.mixins.get_one(item_id)
 
     @router.put("/{item_id}", response_model=MultiPurposeLabelOut)
     def update_one(self, item_id: UUID4, data: MultiPurposeLabelUpdate):
--- a/tests/integration_tests/user_household_tests/test_shopping_list_labels.py
+++ b/tests/integration_tests/user_household_tests/test_shopping_list_labels.py
@@ -1,4 +1,5 @@
 import random
+from uuid import uuid4
 
 from fastapi.testclient import TestClient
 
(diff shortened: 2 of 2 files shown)


michael-genson (approved):
Thanks!

PR #8239, opened 2026-08-29
fix: clamp recipe finder missing item limits

### LINKED ISSUE #8238: [BUG] - Recipe Finder allows a negative value for Max Missing Tools
### First Check

The Recipe Finder accepts and retains a negative value for the **Max Missing Tools** setting.

A negative number of missing tools is not a valid limit. Allowing it may produce confusing or incorrect recipe results because the UI does not explain how a negative limit is interpreted.

The behavior is also inconsistent with the adjacent **Max Missing Ingredients** field:

- Entering `-1` for **Max Missing Ingredients** is normalized to `0`.
- Entering `-1` for **Max Missing Tools** remains `-1`, even after pressing Enter and moving focus away from the field.

The field is visually marked as invalid, but the value is retained and no validation message explains how to correct it.

**Actual behavior**

The field continues to display `-1`. No validation message is shown, and the invalid value remains in the Recipe Finder settings.

**Expected behavior**

**Max Missing Tools** should only accept non-negative integers. A negative value should be clamped to `0`, rejected in favor of the previous valid value, or accompanied by a clear validation message and prevented from being applied.

Both maximum-missing fields should use consistent validation behavior.

Recipe Finder showing Max Missing Tools set to -1

### Steps to Reproduce

1. Sign in to Mealie.
2. Navigate to **Recipes → Recipe Finder**.
3. Expand **Settings**.
4. Enter `-1` in **Max Missing Tools**.
5. Press Enter.
6. Move focus away from the input.

**Result:** The field still displays `-1`.

**Control check:** Performing the same steps in **Max Missing Ingredients** normalizes the value to `0`.

No browser console warnings or errors were observed while reproducing the issue on the public demo site. Server-side logs are not available for the public demo deployment.

### Mealie Version

Public demo site running the nightly build (tested 2026-08-29). The exact build tag is not exposed to the demo user.

### Deployment

Other (please specify below)

Reproduced consistently at https://demo.mealie.io/g/home/recipes/finder using the public demo account in a Chromium-based browser.

YibingZhang325:

- clamps Recipe Finder missing-food and missing-tool limits before they are used by the finder
- normalizes previously saved negative limits when the page initializes
- uses the same normalization path for both settings so their behavior stays consistent
- adds unit tests for the normalization helper, covering positive, zero, negative, null, and undefined values

This fixes the case where `Max Missing Tools` could retain `-1` even though the input has a minimum of zero. There is no visual layout change; the before state is documented with a screenshot in the linked issue.

Fixes #8238

## Testing

- `npm run test:ci` — 32 test files and 321 tests passed
- `eslint .`

## AI / LLM Assistance

OpenAI Codex assisted with the implementation. I reviewed the final diff.


michael-genson (approved):
Thanks!

PR #8250, opened 2026-08-31
feat: warn before importing duplicate recipe URLs

evelrin:

Accidentally importing the same source URL currently creates another recipe with a suffixed name. This PR adds a non-blocking duplicate warning to the single-URL import flow.

- Adds a group-scoped endpoint that looks up a recipe by an equivalent `orgURL`.
- Normalizes both the submitted and stored URLs at comparison time, removing fragments and known tracking parameters while retaining meaningful query parameters; no data migration is required.
- Checks the source URL before starting a URL import.
- Shows a dialog that lets the user open the existing recipe, cancel, or import another copy anyway.
- Keeps the existing import endpoints unchanged, so API clients and intentional duplicate imports remain supported.
- Adds integration coverage for matches in the same group, visibility across households in that group, isolation from other groups, tracking-parameter normalization, preservation of meaningful query parameters, and URLs without a match.

Related discussion: https://github.com/mealie-recipes/mealie/discussions/8249

The warning is currently limited to the interactive single-URL import page. Bulk imports and direct API calls retain their existing behavior.

Only the `en-US` locale was edited, following the project's Crowdin localization workflow.

## Testing

- `ruff check` on the changed Python files: passed
- `ruff format --check` on the changed Python files: passed
- `mypy` on the changed backend files: passed
- ESLint on the changed Vue and TypeScript files: passed
- Focused duplicate URL integration tests: 3 passed
- Full production Docker image build, including Nuxt static generation: passed
- Manual local verification against a copied SQLite data directory:
  - an existing source URL displays the warning;
  - **Open existing recipe** navigates to the correct original recipe;
  - no duplicate is created unless **Import anyway** is selected.

## AI / LLM Assistance

OpenAI Codex was used to inspect the repository, draft the implementation and tests, and assist with local verification. The resulting changes were linted, type-checked, built in the project's production Docker image, and tested against an isolated copy of an existing Mealie SQLite database.


michael-genson:
Thanks for the PR, but I don't think we're going to implement this at this time. There's a lot of volatility with URLs which makes this a pretty heavy maintenance burden which is more likely to either flag false positives or miss a lot of false negatives.

We already have handling for duplicate names which more-or-less covers the spirit of this PR.

PR #8260, opened 2026-09-02
fix: refetch organizer data pages on visit to avoid stale data

### LINKED ISSUE #8224: [BUG] - Organizer data pages (Tags/Categories/Foods/Tools) show stale data until hard refresh

lotusk:

The Tags, Categories, Foods, and Tools data management pages (`pages/group/data/{tags,categories,foods,tools}.vue`) read from module-scope singleton stores (`use-tag-store.ts`, `use-category-store.ts`, `use-food-store.ts`, `use-tool-store.ts`) that only fetch once per SPA session, guarded by an `initialized` flag. Revisiting the page never triggers a refetch, and nothing outside each store's own create/update/delete actions invalidates it. As a result, if a tag/category/food/tool is changed from a different page (e.g. tagging a recipe from the recipe editor), the organizer data pages keep showing stale values — most visibly the `recipeCount` column — until a hard page reload resets the module state. This adds an `onMounted` hook to each of the four pages that calls the store's existing `actions.refresh()`, so navigating to the page always pulls current data. The shared store factory (`use-store-factory.ts`) itself is untouched — these stores are also used by 13+ other components (recipe editor, shopping list, recipe browsing pages, etc.) where refetching on every mount would be wasteful and isn't needed, so the fix is scoped to only the four pages where "always show current data" is the actual expectation.

Fixes #8224

`foods.vue` already had an unrelated `onMounted` hook (for setting the seed-dialog locale); added a second one next to the store declaration rather than merging into the existing one, to keep it colocated with `foodStore` like the other three pages.

## Testing

Reproduced the staleness locally before the fix: noted a tag's `recipeCount` on the Tags page, tagged an existing recipe with that same tag from the recipe editor in the same tab, navigated back (browser back — confirmed the app doesn't use `<KeepAlive>` around pages, so this remounts the component same as a link click), and saw the count was unchanged until a hard refresh. Applied the fix and repeated the same steps — the count now updates correctly on navigating back, no hard refresh needed. Repeated for all four pages.

## AI / LLM Assistance

Used an AI coding assistant to trace the root cause to the shared store factory's `initialized` guard, scope the fix to the four affected pages without touching the shared factory (used elsewhere in the app), and verify the reproduction before and after the fix.


michael-genson (approved):
Thanks!

PR #8265, opened 2026-09-02
feat: ingredient substitutions

michael-genson:

_(REQUIRED)_

Adds the ability to specify ingredient substitutions on two levels:
1. on the food (as a common substitution)
2. on the recipe ingredient (as an explicit substitution for that recipe)

Also wires in substitution logic into the recipe finder, and adds support for populating substitutions via the NLP parser and the AI parser.

---

You can add substitutions for foods via the data management page:
Screenshot 2026-09-03 at 1 02 14 PM

Substitutions are one-way, but you can add the bidirectional substitution via this dialog. As an example for why they're one way, consider substituting water for milk - if a recipe calls for water, you _definitely_ can't use milk, but water _can_ often be used in place of milk.

Note that substitutions can be a food, a note, or a food and a note. This is the same logic we use for ingredients (and consequently ingredient substitutions).

---

Recipes have a similar UI in edit mode:
image

And are rendered with a small button:
Screenshot 2026-09-02 at 4 37 16 PM

Note how recipe-level and food-level substitutions appear differently:
Screenshot 2026-09-02 at 4 37 58 PM

They also print:
Screenshot 2026-09-02 at 4 39 04 PM

Though that can be turned off:
Screenshot 2026-09-02 at 4 39 21 PM

---

The recipe finder also renders them specially:
Screenshot 2026-09-02 at 4 38 33 PM

And can be toggled off:
Screenshot 2026-09-02 at 4 38 44 PM

_(REQUIRED)_

Partially implements #2824
Partially implements #3717 
Partially implements #7635 

Are the UI elements intuitive enough? Specifically the data management page deserves some extra attention because it has the potential to be confusing if you're not familiar with substitutions (and no one is since they don't exist yet...).

## Testing

Added the proper tests, tested E2E manually via the frontend.

## AI / LLM Assistance

_(REQUIRED)_

Extensive prompting with Claude, including a spec document for each component.


miah120:
Ooh this looks cool! Is there a reason the data manager page uses a dialog action to add a new substitution while the recipe level substitutions have the add button at the bottom of the list? 

Imo I think the one on the recipe page feels more intuitive if they can be consistent.

michael-genson:
It matches the alias dialog, but I agree I'm not really happy with it. I'll try implementing something similar to the recipe page

michael-genson:
I unified the UIs so the data management page uses the add button like the recipe page does:
Screenshot 2026-09-03 at 1 02 14 PM

p0lycarpio:
It works like a charm!
From an accessibility standpoint, I think the  actions buttons and the icon next to the original ingredient name in the recipe are too small.
Also, when testing on a screen the size of a smartphone, the delete button wraps to the next line, and the actions on the ingredient page aren't scrollable.

michael-genson:
@p0lycarpio all addressed. Here's the new button size:
Screenshot 2026-09-03 at 10 32 08 PM

p0lycarpio (dismissed):
LGTM, thanks!

PR #8274, opened 2026-09-03
feat: Show calendar hint when adding to meal plan

miah120:
When selecting a date for meal planning, dates that already have a meal planned will show up with a dot under it. 
This allows users a quick way to see what days need a plan, and which are already covered. 

| Before | After |
|--------|--------|
| image | image | 

NA

## AI / LLM Assistance
No AI used


p0lycarpio (approved):
A small but very useful change. Thanks for that!

csmit39:
We have been dying for this change. Thank you @miah120!

PR #8275, opened 2026-09-03
feat: search the items on a shopping list

PulfordJ:

A shopping list has no way to find a single item on it. Once a list grows past a
screenful this becomes a real problem for checked items in particular: they
collapse into one section whose only controls are "Uncheck All Items" and
"Delete Checked", so re-adding something you previously ticked off means
scrolling the whole section looking for it.

This adds a search field to the shopping list. A list already arrives in a
single request, so nothing new is fetched — the items in memory are matched and
non-matching ones are hidden.

I moved from a Google Keep shopping list to mealie and sometimes I don't need to buy all the ingredients so this is a feature I use in Google Keep often and I would love to bring it into mealie.

Changes by file:

- **`use-shopping-list-search.ts`** (new) — owns the search state. Matching is
  delegated to the existing `useSearch`, so the shopping list ranks and tolerates
  typos exactly like every other search box in Mealie. List items are adapted to
  `ISearchableItem`: food name and plural name, the item's note (a food-backed
  item can still carry one), food aliases, unit name/abbreviation, and the label
  name — so searching a label narrows a long list to a single aisle.
- **`use-shopping-list-page.ts`** — wires the composable up over every item on
  the list, checked and unchecked alike, and derives the visible checked count
  and a "did anything match" flag.
- **`shopping-lists/[id].vue`** — adds the field, hides non-matching items,
  drops labels that have no matches rather than leaving empty headings, shows
  the existing `search.no-results` message when nothing matches, and expands the
  checked section automatically while searching so a previously-checked item can
  be found and unchecked directly.

Two decisions worth calling out:

- **Only the match *set* from `useSearch` is used, not its order.** Its ranking
  would fight the label grouping and the user's aisle ordering, so items stay
  where they are and are hidden in place.
- **Non-matching items are hidden rather than filtered out of the arrays.**
  `ShoppingListItem` replaces the whole item object when it is checked off
  (`model.value = { ...model.value, checked: ... }`), and items are rendered with
  `v-model` bound into their backing array. Rendering a filtered copy would write
  that update into a throwaway array and lose it. This is commented in the
  composable so it doesn't get "simplified" back into a bug later.

Reordering is disabled while a search is active: the visible items are a subset,
so persisting a drag would write an incomplete order.

No new translation strings — this reuses `search.search` and `search.no-results`.

None that I could find — I searched open issues for shopping list search/filter
and didn't turn up an existing report. Happy to link one if a maintainer knows of it.

## Testing

- Added `use-shopping-list-search.test.ts` — 8 cases via `vitest`, covering the
  empty-query passthrough, note-only items, food-backed items, a note on a
  food-backed item, food aliases, label-name matching, a query that matches
  nothing, and `clearSearch`. It drives `debouncedSearch` directly to stay
  synchronous, the same approach `use-search.test.ts` uses.
- The template changes were exercised against a live instance holding a ~130
  item list, via a backport of the same markup to the 3.16.0 release I run at
  home.

I (without claude) have attached the before/after images in the comment below https://github.com/mealie-recipes/mealie/pull/8275#issuecomment-5530458380 .

## AI / LLM Assistance

Substantial, and I'd rather be upfront about it. This PR — the implementation,
the tests, and this description — was written by Claude (Anthropic's Claude Code)
working interactively with me. I set the goal, reviewed the output, and am
running the feature on my own instance.

For what it's worth about the process: the model read the existing shopping-list
composables before writing anything, and reworked its own first attempt to
delegate to `useSearch` after finding that composable. Two bugs it caught in its
own draft during review were a `v-if` sharing an element with `v-for` (which
wouldn't have compiled, since `v-if` is evaluated first and the loop variable
isn't in scope) and Vuetify's `clearable` emitting `null` into a `.trim()` call.
The unit tests were run locally and pass; the UI was not visually reviewed by
the model.

Happy to make whatever changes you'd like, and no hard feelings if the project
would rather not take LLM-authored contributions — I'd sooner ask than waste a
reviewer's time.

### DIFF
--- a/frontend/app/pages/shopping-lists/[id].vue
+++ b/frontend/app/pages/shopping-lists/[id].vue
@@ -182,61 +182,79 @@
         />
       </div>
 
+      <v-text-field
+        :model-value="search"
+        :label="$t('search.search')"
+        :prepend-inner-icon="$globals.icons.search"
+        clearable
+        hide-details
+        density="compact"
+        variant="solo"
+        flat
+        single-line
+        @update:model-value="value => search = value ?? ''"
+        @click:clear="clearSearch"
+      />
+
       <TransitionGroup name="scroll-x-transition">
-        <BaseExpansionPanels v-for="(value, key) in itemsByLabel" :key="key" :v-model="0" start-open>
-          <v-expansion-panel class="shopping-list-section">
-            <v-expansion-panel-title
-              :color="getLabelColor(key)"
-              class="body-1 font-weight-bold section-title"
-            >
-              {{ key }}
-            </v-expansion-panel-title>
-            <v-expansion-panel-text eager>
-              <VueDraggable
-                :model-value="value"
-                handle=".handle"
-                :delay="250"
-                :delay-on-touch-only="true"
-                @start="loadingCounter += 1"
-                @end="loadingCounter -= 1"
-                @update:model-value="updateIndexUncheckedByLabel(key.toString(), $event)"
+        <template v-for="(value, key) in itemsByLabel" :key="key">
+          <BaseExpansionPanels v-if="hasMatches(value)" :v-model="0" start-open>
+            <v-expansion-panel class="shopping-list-section">
+              <v-expansion-panel-title
+                :color="getLabelColor(key)"
+                class="body-1 font-weight-bold section-title"
               >
-                <TransitionGroup name="scroll-x-transition">
-                  <ShoppingListItem
-                    v-for="(item, index) in value"
-                    :key="item.id"
-                    v-model="value[index]"
-                    class="my-2 w-auto"
-                    :edit="editingItem === item.id"
-                    :labels="allLabels || []"
-                    :units="allUnits || []"
-                    :foods="allFoods || []"
-                    :recipes="recipeMap"
-                    @checked="(item) => {
-                      saveListItem(item);
-                      itemCheckedToast(item);
-                    }"
-                    @save="(item) => {
-                      editingItem = undefined;
-                      saveListItem(item);
-                    }"
-                    @delete="deleteListItem(item)"
-                    @view="editingItem = undefined"
-                    @edit="editingItem = item.id"
-                  />
-                </TransitionGroup>
-              </VueDraggable>
-            </v-expansion-panel-text>
-          </v-expansion-panel>
-        </BaseExpansionPanels>
+                {{ key }}
+              </v-expansion-panel-title>
+              <v-expansion-panel-text eager>
+                <VueDraggable
+                  :model-value="value"
+                  handle=".handle"
+                  :delay="250"
+                  :delay-on-touch-only="true"
+                  :disabled="isSearching"
+                  @start="loadingCounter += 1"
+                  @end="loadingCounter -= 1"
+                  @update:model-value="updateIndexUncheckedByLabel(key.toString(), $event)"
+                >
+                  <TransitionGroup name="scroll-x-transition">
+                    <template v-for="(item, index) in value" :key="item.id">
+                      <ShoppingListItem
+                        v-if="matchesSearch(item)"
+                        v-model="value[index]"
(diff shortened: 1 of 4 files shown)


PulfordJ:
(In a shopping list) 

Before (No search bar): 

image


After (with search bar):

image

With search for "tomato"

image

With the small tomatoes then unchecked:

image

frontend/app/pages/shopping-lists/[id].vue
```diff
@@ -182,61 +182,79 @@
...
       <!-- Checked Items -->
-      <v-expansion-panels flat>
-        <v-expansion-panel v-if="listItems.checked && listItems.checked.length > 0">
+      <v-expansion-panels v-model="checkedPanel" flat>
+        <v-expansion-panel v-if="listItems.checked && hasMatches(listItems.checked)">
           <v-expansion-panel-title class="border-solid border-thin py-1">
             <div class="d-flex align-center flex-0-1-100">
               <div class="flex-1-0">
-                {{ $t('shopping-list.items-checked-count', listItems.checked ? listItems.checked.length : 0) }}
+                {{ $t('shopping-list.items-checked-count', visibleCheckedCount) }}
```
Comment (michael-genson) [inline frontend/app/pages/shopping-lists/[id].vue]: While a search is active this shows the filtered count, but "Uncheck All Items" and "Delete Checked" still act on every checked item, so someone could delete items the search is hiding. Can we disable those two buttons while `isSearching` is true?

frontend/app/pages/shopping-lists/[id].vue
```diff
@@ -182,61 +182,79 @@
         />
       </div>
 
+      <v-text-field
```
Comment (michael-genson) [inline frontend/app/pages/shopping-lists/[id].vue]: Nice work on this. Could we make the search take up less space on the page? Two things I'd like: move it into the existing button group in the page title as a search icon that toggles the field (similar to what `AppHeader.vue` does on narrow screens), and only offer it once the list has more than 20 items, since that's where finding something gets hard. Happy to adjust that number if it feels off in practice.

michael-genson (changes requested):

frontend/app/pages/shopping-lists/[id].vue
Comment (PulfordJ) [reply frontend/app/pages/shopping-lists/[id].vue]: Good point, done!

frontend/app/pages/shopping-lists/[id].vue
Comment (PulfordJ) [reply frontend/app/pages/shopping-lists/[id].vue]: Thanks for the review, I've adjusted this per your guidance. 20 items seemed reasonable to me too

PulfordJ:
Thanks for the review @michael-genson ! I've pushed the changes and rebased onto the latest mealie-next, which clears the conflicts with the condensed-view work in [#8402](https://github.com/PulfordJ/dotfiles-media-server/issues/8402).

Search takes less space

- The search field is hidden by default. A search icon in the page title's button group toggles it and puts the cursor in the field. The icon is highlighted while the field is open.
- The icon only appears once the list has more than 20 items. That's a single constant (SEARCH_MIN_LIST_SIZE), so it's easy to change.
- Closing the field also clears the search, so items can't stay hidden by a search you can no longer see.
- If the list drops to 20 items or fewer while the field is open, the icon stays until the field is closed, so it can't get stuck open.

Bulk actions while searching

- "Uncheck All Items" and "Delete Checked" are disabled while a search is active.
- I also disabled "Check All" in the title while searching, because it has the same problem with hidden unchecked items. Happy to revert that if you'd rather keep it to the two you mentioned.

Testing

- Added unit tests for the threshold and the open/close behaviour. The shopping-list and search tests pass, and pnpm lint --max-warnings=0 is clean.
- I've backported the same changes to the 3.16 build I run at home and am using them on my own list.

Some screenshots

Showing "Uncheck all items" and "Delete Checked" are hidden (the latter) when searching. I also greyed out "check all items" when a search is active for consistency (happy to change this if you disagree): 

image

No search icon when less than 20 items present: 

image

Search box hidden until the magnifying glass is shown: 

image

Issue #8271, opened 2026-09-03
[BUG] - Some recipe images are no longer displayed


flipzoom:
### First Check

- [x] This is not a feature request.
- [x] I added a very descriptive title to this issue (title field is above this).
- [x] I used the GitHub search to find a similar issue and didn't find it.
- [x] I searched the Mealie documentation, with the integrated search.
- [x] I already read the docs and didn't find an answer.
- [x] This issue can be replicated on the demo site (https://demo.mealie.io/).

### What is the issue you are experiencing?

Since updating to the latest version, images are no longer displayed for some recipes. Neither in the overview nor on the details page. 

<img width="2542" height="1706" alt="Image" src="https://github.com/user-attachments/assets/148475ae-bb02-4d02-bff8-805ff985b9ad" />

The problem also occurs on the demo site. Using a different browser or clearing the cache does not resolve the issue.

There are no relevant server logs. The browser console only shows the following error: 

<img width="1356" height="259" alt="Image" src="https://github.com/user-attachments/assets/24a914eb-8695-481e-b078-9800d17163df" />

### Steps to Reproduce

Open the demo and you’ll see that the recipe images are missing.

<img width="3176" height="1804" alt="Image" src="https://github.com/user-attachments/assets/0377fe20-12d0-499f-9c71-9fb5adade46f" />

<img width="3256" height="1373" alt="Image" src="https://github.com/user-attachments/assets/bcf93a61-ae2a-4280-9cfb-e4f4a281fb79" />

### Please provide relevant logs

There aren't any.

### Mealie Version

**Details**
Version: v3.25.0
Build: 932fcae7ef3c3979e7806ab2a5b09e7cb83ae48f
Anwendungsmodus: Produktivumgebung
Demostatus: Keine Demo
API-Port: 9000
API Dokumentation: Aktiviert
Datenbanktyp: sqlite
Standardhaushalt: Family
Rezept Scraper Version: 15.12.0

**Checks**
Sichere Website: Ja
Serverseitige Basis-URL: Ja
LDAP deaktiviert: Nein
OIDC deaktiviert: Nein
E-Mail konfiguriert: Ja

### Deployment

Docker (Linux)

ImanuelBertrand:
I'm working on this

ImanuelBertrand:
PR created: #8276

ImanuelBertrand:
Merged 👍

michael-genson:
[...]

PR #8276, opened 2026-09-03
fix: fix missing images / keep recipes.image in sync with the images on disk

ImanuelBertrand:

`recipes.image` is a cache key, but the frontend now reads its presence as the answer to "does this recipe have a picture?" (#8075). That turned a column which had long been allowed to drift out of sync with the filesystem into the thing that decides whether an image is rendered at all, and recipes with a picture but no key started showing the placeholder instead. Nothing is logged, because no request is made. Reported in #8271.

Reconcile the column with the disk once, and stop the four paths that were writing values it could not back up:

- `RecipeService.update_one` preserves the stored image when the payload does not mention it. Updates are a full overwrite, so any client round-tripping a recipe without echoing `image` back silently cleared it while the files stayed on disk. This is the one that reached users. An explicit `"image": null` still clears.
- `finalize_scraped_recipe` only stamps a key once a file has actually landed. It used to stamp one unconditionally and fall back to the truthy `NO_IMAGE` placeholder on failure, so a scrape that downloaded nothing still left the recipe asking the media route for a 404 on every render - the complaint in #4804 that #8075 set out to fix.
- `POST /{slug}/image` returns 400 rather than reporting success with a key for an image that was never downloaded. `RecipeDataService.scrape_image` now returns the written path so this is detectable.
- The migration importers no longer persist the source archive's path or URL as though it were a cache key; `import_image`/`scrape_image` record one once a file lands. The migrators still read `.image` off the validated recipes to locate files inside the archive, so it is dropped from the persisted copy only.

The backfill skips its clearing half, with a warning, when recipes reference images but no image files are found at all: an unmounted data volume is indistinguishable from "nothing has an image", and clearing on that reading would drop every reference in the database.

`RepositoryRecipes.update_image` now writes a real cache key instead of `randint(0, 255)` into a `String` column, since the migration importers needed exactly that primitive.

#8271 

None

## Testing

Tests succeed

## AI / LLM Assistance

Claude Code (Opus 5)

### DIFF
--- a/mealie/alembic/versions/2026-09-03-16.12.44_4b91d3a7c0e2_backfill_recipe_image_column.py
+++ b/mealie/alembic/versions/2026-09-03-16.12.44_4b91d3a7c0e2_backfill_recipe_image_column.py
@@ -1,0 +1,91 @@
+"""backfill recipe image column from disk
+
+Revision ID: 4b91d3a7c0e2
+Revises: f2191b69db2e
+Create Date: 2026-09-03 16:12:44.183921
+
+`recipes.image` is a cache key, but the frontend now reads its presence as the answer to
+"does this recipe have a picture?". That made rows which had drifted out of sync with the
+files on disk suddenly visible: a recipe with an image but no key renders the placeholder
+and never requests the file, and a recipe with a key but no image asks the media route for
+a 404 on every render. This reconciles the column with what is actually on disk, once.
+
+"""
+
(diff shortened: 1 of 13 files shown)


p0lycarpio (approved):
Tested with the latest commit 3ad8f12cf349cc624e75f04050045a4961cdc04a and the `directory_name` method. I confirm that it works. Thanks for the fix!

PR #8281, opened 2026-09-03
fix: QueryFilter improvements

p0lycarpio:

_(REQUIRED)_

This PR improves the filter search in several ways :

- On the recipe search page, you no longer need to select an ingredient to apply other filters. You can now use the “advanced” filters to find recipes even when you don't necessarily want to create a dedicated cookbook.
- The rating filter has been improved following #7966. As the only numerical filter, the rating ranges from 0 to 5.
- Warnings and deprecations have been fixed.

_(REQUIRED)_

N/A

## AI / LLM Assistance

_(REQUIRED)_

None


michael-genson (approved):
LGTM

PR #8285, opened 2026-09-04
feat: Add to shopping list improvements

miah120:
This PR introduces a setting for favorite (aka default) shopping list. This list will be assumed to be the one to add to when using the add to list option from a recipe or meal plan. 

[Screencast_20260904_110651.webm](https://github.com/user-attachments/assets/cfc0641d-7e00-49a0-bd18-71ad00058789)

Users can set their favorite via the shopping lists page: 
image

By default, no list is selected, and everything matches current behavior. 

Additionally, if no lists are available when a user goes to add a recipe to a shopping list, they will be able to create one from the dialog. 

[Screencast_20260904_111000.webm](https://github.com/user-attachments/assets/9af62119-62c7-41c5-81a8-de13fbb9bea2)

NA

## AI / LLM Assistance
No AI used


mealie/db/models/users/users.py
```diff
@@ -105,7 +105,13 @@ class User(SqlAlchemyBase, BaseMixins):
     mealplans: Mapped[list["GroupMealPlan"]] = orm.relationship(
         "GroupMealPlan", order_by="GroupMealPlan.date", **sp_args
     )
-    shopping_lists: Mapped[list["ShoppingList"]] = orm.relationship("ShoppingList", **sp_args)
+    shopping_lists: Mapped[list["ShoppingList"]] = orm.relationship(
+        "ShoppingList", **sp_args, primaryjoin="User.id==ShoppingList.user_id"
+    )
+    favorite_shopping_list_id: Mapped[GUID | None] = mapped_column(ForeignKey("shopping_lists.id"))
```
michael-genson: This creates a circular FK between `users` and `shopping_lists`. `AIProviderSettings.default_provider_id` (`ai_providers.py:111`) has the same shape and uses `use_alter=True` on the FK plus `post_update=True` on the relationship — can we match that? A test that deletes a favorited list and checks the favorite goes null would be good too.

mealie/schema/user/user.py
```diff
@@ -116,6 +116,7 @@ class UserBase(MealieModel):
     group: str | None = None
     household: str | None = None
     advanced: bool = False
+    favorite_shopping_list_id: UUID4 | None = None
```
michael-genson: Since this is on the user update payload, any list id is accepted right now. Shopping lists are household-scoped, so favoriting another member's list in your own household is fine, but can we reject an id from a different household? A test for that case too, please.

frontend/app/pages/shopping-lists/index.vue
```diff
@@ -162,6 +166,17 @@ const { data: shoppingLists } = useAsyncData(useAsyncKey(), async () => {
...
 
+function updateFavorite(id: string) {
+  if (!auth.user.value) return;
+  if (auth.user.value.favoriteShoppingListId === id) {
+    auth.user.value.favoriteShoppingListId = null;
+  }
+  else {
+    auth.user.value.favoriteShoppingListId = id;
+  }
+  userApi.users.updateOne(auth.user.value?.id, auth.user.value);
```
michael-genson: This isn't awaited and nothing handles a failure, so the heart stays filled even if the save didn't go through. Can we await it and revert the local change if it fails?

michael-genson (changes requested):
Thanks! A few things below. This will also need a rebase — the migration's `down_revision` is behind `mealie-next` now.

PR #8289, opened 2026-09-04
fix: clarify migrations are for other applications, not Mealie upgrades

### LINKED ISSUE #6114: [BUG] - Phrasing for "Mealie Pre v1.0" migrations in docs and UI
### First Check

Hello! I was looking for how to migrate recipes between Mealie instances, and it seems the UI and the docs assume that the only kind of supported migration is from pre v1.0 instances. That makes sense because that was the only supported migration at that time, but it seems the zip file format is pretty stable now.

The easy fix would be to change the UI and docs to imply that _any_ version of mealie can migrate to any newer version, but that might not be 100% true, so I wanted to put the question to the community of what would be accurate phrasing :)

### Steps to Reproduce

1. Go to https://demo.mealie.io/group/migrations/ or https://docs.mealie.io/documentation/getting-started/migrating-to-mealie-v1/
2. Pretend you're clueless like me and search for a way to import data from my v3 mealie instance 😅

n/a

### Mealie Version

v3.1.2

### Deployment

Docker (Linux)

_No response_

henry1113nz:

In Mealie, "migration" means importing data from a *different application*. Moving
data between Mealie instances, or upgrading one, is a backup and restore. Nothing on
the migrations page or in the docs says so, and the migration type selector defaults
to "Mealie Pre v1.0", so someone looking for a way to move data from one current
Mealie instance to another lands on the migration tool and reasonably assumes it is
the right place. That is what happened in #6114.

This PR does not change any behaviour, only wording.

- `frontend/app/lang/messages/en-US.json`
    - `migration.recipe-data-migrations-explanation` — the migrations page subtitle now
      points readers at the backup and restore tools for Mealie-to-Mealie transfers.
    - `migration.mealie-pre-v1.description-long` — states that this option is only for
      instances older than v1.0, and that a v1.0+ backup should be restored instead.
- `docs/docs/documentation/getting-started/migrating-to-mealie-v1.md`
    - Adds an admonition at the top of the guide describing which versions it covers,
      following @michael-genson's explanation in the issue.

Only `en-US.json` is edited; the other locales are managed by Crowdin.

Text before and after (the change is text-only, so there is nothing visual to show):

| string | before | after |
| --- | --- | --- |
| migrations page subtitle | "...This is a great way to get started with Mealie." | "...Moving data between Mealie instances, or restoring an earlier Mealie backup, is done with the backup and restore tools instead." |
| pre-v1.0 option | "...Note that only recipes can be imported from the export." | "...This applies only to instances older than v1.0. A backup taken from v1.0 or later should be restored with the backup and restore tools." |

Fixes #6114

**One wording question I did not want to decide on my own.** There is a third string
that contributes to the same confusion:

```
"coming-from-another-application-or-an-even-older-version-of-mealie":
  "Coming from another application or an even older version of Mealie? ..."
```

"an even older version of Mealie" is vague in exactly the way the issue describes, and
"a version of Mealie older than v1.0" would be precise. I left it alone because the key
name is the English sentence itself: changing only the value makes the key misleading,
while renaming the key means touching `EndPageContent.vue` and invalidating the existing
translations for that string. Both seem like your call rather than mine. Happy to add
either version if you have a preference.

Separately, I noticed `migration.mealie-text`, `migration.paprika-text`,
`migration.nextcloud-text` and `migration.chowdown-text` duplicate the corresponding
`description-long` strings and have no remaining references in the frontend. I have left
them untouched to keep this PR to the reported issue, but I am happy to open a separate
PR removing them if that would be useful.

## Testing

- `mkdocs build` completes with no new warnings, and the new link resolves to
  `../usage/backups-and-restoring/`; I confirmed the target page is generated and that
  the admonition renders as intended.
- `en-US.json` still parses, and the added text is ASCII only, matching the rest of the
  file (my first draft used an em dash escape, which I replaced).
- No behavioural change, so no tests were added. The strings are referenced from
  `frontend/app/pages/group/migrations.vue`, which reads them through `$t` and is
  unaffected by their content.
- The `--strict` docs build aborts on a pre-existing `demo_url` config warning, unrelated
  to this change.

## AI / LLM Assistance

I chose this issue and read the discussion myself, and I confirmed against current
`mealie-next` that the wording described in the issue is still present before starting.
An LLM was used to locate the relevant strings and their usages, to draft the replacement
wording and the docs admonition, and to run the verification described above. I reviewed
the resulting diff, and the decisions about scope — which strings to change, and which to
raise with you instead of changing — were mine.


henry1113nz:
Rebased this onto current `mealie-next`.

There's one decision in here I deliberately left to you rather than guessing at, and looking
back I don't think I flagged it clearly enough when I opened this.

There's a third string that feeds the same confusion #6114 describes:

    "coming-from-another-application-or-an-even-older-version-of-mealie":
      "Coming from another application or an even older version of Mealie? ..."

"an even older version of Mealie" is vague in exactly the way the issue is about - "a version
of Mealie older than v1.0" would say what's actually meant. I left it alone because the key
name *is* the English sentence: changing only the value leaves the key describing text that no
longer exists, and renaming the key means touching `EndPageContent.vue` and invalidating the
string across the 42 locale files that currently carry it.

Either of those seemed like your call rather than mine. Happy to make the change whichever way
you'd prefer, or to leave it as is if the current wording doesn't bother you.

frontend/app/lang/messages/en-US.json
```diff
@@ -364,7 +364,7 @@
       "title": "DVO Cook'n X3"
     },
     "recipe-data-migrations": "Recipe Data Migrations",
-    "recipe-data-migrations-explanation": "Recipes can be migrated from another supported application to Mealie. This is a great way to get started with Mealie.",
+    "recipe-data-migrations-explanation": "Recipes can be migrated from another supported application to Mealie. This is a great way to get started with Mealie. Moving data between Mealie instances, or restoring an earlier Mealie backup, is done with the backup and restore tools instead.",
     "coming-from-another-application-or-an-even-older-version-of-mealie": "Coming from another application or an even older version of Mealie? Check out migrations and see if your data can be imported.",
```
michael-genson, line 368: Good catch, thanks for flagging it. Let's rename the key to something generic like `migration.migration-description` and update the value at the same time (e.g. "a version of Mealie older than v1.0"). We're moving toward generic key names, even though a lot of the app still uses sentence-style keys. You'll need to update the usage in `EndPageContent.vue` too. Don't worry about invalidating the other locales; Crowdin will pick the new string up.

michael-genson (changes requested):

frontend/app/lang/messages/en-US.json
henry1113nz: > Good catch, thanks for flagging it. Let's rename the key to something generic like `migration.migration-description` and update the value at the same time (e.g. "a version of Mealie older than v1.0"). We're moving toward generic key names, even though a lot of the app still uses sentence-style keys. You'll need to update the usage in `EndPageContent.vue` too. Don't worry about invalidating the other locales; Crowdin will pick the new string up.

Done - renamed to `migration.migration-description` and updated the usage in
`EndPageContent.vue`. The value now reads "Coming from another application, or a version of
Mealie older than v1.0? Check out migrations and see if your data can be imported."

I left the old key in the other locale files rather than stripping it, since you mentioned
Crowdin will pick the new string up.

Also rebased onto current `mealie-next`. I put the rename in a separate commit so the change
since your review is easy to see.

henry1113nz:
Done. Renamed it to `migration.migration-description` and updated the usage in
`EndPageContent.vue`. The text now reads "Coming from another application, or a version of
Mealie older than v1.0? Check out migrations and see if your data can be imported."

Left the old key in the other locale files since you said Crowdin would pick the new one up.

I also rebased onto current `mealie-next`, and put the rename in its own commit so it's easy
to see what changed since your review.

michael-genson (approved):
Thanks for the rename and the clearer wording!

PR #8292, opened 2026-09-05
feat: convert between metric/imperial units

michael-genson:

_(REQUIRED)_

Enables toggling between as-written, metric, and imperial units:
Screenshot 2026-09-04 at 9 49 52 PM

Translation info is generated through `pint` which is what the backend uses for unit management. This relies on ingredients being parsed and standardized (most units are automatically standardized, but users can manage standardization manually from the data management page as well).

This PR also copies the backend seed data to the frontend during Docker build time so we can leverage the unit translations. This also paves a pattern for future food translation, too.

_(REQUIRED)_

Closes https://github.com/mealie-recipes/mealie/discussions/3881
Replaces https://github.com/mealie-recipes/mealie/pull/8202

This PR intentionally makes no attempt to mess with anything in the instructions and focuses only on parsed ingredients. This can be a future enhancement if we want, but seems like a pain.

## Testing

Manually tested on local dev and in a prod build to double check the seed data copying works as expected.

## AI / LLM Assistance

_(REQUIRED)_

Heavy Claude implementation.


p0lycarpio:
Great work! I wasn't able to thoroughly test the feature because I only use the metric system in my recipes, but from my perspective, it works.
For users like me who use only one unit, would it be possible to disable the conversion at the group or instance level? If not, the temporary workaround is to unset the standardization data for each unit.

michael-genson:
Thanks! Right now there is no setting, but the default behavior is no conversion (i.e. "As written"). We could add a preference or something to hide the switch, but I'm not sure it's worth the trouble.

dev/code-generation/unit_systems.py
```diff
@@ -0,0 +1,268 @@
...
+
+
+class UnitDimension(StrEnum):
+    MASS = "mass"
+    VOLUME = "volume"
+
+
+class UnitSystem(StrEnum):
+    METRIC = "metric"
+    US = "us"
```
p0lycarpio, line 36: I think “Imperial” would be more appropriate.

frontend/app/components/Domain/Recipe/RecipeUnitSystemButton.vue
```diff
@@ -0,0 +1,82 @@
+<template>
+  <v-menu
+    v-if="hasConvertibleIngredients"
```
p0lycarpio: What do you think about disabling the button instead of hiding it when it's not applicable?

p0lycarpio (commented):
I think a button could be added later on, because I'm convinced that many users don't need to convert their recipes. Maybe we should ask discussion #3881.
Just a couple of comments ;)

frontend/app/components/Domain/Recipe/RecipeUnitSystemButton.vue
michael-genson: Yeah I think that's better

dev/code-generation/unit_systems.py
michael-genson: There's actually a real (albeit silly) reason for this: technically "Imperial" is a different measuring system. When people say "Imperial" they actually mean "US Imperial". The distinction is important because Pint (where we're getting our conversion data from" does _actually_ have all three (metric, imperial, and US imperial).

Realistically no one uses imperial (as opposed to US imperial) but internally it's worth keeping consistent.

michael-genson:
Posted to https://github.com/mealie-recipes/mealie/discussions/3881, we'll see what people think. I also won't use this on my instance but maybe some of my users will ¯\\\_(ツ)_/¯

fbd-ss:
I wouldn't mind a task that can go through all recipes and convert to selected measurement unit, either interactively or unattended.  I disagree most people won't need this, I am betting anyone who downloads from url will use this.  Having a choice to convert per recipe would be better than per item in my use case.

michael-genson:
This is automatic as long as you define your unit conversions (which is automatically done by Mealie in most scenarios). You just need to parse your recipes if you aren't already.

p0lycarpio (dismissed):
LGTM

michael-genson:
Made a couple changes:
- implemented some smart rounding to make converted values look more "human"
- forced everything to be abbreviated when converting (since that's convention) as long as the translation is available

PR #8297, opened 2026-09-05
fix: filter related fields with EXISTS instead of joins

### LINKED ISSUE #7828: [BUG] - display of cookbook sometimes shows only a subset of the recipes selected by the filters
### First Check

Cookbooks do not always display all the recipes that should be shown according to the set filters. When this happens the recipes actually displayed are a subset, varying according to sort criteria (ie if alphabetical, the first 70 or so starting A are shown, if reverse alphabetical, a similar number from ZZZZ), the number of recipes shown varies. This seems to happen sometimes when the number of recipes in a cookbooks is large. It also seems more likely to occur when the cookbook filters exclude only a small proportion of all recipes.

### Steps to Reproduce

1. Create a cookbook which should show very many recipes.
2. Check if the number of recipes is correct.
3. If the number is not correct, change to sort order and notice that the number of and actual recipes displayed change.
4. If this doesn't work, try again!

I have set up examples on the demo site.

### Mealie Version

Version
[v3.20.1](https://github.com/mealie-recipes/mealie/releases/tag/v3.20.1)
Build
[a562409e43964d3734eabd64e4a2fd64bffaab4d](https://github.com/mealie-recipes/mealie/commit/a562409e43964d3734eabd64e4a2fd64bffaab4d)

It was an issue on v3.19 and v3.20.

### Deployment

Docker (Synology), Other (please specify below)

Portainer on Synology NAS.

derekdiliu:

Cookbooks can silently show only a fraction of the recipes their filters select, and some filters return nothing at all. Both come from how `QueryFilterBuilder` handles attribute strings that traverse a relationship, such as `tags.name`.

For each filter component the builder called `query.join(...)` on the relationship. A join against a to-many relationship returns one row per related record, so a recipe with three matching tags occupied three rows. `LIMIT` and `OFFSET` then counted those duplicates. The total is computed from a `DISTINCT` subquery and stays correct, which is why the reported count looks right while the page itself is short, and why the number of visible recipes changes with the sort order.

Two components filtering the same relationship joined the same table twice with no alias, so both conditions had to hold on one joined row. `tag is A AND tag is B` could never match anything.

Filter elements are now wrapped in `EXISTS` subqueries, one per relationship traversed by the attribute string. Each component becomes independent and no rows are duplicated, so `LIMIT`/`OFFSET` operate on distinct records. Ordering still joins, because ordering by a related column needs the join.

The pagination code itself is untouched. Adding `DISTINCT` there was the other obvious option, but it breaks on Postgres wherever a search orders by an expression that is not in the select list, which is the case for fuzzy search.

Fixes #7828

Two spots deserve a close look.

`NOT IN` used to build its own subquery off the join to express "none of the related rows match". That subquery depended on the join being present, so it is now a negated `EXISTS`. The meaning is unchanged but the implementation is entirely different.

`CONTAINS ALL` already built one `.any()` per value by hand. It now goes through the same wrapper, which keeps a separate `EXISTS` per value and avoids nesting one inside another.

This changes the shape of the SQL for every filtered list endpoint, not just cookbooks. I have not benchmarked `EXISTS` against the previous joins; on the reporter's roughly 1300 recipe library this is unlikely to matter, but you may have a view on larger instances.

## Testing

Two regression tests in `tests/unit_tests/repository_tests/test_pagination.py`, one per symptom. Both fail on `mealie-next` and pass with this change:

- `test_pagination_filter_multi_tag_page_is_not_short` asks for 5 records and gets 2 before the fix, with `total` reporting 5 either way.
- `test_pagination_filter_two_tag_conditions_anded` returns nothing before the fix.

Worth noting that almost every existing filter test in that file uses `per_page=-1`, which skips `LIMIT` entirely, so the pagination half of this bug was invisible to the suite.

Full backend suite on SQLite: 2788 passed, 16 skipped. Repository tests on Postgres 15: 159 passed. I could not complete a full Postgres run locally, so I am relying on CI for that matrix.

## AI / LLM Assistance


michael-genson:
Looks good, two tests I’d like to add:
- a filter on a to-one relationship, since `_wrap_in_relationships` branches on uselist and both new tests only exercise the `.any()` side.
- a `NOT IN` case on a related field. A recipe with two tags where only one is excluded would pin the “none of the related rows match” semantics.

derekdiliu:
Both added in 296a5c6. The to-one case filters shopping list items by `label.name`, since `ShoppingListItem.label` is a nullable to-one and so goes through `.has()`. The `NOT IN` case is the one you described.

Both also pass against `mealie-next`, which is deliberate: they pin behaviour the refactor has to preserve rather than behaviour it introduces.

michael-genson (approved):
Thanks so much!

PR #8309, opened 2026-09-06
fix: coerce null food/unit description to empty string

### LINKED ISSUE #8308: [BUG] - A single food with a NULL description crashes the entire foods list endpoint (500), silently shown as "no data"
### First Check

`IngredientFood.description` (mealie/schema/recipe/recipe_ingredient.py, `UnitFoodBase.description: str = ""`) requires a non-null string when serializing a food for API responses. But the underlying DB column, `ingredient_foods.description`, is nullable (`mapped_column(String)` with no `nullable=False`, see mealie/db/models/recipe/ingredient.py).

If any single row in a group's `ingredient_foods` table has `description = NULL`, then `GET /api/foods` (and any other endpoint that lists/paginates foods for that group) throws a Pydantic ValidationError while building the response list, because `repos/repository_generic.py::page_all` validates every row into the schema in one list comprehension — one bad row raises for the whole batch, returning a 500 for the entire request.

The frontend's food store (`frontend/app/composables/store/use-food-store.ts` via `use-store-factory.ts`) does not surface this error to the user in any way. The Data Management > Foods page just renders "No data available" / "0-0 of 0", identical to what an empty table looks like, giving zero indication that anything is wrong. This is very easy to mistake for lost data, since ALL foods in the group disappear from the list at once, not just the one bad row.

How a NULL description can legitimately end up in the table: `AlchemyExporter.restore()` (mealie/services/backups_v2/alchemy_exporter.py) restores a full database JSON dump directly via SQLAlchemy Core, bypassing the Pydantic `CreateIngredientFood` schema (which is what normally forces `description` to default to `""`). Any backup/dump that has `description: null` for a food row (e.g. from an older schema version, a partial export, or one produced by external tooling) will reproduce this crash silently after restore.

### Steps to Reproduce

1) Get direct DB access (or produce a backup/dump file) for a Mealie instance.
2) Set any single row's ingredient_foods.description to NULL (e.g. UPDATE ingredient_foods SET description = NULL WHERE id = '<some-food-id>';), or restore a backup containing such a row.
3) Open Data Management > Foods in the UI, or call GET /api/foods.
   -> Observe: the entire foods list for that group returns 500 / renders as "No data available", even though most rows are fine.

500 Internal Server Error. Backend traceback:

pydantic_core._pydantic_core.ValidationError: 1 validation error for IngredientFood
description
  Input should be a valid string [type=string_type, input_value=None, input_type=NoneType]

Raised from mealie/repos/repository_generic.py, page_all(), in:
    items=[eff_schema.model_validate(s) for s in data],

### Mealie Version

mealie-next

### Deployment

Other (please specify below)

Reproduced in a local development environment (task py + task ui running natively, Postgres/Mailpit via docker-compose.dev.yml). Confirmed via direct API call (GET /api/foods) returning 500, and traced to the exact Pydantic validation error above. Fixed locally for testing by setting the offending row's description back to an empty string, after which the endpoint returned all rows normally.

chiliec:

The `ingredient_foods.description` and `ingredient_units.description` DB columns are nullable (`mapped_column(String)`, no `nullable=False`), but the shared `UnitFoodBase.description` schema field is typed `str` (default `""`). So if any single row has `description = NULL`, validating it raises a Pydantic `ValidationError`.

Because `page_all` validates every row into the schema in one comprehension, one bad row raises for the **whole** batch — `GET /api/foods` (and the units list, and any paginated foods/units endpoint) returns a 500. The frontend food store surfaces this as an empty "No data available" list, so it looks like all foods silently vanished.

A NULL can legitimately land in the column via `AlchemyExporter.restore()`, which writes dump rows through SQLAlchemy Core and bypasses the `CreateIngredientFood` schema that otherwise defaults `description` to `""`.

- `mealie/schema/recipe/recipe_ingredient.py`: add a `description` before-validator on `UnitFoodBase` that coerces `None` to `""`. This covers both foods and units, since both inherit `UnitFoodBase` and both DB columns are nullable. Non-null values pass through unchanged, so the serialized output is unchanged for every existing row.

Fixes #8308

## Testing

Added `tests/unit_tests/schema_tests/test_ingredient_food.py` covering:
- food with `NULL` description validates to `""`
- food preserves a non-null description
- unit with `NULL` description validates to `""`

Ran with the project's Python 3.12 + pydantic 2.13.4:

```
$ python -m pytest tests/unit_tests/schema_tests/test_ingredient_food.py
3 passed
```

Verified genuine RED→GREEN: reverting the new validator makes the two null-description tests fail with the exact `Input should be a valid string [input_value=None]` error from the report; restoring it makes them pass. `ruff check` and `ruff format --check` are clean on both files.

## AI / LLM Assistance

An LLM assisted in drafting the validator and the regression test. The root-cause analysis, the RED→GREEN verification, and the ruff/pytest runs above were all executed and reviewed before submitting.


chiliec:
Heads-up on CI: the `End-to-End Tests / test` job is red, but the failure is in the OIDC login e2e spec (`login-oidc-sequential-login-webkit`) — the OAuth flow times out waiting to redirect from `/login/?code=...` to `/g/home`, unrelated to this PR. This change only adds a backend Pydantic `description` validator + a schema unit test and touches no auth/OIDC/frontend code. Both `Backend Server Tests / tests (sqlite)` and `(postgres)` — which run the new test — passed. Looks like a flaky OIDC run; a re-run should clear it.

michael-genson (approved):
Thanks!

PR #8327, opened 2026-09-07
Fix recipe scraper issue

svathsa21:
## Description

Fixed the recipe scraper issue that related to recipe step ingredient references.

### Changes made
- Updated Recipe Step schema to support ingredient references correctly.
- Updated the scraper cleaner logic.
- Added unit tests for the cleaner functionality.
- Ran the project pre-commit checks successfully, including Ruff formatting and checks.

###Testing
- Added tests covering the updated scraper cleaning behaviour.
- Ruff checks and formatting passed successfully.


michael-genson (commented):
Doesn't this now skip the instructions cleaner?

Rouzax:
Thanks for picking this up, it has been open a while. I have been in the same code this week, so three things that might save you a round trip:

1. Setting `title` on every step of a section will render the heading above every step rather than once per section. `RecipePageInstructions.vue` draws a section banner for each step where `hasSectionTitle(step.title)` is true, and `toggleCollapseSection` treats a titled step as the start of a new section, so the heading needs to land on the first step of its section only.

2. The test file added here is empty and its path is `tests/unit_tests/services_testsscraper_tests/` (missing a separator before `scraper_tests`), so nothing in it runs.

3. Two shapes get lost by the rewritten comprehension: `itemListElement` is sometimes a single step dict rather than a list (rezeptwelt.de, and the fixture already in `tests/data/html/`), and sites mix loose steps with sections in the same `recipeInstructions` list, which the `HowToSection` pattern does not match at all.

Worth knowing before you invest more in this: even with the cleaner fixed, a sectioned recipe imported by URL or by "Import from HTML or JSON" is unchanged, because `clean_scraper()` calls `scraped_data.instructions()` before the cleaner ever sees the structure, and `recipe_scrapers` flattens sections into text at that point (emitting the section name as its own line). The cleaner's `HowToSection` branch is only reachable from the migration importers. I wrote up the details in #6887.

I opened #8331 with both layers and tests; happy to close mine in favour of yours if you would rather carry it.

michael-genson:
Hey there, thanks for picking this up! I'm going to close this one in favor of #8331, which fixes both places the headings get dropped (the cleaner, and the scraper flattening sections before the cleaner sees them) and includes tests.

PR #7785, opened 2026-06-21
fix: preserve identity fields and fix slug when saving JSON-edited recipes

audiophonicz:
## What this PR does / why we need it:

Fixes several silent failure modes when saving a recipe whose JSON was pasted into the editor without full identity/organizer fields (e.g. AI-generated recipes).

- `RecipePage.vue` — use `route.params.slug` (URL slug) instead of `recipe.value.slug` for the PUT endpoint; `recipe.value.slug` is `undefined` when pasted JSON omits it, causing saves to go to `PUT /api/recipes/undefined`
- `repository_recipes.py` — preserve `id`, `user_id`, `household_id`, `group_id` from the existing DB entry on update; generate slug from name if missing; resolve `tags`, `recipe_category`, and `tools` by name within the group, creating them if they don't exist, so organizers don't require a pre-populated `id`/`slug`/`groupId`
- `cleaner.py` — preserve `title` and `summary` on instruction steps instead of stripping them
- `scraper_strategies.py` — read `title` and `summary` from instruction dicts instead of hardcoding empty strings
- `recipe.py` — make `slug` optional on `RecipeTag` (auto-generated from name); make `id` optional on `RecipeTool`

## Which issue(s) this PR fixes:

Fixes #6887
Fixes #6406
Fixes #5865
Fixes #4799

## Testing:

Verified on v3.19.2 by pasting AI-generated JSON with no `id`, `slug`, `groupId`, `userId`, `householdId`, and with categories/tags/tools containing only `name` fields. Recipe saves correctly, all fields persist on refresh, and organizers are created automatically.

<details>
<summary>JSON used for testing:</summary>

```json
{
  "name": "Smoked Brisket Test",
  "description": "Test recipe to validate all previously untested import fields: categories, tags, tools, ingredient section headers, and markdown in instruction text.",
  "recipeServings": 12,
  "recipeYieldQuantity": 0,
  "recipeYield": "12 servings",
  "totalTime": "14 hours",
  "prepTime": "30 minutes",
  "cookTime": null,
  "performTime": "13 hours 30 minutes",
  "recipeCategory": [
    { "name": "Dinner" },
    { "name": "BBQ" }
  ],
  "tags": [
    { "name": "smoked" },
    { "name": "beef" },
    { "name": "BBQ" },
    { "name": "low-and-slow" }
  ],
  "tools": [
    { "name": "Smoker" },
    { "name": "Meat thermometer" },
    { "name": "Butcher paper" }
  ],
  "rating": null,
[...]

Issue #6406, opened 2025-10-20
[BUG] - JSON Import of recipeInstructions Issues


rjgrandy:
### First Check

- [x] This is not a feature request.
- [x] I added a very descriptive title to this issue (title field is above this).
- [x] I used the GitHub search to find a similar issue and didn't find it.
- [x] I searched the Mealie documentation, with the integrated search.
- [x] I already read the docs and didn't find an answer.
- [x] This issue can be replicated on the demo site (https://demo.mealie.io/).

### What is the issue you are experiencing?

I used Gemini to convert a blue apron recipe to JSON for import. I can't get mealie to use the "summary" names that Gemini uses in the JSON. Also the markdown I use in the "text" field gets removed when I import. If I go back and edit the recipe, show the json and then update those areas, the recipe shows appropriately.

Is there something I am not doing right?

My JSON is below:

```
{
  "name": "Black Pepper Beef with Bok Choy & Garlic Rice",
  "slug": "black-pepper-beef-with-bok-choy-garlic-rice",
  "image": null,
  "recipeServings": 2,
  "recipeYieldQuantity": null,
  "recipeYield": "",
  "totalTime": "25-35 minutes",
  "prepTime": null,
  "cookTime": null,
  "performTime": null,
  "description": "Black pepper is a go-to seasoning in countless recipes, but it takes center stage in tonight's beef stir-fry. We're combining coarsely ground black pepper with sweet chili sauce and citrusy ponzu, for aromatic flavor with just the right amount of kick. Crisp bok choy and bell pepper help round out the stir-fry, served over a bed of garlic-seasoned rice.",
  "recipeCategory": [],
  "tags": [],
  "tools": [],
  "rating": null,
  "orgURL": "",
  "dateAdded": null,
  "dateUpdated": null,
  "createdAt": null,
  "updatedAt": null,
  "lastMade": null,
  "recipeIngredient": [
    {
      "quantity": 10,
      "unit": {
        "name": "ounce",
        "pluralName": "ounces",
        "description": "",
        "extras": {},
        "fraction": true,
        "abbreviation": "oz",
        "pluralAbbreviation": null,
        "useAbbreviation": false,
        "aliases": []
      },
      "food": {
        "name": "thinly sliced beef",
        "pluralName": "thinly sliced beef",
        "description": "",
        "extras": {},
        "aliases": [],
        "label": null
      },
      "note": null,
      "display": "10 ounces thinly sliced beef",
      "title": null,
[...]

PR #8331, opened 2026-09-07
fix: keep instruction section and step headings when importing

Rouzax:

Recipes whose instructions carry headings lose them on import. A `HowToSection` grouping steps loses the grouping and each section heading arrives as a bogus instruction step, and a step that names itself loses that name. `RecipeStep` already has fields for both, `title` for the section heading ("This is the section title!!!" in `mealie/schema/recipe/recipe_step.py`) and `summary` for the step's own heading, and the recipe page already renders both, so all that was missing was carrying them across.

There are two places they get dropped, and fixing only one of them changes nothing for the import paths people actually use.

- `mealie/services/scraper/cleaner.py` - `clean_instructions()` flattened every `HowToSection` into a bare list of steps and discarded `name`. The section branch is now reached whenever any entry in the list is a section rather than only when the first one is, because sites do mix loose steps and sections in the same list (a recipe that opens with a couple of ungrouped steps and only then starts grouping them); the old pattern matched such a list as a plain list of steps and silently dropped every section in it. It now keeps the name as `title` on the **first** step of each section, which is how the frontend groups the steps that follow, and fills a step's `summary` from either the `summary` key Mealie writes itself or schema.org's `HowToStep.name`. Sites fill that name with a copy of the step text as often as with a real heading, so a name the text already opens with is dropped (the rule recipe_scrapers applies to the same field, compared case insensitively and after decoding entities), and so is a name truncated mid word, which is what yummly.com publishes ("Step 1: Preheat oven to 425…"). Section names are run through `clean_string()` like any other scraped string, and are read from `Name` as well as `name`, which is the spelling atelierdeschefs.fr publishes. The single element base case was removed because the list case below it already matches a one element list, and returning the object untouched skipped sanitization. Two shapes that used to raise `TypeError` are now handled: a lone section passed as a bare dict, and a single step passed as a bare dict (rezeptwelt.de publishes recipes this way, which is why `tests/utils/recipe_data.py` carried a "can't parse steps" comment; that recipe now imports as one real step titled "Waffelteig" instead of a stringified Python dict).
- `mealie/services/scraper/scraper_strategies.py` - `clean_scraper()` asked `recipe_scrapers` for `instructions()` first, which renders instructions as plain text and emits each section name as a line of its own, so the cleaner never saw a `HowToSection` on the URL or the html-or-json path at all. Both parses now run and `prefer_structured_instructions()` picks between them, and `RecipeStep(title="", ...)` no longer hardcodes the headings away.

The picking rule is the part worth reviewing. The structured parse is used only when it gains a heading, and only when every piece of text the scraper's own parse produced is still accounted for, as either a step or a heading. That second condition matters because a site specific scraper reads the instructions off the page instead of out of the JSON-LD, and its text can differ from, or beat, what the page publishes: Cookomix, Magimix, SunBasket and ThePlantBasedSchool all produce better text than their own structured data. The rule keeps the structured data a refinement of the same content and never a replacement for better content. The single exception is a step dict that reached us as its own `repr`, which happens when a site nests one step where the spec wants a list (hhursev/recipe-scrapers#2006); that is not instruction text, so it does not count as content to preserve.

### Before

Importing a recipe with four `HowToSection`s (cookbook.pfeiffer.net.au/biryani-chicken-rajasthani) turns every heading into a step of its own, so "to marinate" is step 1 and the real first step is step 2:

The same thing in the JSON from #6887, pasted into "Import from HTML or JSON":

```
[{"text": "Prep Work"}, {"text": "Dice the onion."}, {"text": "Mince the garlic."}, {"text": "Cooking"}, {"text": "Brown the beef."}]
```

### After

The headings become section titles on the first step of each section, which the recipe page renders as a section banner, and the steps are numbered as the site numbers them:

Same import, as JSON:

```
[{"title": "Prep Work", "text": "Dice the onion."}, {"text": "Mince the garlic."}, {"title": "Cooking", "text": "Brown the beef."}]
```

The same applies to a step's own heading. Pasting a schema.org recipe whose steps carry `summary`, which is what Mealie's own JSON editor and AI generated recipes produce, drops every heading and the cards fall back to "Step: N":

After, the step keeps the heading it was given:

A recipe exported from Mealie and pasted straight back in now keeps its section headings, its step headings and its text exactly, **provided the pasted JSON declares `@context`/`@type`**. Mealie's export does not, and `recipe_scrapers` rejects it as "no schema found" before any of this code runs, so the round trip in #6887 still needs that separate fix. Not addressed here.

Contributes to #6887: both instruction rows of its table, `recipeInstructions[].title` and `recipeInstructions[].summary`. Also covers the instruction half of #6406. The remaining rows of #6887 (categories, tools) and the recipe saving problems in #7785 (identity fields on update, organizers by name, the `PUT /api/recipes/undefined` slug bug) are deliberately left alone here.

The judgement call worth a second opinion is `prefer_structured_instructions()`, described above. Restricting the structured read to the generic wild mode scraper was the obvious alternative, but measuring it against the corpus showed that would give up 250 of the 258 improved pages to avoid 11 regressions, where the content comparison avoids all 11 and gives up only those same 11.

Related, not addressed here: when `instructions()` raises and `try_get_default` falls back to the raw structured data, an unparseable shape still raises `TypeError` out of `clean_scraper`. That behaviour predates this PR and is unchanged by it. Upstream `recipe_scrapers` has an open issue for the shapes that trigger it (hhursev/recipe-scrapers#2006).

There is no grouped instruction API in `recipe_scrapers` to lean on: `ingredient_groups()` exists, the instruction equivalent has been open as hhursev/recipe-scrapers#1225 since 2024, and hhursev/recipe-scrapers#2046 shows the flattening is intentional there, so reading the schema data directly is the practical route.

## Testing

- `tests/unit_tests/services_tests/scraper_tests/test_cleaner_parts.py` - fifteen new `clean_instructions` cases: named sections, the `item` key spelling, name cleaning, an empty name, a named section with no steps, loose steps mixed with sections in one list, a bare section dict wrapping a single step dict, a bare step dict, a preserved step summary, a section heading and step summary side by side, the capitalised `Name` key some sites use, a `HowToStep.name` becoming a step summary, a name that merely repeats the text being dropped, the same repetition hidden behind HTML entities, and a name truncated mid word. The shared expectation was made per case so sectioned cases can assert their own output.
- `tests/unit_tests/services_tests/scraper_tests/test_scraper_strategies.py` (new) - drives `RecipeScraperPackage.clean_scraper()` over schema.org JSON the way the html-or-json route does: section headings become step titles, headings are not imported as steps, step names and summaries survive, flat recipes are untouched, a single step section dict imports, an unparseable section falls back to the scraper's own output, and a site whose scraper splits a one step method keeps that split. `prefer_structured_instructions()` is also unit tested directly.
- `task py:test` (full suite), `task py:lint`, `task py:mypy` and `ruff format --check` all pass locally.
- Swept the whole `recipe_scrapers` fixture corpus, 1104 real pages across ~600 sites, driving each page through its own site scraper and comparing the shipped output with this branch. **255 pages gain headings (445 section headings, 541 step headings), no page loses a single piece of instruction text, and no heading merely repeats or truncates its own step.** Two earlier versions of the picking rule did lose text, on 16 and then 11 pages, and an earlier version of the heading rules produced 16 junk headings, which is what the sweep was for.
- Checked against live pages as well as the synthetic fixtures. `cookbook.pfeiffer.net.au/biryani-chicken-rajasthani` goes from 21 steps (4 of them headings) to 17 steps under 4 headings; `amateurprochef.com` (6 sections, one of them an empty placeholder with no `itemListElement`) goes from 16 steps to 10 under 5 headings, the placeholder skipped; `leukerecepten.nl` (a site with its own scraper, one section) keeps all 8 steps and gains the heading; `cookbook.pfeiffer.net.au/chicken-pepper` is the mixed list case above, going from 9 steps to 7 with both loose steps intact.

## AI / LLM Assistance

Investigation, implementation, tests and this description were produced with LLM assistance (Claude Code), reviewed by me, and validated by running the test suite, linter and type checker locally.

### DIFF
--- a/mealie/services/scraper/cleaner.py
+++ b/mealie/services/scraper/cleaner.py
@@ -150,6 +150,38 @@ def is_how_to_section(entry: typing.Any) -> bool:
     return isinstance(entry, dict) and "HowToSection" in (entry.get("@type"), entry.get("type"))
 
 
+def _step_heading(instruction: dict) -> str:
+    """Return the step's own heading, which Mealie stores as `summary`.
+
+    schema.org calls it `HowToStep.name`, but sites routinely fill that with a copy of the
+    text, or with the text truncated, so a name the text already opens with is dropped
+    rather than shown twice. This is the rule recipe_scrapers applies to the same field.
+    """
+    if summary := instruction.get("summary"):
+        return clean_string(summary)
+
+    name = instruction.get("name")
+    if not isinstance(name, str) or not name.strip():
+        return ""
+
+    # compare the cleaned forms: a name carrying HTML entities would otherwise sail past
+    # the check against text that has already had them decoded
+    heading = clean_string(name).strip()
+    if not heading:
+        return ""
+
+    # sites also truncate the text into the name (yummly's "Step 1: Preheat oven to 425…"),
+    # and a heading cut off mid word is worse than no heading at all
+    if heading.endswith(("...", "\u2026")):
+        return ""
+
+    text = clean_string(instruction.get("text") or "")
+    if text.casefold().startswith(heading.rstrip(". \u2026").casefold()):
+        return ""
+
+    return heading
+
+
 def clean_instructions(steps_object: list | dict | str, default: list | None = None) -> list[dict]:
     """
     instructions attempts to parse the instructions field from a recipe and return a list of
@@ -197,7 +229,7 @@ def clean_instructions(steps_object: list | dict | str, default: list | None = N
 
                 # a section heading lives on the first step of the section (RecipeStep.title),
                 # which is how the frontend groups the steps that follow it
-                if section_title := clean_string(entry.get("name", "")):
+                if section_title := clean_string(entry.get("name") or entry.get("Name") or ""):
                     section_steps = [section_steps[0] | {"title": section_title}, *section_steps[1:]]
 
                 instructions.extend(section_steps)
@@ -215,7 +247,7 @@ def clean_instructions(steps_object: list | dict | str, default: list | None = N
             return [
                 {"text": _sanitize_instruction_text(instruction["text"])}
                 | ({"title": instruction["title"]} if instruction.get("title") else {})
-                | ({"summary": instruction["summary"]} if instruction.get("summary") else {})
+                | ({"summary": heading} if (heading := _step_heading(instruction)) else {})
                 for instruction in steps_object
                 if "text" in instruction and instruction["text"].strip()
             ]
--- a/mealie/services/scraper/cleaner.py
+++ b/mealie/services/scraper/cleaner.py
@@ -3,7 +3,6 @@ import functools
 import html
 import json
 import numbers
-import operator
 import re
 import typing
 from datetime import datetime, timedelta
@@ -146,6 +145,11 @@ def clean_image(image: str | list | dict | None = None, default: str = NO_IMAGE)
             return [default]
 
 
+def is_how_to_section(entry: typing.Any) -> bool:
+    """schema.org groups steps with `@type: HowToSection`; some sites spell the key `type`."""
+    return isinstance(entry, dict) and "HowToSection" in (entry.get("@type"), entry.get("type"))
+
+
 def clean_instructions(steps_object: list | dict | str, default: list | None = None) -> list[dict]:
     """
     instructions attempts to parse the instructions field from a recipe and return a list of
@@ -155,14 +159,50 @@ def clean_instructions(steps_object: list | dict | str, default: list | None = N
         TypeError: If the instructions field is not a supported type a TypeError is raised.
 
     Returns:
-        list[dict]: An ordered list of dictionaries with the keys `text`
+        list[dict]: An ordered list of dictionaries with the key `text`, plus `title` on the
+        first step of a named HowToSection, which is where Mealie stores a section heading
     """
     if not steps_object:
         return default or []
 
     match steps_object:
-        case [{"text": str()}]:  # Base Case
-            return steps_object
+        case [*_] if any(is_how_to_section(step) for step in steps_object):
+            # HowToSections should have the following layout,
+            # {
+            #  "@type": "HowToSection",
+            #  "name": "Section A",
+            #  "itemListElement": [
+            #    {
+            #      "@type": "HowToStep",
+            #      "text": "Instruction A"
+            #    },
+            # }
+            #
+            # Some sites (e.g. NYT Cooking) emit empty HowToSection placeholders
+            # with no itemListElement key, or use "item" per the schema.org spec.
+            # Use .get() with both fallbacks so those sections are skipped gracefully.
+            steps_object = typing.cast(list, steps_object)
+
+            instructions: list[dict] = []
+            for entry in steps_object:
+                if not is_how_to_section(entry):
+                    # a recipe can open with a few loose steps and only then start
+                    # grouping them, so both kinds share the one list
+                    instructions.extend(clean_instructions([entry]))
+                    continue
+
+                section_steps = clean_instructions(entry.get("itemListElement", entry.get("item", [])))
+                if not section_steps:
+                    continue
+
+                # a section heading lives on the first step of the section (RecipeStep.title),
+                # which is how the frontend groups the steps that follow it
+                if section_title := clean_string(entry.get("name", "")):
+                    section_steps = [section_steps[0] | {"title": section_title}, *section_steps[1:]]
+
+                instructions.extend(section_steps)
+
+            return instructions
         case [{"text": str()}, *_]:
             # The is the most common case. Most other operations eventually resolve to this
             # match case before being converted to a list of instructions
@@ -178,6 +218,15 @@ def clean_instructions(steps_object: list | dict | str, default: list | None = N
                 for instruction in steps_object
                 if "text" in instruction and instruction["text"].strip()
             ]
+        case {"text": str()}:
+            # A single step is sometimes passed as a bare dict rather than a list of one
+            #
+            # {"@type": "HowToStep", "text": "Instruction A"}
+            #
+            return clean_instructions([steps_object])
+        case {"@type": "HowToSection"} | {"type": "HowToSection"}:
+            # Likewise, a recipe with only one section may pass that section on its own
+            return clean_instructions([steps_object])
         case {0: {"text": str()}} | {"0": {"text": str()}} | {1: {"text": str()}} | {"1": {"text": str()}}:
             # Some recipes have a dict with a string key representing the index, unsure if these can
             # be an int or not so we match against both. Additionally, we match against both 0 and 1 indexed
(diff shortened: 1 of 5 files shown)


Rouzax:
While validating this I checked the round trip #6887 asks for, exporting a recipe from Mealie and pasting it straight back into "Import from HTML or JSON". With this PR the instructions survive it exactly, section headings, step headings and text, but only if the pasted JSON declares `@context`/`@type`. Mealie's own export does not, so `recipe_scrapers` finds no schema.org Recipe and the paste fails with `BAD_RECIPE_DATA` before any of this code runs.

That looks like a separate two line fix in `create_recipe_from_html_or_json`, giving the pasted object `@context`/`@type` when it has none, and it would make "Mealie can import its own export" true. It touches schema detection for every pasted payload though, so I did not want to widen this PR on my own judgement.

Would you rather I fold that into this PR, or open it separately once this lands?

michael-genson (approved):
Thanks for the thorough write-up and the corpus sweep! Let's keep the `@context`/`@type` fix for pasted exports in a separate PR.

Rouzax:
@michael-genson before I open that follow-up PR I actually tested the two line fix, and it does not work. I had only checked half of it when I wrote the comment above, so I want to correct that, and I think the real problem is a bigger and simpler one: Mealie cannot read its own exports.

**The two keys do not fix the paste box.** I exported a recipe the way Mealie exports one, added `@context` and `@type`, and pasted it back. The instructions come through fine, which is the part I checked last time. The ingredients arrive as the raw object printed out as text, curly braces and all, and that is what you would see in the recipe:

```
ingredients: note="{'quantity': 1.0, 'unit': {'id': '7fbff8f1-...', 'name': 'cup',
                    'pluralName': 'cups', 'description': '', 'extras': {}, ...},
                    'food': {...}, 'note': 'warmed', 'display': '1 cup tomato sauce warmed',
                    'title': 'For the sauce', ...}"
```

Mealie writes each ingredient as an object, but schema.org expects a plain string, and `recipe_scrapers` turns each object into text before Mealie ever sees it. `clean_ingredients` does know how to handle objects, but it never gets the chance, because `clean_scraper` only looks at the raw data when the scraper hands back nothing, and here it hands back a list of text. Any site that writes ingredients as objects hits this today, so it is not really about our export.

**It would also break something that works now.** With an AI provider configured, pasting an unmarked export already succeeds: `recipe_scrapers` finds no recipe, the AI scraper takes over, and you get HTTP 201 and a usable recipe. Add the two keys and the normal scraper wins instead, so those users would lose a working import and get the mangled ingredients above. Without AI, the same paste gives a clean `BAD_RECIPE_DATA`. Either way the two lines make things worse.

**And the loss is much wider than ingredients.** Here is the same recipe taken through both routes, field by field:

| | zip import | paste the JSON export |
|---|---|---|
| name, description, servings | kept | kept |
| yield text, original URL | kept | lost |
| prep / cook / total times | kept | reformatted, cook time lost |
| categories, tags, tools | kept | all lost |
| nutrition | kept | kept |
| notes | kept | kept |
| recipe settings | kept | reset to defaults |
| extras | kept, but see the bug at the end | lost |
| ingredient amount, unit, food | kept | lost, replaced by the object as text |
| ingredient section heading | kept | lost |
| step heading and summary | kept | kept |
| step to ingredient links | kept | lost |
| comments | not carried | not carried |

Categories and tools are items 2 and 3 of #6887, and they stay broken with the two keys in place. That column is not a gap we can close by patching the web scraper either. It builds each step from a title, a summary and some text, so the step to ingredient links are thrown away, and ingredients get fresh internal ids, so the links could not be rebuilt anyway. Making that path faithful would mean rewriting the zip importer inside the web scraper.

**The importer we need already exists and is nearly perfect.** As the table shows, importing an exported zip keeps everything, including structured ingredients and the link from a step to its ingredient. The image in the zip survives too. It has one catch: you cannot use it on your own instance. Export a recipe as a zip and import it straight back and you get HTTP 500, `UNIQUE constraint failed: recipes.id`, because the importer reuses the id from the file and that recipe is already there. Change only the id and it imports perfectly.

So the ask in #6887, export a recipe and import it back and get the same recipe, is failing on a duplicate database id rather than on anything to do with schema.org.

**What I think the actual bug is.** Mealie offers "raw" JSON as a download in the recipe menu, and there is no way to get that file back in. The paste box sends it to the web scraper, which is built for other people's websites and throws away most of what is in our own file. We already have an importer that reads Mealie files properly, we just do not point anything at it.

So I would like to propose this instead of the two lines:

1. Give the recipe a new id when the one in the file is already taken. Small, and it makes zip export and re-import work on your own instance. I checked it is narrow: when the id is free it is kept exactly as today, so this only changes the case that currently 500s. Duplicate names already work this way, where `my-recipe` becomes `my-recipe-1`. Only one place in the code calls this importer, so nothing else depends on the current behaviour.
2. Recognise a Mealie JSON export in the paste box and send it to that same importer rather than the web scraper. Mealie would then read back both of the files it hands out. It reuses code that already forces the recipe into the current user's group and household, so the multi tenancy side is handled. It needs option 1 first. I have not built the detection yet, so this part is a proposal rather than something I have measured.

Two smaller options if you would rather keep the blast radius tiny, though neither gets us a faithful round trip:

3. Add the two keys only when the paste has `recipeIngredient` or `recipeInstructions`, and read the ingredients from the raw data when they are objects, using their `display`, `note` and `title`. That makes a pasted export readable rather than faithful, and it would have to leave AI users alone so it does not break them. It would pick up ingredient section headings, item 4 of #6887. There is precedent for reading our own fields here, since notes are already read straight from the raw data.
4. Add a second export format next to `raw` that writes proper schema.org JSON, with ingredients as plain strings and instructions as `HowToSection` and `HowToStep`. Nothing breaks and other apps could read it, but it loses detail on purpose.

I would drop the idea of marking the raw export itself, which is what I originally suggested. That file has ingredients, categories, tags and tools as objects and an internal image reference rather than a URL, so labelling it a schema.org Recipe would mean publishing something that claims to be a standard recipe and is not. While it is unmarked, nothing outside Mealie tries to read it. For reference, `recipe_scrapers` needs both keys, either one alone is still rejected.

**One unrelated bug I ran into.** Importing a zip writes `group_id` and `household_id` into the recipe's `extras`. The import walks every nested object and stamps those two fields onto each one, and `extras` is a free-form field, so they stick. Export `{'my_key': 'my_value'}` and it comes back as `{'my_key': 'my_value', 'group_id': '...', 'household_id': '...'}`. The exported file itself is clean, `settings` is fine because unknown fields get dropped there, and pasting does not do it. I could not find an existing issue for this or for the duplicate id, and I am happy to open both if you would like them tracked.

michael-genson:
Thanks for digging into this, and for correcting the earlier comment! Agreed that the two-line fix isn't worth it.

Let's do option 1 (new id when the one in the file is already taken) and the `extras` bug together in one PR. Both are small fixes to the zip importer and both come up on the same export → import round trip, so they fit in one PR. You don't need to open issues for them; linking #6887 in that PR is enough.

And if you're willing, once that's merged, open a separate PR for option 2 (routing a pasted Mealie export to the zip importer). It changes how every pasted payload is detected, so I'd rather review it on its own.

Rouzax:
@michael-genson starting on option 2 (routing a pasted Mealie export to the zip importer), and before I build the detection I'd like your call on which JSON shapes it should accept. One correction to my earlier comment first: the recipe menu has no "raw" JSON download. Its Download item is the zip, and the only raw JSON export is the API's `GET /api/recipes/{slug}/exports?template_name=raw`, which nothing in the frontend calls.

That leaves three shapes a user can realistically end up with:

- **The JSON inside the Download zip or the bulk export.** snake_case keys (`recipe_ingredient`, `group_id`), written by `model_dump_json()`.
- **The recipe page's JSON editor, or the API raw export.** camelCase keys (`recipeIngredient`, `groupId`, `orgURL`), which is `by_alias=True`. I expect this is what people actually copy and paste.
- **Hand-written or AI-generated JSON in Mealie's field names.** No ids, often only partly filled in. This is the use case behind #6887.

The options as I see them:

1. **Zip JSON only.** Smallest change, but the JSON editor copy keeps failing.
2. **Any complete Mealie recipe dump, snake_case or camelCase.** Keys are normalised to snake_case before `clean_recipe_dict`, because it only recognises snake_case today: fed camelCase, it skips the category and tag checks and can link another group's categories. Detection stays strict: no `@context`/`@type`, a `slug` and group id present, and ingredients as objects with a `referenceId`. That keeps `test_create_by_html_or_json_stream_invalid_data[missing-schema-declaration]` failing as it does now.
3. **Option 2 plus loose Mealie-shaped JSON.** The biggest win for #6887, but detection becomes guesswork, and it overlaps with plain JSON that we deliberately reject today.

My preference is 2. Whichever you pick, the paste carries no image or asset files, so `image` is cleared and `assets` and `comments` are dropped. The "Import keywords as tags" and "Import categories" checkboxes would apply the same way they do on the scraper path. If a paste is recognised as a Mealie export but fails to import, I'd return a clear error rather than fall back to the scraper or AI, since the fallback only produces the mangled result this is meant to avoid. Does 2 work for you, or would you rather keep it narrower or go wider?

PR #8332, opened 2026-09-07
fix: aggregate to-many orderings instead of joining

### LINKED ISSUE #8302: [BUG] - Ordering by a "to-many" field produces bad results
### First Check

Similar to the issue reported in https://github.com/mealie-recipes/mealie/issues/7828 and fixed by https://github.com/mealie-recipes/mealie/pull/8297, when using an `orderBy` on a "to-many" field (e.g. tag name) results will contain duplicate rows.

In practice this isn't a big deal since ordering by a "to-many" field is weird and also not directly supported in the UI, but it is technically allowable by the API so the behavior should be fixed.

We should either:
1) fix the behavior similar to how it's fixed in https://github.com/mealie-recipes/mealie/pull/8297 (ideal)
2) raise a `ValueError` when trying to order this way

### Steps to Reproduce

1) Have some recipes with some tags
2) Order by tag name (`tags.name`) via the API
3) Observe duplicate rows

N/A

### Mealie Version

nightly

### Deployment

Docker (Synology)

_No response_

derekdiliu:

Ordering a list by a field on a to-many relationship, for example `orderBy=tags.name`, returns short pages. The ordering path joined the relationship, and a join against a to-many returns one row per related record, so `LIMIT` and `OFFSET` counted duplicates. The total is computed before the ordering is applied and stays correct, so the response reports more records than it contains.

Same shape as #7828, on the path #8297 left joining because ordering needs a value to sort on and `EXISTS` cannot provide one.

Orderings that traverse a to-many relationship now use a correlated subquery reducing the related values to one per record: the lowest when ascending, the highest when descending. That is the value the ordering would have surfaced from the joined rows anyway. To-one relationships keep joining, since they cannot duplicate a record.

`get_model_and_model_attr_from_attr_string` no longer takes a `query`. Filtering stopped using it in #8297, and ordering now collects the relationships as it walks them, so nothing was left passing one in.

Fixes #8302

`aggregate_over_relationships` follows the implementation from #8312, and the ascending/descending test is that PR's design. I first wrote the subquery with `.join()` and `.correlate()` and got a global aggregate rather than a per-row one, which reordered `test_pagination_shopping_list_items_with_labels`. Putting the join conditions in the `WHERE` clause, as #8312 did, correlates properly.

Worth flagging on the test data: I could not get the short-page test to fail on `mealie-next` while every recipe carried the same tags, because ordering by tag name groups the recipes under their shared first tag and the duplicates never enter the page window. Giving each recipe its own tags makes one recipe's rows sort together, and the page then returns two records instead of five.

## Testing

- `test_pagination_order_by_to_many_field_page_is_not_short` asks for five records and gets two on `mealie-next`.
- `test_pagination_order_by_to_many_field`, parametrised over both directions, pins that ascending sorts by each record's lowest related value and descending by its highest, with unrelated records last. It passes before and after, which is intended, since the aggregate has to produce the same order the join did.

Three tests in `test_query_filter_builder.py` were updated for the dropped return value.

## AI / LLM Assistance


michael-genson (commented):
This is similar to the approach in my closed PR. The major issue I have with this is that it makes the query filter builder API way more complex. Is there a way to do this without overcomplicating the QFB API?

derekdiliu:
Fair point. The repo layer ends up importing the chain type plus two helpers, and deciding for itself whether the chain has a to-many in it. I'd fold that into one call: 
query, order_attr = QueryFilterBuilder.get_order_attr( query, attr_string, model, descending=..., column_aliases=... ) 

Chain, to-many check and the join/aggregate choice stay inside. The two helpers go private, and add_order_attr_to_query drops back to just direction and null position. While I'm in there the alias lookup and lowering can move next to _transform_model_attr, so filtering and ordering stop keeping separate copies. 

Smaller surface than this PR, and than before it. Sound right, or would you rather ordering didn't go through the builder at all?

derekdiliu:
Done in 62c1a23. The repo layer is down to one call:

```python
query, order_attr = QueryFilterBuilder.get_order_attr(
    query, order_by, self.model, descending=..., column_aliases=...
)
```

The chain type is no longer imported there, the two helpers are private, and `add_order_attr_to_query` is back to just direction and null position. The alias lookup and lowering moved next to `_transform_model_attr`, so filtering and ordering share one copy instead of two.

The ordering tests are untouched, which I think is the useful part: nothing about the external behaviour moved.

mealie/services/query_filter/builder.py
```diff
@@ -323,7 +317,73 @@ def get_model_and_model_attr_from_attr_string[Model: SqlAlchemyBase](
...
+        order_attr = cls._transform_model_attr(order_attr, order_attr.type)
+
+        if not relationships:
+            return query, order_attr
+
+        if any(uselist for _, uselist in relationships):
+            return query, cls._aggregate_over_relationships(order_attr, relationships, model, descending=descending)
+
+        # "to-one" relationships cannot duplicate records, so joining is cheaper than aggregating
+        return cls._join_relationships(query, relationships), order_attr
```
michael-genson: Route to-one through `_aggregate_over_relationships` as well and delete `_join_relationships`. The FK makes the `min`/`max` a no-op, so the branch isn't earning itself.

With it gone, `get_order_attr` drops the `query` parameter and the tuple return:

```python
order_attr = QueryFilterBuilder.get_order_attr(order_by, self.model, descending=..., column_aliases=...)
```

This also stops the fix depending on `uselist`, which is an ORM declaration rather than a schema guarantee — `RecipeModel.nutrition`, `Group.preferences` and `Group.ai_provider_settings` are all `uselist=False` over non-unique FKs.

mealie/services/query_filter/builder.py
```diff
@@ -257,12 +257,12 @@ def get_model_and_model_attr_from_attr_string[Model: SqlAlchemyBase](
...
         *,
-        query: sa.Select | None = None,
         collect_relationships: RelationshipChain | None = None,
-    ) -> tuple[type[SqlAlchemyBase], InstrumentedAttribute, sa.Select | None]:
+    ) -> tuple[type[SqlAlchemyBase], InstrumentedAttribute]:
         """
         Take an attribute string and traverse a database model and its relationships to get the desired
-        model and model attribute. Optionally provide a query to apply the necessary table joins.
+        model and model attribute. Optionally collect the relationships traversed on the way, which can then
+        be applied to a query with `join_relationships` or `aggregate_over_relationships`.
```
michael-genson: These are `_join_relationships` and `_aggregate_over_relationships` now.

michael-genson (changes requested):
Thanks, I like this much better. Two things:

derekdiliu:
Done in d28bfc0. `_join_relationships` is gone and every related ordering goes through the aggregate, so `get_order_attr` no longer takes or returns the query:

```python
order_attr = QueryFilterBuilder.get_order_attr(order_by, self.model, descending=..., column_aliases=...)
```

Docstrings point at the private names now. Good catch on `uselist`, the to-one shortcut was only as safe as the ORM declaration.

michael-genson (approved):
Looks great, thanks!

PR #8336, opened 2026-09-08
fix: auto_init merges model_config.exclude with default {"id"} instead of replacing it

### LINKED ISSUE #8088: [BUG] PUT /api/foods/{id} (and any other model with a custom model_config.exclude) can null the row's own primary key mid-update
## Summary

`auto_init._get_config()` overwrites the default primary-key exclusion (`{"id"}`) instead of
merging with it, the moment a model declares its own `model_config = ConfigDict(exclude={...})`.
Concretely, this makes **any `PUT /api/foods/{id}` that changes anything** — including a
completely ordinary full-replace update from a well-behaved client — fail with a database
integrity error, because the food's own primary key gets set to `None` partway through the
update.

## Reproduction

1. `GET /api/foods/{id}` for any existing food, note its `aliases`.
2. `PUT /api/foods/{id}` with a `CreateIngredientFood` body built the normal way — `name`,
   `pluralName`, `description`, `labelId`, and `aliases` all round-tripped from what was just
   fetched (adding one new alias to the list is enough to trigger it, but any change to any
   field reproduces this the same way).
3. Server returns `409` with a Postgres integrity error:
   ```
   {"detail":{"message":"Database integrity error","error":true,"exception":"(psycopg2.errors.NotNullViolation) null value in column \"id\" of relation \"ingredient_foods\" violates not-null constraint\nDETAIL: Failing row contains (..., null, <group_id>, <food name>, ...)"}}
   ```

Hit this for real via a client doing a completely ordinary "add a couple of aliases to an
existing food, round-trip everything else unchanged" `PUT` — not a hypothetical or a fuzzed
request.

## Root cause

`mealie/db/models/_model_utils/auto_init.py`, `_get_config()`:

```python
def _get_config(relation_cls: type[SqlAlchemyBase]) -> AutoInitConfig:
    cfg = AutoInitConfig()  # cfg.exclude starts as _default_exclusion() -> {"id"}
    cfgKeys = cfg.model_dump().keys()
    try:
        class_config: ConfigDict = relation_cls.model_config
    except AttributeError:
        return cfg
    for attr in class_config:
        if attr in cfgKeys:
            setattr(cfg, attr, class_config[attr])   # <-- replaces cfg.exclude entirely
    return cfg
```

`AutoInitConfig.exclude` defaults to `{"id"}` specifically so a row's primary key is never
touched by `auto_init`. But when a model declares its own `model_config.exclude` — usually to
keep `auto_init` from mishandling one specific relationship it wants to set up by hand —
`setattr(cfg, "exclude", class_config["exclude"])` throws the default away and replaces it
wholesale. `IngredientFoodModel` does exactly this:

```python
# mealie/db/models/recipe/ingredient.py
model_config = ConfigDict(
    exclude={
        "households_with_ingredient_food",
    }
)
```

— intending to add one exclusion, but actually ending up with `exclude = {"households_with_ingredient_food"}`, `"id"` no longer protected at all.

Then `BaseMixins.update()` (`mealie/db/models/_model_base.py`) does:

```python
def update(self, *args, **kwargs):
    self.__init__(*args, **kwargs)
    ...
```

— re-invoking the `@auto_init()`-wrapped `__init__` on an **already-persisted** row. Every
`Create*`/`Save*` schema for these models carries an `id` field that defaults to `None`
(`UnitFoodBase.id: UUID4 | None = None`), and a client normally never sends `id` in the body at
all — but Pydantic's `model_dump()` includes every declared field regardless, so `kwargs["id"]`
is `None` here even though the client's JSON never mentioned `id`. With `"id"` no longer in
`exclude`, `auto_init`'s kwarg loop hits:

```python
if key in model_columns:
    setattr(self, key, val)   # self.id = None, on a row that already exists
    continue
```

— nulling the row's primary key in memory. SQLAlchemy then can't resolve the flush back to an
`UPDATE` against the existing row, and Postgres rejects the resulting write for violating the
`NOT NULL` constraint on the primary key.

## Blast radius beyond foods

The same `model_config.exclude` pattern, with the same missing `"id"`, exists on several other
models — meaning the same failure mode is very likely reachable through each of their own
update paths too, not just foods:

- `RecipeToolModel` (`mealie/db/models/recipe/tool.py`) — `exclude={"households_with_tool"}`
- `GroupModel` (`mealie/db/models/group/group.py`)
- `HouseholdModel` (`mealie/db/models/household/household.py`)
- `UserModel` (`mealie/db/models/users/users.py`)
- `ReportModel` (`mealie/db/models/group/report.py`) — `exclude=["entries"]`

Notably, a few other models already work around this ad hoc by manually re-adding `"id"` to
their own `exclude` set — which only makes sense as a workaround for exactly this bug:

- `RecipeInstructionModel` (`mealie/db/models/recipe/instruction.py`):
  `exclude={"id", "ingredient_references"}`
- `ShoppingListItemModel` / `ShoppingListModel`
  (`mealie/db/models/household/shopping_list.py`): `exclude={"id", "recipe"}` /
  `exclude={"id", "list_items"}`

That inconsistency — some models remembering to re-add `"id"`, most not — is itself evidence
this should be fixed once at the source rather than per-model.

## Fix

Merge a model's own `exclude` with the default instead of replacing it, in `_get_config()`:

```python
for attr in class_config:
    if attr not in cfgKeys:
        continue
    if attr == "exclude":
        cfg.exclude = cfg.exclude | set(class_config[attr])
    else:
        setattr(cfg, attr, class_config[attr])
```

This fixes it for every model in one place — including the ones that don't currently know they
need to manually re-add `"id"` — and the manual `"id"` re-additions on
`RecipeInstructionModel`/`ShoppingListItemModel`/`ShoppingListModel` become harmless redundancy
rather than a required workaround.

Fixed in: https://github.com/blue-mtn-dog/mealie/commit/c17a175a

## Environment

- Mealie version: 3.22.0 (also confirmed present, unpatched, on the `mealie-next` branch as of
  2026-08-14)
- Reproduced against a self-hosted instance (Postgres backend)

blue-mtn-dog:
## Summary

Fixes #8088.

`_get_config()` in `mealie/db/models/_model_utils/auto_init.py` replaced `AutoInitConfig`'s default primary-key exclusion (`{"id"}`) with a model's own `model_config.exclude`, instead of merging the two. Any model declaring `model_config = ConfigDict(exclude={...})` without separately re-listing `"id"` loses primary-key protection during `BaseMixins.update()`, which re-invokes `__init__(**kwargs)` on an already-persisted row. Every `Create*`/`Save*` schema for these models carries an `id` field defaulting to `None` when the client doesn't send one, so `auto_init` then does `setattr(self, "id", None)` on a live row — nulling its primary key mid-update and failing the next flush with a `NOT NULL` constraint violation.

Confirmed concretely on `IngredientFoodModel`: any `PUT /api/foods/{id}` that changes anything fails this way. `RecipeToolModel`, `GroupModel`, `HouseholdModel`, `UserModel`, and `ReportModel` declare `exclude` the same vulnerable way. `RecipeInstructionModel` and `ShoppingListItemModel`/`ShoppingListModel` already work around this ad hoc by manually re-adding `"id"` to their own `exclude` — this fixes it at the source so no model needs to remember to do that, per the suggestion in the issue's follow-up comment.

## Fix

`_get_config()` now unions a model's own `exclude` with the default instead of replacing it wholesale. Every other `AutoInitConfig` attribute keeps its existing replace-on-match behavior.

## Tests

Added `tests/unit_tests/test_auto_init_config.py` — direct unit coverage of `_get_config()`:
- No `model_config` → default `{"id"}` exclusion still applies.
- Custom `exclude` → merged with `{"id"}`, not replacing it.
- Custom `exclude` that already includes `"id"` → no duplication.
- An unrelated config key (`get_attr`) → default `exclude` untouched.

This was requested explicitly in the issue's follow-up comment, alongside independent confirmation of the root cause.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/mealie/db/models/_model_utils/auto_init.py
+++ b/mealie/db/models/_model_utils/auto_init.py
@@ -39,7 +39,15 @@ def _get_config(relation_cls: type[SqlAlchemyBase]) -> AutoInitConfig:
         return cfg
     # Map all matching attributes in Config to all AutoInitConfig attributes
     for attr in class_config:
-        if attr in cfgKeys:
+        if attr not in cfgKeys:
+            continue
+
+        if attr == "exclude":
+            # A model's own `exclude` adds to the default exclusions (e.g. "id")
+            # rather than replacing them -- otherwise a model that declares
+            # `exclude` loses primary-key protection during `auto_init`.
+            cfg.exclude = cfg.exclude | set(class_config[attr])
+        else:
             setattr(cfg, attr, class_config[attr])
 
     return cfg
--- a/tests/unit_tests/test_auto_init_config.py
+++ b/tests/unit_tests/test_auto_init_config.py
@@ -1,0 +1,41 @@
+from pydantic import ConfigDict
+
+from mealie.db.models._model_utils.auto_init import _get_config
+
+
+class _NoConfig:
+    pass
+
+
+class _CustomExcludeOnly:
+    model_config = ConfigDict(exclude={"households_with_ingredient_food"})
+
+
+class _CustomExcludeWithId:
+    model_config = ConfigDict(exclude={"id", "recipe"})
+
+
+class _OtherConfigKeyOnly:
+    model_config = ConfigDict(get_attr="slug")
+
+
+def test_get_config_defaults_to_excluding_id_when_no_model_config():
+    cfg = _get_config(_NoConfig)
+    assert cfg.exclude == {"id"}
+
+
+def test_get_config_merges_custom_exclude_with_default_id_exclusion():
+    """Regression test for GH issue #8088."""
+    cfg = _get_config(_CustomExcludeOnly)
+    assert cfg.exclude == {"id", "households_with_ingredient_food"}
+
+
+def test_get_config_does_not_duplicate_id_when_custom_exclude_already_has_it():
+    cfg = _get_config(_CustomExcludeWithId)
+    assert cfg.exclude == {"id", "recipe"}
+
+
+def test_get_config_preserves_default_exclude_for_unrelated_config_keys():
+    cfg = _get_config(_OtherConfigKeyOnly)
+    assert cfg.get_attr == "slug"
+    assert cfg.exclude == {"id"}


michael-genson (approved):

michael-genson:
In the future please use the GitHub PR template

Issue #3505, opened 2024-04-21
Recipe with a single 'space' as the title returns "404 page not found" - reproduced on the demo site


dino9848:
### First Check

- [X] This is not a feature request.
- [X] I added a very descriptive title to this issue (title field is above this).
- [X] I used the GitHub search to find a similar issue and didn't find it.
- [X] I searched the Mealie documentation, with the integrated search.
- [X] I already read the docs and didn't find an answer.
- [X] This issue can be replicated on the demo site (https://demo.mealie.io/).

### What is the issue you are experiencing?

Create a manual recipe on the demo site by creating a recipe titled with just a single space creates a 404 error and no way to delete or edit the issue

### Steps to Reproduce

1) Click Create  - select the 'create a recipe manually'
2) enter a single space in the recipe name box and click "Create"

It creates a recipe without any option to delete from the GUI.  Clicking on the edit option creates the 404 error

Note:  If you do this twice, the second recipe created with automatically be named "(1)" and it does allow editing and deleting.
Best regards and thanks for a wonderful app.

### Please provide relevant logs

No need.  It happens on every version I tried.   Can even be reproduced on your demo site.  Would love to be able to correct my database that has one of these 'no-name can't delete'  recipes

### Deployment

Docker (Linux)

### Additional Deployment Details

Running on current CasaOS

dino9848:
Able to delete from the data management admin account.  Regular user can't delete.

boc-the-git:
Solution here is to stop a recipe being able to be created when the title (removing whitespace) is null

nephlm:
The route seems to call `RecipeService.create_one()` which has the following code:

https://github.com/mealie-recipes/mealie/blob/mealie-next/mealie/services/recipe/recipe_service.py#L121-L123

```
    def create_one(self, create_data: Recipe | CreateRecipe) -> Recipe:
        if create_data.name is None:
            create_data.name = "New Recipe"
```
It seems like the correct thing might be just to change that to:

```
    def create_one(self, create_data: Recipe | CreateRecipe) -> Recipe:
        if not create_data.name:
            create_data.name = "New Recipe"
```
Later code handles deduping if there are previous "New Recipe".  Though the proposed changes don't handle leading/trailing spaces so it will have to be a little more complicated than that.

michael-genson:
We should have a "clean string" method somewhere that handles things like that, probably with the migration code. Although honestly we should probably handle that when creating the slug since slugs already clean strings anyway

PR #8343, opened 2026-09-08
fix: default the recipe name when an AI import returns a blank one

adman234:

An AI import fails outright when the provider returns a recipe whose name is empty or contains only
characters that don't survive slugification. `to_recipe` passes the name straight to
`create_recipe_slug`, which raises `SlugError` on an empty slug, and nothing in the workflow catches
it — so the whole import is lost even though the ingredients and instructions usually came back
intact.

```
File ".../import_workflow/recipe_conversion.py", line 44, in to_recipe
    slug=create_recipe_slug(openai_recipe.name),
File ".../schema/recipe/recipe.py", line 55, in create_recipe_slug
    raise SlugError("Recipe name cannot be empty or contain only special characters")
mealie.core.exceptions.SlugError: Recipe name cannot be empty or contain only special characters
```

`OpenAIRecipe.name` is typed `str`, so a provider can't return `None` — but an empty string passes
validation fine, and a junk title shouldn't cost the user the rest of the recipe.

- `mealie/services/recipe/import_workflow/recipe_conversion.py` — add `resolve_name_and_slug`, which
  returns the provider's name when it slugifies, and otherwise falls back to a translated default.
- `mealie/lang/messages/en-US.json` — add `recipe.recipe-defaults.name`, alongside the existing
  `ingredient-note` and `step-text` defaults. Only `en-US` is touched; the other locales are
  Crowdin-managed.
- `tests/unit_tests/services_tests/test_recipe_import_workflow.py` — unit tests for each branch.

Three decisions worth calling out:

- **Scoped to the AI import path, not `create_recipe_slug` itself.** Raising is arguably correct for
  manual creation, and other callers depend on that behaviour.
- **There's a third fallback to an ASCII slug.** Once the default name is translated you can't assume
  it slugifies — `slugify` can return empty for scripts it can't transliterate. Without that fallback
  the patch would swap one crash for a rarer, stranger one.
- **`t(key, default)` rather than `t(key)`**, so this keeps working in a locale whose file hasn't
  picked the key up from Crowdin yet. Verified: de-DE and fr-FR currently fall through to
  "New Recipe" rather than rendering a raw key.

It lives in `to_recipe` because, per that function's own docstring, it's shared by every step that
produces a draft recipe — so this also covers a recipe that imported fine and then came back from the
translation step with a blank name.

No open issue matches this traceback, so there's nothing to close. For context:

- Discussion #6986 — maintainer note on self-hosted models and AI import reliability. A self-hosted
  OpenAI-compatible endpoint is what surfaced this, but the crash is a missing guard in Mealie: any
  provider returning a junk title should degrade gracefully.
- Issue #3505 — blank names via *manual* creation (a different code path). Relevant because a
  maintainer there suggested the empty-name case is best handled where slugs are created, since slugs
  already clean strings. That's the principle this follows.

There's an existing string `recipe.import-errors.provider-returned-nothing` ("The AI provider did not
return a recipe"), and a reasonable objection is that a blank name means the provider failed, so the
user should get that error rather than a silent "New Recipe".

The argument for defaulting: the ingredients and instructions typically survive intact, so discarding
the whole import over a missing title throws away good data the user would have to re-import to
recover, and renaming is one click. Mealie also already defaults a blank name to "New Recipe"
elsewhere (`create_one` in the recipe repository, per #3505), so this is consistent with existing
behaviour rather than new precedent.

If you'd rather have the error path, the change is small either way — swap the fallback for a raised
import error using that existing string. Happy to do that instead.

## Testing

Seven unit tests covering: a normal name passing through with its expected slug; empty and
whitespace-only names; a punctuation-only name; an emoji-only name; a translated default; and a
stub translator returning an unsluggable string, which exercises the ASCII fallback.

Verified end to end as well, not just at the unit level: the stock 3.25.1 image reproduces the
traceback above, and with the patch applied the same input yields `name="New Recipe"`,
`slug="new-recipe"` with both ingredients and both instructions preserved — checked across en-US,
de-DE and fr-FR.

Full `tests/unit_tests` suite passes. Ruff (format + lint) and mypy are clean.

## AI / LLM Assistance

This PR was written by Claude Code (Claude Opus 5) working from my direction, and I've reviewed it
before submitting. The bug came from my own instance — a self-hosted OpenAI-compatible endpoint
returning a recipe with no title — and I supplied the traceback and the intended approach, including
the decision to scope the change to the import path rather than to `create_recipe_slug`. The
implementation, the tests, and the before/after verification against a real container were produced
in that session.

### DIFF
--- a/mealie/lang/messages/en-US.json
+++ b/mealie/lang/messages/en-US.json
@@ -8,6 +8,7 @@
         "recipe-image-deleted": "Recipe image deleted",
         "recipe-defaults": {
             "ingredient-note": "1 Cup Flour",
+            "name": "New Recipe",
             "step-text": "Recipe steps as well as other fields in the recipe page support markdown syntax.\n\n**Add a link**\n\n[My Link](https://demo.mealie.io)\n"
         },
         "servings-text": {
--- a/mealie/services/recipe/import_workflow/recipe_conversion.py
+++ b/mealie/services/recipe/import_workflow/recipe_conversion.py
@@ -1,5 +1,6 @@
 from pydantic.alias_generators import to_camel
 
+from mealie.core.exceptions import SlugError
 from mealie.schema.openai.recipe import OpenAIRecipe
 from mealie.schema.recipe.recipe import Recipe, create_recipe_slug
 from mealie.schema.recipe.recipe_ingredient import RecipeIngredient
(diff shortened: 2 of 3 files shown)


michael-genson (approved):
Thanks!

Issue #7957, opened 2026-07-27
[BUG] - Gemini OpenAI-compatible endpoint authentication fails when using API key field


niker:
### First Check

- [x] This is not a feature request.
- [x] I added a very descriptive title to this issue (title field is above this).
- [x] I used the GitHub search to find a similar issue and didn't find it.
- [x] I searched the Mealie documentation, with the integrated search.
- [x] I already read the docs and didn't find an answer.
- [ ] This issue can be replicated on the demo site (https://demo.mealie.io/).

### Description

When configuring Mealie to use Google's Gemini API through the OpenAI-compatible endpoint, authentication fails when the API key is entered into Mealie's built-in API key field.

The same API key and endpoint work correctly when the `Authorization: Bearer <API_KEY>` header is added manually through the custom headers field.

This suggests that Mealie's OpenAI-compatible provider is not passing the API key in the expected format.

### Steps to Reproduce

- Model: `gemini-3.1-flash-lite`
- Base URL: `https://generativelanguage.googleapis.com/v1beta/openai/`
- API Key: `correct`

Results in:

```
openai.PermissionDeniedError: Error code: 403

Requests to this API generativelanguage.googleapis.com method
google.ai.generativelanguage.v1main.GenerativeService.GenerateContent
are blocked.

reason: API_KEY_SERVICE_BLOCKED
```

### Workaround

Adding the authentication manually as a custom header fixes the issue:

Header Key:
```
Authorization
```

Header Value:
```
Bearer <API_KEY>
```

After adding this header, Mealie successfully communicates with Gemini using the OpenAI-compatible endpoint.

The API key can be set to a dummy value.

This should at least be mentioned in the docs as the latest flash lite is currently one of the most cost effective models available.

### Please provide relevant logs

[...]

PR #8344, opened 2026-09-08
fix: apply a submitted API key when updating an AI provider

adman234:

An AI provider's API key can't be changed after the provider is created. Submitting a new key
appears to succeed — the update returns 200 and the other edited fields are saved — but the
provider keeps authenticating with its original key. The only way to correct a mistyped or
rotated key today is to delete the provider and create it again, and nothing in the UI or the
response indicates the submitted key was dropped.

`AIProviderCreate.api_key` is declared `Field("", exclude=True)`, so it is absent from
`model_dump()`. `GroupRepositoryAIProvider.update` dumps the schema and then treats the now-missing
key as "the client didn't send one", so its merge-the-stored-key branch fires on every update and
writes the old key back over the submitted one. `create` already works around the exclusion by
carrying the key across the dump by hand; this does the same in `update`.

- `mealie/repos/repository_ai_provider.py` — preserve `api_key` across `model_dump()` in `update`,
  mirroring what `create` does.
- `tests/integration_tests/user_group_tests/test_group_ai_providers.py` — two regression tests, one
  per branch of the merge behaviour.

Preserving the stored key when the update omits one is deliberate and unchanged: the key is
write-only, so `GroupAIProviderDialog` never receives it to send back, and leaving the field blank
while editing other fields must not wipe it.

No API, schema, or migration changes; the fix is three lines inside the repository.

No open issue that I could confirm matches this. #7957 ("Gemini OpenAI-compatible endpoint
authentication fails when using API key field") may share this root cause — if the reporter had
ever saved that provider and later corrected the key, they would have been authenticating with the
original value — but I haven't been able to verify that against their setup, so I'm not claiming it
closes that issue. Happy to add a `Fixes` line if a maintainer confirms the link.

The write-once behaviour is easy to miss because it's invisible from the API surface — `api_key` is
excluded from every response, so no existing test could observe that the update was dropped. The new
tests read the key back off the repository rather than the response for that reason.

Introduced alongside the provider CRUD in #7650; this is not a regression from a recent change.

Worth deciding separately: `update` silently ignoring an unusable field is the kind of thing that
would be better caught at the schema boundary. I kept the fix minimal rather than reworking how
`exclude=True` interacts with the repository layer, but I'm happy to take that further if you'd
prefer a more structural fix.

## Testing

- `test_update_provider_rotates_the_api_key` — creates a provider, PUTs a new key, asserts the
  stored key changed. Fails on `mealie-next` without the fix, passes with it.
- `test_update_provider_without_an_api_key_keeps_the_stored_one` — PUTs an update with no key,
  asserts the original survives. Passes both with and without the fix; it guards the behaviour this
  change must not break.

Also ran the surrounding suites against the change: `tests/integration_tests/user_group_tests/test_group_ai_providers.py`,
`tests/integration_tests/admin_tests/test_admin_ai_providers.py`,
`tests/unit_tests/schema_tests/test_ai_providers.py` and
`tests/unit_tests/services_tests/test_openai_service.py` — all pass. Ruff (format + lint) and mypy
are clean.

## AI / LLM Assistance

This PR was written by Claude Code (Claude Opus 5) working from my direction, and I've reviewed it
before submitting. The bug was found while debugging a 401 from a self-hosted OpenAI-compatible
endpoint on my own instance: I kept re-entering the API key with no effect, and tracing the update
path through `repository_ai_provider.py` explained why. The diagnosis, the fix and the tests were
produced in that session; the failing-without-the-fix / passing-with-it check above was run to
confirm the tests actually exercise the bug rather than just passing.

### DIFF
--- a/mealie/repos/repository_ai_provider.py
+++ b/mealie/repos/repository_ai_provider.py
@@ -27,7 +27,11 @@ class GroupRepositoryAIProvider(GroupRepositoryGeneric[AIProviderOut, AIProvider
 
     def update(self, match_value: str | int | UUID4, new_data: AIProviderCreate | dict):
         if isinstance(new_data, AIProviderCreate):
+            # api_key is excluded from the schema's dump, so carry it over by hand, the same
+            # way create does. Without this every update looks like it omitted the key.
+            api_key = new_data.api_key
             new_data = new_data.model_dump()
+            new_data["api_key"] = api_key
 
         # Merge existing API key into new data
         if not new_data.get("api_key"):
--- a/tests/integration_tests/user_group_tests/test_group_ai_providers.py
+++ b/tests/integration_tests/user_group_tests/test_group_ai_providers.py
@@ -76,6 +76,53 @@ def test_update_provider(api_client: TestClient, unique_user: TestUser):
         api_client.delete(api_routes.groups_ai_providers_providers_provider_id(provider.id), headers=unique_user.token)
 
 
+def test_update_provider_rotates_the_api_key(api_client: TestClient, unique_user: TestUser):
+    """A submitted key must actually replace the stored one.
+
+    The key is write-only, so the response can't confirm this; read it back off the repository.
+    """
+    provider = unique_user.repos.group_ai_providers.create(
+        AIProviderCreate(name=random_string(), model="gpt-4o", api_key="original-key")
+    )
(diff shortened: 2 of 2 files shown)


michael-genson (approved):
Thanks!

PR #8349, opened 2026-09-09
fix: skip mounted directories during backup restore

### LINKED ISSUE #5524: [BUG] Restoring from a malformed backup breaks the app
### First Check

My Mealie docker crashed.  I was able to get a fresh copy of Mealie to run.  Setup went fine, but when it came time to restore my my recent back, that restore had and error, and which broke Mealie complete.  Unable to login anymore.  I tried several previous backup, all caused the same behavior.

I was finally able to get a backup from earlier in Feb to work, but it's missing several dozen recipes (expected from an old backup).

I immediately made a new back of the restored Feb backup and this brand new backup also breaks Mealie.

I did try my backup on the demo site, and now the demo site is broken as well.

Is there any way to determine why my backups are breaking Mealie?

The backups are too large to attach, but can be located at the link below.

The one dated in Feb is the one that works.  The ones from May and June break.

https://drive.google.com/drive/folders/11R6Z8ACnKUFTs9fweLBd-4u3BpevnBPX?usp=sharing

### Steps to Reproduce

1) Upload the backup to Mealie
2) Hit restore Backup
3) suffer

[ERROR|admin_backups|L112] 2025-06-12T14:46:54: [Errno 16] Device or resource busy: PosixPath('/app/data/mealie-pgdata')
  File "/opt/mealie/lib/python3.12/site-packages/mealie/routes/admin/admin_backups.py", line 105, in import_one
    backup.restore(file)
  File "/opt/mealie/lib/python3.12/site-packages/mealie/services/backups_v2/backup_v2.py", line 106, in restore
    self._copy_data(contents.data_directory)
  File "/opt/mealie/lib/python3.12/site-packages/mealie/services/backups_v2/backup_v2.py", line 67, in _copy_data
    shutil.rmtree(self.directories.DATA_DIR / f.name)
  File "/usr/local/lib/python3.12/shutil.py", line 759, in rmtree
    _rmtree_safe_fd(stack, onexc)
  File "/usr/local/lib/python3.12/shutil.py", line 703, in _rmtree_safe_fd
    onexc(func, path, err)
  File "/usr/local/lib/python3.12/shutil.py", line 662, in _rmtree_safe_fd
    os.rmdir(name, dir_fd=dirfd)
OSError: [Errno 16] Device or resource busy: PosixPath('/app/data/mealie-pgdata')
[INFO|httptools_impl|L476] 2025-06-12T14:46:54: 50.107.56.136:0 - "POST /api/admin/backups/mealie_2025.06.12.18.46.21.zip/restore HTTP/1.1" 500
[INFO|httptools_impl|L476] 2025-06-12T14:47:00: 50.107.56.136:0 - "GET / HTTP/1.1" 200

### Mealie Version

v2.8.0

### Deployment

Docker (Linux)

_No response_

tiantongluo66-crypto:

This PR fixes #5524, where restoring a backup can fail when the backup contains a directory whose destination is a mounted directory inside Mealie's data directory.

The issue occurs in `BackupV2._copy_data()`. The original implementation removes each destination directory with `shutil.rmtree()` before copying the restored directory into place. If the destination is a mount point, such as `/app/data/mealie-pgdata`, attempting to remove it can raise:

`OSError: [Errno 16] Device or resource busy`

This interrupts the restore process and may leave the application in an inconsistent state because database restoration has already started before the data directory is restored.

Changes made in this PR:

* **`mealie/services/backups_v2/backup_v2.py`**

  * Creates the destination path once for each item being restored.
  * Checks `Path.is_mount()` before removing a destination directory.
  * Logs a warning and skips the directory if the destination is a mount point.
  * Only calls `shutil.rmtree()` when the non-mounted destination actually exists.
  * Keeps the existing restore behaviour for normal files and directories.

* **`tests/unit_tests/services_tests/backup_v2_tests/test_backup_v2.py`**

  * Adds `test_copy_data_skips_mounted_directories`.
  * Simulates a mounted destination directory.
  * Verifies that mounted directories are neither removed with `shutil.rmtree()` nor overwritten with `shutil.copytree()`.

The mount-point check is based on the destination path rather than a specific directory name such as `mealie-pgdata`, so the protection also applies to other mounted directories that may exist inside the Mealie data directory.

Fixes #5524

The main change is in `BackupV2._copy_data()`. In particular, review of the `Path.is_mount()` check and the decision to skip mounted destinations would be appreciated.

The intention is to keep this fix limited to preventing restore operations from deleting or overwriting mounted directories while preserving the existing behaviour for regular directories.

## Testing

A regression test, `test_copy_data_skips_mounted_directories`, was added to reproduce the relevant behaviour in isolation.

The test:

* creates a temporary backup data directory containing a directory representing `mealie-pgdata`;
* mocks `Path.is_mount()` so the corresponding destination is treated as a mount point;
* mocks `shutil.rmtree()` and `shutil.copytree()`;
* verifies that neither operation is performed on the mounted destination.

The regression test was first run against the original implementation. It failed as expected because the original code attempted to remove the mounted destination:

`AssertionError: Expected 'rmtree' to not have been called. Called 1 times.`

After applying the mount-point check, the same regression test passed:

`1 passed`

The complete `backup_v2` unit test file was also run to check for regressions in the existing backup and restore functionality:

`4 passed in 145.52s`

Tests were run using Python 3.12 and pytest 9.1.1 in the project's Docker/Linux environment.

## AI / LLM Assistance

AI/LLM assistance was used to help analyse the reported error, discuss possible implementation approaches

The code changes were reviewed locally, and the regression test and existing `backup_v2` tests were executed locally to verify the behaviour before submission.

### DIFF
--- a/mealie/services/backups_v2/backup_v2.py
+++ b/mealie/services/backups_v2/backup_v2.py
@@ -77,15 +77,25 @@ class BackupV2(BaseService):
 
     def _copy_data(self, data_path: Path) -> None:
         for f in data_path.iterdir():
+            destination = self.directories.DATA_DIR / f.name
+
             if f.is_file():
                 if f.name not in self.RESTORE_FILES:
                     continue
 
-                shutil.copyfile(f, self.directories.DATA_DIR / f.name)
+                shutil.copyfile(f, destination)
                 continue
 
-            shutil.rmtree(self.directories.DATA_DIR / f.name)
-            shutil.copytree(f, self.directories.DATA_DIR / f.name)
+            if destination.is_mount():
+                self.logger.warning(
+                    f"Skipping mounted data directory during restore: {destination}"
+                )
+                continue
+
+            if destination.exists():
+                shutil.rmtree(destination)
+
+            shutil.copytree(f, destination)
 
         # since we copied a new .secret, AppSettings has the wrong secret info
         self.logger.info("invalidating appsettings cache")
--- a/tests/unit_tests/services_tests/backup_v2_tests/test_backup_v2.py
+++ b/tests/unit_tests/services_tests/backup_v2_tests/test_backup_v2.py
@@ -1,5 +1,7 @@
 import statistics
+from pathlib import Path
 from typing import Any
+from unittest.mock import patch
 
 from sqlalchemy.orm import Session
 
@@ -61,6 +63,24 @@ def test_database_restore():
 
     for s1, s2 in zip(snapshop_1, snapshop_2, strict=False):
         assert snapshop_1[s1].sort(key=dict_sorter) == snapshop_2[s2].sort(key=dict_sorter)
+
+
+def test_copy_data_skips_mounted_directories(tmp_path: Path):
+    backup_v2 = BackupV2()
+
+    source_dir = tmp_path / "source"
+    mounted_dir = source_dir / "mealie-pgdata"
+    mounted_dir.mkdir(parents=True)
+
+    with (
+        patch.object(Path, "is_mount", return_value=True),
+        patch("mealie.services.backups_v2.backup_v2.shutil.rmtree") as mock_rmtree,
+        patch("mealie.services.backups_v2.backup_v2.shutil.copytree") as mock_copytree,
+    ):
+        backup_v2._copy_data(source_dir)
+
+    mock_rmtree.assert_not_called()
+    mock_copytree.assert_not_called()
 
 
 def _5ab195a474eb_add_normalized_search_properties(session: Session):


elliot-ultra:
image



I ran into this issue too — when I restored a backup on my Macbook, all the recipe images were gone. Even worse, the second time I tried to restore, the page just showed a `502 Internal Server Error`. This PR should address that.

One thing I noticed: the restore process doesn't have any transaction or data-completeness check. If it's interrupted, the data ends up incomplete and the app breaks. A concrete example I hit: restoring a backup whose alembic_version isn't a migration that exists in the currently running codebase makes AlchemyExporter.restore() raise a raw alembic.util.exc.CommandError at command.upgrade(alembic_cfg, alembic_version) (alchemy_exporter.py:213). BackupSchemaMismatch (backup_v2.py:15), which is supposed to give a clean 400 for exactly this case, is never actually raised anywhere in the codebase, so this falls through to a generic 500, and by that point drop_all() has already wiped the database, leaving the app broken with no way back.

It might be worth doing something like a transaction here:
1. Rename the old database file and data folders instead of deleting them (e.g. mealie.db → mealie.db.backup, recipes/ → recipes.backup/).
2. Write the new database and data folders in their place.
3. If an exception is raised at any point, rename the backups back to restore the previous state.

Only delete the .backup files/folders once the whole restore has completed successfully.

mealie/services/backups_v2/backup_v2.py
```diff
@@ -77,15 +77,25 @@ def backup(self) -> Path:
...
                 if f.name not in self.RESTORE_FILES:
                     continue
 
-                shutil.copyfile(f, self.directories.DATA_DIR / f.name)
+                shutil.copyfile(f, destination)
                 continue
 
-            shutil.rmtree(self.directories.DATA_DIR / f.name)
-            shutil.copytree(f, self.directories.DATA_DIR / f.name)
+            if destination.is_mount():
```
michael-genson: Skipping means a legitimately mounted data directory (say `recipes/`, where images live) never gets restored, and the restore still reports success after the database has been replaced. Could we copy into the destination instead of replacing it — clear its contents, then `shutil.copytree(f, destination, dirs_exist_ok=True)`? That works whether or not it's a mount, and avoids `rmtree` on the mount point.

michael-genson (changes requested):
Thanks! One more thing outside the diff: in `backup()`, `EXCLUDE_DIRS` only matches `data_file.parent.name`, so nested files (e.g. `mealie-pgdata/base/...`) still get zipped into the backup — which is what makes #5524 happen in the first place. Can we fix that here too, e.g. by checking the path parts against `EXCLUDE_DIRS`?

Issue #6633, opened 2025-12-01
[BUG] - Editing recipe concurrently (two users/devices) overwrites first set of changes without warning or recourse


lehnerpat:
### First Check

- [x] This is not a feature request.
- [x] I added a very descriptive title to this issue (title field is above this).
- [x] I used the GitHub search to find a similar issue and didn't find it.
- [x] I searched the Mealie documentation, with the integrated search.
- [x] I already read the docs and didn't find an answer.
- [x] This issue can be replicated on the demo site (https://demo.mealie.io/).

### What is the issue you are experiencing?

When the recipe is edited at the same time in two places (two sessions / two devices / two users), the second one to save their changes overwrites the first one's, without any warning or any way to recover the lost changes.

This is exacerbated by the PWA having no easy way to reload a recipe page, and recipes having no version history.

### Steps to Reproduce

1. Open a recipe and enter edit mode, and make a change (let's say, set the description to "A") -- don't save yet
2. In another session (e.g. another browser session, another device, or using another user account in another session), open the same recipe and enter editing mode, and make another change (let's say, add an ingredient "Apples") -- don't save yet
3. In the first session, save your changes
4. In the second session, save your changes

**Expected result:**
Mealie detects the conflict in the second session and prevents the user from unknowingly overwriting the changes (at least refuse to save; ideally, provide more information, possibly a way to resolve the conflicts).

**Actual result:**
Mealie loses the first session's changes without any information to the user. Since there is no version history for recipes, the changes are irrevocably lost :(

### Please provide relevant logs

--

### Mealie Version

3.5.0

### Deployment

Docker (Linux)

lehnerpat:
As an idea for how to detect whether the frontend is trying to save changes based on an outdated version, I would suggest considering optimistic locking:

* add a revision counter to the recipe entity
* increment this revision counter on each change (in the backend)
* when a client wants to update an existing recipe, it must include the revision on which its changes are based
* if that revision isn't the latest one that the server has (i.e. if another change has happened in the meantime), the server rejects the change (instructing the client to fetch the latest version and try again)
  * obviously the frontend must then deal with this error properly; at least inform the user, but ideally prevent the loss of the changes the user was trying to save (in the example above, this would be the second set of change; we don't want to trade losing one set of changes for losing another set of changes)

[...]

PR #8352, opened 2026-09-09
fix: prevent stale recipe edits from overwriting newer changes

Frank-Ping:

When two users edit the same recipe, a later save can silently overwrite changes saved by the other user. This PR adds timestamp-based optimistic concurrency checks.

- `repository_recipes.py`: atomically checks `updatedAt` before updating a recipe and rolls back bulk updates if a conflict occurs.
- `exceptions.py` and `recipe_crud_routes.py`: introduce a recipe conflict exception, return HTTP 409, and support timestamp preconditions for image operations.
- `RecipePage.vue`: submits the original timestamp and preserves unsaved drafts when saving fails.
- `use-image-edit.ts`, image components, and the recipe API client: coordinate image updates with the editor’s timestamp without replacing its draft.
- `axios.ts` and `en-US.json`: display a persistent conflict warning and suppress the duplicate toast for handled 409 responses.
- Backend and frontend tests cover stale updates, bulk rollback, concurrent requests, and image-edit version handling.

The implementation uses the existing update timestamp, so no database migration or additional dependency is required.
图片

Related to #6633.

Feedback would be particularly useful on the concurrency contract and backward compatibility:

- Clients that omit timestamps retain existing behavior and are not protected against stale writes.
- Saving unchanged content still advances the timestamp and can cause another editor’s save to conflict.
- Image downloads hold a database write transaction; the performance implications should be reviewed.

## Testing

- Passed 71 backend tests across the recipe CRUD, image assets, and new conflict test modules using an isolated Linux container with SQLite.
- Concurrent API requests using the same timestamp resulted in one successful save and one HTTP 409 response.
- Passed four frontend tests covering image-edit timestamp synchronization, draft preservation, duplicate-request prevention, and error recovery.
- Passed Ruff, targeted ESLint checks, and `git diff --check`.
- Successfully generated the production frontend using Node.js 24 and pnpm.
- Manually tested two browser sessions in Docker and observed a stale save returning HTTP 409 after the other session saved.

## AI / LLM Assistance

GPT-6 assisted with implementation, regression tests, and this PR description. I reviewed the AI-generated implementation and code, manually reproduced the issue, and tested the two-user editing scenario in Docker.

### DIFF
--- a/frontend/app/components/Domain/Recipe/RecipeImageUploadBtn.vue
+++ b/frontend/app/components/Domain/Recipe/RecipeImageUploadBtn.vue
@@ -81,6 +81,7 @@
 </template>
 
 <script setup lang="ts">
+import { useImageEdit } from "~/composables/recipe-page/use-image-edit";
 import { alert } from "~/composables/use-toast";
 import { useUserApi } from "~/composables/api";
 
@@ -98,6 +99,7 @@ const emit = defineEmits<{
 
 const i18n = useI18n();
 const api = useUserApi();
+const editImage = useImageEdit();
 
 const url = ref("");
 const loading = ref(false);
@@ -112,7 +114,8 @@ function uploadImage(fileObject: File) {
 async function deleteImage() {
   loading.value = true;
   try {
-    await api.recipes.deleteImage(props.slug);
+    const result = await editImage(config => api.recipes.deleteImage(props.slug, config));
+    if (!result || result.error) return;
     emit(DELETE_EVENT);
     menu.value = false;
   }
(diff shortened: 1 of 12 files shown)


michael-genson:
Thanks for tackling this, and for the thorough tests. Before going further on the implementation, I want to talk through the approach.

For a recipe app, I think the realistic case is two people editing hours apart on different devices, not two saves landing in the same second. That case can be handled in the editor: before saving, fetch the recipe's current `updatedAt`, and if it changed, show a dialog offering to overwrite or reload, keeping the draft either way. No API contract change, and you've already built most of the UI for it here.

That does leave a window where two people saving at the same moment still overwrite each other, which I'm comfortable with.

What do you think?

PR #8356, opened 2026-09-10
fix: make TypeScript generation deterministic

### LINKED ISSUE #8222: [BUG] - dev/code-generation types generator produces non-deterministic output depending on model processing order

justadityaraj:
### Problem / Context

`pydantic2ts` temporarily changes every exported Pydantic model's `extra` setting to `forbid`, but it does not restore models whose original setting was unset. Shared models therefore retain state from whichever schema module was generated first, making later TypeScript output depend on module processing order.

### Changes

- Load each schema module before generation so all exported models are known up front.
- Keep non-`allow` models closed for the complete generation pass, then restore every original model config even when generation fails.
- Build dotted module names from `Path.parts`, which also makes the preload work with Windows paths.
- Add a regression covering the dependency's config mutation and restoration.

### User impact

Generated TypeScript definitions no longer gain or lose index signatures based on schema-module order, so `task dev:generate` produces stable output and avoids unrelated generated-file changes.

Fixes #8222

## Verification

- Regression test: 1 passed; the test failed on the base revision because the preservation boundary did not exist.
- Ruff lint and format checks: passed for both changed files.
- MyPy: passed for `dev/code-generation/gen_ts_types.py`.
- Complete TypeScript generator: passed for all 12 schema modules; generated definitions matched the committed files with no semantic diff.
- `git diff --check`: passed.
- The repository-wide pytest bootstrap was attempted but cannot import the locked `python-ldap` dependency on Windows because its native extension requires MSVC/OpenLDAP.

## AI / LLM Assistance

AI assistance was used to help investigate the generator state leak, implement the fix, and draft the regression test. I reviewed the complete diff and independently ran all verification listed above.

### DIFF
--- a/dev/code-generation/gen_ts_types.py
+++ b/dev/code-generation/gen_ts_types.py
@@ -1,9 +1,14 @@
 import re
 import subprocess
+from collections.abc import Iterable, Iterator
+from contextlib import contextmanager
+from importlib import import_module
 from pathlib import Path
 
 from jinja2 import Template
+from pydantic import BaseModel
 from pydantic2ts import generate_typescript_defs
+from pydantic2ts.cli.script import extract_pydantic_models
 from utils import log
 
 # ============================================================
@@ -30,6 +35,20 @@ export {};
 
 CWD = Path(__file__).parent
 PROJECT_DIR = Path(__file__).parent.parent.parent
+
+
+@contextmanager
+def forbid_extra_fields(models: Iterable[type[BaseModel]]) -> Iterator[None]:
+    model_configs = {model: model.model_config.copy() for model in models}
+    try:
+        for model in model_configs:
+            if model.model_config.get("extra") != "allow":
+                model.model_config["extra"] = "forbid"
+        yield
+    finally:
+        for model, config in model_configs.items():
+            model.model_config.clear()
+            model.model_config.update(config)
 
 
 def generate_global_components_types() -> None:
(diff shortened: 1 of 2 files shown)


michael-genson:
Thanks for this! I've updated the test to fix some import issues but otherwise left the PR as-is.

justadityaraj:
Thanks!

PR #8357, opened 2026-09-10
fix: Improve OpenAI response JSON preprocessing for local LLMs

anglim3:

OpenAI-compatible local servers (for example Ollama) often wrap JSON in markdown code fences, add bold markers, surround the payload with extra text, or return a bare JSON array instead of the expected object. Mealie currently only strips null characters before `model_validate_json`, so those responses fail to parse.

This change keeps `_preprocess_response` generic and provider-agnostic:

- Strip markdown code fences (`` ``` `` / `` ```json ``)
- Strip markdown bold markers
- Strip surrounding whitespace
- Extract the outermost JSON object or array when it is surrounded by text
- If the response is a bare JSON array and the schema has a list field, wrap it as `{field_name: arr}`

Rationale: this helps OpenAI-compatible local servers that wrap JSON in markdown, lack structured-output `parse()`, or otherwise emit slightly non-strict JSON text. No prompt, UX, or provider-selection behavior is changed.

File changed:
- `mealie/schema/openai/_base.py` only

None. Compatibility improvement for OpenAI-compatible local servers.

## Testing

Unit coverage of the existing null-stripping path is unchanged. Additional preprocessing cases (fenced JSON, surrounding text, bare arrays) were verified against the OpenAI schema models.

## AI / LLM Assistance

An LLM coding assistant implemented this change from a written specification. The diff is limited to JSON response preprocessing in `mealie/schema/openai/_base.py`.

### DIFF
--- a/mealie/schema/openai/_base.py
+++ b/mealie/schema/openai/_base.py
@@ -6,7 +6,6 @@ from pydantic import BaseModel
 from mealie.core.root_logger import get_logger
 
 RE_NULLS = re.compile(r"[\x00\u0000]|\\u0000")
-RE_CODE_FENCES = re.compile(r"```(?:json)?", re.IGNORECASE)
 
 logger = get_logger()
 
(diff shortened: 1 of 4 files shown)


michael-genson:
Similar to https://github.com/mealie-recipes/mealie/pull/8133, but willing to accept this one over that one since it's simpler. Can we add tests for the JSON parsing in all the scenarios you mentioned in the PR body?

anglim3:
Added unit tests at `tests/unit_tests/schema_tests/test_openai_base_preprocess.py` covering the preprocessing scenarios from the PR body: markdown fences, bold markers, surrounding text, bare array wrapping, null stripping, and empty/none responses.

mealie/schema/openai/_base.py
```diff
@@ -18,13 +19,49 @@ class OpenAIBase(BaseModel):
...
+
     @classmethod
     def _preprocess_response(cls, response: str | None) -> str:
         if not response:
             return ""
 
         response = re.sub(RE_NULLS, "", response)
-        return response
+
+        # Local OpenAI-compatible servers (e.g. Ollama) often wrap JSON in markdown.
```
michael-genson: This never receives the malformed responses. `_get_raw_response` calls `client.chat.completions.parse(response_format=...)`, and the SDK validates `message.content` against the schema inside that call. So fenced or wrapped JSON fails there, and this preprocessing only ever runs on responses that are already valid. Can we switch `_get_raw_response` to `chat.completions.create` with the schema passed as `response_format`, so the raw content reaches `parse_openai_response`? Can we also add a test that goes through `get_response`, since the current tests call `parse_openai_response` directly?

mealie/schema/openai/_base.py
```diff
@@ -18,13 +19,49 @@ class OpenAIBase(BaseModel):
...
+        response = RE_CODE_FENCES.sub("", response)
+        response = response.replace("**", "")
```
michael-genson: Can we drop the `**` stripping? It runs on every response, OpenAI included, and removes bold from string values, like the markdown in `OpenAICompiledSource.content`. The fence regex has the same problem with backticks inside values. Could we only strip a fence at the start and end of the response?

michael-genson (changes requested):
Thanks for this! A couple of changes needed:

anglim3:
Thanks for the review.

`_get_raw_response` now uses `chat.completions.create` with the schema still passed as `response_format`, so structured outputs stay in place but raw `message.content` reaches `parse_openai_response`. Added a `get_response` test that feeds fenced JSON through that path.

Preprocessor no longer strips `**` globally. Fence handling only removes an opening/closing fence at the start and end of the response, so backticks or bold inside JSON string values are left alone.

michael-genson:
Hey there, I've gone ahead and merged https://github.com/mealie-recipes/mealie/pull/8133 which does more or less the same thing this PR does, so I'm going to close it. Thanks for your work on this either way!

PR #8358, opened 2026-09-10
fix: Prefer json_object chat completions for broader OpenAI-compatible providers

anglim3:

`client.chat.completions.parse` with a Pydantic `response_format` depends on structured-output `parse()` support. Many OpenAI-compatible local servers (for example Ollama) do not implement that endpoint and instead accept `chat.completions.create` with `response_format={"type": "json_object"}`. Some of those servers also put the usable text in a `reasoning` field when `message.content` is empty.

This change updates `_get_raw_response` / `get_response` only:

- Call `client.chat.completions.create` with `response_format={"type": "json_object"}`
- Inject the response schema JSON into the system prompt so the model still has the expected shape
- If `message.content` is empty and `message.reasoning` is present, use the reasoning text before `parse_openai_response`

Rationale: this helps OpenAI-compatible local servers that wrap JSON in markdown, lack structured-output `parse()`, or put text in a reasoning field. No ingredient UX, auto-create, or confidence-score behavior is changed.

File changed:
- `mealie/services/openai/openai.py` only

None. Compatibility improvement for OpenAI-compatible local servers.

## Testing

Existing OpenAI service tests that mock `_get_raw_response` or `get_response` remain valid. The new completion path and reasoning fallback were checked against the service call shape.

## AI / LLM Assistance

An LLM coding assistant implemented this change from a written specification. The diff is limited to the chat-completion request/response handling in `mealie/services/openai/openai.py`.

### DIFF
--- a/mealie/services/openai/openai.py
+++ b/mealie/services/openai/openai.py
@@ -278,11 +278,13 @@ class OpenAIService(BaseService):
         self, prompt: str, content: list[dict], response_schema: type[T], provider: AIProviderOut
     ) -> ChatCompletion:
         client = self.get_client(provider)
-        return await client.chat.completions.parse(
+        schema_json = json.dumps(response_schema.model_json_schema())
+        system_content = f"{prompt}\n\nRespond with a JSON object that matches this schema:\n{schema_json}"
+        return await client.chat.completions.create(
             messages=[
                 {
                     "role": "system",
-                    "content": prompt,
+                    "content": system_content,
                 },
                 {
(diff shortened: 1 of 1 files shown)


michael-genson:
Ollama supports structured outputs the way we've implemented it: https://docs.ollama.com/capabilities/structured-outputs#generating-structured-json-with-a-schema

This change would break _all_ providers by relying on models to produce valid schema via prompting rather than via the spec, which is exactly what structured outputs is designed for.

PR #8359, opened 2026-09-10
fix: remove stacked variant on number inputs

p0lycarpio:

_(REQUIRED)_

This PR removes the "stacked" attribute from numeric fields, which displays the increase and decrease buttons side by side.
Stacked buttons have the drawback of being less accessible due to their small size, especially with "compact" fields such as those for ingredients.

The size of some fields has been slightly adjusted

_(REQUIRED)_

None

## AI / LLM Assistance

_(REQUIRED)_

`Ctrl + F`

### DIFF
--- a/frontend/app/components/Domain/Group/GroupAIProviderDialog.vue
+++ b/frontend/app/components/Domain/Group/GroupAIProviderDialog.vue
@@ -59,7 +59,6 @@
           type="number"
           :min="0"
           hide-details
-          control-variant="stacked"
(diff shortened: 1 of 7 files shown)


michael-genson (approved):
> `Ctrl + F`

👌 

LGTM. I originally implemented stacked because we used to implement it manually before the VueJS native version existed. I personally haven't had issues with the size but I can definitely see how it can be an issue.

PR #8373, opened 2026-09-12
feat: private recipes per author with server-side visibility

marcobernagit:
## Summary

Recipes in Mealie belong to the household, so every member sees every recipe. This adds an opt-in per-recipe `private` flag: a private recipe is visible only to its author, while group admins keep full visibility.

## Motivation / use case

- Every family member has their own account and adds their own recipes.
- Most recipes are shared with the household.
- Some recipes should stay private until the author decides to share them.

## Changes

**Backend**
- `RecipeSettings` gains a `private` boolean (model, schema, Alembic migration).
- `RepositoryRecipes` becomes privacy-aware when scoped to a user: `by_user(user_id, is_admin)` plus a visibility predicate applied in `page_all`, `get_one`, `get_by_slug` and `get_random`. `RecipeSuggestionMixin.find_suggested_recipes` is filtered too.
- `RecipeService` scopes `group_recipes` to the current user, so detail, duplicate, update/patch and image operations enforce visibility. Bulk actions and exports are filtered as well.
- Non-owners get a `404` (no existence leak); exports return a clean `404`.

**Frontend**
- "Private (only visible to me)" switch in the recipe settings (enabled only for the author).
- Author shown on cards and recipe detail.
- "My recipes" filter in the recipe list.

**Tests**
- `tests/integration_tests/user_recipe_tests/test_recipe_private.py` covers author / other same-group user / group admin visibility and denies update, delete and export to non-owners.

## Notes

- Only `en-US` locale strings were added.
- The migration adds `recipe_settings.private` with `server_default=false`.
- I ran `ruff` and the new/related tests locally.

## Disclosure: this was vibe-coded by a non-coder

I'm **not a developer**. This feature was built with AI assistance ("vibe coding"), and the fork exists to run it in a small family setup. I tried to follow the project conventions, added tests and kept the change as small as I could, but I can't guarantee it matches the maintainers' standards: please review it critically and tell me what to change, or feel free to take it over. No hard feelings if it's out of scope for the project.


michael-genson:
Hey there, appreciate the work on this one, but I'm going to go ahead and close it. In general PRs need to be closer to ready even when using an LLM. I'm also not convinced this is a feature we want to implement since we already have support for private recipes through households.

marcobernagit:
Tks Michael for the quick response. I understand your point, no surprise.
Just thought it would have been fair to let you know about it.
Nice product, anyway.

Have a great day

Il giorno sab 12 set 2026 alle ore 18:00 Michael Genson <
***@***.***> ha scritto:

PR #8376, opened 2026-09-12
fix: pin all resolved addresses in one curl RESOLVE entry so IPv4 fallback works

### LINKED ISSUE #8375: [SCRAPER] -  safehttp recipe scraper fails when host has IPv6 records but container has no IPv6 connectivity
### First Check

- [x] I have verified that this issue _is not_ related to the underlying library
[hhyrsev/recipe-scrapers](https://github.com/hhursev/recipe-scrapers) by **1)** checking
the [debugger](https://demo.mealie.io/g/home/r/create/debug) and data is returned, **2)**
verifying that there _are_ errors in the log related to application level code, or
**3)** verified that the site provides recipe data, or is otherwise supported by
[hhyrsev/recipe-scrapers](https://github.com/hhursev/recipe-scrapers)

- [x] This issue can be replicated on the demo site (https://demo.mealie.io/)

### Please provide 1-5 example URLs that are having errors

I'm running Mealie `nightly` and URL recipe imports began failing with:

```text
httpx.ConnectError: Failed to perform, curl: (7) Failed to connect to healthfulblondie.com:443 after 0 ms: Could not connect to server
```

The container itself has normal IPv4 connectivity.

### Reproduction

From inside the Mealie container:

```bash
curl https://healthfulblondie.com
```

works.

Direct Python `httpx` also works:

```python
import httpx
r = httpx.get("https://healthfulblondie.com", follow_redirects=True)
print(r.status_code, len(r.content))
```

Result:

```text
200 771090
```

Direct `curl_cffi` requests using default, Chrome, Chrome 136, and Safari impersonation all work as well.

However, calling Mealie's scraper directly:

```python
import asyncio
from mealie.services.scraper.fetch import safe_scrape_html

async def main():
    html = await safe_scrape_html("https://healthfulblondie.com")
    print(len(html))

asyncio.run(main())
```

fails with the same curl error 7.

### Cause

`healthfulblondie.com` resolves inside the container to:

```text
IPv4:
104.21.35.84
172.67.216.54

IPv6:
2606:4700:3035::ac43:d836
2606:4700:3037::6815:2354
```

Testing the addresses individually with `curl --resolve` confirms that both IPv4 addresses work, while the IPv6 addresses fail because the container does not have working IPv6 connectivity.

The problem appears to be in:

```text
mealie/pkgs/safehttp/transport.py
```

`_SafeTransportMixin._validate()` resolves all addresses with `socket.getaddrinfo()` and returns all resolved addresses to curl using `CurlOpt.RESOLVE`:

```python
return [f"{host}:{port}:{ip}" for ip in ips]
```

This causes `AsyncSafeTransport` to fail when IPv6 DNS records exist but IPv6 connectivity is unavailable, even though the destination is fully reachable over IPv4.

Normal curl succeeds because it can handle address-family selection/fallback itself, whereas the safe transport pins the DNS results.

### Confirmed workaround

Changing the returned pins to IPv4 only:

```python
return [
    f"{host}:{port}:{ip}"
    for ip in ips
    if isinstance(ip, ipaddress.IPv4Address)
]
```

immediately fixes `safe_scrape_html()` and restores recipe URL importing.

This is only a temporary workaround; ideally `AsyncSafeTransport` should retain SSRF/DNS-rebinding protection while gracefully falling back between validated IPv6 and IPv4 addresses when IPv6 connectivity is unavailable.

### Please provide your logs for the Mealie container `docker logs <container-id> > mealie.logs`

httpx.ConnectError: Failed to perform, curl: (7) Failed to connect to healthfulblondie.com:443 after 0 ms: Could not connect to server. See https://curl.se/libcurl/c/libcurl-errors.html first for more details.

### Deployment

Unraid

chiliec:

Recipe imports fail with `curl: (7) Failed to connect ... after 0 ms` on hosts that resolve to both IPv6 and IPv4 addresses when the container has no IPv6 connectivity (very common for Cloudflare-fronted recipe sites behind Docker's default IPv4-only network).

Cause: `_SafeTransportMixin._validate()` returns one `CURLOPT_RESOLVE` entry per address (`host:port:ipA`, `host:port:ipB`, ...). curl treats each `HOST:PORT:ADDRESS` entry as a full replacement of the DNS-cache entry for that host+port, so only the **last** address survives and curl never tries the others. When the last one is an IPv6 address, the connection fails immediately with no fallback. (The pre-existing `_patch_resolver` tests only used a single IPv4 address, so this never showed up.)

Fix (`mealie/pkgs/safehttp/transport.py`):
- Emit a single `HOST:PORT:ADDR1,ADDR2,...` entry — the multi-address form curl documents — so curl's own address-family fallback (Happy Eyeballs) works while the connection is still pinned to exactly the addresses we validated. SSRF / DNS-rebinding protection is unchanged: every address is still validated before pinning, and curl still can't re-resolve.
- IPv6 addresses are wrapped in `[...]` as the curl docs require for the multi-address form.

Tests (`tests/unit_tests/pkgs/safehttp/test_transport.py`):
- updated `test_pin_covers_all_resolved_addresses` to the single-entry shape
- new `test_pin_brackets_ipv6_addresses`
- new `test_async_transport_falls_back_to_reachable_pinned_address` (parametrised: IPv4 first / IPv6 first) — resolves a host to an unreachable documentation-range IPv6 address plus `127.0.0.1` and asserts the request to a local server still succeeds

Fixes #8375

I did not go the IPv4-only route suggested as a workaround in the issue, since that would break IPv6-only hosts. Letting curl pick among the validated addresses keeps both the pin and dual-stack behaviour.

## Testing

Reproduced against curl_cffi 0.16.2 with a local HTTP server: pinning `['h:port:127.0.0.1', 'h:port:2001:db8::1']` fails with curl error 7 (only the last entry is used), while `['h:port:[2001:db8::1],127.0.0.1']` connects fine.

```
uv run pytest tests/unit_tests/pkgs/safehttp/ -q        # 81 passed
ruff check / ruff format --check / mypy on the touched files   # clean
```

RED→GREEN: with the fix stashed, `test_pin_brackets_ipv6_addresses` and the fallback case where the IPv6 address comes last fail (curl error 7, since only the last RESOLVE entry is used); with the fix all 81 tests pass.

## AI / LLM Assistance

This PR was prepared with an AI assistant (Claude) under my direction; I reviewed the diff and ran the tests myself.

### DIFF
--- a/mealie/pkgs/safehttp/transport.py
+++ b/mealie/pkgs/safehttp/transport.py
@@ -74,6 +74,10 @@ def _matches(host: str, ips: Sequence[IPAddress], hosts: set[str], networks: lis
     return False
 
 
+def _resolve_address(ip: IPAddress) -> str:
+    return f"[{ip}]" if isinstance(ip, ipaddress.IPv6Address) else str(ip)
+
+
 class _SafeTransportMixin:
     """
     Shared SSRF protection for the sync and async curl transports.
@@ -161,7 +165,7 @@ class _SafeTransportMixin:
         if literal is not None:
             # curl connects straight to the literal address; nothing to pin.
             return None
-        return [f"{host}:{port}:{ip}" for ip in ips]
+        return [f"{host}:{port}:{','.join(_resolve_address(ip) for ip in ips)}"]
 
     def _warn(self, request: httpx.Request, reason: str) -> None:
         if self._log:
--- a/tests/unit_tests/pkgs/safehttp/test_transport.py
+++ b/tests/unit_tests/pkgs/safehttp/test_transport.py
@@ -69,7 +69,14 @@ def test_pin_covers_all_resolved_addresses(monkeypatch):
     _patch_resolver(monkeypatch, ["93.184.216.34", "93.184.216.35"])
     transport = AsyncSafeTransport()
     resolve = transport._validate(_request("http://example.test/"))
-    assert resolve == ["example.test:80:93.184.216.34", "example.test:80:93.184.216.35"]
+    assert resolve == ["example.test:80:93.184.216.34,93.184.216.35"]
+
+
+def test_pin_brackets_ipv6_addresses(monkeypatch):
+    _patch_resolver(monkeypatch, ["2606:4700:4700::1111", "93.184.216.34"])
+    transport = AsyncSafeTransport()
+    resolve = transport._validate(_request("https://example.test/"))
+    assert resolve == ["example.test:443:[2606:4700:4700::1111],93.184.216.34"]
 
 
 def test_rejects_when_any_resolved_address_is_unsafe(monkeypatch):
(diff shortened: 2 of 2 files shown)


michael-genson:
Hey there, thanks for this. I didn't see this PR when I opened https://github.com/mealie-recipes/mealie/pull/8376, I pulled a few of your changes into my PR. Gonna close this one in favor of https://github.com/mealie-recipes/mealie/pull/8376, appreciate your work on this!

PR #8378, opened 2026-09-13
fix: skip unparseable nutrition values during unit conversion

### LINKED ISSUE #8377: [BUG] - KeyError: sodiumContent crashes /api/recipes/create/ai when AI-provided nutrition text does not match the expected pattern
### First Check

- [ ] This issue can be replicated on the demo site (https://demo.mealie.io/). (not tried — this only reproduces with a custom AI provider whose nutrition output doesn't hit the regex below; demo.mealie.io uses a different provider so it may not trigger)

`POST /api/recipes/create/ai` (and the "Import with AI" UI) can fail with an uncaught `KeyError: 'sodiumContent'` (or `'cholesterolContent'`), returning a generic 400 "Something went wrong while creating the recipe" with no useful detail to the user.

Root cause is in `clean_nutrition` in `mealie/services/scraper/cleaner.py`:

```python
output_nutrition = {}
for key, val in nutrition.items():
    with contextlib.suppress(AttributeError, TypeError):
        if matched_digits := MATCH_DIGITS.search(val):
            output_nutrition[key] = matched_digits.group(0).replace(",", ".")

for key in ["sodiumContent", "cholesterolContent"]:
    if val := nutrition.get(key, None):
        if isinstance(val, str) and "m" not in val and "g" in val:
            with contextlib.suppress(AttributeError, TypeError):
                output_nutrition[key] = str(float(output_nutrition[key]) * 1000)
```

The second loop checks `nutrition.get(key)` (the **raw** AI-supplied dict) to decide whether to run the mg-conversion, but then reads `output_nutrition[key]` (the **cleaned** dict) to do the actual conversion. `output_nutrition` only gets a `key` entry if the first loop's `MATCH_DIGITS.search(val)` succeeded. If the raw value exists, contains `"g"` and not `"m"`, but didn't match `MATCH_DIGITS` in the first loop (or was skipped there via the suppressed `AttributeError`/`TypeError`), then `output_nutrition[key]` doesn't exist — and unlike the first loop, this block only suppresses `AttributeError`/`TypeError`, not `KeyError`, so it propagates all the way up and fails the whole recipe-creation request.

This is intermediate/AI-response-shape-dependent rather than deterministic: the same input image, re-submitted, succeeded on one attempt and failed with this traceback on another, since it depends on the exact string the model returns for `sodiumContent`/`cholesterolContent`.

### Steps to Reproduce

1. Configure a custom OpenAI-compatible AI provider (Settings → AI Providers) — I used a local Ollama instance (`llama3.2:1b` as Default Provider, `moondream` as Image Provider), but the bug isn't specific to Ollama — any provider that returns a `sodiumContent`/`cholesterolContent` string containing `"g"` but not `"m"`, in a form `MATCH_DIGITS` doesn't match, will hit it.
2. Go to Create → Import with AI, and import a recipe from an image (or `POST /api/recipes/create/ai` with `images=<file>` directly).
3. Depending on the model's exact nutrition-field output, the request either succeeds or fails with a 400 and the traceback below. Retrying the same image sometimes succeeds, sometimes fails, since the AI's raw output text isn't deterministic between calls.

```
ERROR - Error in streaming recipe creation
  File ".../mealie/routes/recipe/recipe_crud_routes.py", line 278, in run
    slug = await create(on_progress)
  File ".../mealie/routes/recipe/recipe_crud_routes.py", line 414, in create
    recipe = await ai_service.create_from_ai(
  File ".../mealie/services/recipe/ai_recipe_service.py", line 51, in create_from_ai
    recipe_data = await self.build_recipe(
  File ".../mealie/services/recipe/ai_recipe_service.py", line 104, in build_recipe
    result = await RecipeImportWorkflow().run(ctx)
  File ".../mealie/services/recipe/import_workflow/workflow.py", line 43, in run
    await step.run(ctx)
  File ".../mealie/services/recipe/import_workflow/steps/build_recipe.py", line 38, in run
    ctx.draft_recipe = cleaner.clean(to_recipe(ctx, response), ctx.translator)
  File ".../mealie/services/recipe/import_workflow/recipe_conversion.py", line 61, in to_recipe
    nutrition=convert_nutrition(openai_recipe),
  File ".../mealie/services/recipe/import_workflow/recipe_conversion.py", line 20, in convert_nutrition
    cleaned = cleaner.clean_nutrition(raw)
  File ".../mealie/services/scraper/cleaner.py", line 600, in clean_nutrition
    output_nutrition[key] = str(float(output_nutrition[key]) * 1000)
KeyError: 'sodiumContent'
```

Client-visible response:
```json
{"detail":{"message":"Something went wrong while creating the recipe. Please try again","error":true,"exception":null}}
```

### Suggested fix

In the `sodiumContent`/`cholesterolContent` loop, read/write consistently from the same source, e.g.:

```python
for key in ["sodiumContent", "cholesterolContent"]:
    if val := nutrition.get(key, None):
        if isinstance(val, str) and "m" not in val and "g" in val:
            with contextlib.suppress(AttributeError, TypeError, KeyError):
                if key in output_nutrition:
                    output_nutrition[key] = str(float(output_nutrition[key]) * 1000)
```

(or convert directly from `val`/`matched_digits` rather than round-tripping through `output_nutrition`). Either way, at minimum `KeyError` should be added to the suppressed exceptions so a partially-parsed nutrition object degrades gracefully instead of failing the entire recipe import.

### Mealie Version

`ghcr.io/mealie-recipes/mealie:v3.25.1`

### Deployment

Docker (Linux)

Self-hosted, SQLite backend, single-container deployment. AI providers are custom OpenAI-compatible endpoints (self-hosted Ollama for Default/Image providers) rather than OpenAI itself — noting this since it's likely *why* this surfaces: a smaller/self-hosted model is more likely to produce a nutrition value shape that `MATCH_DIGITS` doesn't match than GPT-5 is, but the underlying bug is in `clean_nutrition`'s own dict-consistency, not provider-specific.

YibingZhang325:

AI recipe imports can fail when sodium or cholesterol text contains `g` but no number, such as `trace g`. The nutrition cleaner skips the value during parsing, then raises `KeyError` when converting the missing value to milligrams.

- Only convert sodium and cholesterol values that were successfully parsed. Unparseable fields are omitted, while other nutrition values are preserved.
- Add regression tests for both fields, including malformed values, zero, decimal and comma-decimal gram values.
- Add API tests confirming that AI recipe creation succeeds and preserves valid nutrition when either field cannot be parsed.

Fixes #8377

## Testing

- Confirmed six new regression cases fail with `KeyError` before the fix.
- `task py:test -- tests/unit_tests/services_tests/scraper_tests tests/integration_tests/user_recipe_tests/test_recipe_create_from_ai.py`: 181 passed.
- Backend Ruff lint, formatting, and mypy passed (464 source files). Ruff checks on both changed test files passed.
- `git diff --check`: passed.
- `task py:check` (with `PYTEST_ADDOPTS=--maxfail=1`): formatting, lint and mypy passed; full backend suite finished with 2901 passed and 18 skipped. Reran outside the sandbox because local HTTP server tests require binding a loopback port.
- `CI=true pnpm_config_verify_deps_before_run=false task ui:check`: ESLint passed; 36 test files and 391 tests passed. Used existing frontend dependencies after `nuxt prepare` in the isolated worktree.

Both final check runs completed successfully with no failing tests. The 18 backend skips were:

- 9 LDAP login tests requiring the LDAP service configured in GitHub Actions.
- 2 route-mixin tests requiring `psycopg2` from the PostgreSQL extra.
- 6 recipe-parser cases marked for manual execution when updating the parser.
- 1 bulk-import test marked as a long-running scraper test.

These skipped tests were not executed or claimed as passing.

API tests use mocked AI responses; no live AI provider was required or tested.

## Release notes

Fixed recipe imports failing when sodium or cholesterol nutrition text cannot be parsed.

## AI / LLM Assistance

OpenAI Codex assisted with the implementation. I reviewed the final diff.


YibingZhang325:
I checked the current mealie-next branch again and this issue still appears to be present in clean_nutrition(). The sodiumContent / cholesterolContent conversion still reads output_nutrition[key] even when that key may not have been added during parsing, so malformed AI-provided nutrition values can still raise a KeyError.
The original issue #8377 was closed after the issue-template comment, but I don't see a separate fix merged for the underlying bug. This PR adds the guard and regression coverage for that case.

PR #8213, opened 2026-08-26
fix: ignore blank Paprika ingredient lines

justadityaraj:
## What this PR does / why we need it:

**Context:** Paprika exports can end their ingredient text with blank lines. The migration currently retains those as empty ingredient entries, which can later make ingredient parsing fail.

**Changes:** Drop blank lines while preserving non-blank ingredient text and order.

**User impact:** Paprika recipes with trailing blank ingredient lines import without leaving entries that break ingredient parsing.

## Which issue(s) this PR fixes:

Related to #7929. This addresses blank ingredient lines; recognizing ingredient group headings remains separate.

## Testing

- `pytest --noconftest tests/unit_tests/services_tests/test_paprika_migration.py -q` — 1 passed
- `ruff check mealie/services/migrations/paprika.py tests/unit_tests/services_tests/test_paprika_migration.py` — passed
- `ruff format --check mealie/services/migrations/paprika.py tests/unit_tests/services_tests/test_paprika_migration.py` — passed
- `mypy mealie/services/migrations/paprika.py` — passed
- Full backend pytest run reached 100%; it had four unrelated failures: one randomized webhook boundary assertion, two Windows-only path/file-lock cases, and the LDAP security test under the process-local LDAP import stub used because `python-ldap` is unavailable on Windows.

## AI / LLM Assistance

OpenAI Codex assisted with repository analysis, test drafting, implementation, and review. The resulting diff was reviewed and the checks above were run locally.

Issue #7929, opened 2026-07-22
[BUG] - Paprika 3 imports incorrectly – doesn't trim newlines from ends of ingredient lists, ingredient groupings treated as ingredients


zang74:
### First Check

- [x] This is not a feature request.
- [x] I added a very descriptive title to this issue (title field is above this).
- [x] I used the GitHub search to find a similar issue and didn't find it.
- [x] I searched the Mealie documentation, with the integrated search.
- [x] I already read the docs and didn't find an answer.
- [x] This issue can be replicated on the demo site (https://demo.mealie.io/).

### What is the issue you are experiencing?

Recipes imported from Paprika 3 are often imported with a trailing "\n". Newlines should be trimmed if they are followed by nothing.

[Tri-Level Brownies (Layered Bar Cookies).zip](https://github.com/user-attachments/files/30271436/Tri-Level.Brownies.Layered.Bar.Cookies.zip)

As a result, ingredient lists will not parse. 

As a side note, ingredient groupings are not recognized as sections by either the natural language or brute, even though they're clearly distinct from recipes in exported text.

### Steps to Reproduce

1) Import Paprika 3 Recipe with groupings and/or trailing newlines.
2) Attempt to parse via brute or natural language methods

### Please provide relevant logs

, 'B_NAME_TOK'].
[DEBUG|parser|L71] 2026-07-22T10:27:04: Parsing sentence "90g. butter, melted" using "en" parser.
[DEBUG|preprocess|L199] 2026-07-22T10:27:04: _remove_price_annotations: 90g. butter, melted
[DEBUG|preprocess|L199] 2026-07-22T10:27:04: _replace_en_em_dash: 90g. butter, melted
[DEBUG|preprocess|L199] 2026-07-22T10:27:04: _replace_html_fractions: 90g. butter, melted
[DEBUG|preprocess|L199] 2026-07-22T10:27:04: _replace_unicode_fractions: 90g. butter, melted
[DEBUG|preprocess|L199] 2026-07-22T10:27:04: combine_quantities_split_by_and: 90g. butter, melted
[DEBUG|preprocess|L199] 2026-07-22T10:27:04: _identify_fractions: 90g. butter, melted
[DEBUG|preprocess|L199] 2026-07-22T10:27:04: _split_quantity_and_units: 90 g. butter, melted
[DEBUG|preprocess|L199] 2026-07-22T10:27:04: _remove_unit_trailing_period: 90 g. butter, melted
[...]

PR #8384, opened 2026-09-13
fix: import recipe notes from Paprika archives

lotusk:

Recipes imported from a Paprika archive always came in with no notes, even when the export carried them. The Paprika migrator maps the export's `notes` string into a `RecipeNote` object, but every import path then goes through the shared `clean_notes()` helper in `mealie/services/scraper/cleaner.py`, which only accepts `str` and `dict` and silently `continue`s past anything else — so each note was dropped without an error or a log line. Paprika is the only migrator that emitted model objects here; the other seven all pass strings, which is why only Paprika was affected. This changes the Paprika alias to emit a plain dict, matching what the cleaner expects and what the other migrators already do, and drops the now-unused `RecipeNote` import.

Related to #7929 — that issue also reports ingredient-grouping headers being treated as ingredients, which this PR does not address, so it should stay open. (The trailing-newline half of that issue was already fixed by #8213.) The missing-notes problem specifically was reported in [this comment](https://github.com/mealie-recipes/mealie/issues/7929#issuecomment-3121775386).

I fixed this on the migrator side rather than in `clean_notes()` deliberately: the cleaner is shared by every import path (scraper, all migrators), so widening its accepted types is a larger blast radius than bringing the one outlier back in line with the existing convention. That said, a cleaner that silently discards data it doesn't recognise is arguably worth hardening separately — happy to follow up if you'd prefer that direction instead.

## Testing

Added `test_paprika_imports_notes`, which builds a minimal Paprika archive in memory (a gzipped `.paprikarecipe` inside a zip) with a `notes` field, runs it through the real migration endpoint, and asserts the note survives. Verified it actually catches the bug: the test fails on `mealie-next` without the one-line change and passes with it. Also confirmed against the real Paprika 3 export attached to the issue, and ran the full `tests/integration_tests/recipe_migration_tests/` suite (15 passed), plus `ruff check`, `ruff format --check`, and `mypy` on the touched files.

image

image

After the fix — the Paprika export's notes now come through. Note the ingredient "60ml burnt brandy (see note)" referring to this exact note, which was previously dropped entirely.

## AI / LLM Assistance

Used Claude Code to trace where the notes were being dropped (export file → migrator alias → shared cleaner), write the fix and the regression test, and verify the before/after behaviour.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/mealie/services/migrations/paprika.py
+++ b/mealie/services/migrations/paprika.py
@@ -7,8 +7,6 @@ import zipfile
 from gzip import GzipFile
 from pathlib import Path
 
-from mealie.schema.recipe import RecipeNote
-
 from ._migration_base import BaseMigrator
 from .utils.migration_alias import MigrationAlias
 
@@ -53,7 +51,8 @@ class PaprikaMigrator(BaseMigrator):
             MigrationAlias(
                 key="notes",
                 alias="notes",
-                func=lambda x: [z for z in [RecipeNote(title="", text=x) if x else None] if z],
+                # the cleaner only accepts strings and dicts, so a RecipeNote here is silently dropped
+                func=lambda x: [{"title": "", "text": x}] if x else [],
             ),
             MigrationAlias(
                 key="recipeCategory",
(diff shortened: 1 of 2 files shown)


michael-genson (approved):
Thanks!

PR #8385, opened 2026-09-13
feat: filter cookbooks by food label

Xyolyp:

Cookbooks can filter on individual foods, but not on what a food is. A "Seafood" cookbook means listing every fish you cook with, and it quietly stops being right as soon as a new one is imported.

#8126 added the same filter to meal plan rules and made labels an entry in the `Organizer` enum, so the field type, the picker and the query path already exist. This just exposes `recipe_ingredient.food.label_id` in the cookbook editor as well.

Three files: five lines in `CookbookEditor.vue`, one integration test, and the cookbooks paragraph in the docs, which listed only Categories, Tags and Tools. No backend change, no schema change, no migration.

None. Follow-up to #8126, which left the cookbook editor out to keep that PR to one surface.

Same two limits as #8126, since it is the same filter on a different surface: a food carries exactly one label, so labels used as shopping aisles can't double as ingredient kinds, and ingredients that were never parsed into a food can't match.

## Testing

`test_get_cookbook_recipes_by_food_label` mirrors the tag-based `test_get_cookbook_recipes` next to it: a cookbook filtered on a label returns the three recipes carrying a labelled ingredient and none of the three without it.

Lint and 391 vitest clean, ruff and mypy clean, full pytest green. Also checked in the browser.

## AI / LLM Assistance

The change and the test were written with Claude Code. I reviewed it and checked the editor in the browser before opening.


p0lycarpio (changes requested):
Thanks for this! Can you also add this filter in [recipe finder](https://github.com/Xyolyp/mealie/blob/15d24cc24665456a4654d5b9b79aec5259371bf8/frontend/app/pages/g/%5BgroupSlug%5D/recipes/finder/index.vue#L664) ?

Xyolyp:
Done. The finder had no food filter either, so I added both: `recipe_ingredient.food.id` and `recipe_ingredient.food.label_id`, in the same order as in the cookbook editor.

p0lycarpio (approved):
Thanks!

PR #8397, opened 2026-09-14
fix: allow non-ASCII characters in SMTP credentials

### LINKED ISSUE #5886: [BUG] - Test email failure related to ascii codec
### First Check

Test email fails from the Settings page

### Steps to Reproduce

1. Go to `/admin/site-settings`
2. Enter a test email, e.g. "test@mytest.com"
3. Press "Test"
4. Notice the failure
5. Open console or docker compose logs to see error

`docker compose logs` :

```
mealie  | INFO     2025-08-04T01:56:42 - [62.66.150.239:0] 200 OK "GET /api/users/self HTTP/1.1"
mealie  | ERROR    2025-08-04T01:56:45 - 'ascii' codec can't encode character '\u20ac' in position 37: ordinal not in range(128)
mealie  | ERROR    2025-08-04T01:56:45 - 'ascii' codec can't encode character '\u20ac' in position 37: ordinal not in range(128)
mealie  | INFO     2025-08-04T01:56:45 - [62.66.150.239:0] 200 OK "POST /api/admin/email HTTP/1.1"
```

### Mealie Version

- Docker image: ghcr.io/mealie-recipes/mealie:v3.0.2
- Build: [4744e6371aaa422e129bd83ccba7278ee1f24c78](https://github.com/mealie-recipes/mealie/commit/4744e6371aaa422e129bd83ccba7278ee1f24c78)

### Deployment

Docker (Linux)

Fresh install on Ubuntu 24.04 with docker / sqlite.

docker-compose.yml:

```services:
  mealie:
    image: ghcr.io/mealie-recipes/mealie:v3.0.2 # 
    container_name: mealie
    restart: always
    ports:
        - "9925:9000" # 
    deploy:
      resources:
        limits:
          memory: 500M # 
    volumes:
      - mealie-data:/app/data/
    secrets:
      - smtp-password
    environment:
      ALLOW_SIGNUP: "false"
      PUID: 1005
      PGID: 1005
      API_DOCS: "false"
      TZ: Europe/Copenhagen
      BASE_URL: https://XXXXXXXX
      SECURITY_MAX_LOGIN_ATTEMPS: 3
      DEFAULT_GROUP: Home
      DEFAULT_HOUSEHOLD: Family
      SMTP_HOST: XXXXX.XXXXX.com
      SMTP_PORT: 587
      SMTP_FROM_NAME: Mealie
      SMTP_AUTH_STRATEGY: TLS
      SMTP_FROM_EMAIL: XXXXX@XXXXX.com
      SMTP_USER: XXXXX@XXXXX.com
      SMTP_PASSWORD_FILE: /run/secrets/smtp-password

volumes:
  mealie-data:

secrets:
  smtp-password:
    file: ./secrets/smtp-password.txt
```

lotusk:

Sending any email (including the admin "Send Test Email" button) fails with `'ascii' codec can't encode character '\u20ac' ...` when the SMTP username or password contains a non-ASCII character. The cause is in the standard library: `smtplib.SMTP.auth()` base64-encodes every AUTH response via `.encode('ascii')`, and `auth_cram_md5()` does the same with the password, so `server.login()` raises a `UnicodeEncodeError` before anything reaches the server. This is still the case on Python 3.12.13. Non-ASCII characters in the subject, sender name or body were never affected, because `EmailMessage` already encodes those correctly. Only the credentials were.

- `mealie/services/email/email_senders.py`
  - Adds `UTF8AuthSMTP`, a `smtplib.SMTP` subclass that overrides `auth()` to encode AUTH responses as UTF-8. RFC 4616 already specifies UTF-8 for PLAIN.
  - Also overrides `auth_cram_md5()` for the same reason.
  - Adds `UTF8AuthSMTPSSL`, which applies the same behaviour on top of `SMTP_SSL`.
  - `Message.send` now uses these classes. Nothing else about the sending flow changes.
- `tests/unit_tests/services_tests/test_email_service.py`: adds tests for PLAIN, LOGIN and CRAM-MD5 logins with a non-ASCII username and password.

Fixes #5886

- The `auth()` override mirrors the stdlib implementation line for line; the only change is the encoding. I chose this over monkeypatching `smtplib` globally so the change stays scoped to Mealie's own sender.
- CRAM-MD5 has to be covered as well, not only PLAIN/LOGIN. `login()` tries CRAM-MD5 first whenever the server advertises it, and a `UnicodeEncodeError` is not an `SMTPAuthenticationError`, so `login()` would not fall back to another mechanism.
- The issue's log doesn't say which field contained the `€`. I reproduced the exact error locally with a non-ASCII password (and separately with a non-ASCII username). I also confirmed that a non-ASCII subject, sender name or body does not trigger it. That leaves the credentials as the only path that produces this error.

## Testing

- Added `test_smtp_login_{plain,login,cram_md5}_non_ascii_credentials`. They use a fake server that replays scripted SMTP replies instead of connecting anywhere, and assert the exact base64 AUTH responses sent (UTF-8 encoded credentials, and the correct HMAC digest for CRAM-MD5).
- Verified the tests catch the bug: with the fake server based on stock `smtplib.SMTP`, all three fail with `UnicodeEncodeError`; with the fix they pass.
- `test_email_service.py`: 7 passed. `ruff check`, `ruff format --check` and `mypy` pass on the touched files.
- I have not tested against a real SMTP server.

## AI / LLM Assistance

Used Claude Code to help find the root cause in `smtplib`, reproduce the error, write the fix and the regression tests, and verify that the tests fail without the fix.


mealie/services/email/email_senders.py
```diff
@@ -13,6 +16,60 @@
...
+
+class UTF8AuthSMTP(smtplib.SMTP):
+    """
+    smtplib encodes AUTH credentials as ASCII, so logging in with a username or password
+    containing non-ASCII characters (e.g. "€") raises a UnicodeEncodeError before anything
+    is sent to the server. PLAIN is defined as UTF-8 (RFC 4616) and servers accept UTF-8
+    for LOGIN and CRAM-MD5 as well, so encode credentials as UTF-8 instead.
+    """
+
+    def auth(
```
michael-genson: Nice find on the root cause. Can we avoid duplicating `smtplib.SMTP.auth` though? Since `auth()` encodes whatever the `auth_*` helpers return, overriding those to return a small `str` subclass whose `encode()` always uses UTF-8 gets the same result without copying the challenge loop, and it keeps inheriting any stdlib changes (we're planning to move to 3.14 before long). `auth_cram_md5` would still need its own body since it hashes the password itself.

michael-genson (changes requested):

mealie/services/email/email_senders.py
lotusk: Oh, thank you, I will take a look.

lotusk: Thanks, done. I changed it to the str subclass you suggested.

auth_plain and auth_login now just wrap what super() returns. The auth() override is gone. auth_cram_md5 still has its own body since it hashes the password itself.

I didn't touch the tests, so they also show the bytes on the wire are the same. They pass here and still fail on mealie-next.

michael-genson (approved):
Thanks! I like this much better

PR #8398, opened 2026-09-14
feat: add recipe tag ordering

feilongcheng:

Recipe cards show only the first two tags. Users currently cannot control which tags appear first.

This PR:

- Adds a Reorder Tags dialog to the Tag Data page.
- Allows users to reorder tags by dragging them.
- Saves one tag order for the whole group.
- Adds a `position` field and a database migration.
- Adds an API endpoint for saving the order.
- Returns recipe tags in the saved order.
- Keeps new tags at the end of an existing saved order.
- Adds integration tests for tag ordering.

The tag list is scrollable, so it can also work with many tags.

#### Reorder Tags button

button

#### Reorder Tags dialog

dialog

No issue.

Related discussion: https://github.com/mealie-recipes/mealie/discussions/8093

This PR uses one group-wide tag order. It does not add per-recipe tag pinning.

I would like feedback on the database field and the API endpoint.

## Testing

I manually tested:

- Opening the Reorder Tags dialog.
- Dragging tags into a new order.
- Saving the order.
- Refreshing the page and checking that the order was saved.
- Checking that recipe cards show the first two tags from the saved order.
- Creating a new tag and checking that it appears at the end.

Automated checks:

- 29 backend organizer tests passed.
- 36 frontend test files passed.
- 391 frontend tests passed.
- ESLint and Ruff checks passed.

## AI / LLM Assistance

I used Codex to help inspect the existing code, discuss the design, implement parts of the change, and review the tests. I reviewed and understood the final changes. I also tested the feature manually.


michael-genson:
Hey there, thanks for the PR! I like the goal, but I'm going to close this one. A global tag order is a lot of machinery (new column, migration, endpoint, dialog) for deciding which two tags land on a card, and it applies the same priority to every recipe in the group.

I'd rather see something like a "primary tag" you can set per recipe, so the choice lives with the recipe it affects. If you're interested in that, it's worth sketching in a discussion first so we can agree on the shape before anyone builds it.

PR #8400, opened 2026-09-14
fix: avoid useAsyncData in recipe timeline scroll handler

nzyz995:

The recipe timeline called `useAsyncData` from its infinite-scroll event handler after the component had already mounted. This produced a `NUXT_E3003` warning and registered a new async-data entry with a random key for each paginated request.
- Call the existing timeline request directly from the scroll handler.
- Preserve the loading guard and throttling behavior.
- Reset the loading state with `try/finally`.
- Remove the unused `useAsyncKey` import.

No existing issue. This was discovered while testing the recipe timeline locally.

## Testing

- Opened the recipe timeline and scrolled near the bottom.
- Confirmed the next page of events was loaded.
- Confirmed repeated scrolling continued to load events normally.
- Confirmed `NUXT_E3003` no longer appeared in the Console.

## AI / LLM Assistance

CodeX was used to help analyze the Nuxt lifecycle warning, trace the
infinite-scroll execution path, suggest the fix. I manually applied, reviewed, and tested the change.


michael-genson (approved):
Thanks!

PR #8402, opened 2026-09-14
feat: shopping list condensed view

kylerk:

The original shopping list view, when used on a mobile phone, does not efficiently show the maximum number of items.  I've added a condensed view option, and the option to hide the section labels to the shopping list.  This has changed the shopping list from being able to only show 5 or 6 items on the screen, to sometimes showing up to 20 + items when condensed and labels are hidden.

Labels are extremely useful for getting your list in an organized state, but once your in the store shopping, it can be helpful to hide them to be able to focus on what foods to buy.

image
image

image

While I was developing this change, many small little tweaks to padding, spacing and text wrapping on the condensed page were done to try to maximize it's use.

I haven't found any specific issues with this request.

## Testing

I've been using this updated view for about 2 weeks while using the app for real life shopping list creation and on real shopping trips. I use it on iOS.  My wife has also been using it for the same period on Android.  I occasionally work on the shopping list on my windows PC.

I have not looked into localization support of this feature yet.  If you want to accept the feature,  I will be happy to figure out how to test it with localization.

## AI / LLM Assistance

Claude Fable was used to create this PR.  I have looked over all the changes manually, however I'm not an extremely experienced web dev.


kylerk:
I updated the code to use the label color tint if it has been setup

image

kylerk:
Put back the background coloring to only cover label area in condensed view.  Remove the 50% opacity of it.

image

michael-genson:
Looks great, haven't had time yet to review the code, but my initial reaction - I'm not sure I see a need to maintain both views, would it be better to just make this the only view?

I'm also hesitant to have a "no label" view, we used to have both (labels and no labels) and we received feedback that basically no one ever used the "no label" view so we ditched it. I'm not outright against it, but I'm not convinced we need it.

kylerk:
I agree that condensed view could become the only default view.  I'll make a commit with that implemented.

I think that "hide labels" and "show labels" should remain as options.

I got into using this type of meal planning and shopping list app to help deal with a certain level of cognitive overload, and being able to hide all the labels lets me really focus on the task at hand when in a store.  There are situations where having the labels turned on, could almost 1.5 to 2x the number of lines on screen at a time if most items are in their own category.

This would be solvable by manually adjusting all your categories, but I think that is far to much burden on the user and pushes people to just write their grocery list as notes.

michael-genson:
Can I ask - what do you use labels for if not for organizing your shopping list? Generally that's what the feature is for, for helping you organize which aisle/section each item is in (and also why you can re-order labels, so they match your store's layout).

kylerk:
> Can I ask - what do you use labels for if not for organizing your shopping list? Generally that's what the feature is for, for helping you organize which aisle/section each item is in (and also why you can re-order labels, so they match your store's layout).

The labels help create order and flow of the list automatically, i certainly appreciate them.  And I can adjust it while I'm creating the shopping list at home, but in the store, I don't need the labels.

michael-genson:
Ah okay, I see the use case. For now let's leave off the "no label" view: there are a couple of issues/bugs I see with it that I'd rather address separately. That means dropping the `hideLabels` preference, the three-dot menu toggle, the `.shopping-list--hide-labels` styles, and the two new `en-US` strings. The rest looks really good, love the saved space!

kylerk:
I did a shopping run last night with a full list (and wrangling two kids), and it was a little bit too condensed.

I've added a commit that adds about 7px of space.  Here is the before and after.

image

I've also added a commit to remove the "hide label" work for now.

frontend/app/pages/shopping-lists/[id].vue
```diff
@@ -434,14 +439,96 @@ const {
...
+    height: 32px;
+    margin-left: 0 !important;
+  }
+
+  .shopping-list-item-row .mb-2 {
+    margin-bottom: 0 !important;
+  }
+
+  /* edit is also in the drag handle's menu, so the pencil is just noise here */
+  .shopping-list-item-row .shopping-list-item__edit {
```
Comment (michael-genson) [inline frontend/app/pages/shopping-lists/[id].vue]: Can we keep the pencil? Hiding it doesn't save much space, and the drag handle menu is hard to discover as the only way to edit an item.

frontend/app/pages/shopping-lists/[id].vue
```diff
@@ -149,7 +149,10 @@
     />
 
     <!-- Viewer -->
-    <section v-if="!edit" class="py-2 d-flex flex-column ga-4">
+    <section
+      v-if="!edit"
+      class="py-2 d-flex flex-column ga-1 shopping-list-view"
```
Comment (michael-genson) [inline frontend/app/pages/shopping-lists/[id].vue:154]: On desktop the "Add Item" field is now right up against the first label, since the gap dropped from `ga-4` to `ga-1`. Can we add a bit of margin below the desktop add item block (the `<div v-else>`) so it has some room again? The mobile form already has `my-4`.

michael-genson (changes requested):
Thanks for the updates, this is looking great! Two small tweaks:

frontend/app/pages/shopping-lists/[id].vue
Comment (kylerk) [reply frontend/app/pages/shopping-lists/[id].vue]: Should be resolved in 7ac12e5d407540cebf823849f3d12efe719f68ab

frontend/app/pages/shopping-lists/[id].vue
Comment (kylerk) [reply frontend/app/pages/shopping-lists/[id].vue:154]: Should be resolved in 7ac12e5d407540cebf823849f3d12efe719f68ab

kylerk:
image

With the pencil back.  I'm proposing to give it a little bit of opacity fade to keep it visually quieter.  The final value I commited it 55%, not the 40% shown in the preview as that was too faded out.

michael-genson:
I'm good with the opacity change, we can always switch it back if it becomes a problem. Your updated one is subtle enough. How would you feel about bringing back the bold on the label text? On some combinations (i.e. lighter label colors) it's a little harder to read now.

kylerk:
image

image


I'm good with bolding it.  Pushed it.

michael-genson:
Looks great thanks! I don't see the commit for the bold change, once that's in I'll approve.
When you make the follow up PR for the hide labels option be sure to ping me, I'll take a look when I can!

kylerk:
@michael-genson Bold is pushed.

michael-genson (approved):
Thanks so much for your work on this!

PR #8403, opened 2026-09-14
feat: Crossing out shopping list items should give you a moment to fix a mistake

kylerk:
I built this feature on top of commits from [my condensed view UI change](https://github.com/mealie-recipes/mealie/pull/8402), I'll unmark this from draft if that feature is accepted.

When you are shopping, it is easy to accidentally tap the wrong thing, and have it disappear into the "Items Checked List".  It's a bit difficult and tedious to fix your mistake. There was a toast that popped up that you could use to fix your mistake, but it added cognitive load, and visual clutter to the usage of the app.

This PR includes a simple solution to this problem.  Instead of instantly sending your checked item to the checked list, the item is immediately cross out under your finger, and then waits 5 seconds to be sent to the "Items checked list".  If you want to revert this action, simply uncheck the item during the 5 second wait period, and you've solved your problem.  

Here is a video in action.   I'm using a condensed view from my other PR if you [notice it looks different](https://github.com/mealie-recipes/mealie/pull/8402)

https://github.com/user-attachments/assets/945e9d33-f7d8-46a8-8cb5-db1714f8913e

No known issues indicate this was an issue people dealt with.

## Additional work that is needed

If this PR is accepted, I will figure out if we need to fully remove the old toast translation if it is no longer necessary.

## Testing

I have used this feature in the wild for the past 2 weeks on my phone under real shopping conditions and it has worked great.

## AI / LLM Assistance

This was coded with Claude Fable.  I have limited web dev experience (I'm a game dev), but have look through all the code changes and they appear logical.


michael-genson:
Thanks for the PR, but I don't think we need another solution in addition to the "Undo" toast. I'd be open to improvements on that if you think there is a way to make it easier to use.

kylerk:
> Thanks for the PR, but I don't think we need another solution in addition to the "Undo" toast. I'd be open to improvements on that if you think there is a way to make it easier to use.

Perhaps I wasn't clear enough, this new way of leaving the crossed out item for  5 seconds fully replaces the current toast system.

This new way has significant advantages over the toast system.  If you quickly cross out 3 things on the list, the original toasts overlap on top of each other and you cannot fix the first thing if it was a misclick.

This new approach allows for fixing mis-clicks on multiple items being checked off at the same time.

It's both a cleaner look and more error resistant.

michael-genson:
The app used to do something similar but user feedback indicated it was a poor experience so we removed it. Much later we added the toast.

I'd welcome improvements to the toast undo workflow, but it seems to be popular enough in its current state that we haven't received any negative feedback about it.

PR #8410, opened 2026-09-15
fix: use relative redirects for SPA directory routes

### LINKED ISSUE #7251: [BUG] - Redirect from /login to /login/ uses HTTP Host header, not BASE_URL, which breaks OIDC login
### First Check

The 307 redirect from /login to /login/ uses the HTTP `Host` header rather than the specified BASE_URL. When there is a reverse proxy in front of Mealie that uses a different internal `Host` value/header, this completely breaks the OIDC login flow. This traces back to a shortcoming in uvicorn (https://github.com/Kludex/uvicorn/issues/965 and https://github.com/Kludex/uvicorn/pull/2811) but can be worked around if Mealie just correctly uses the "/login/" URL.

### Steps to Reproduce

1) Make a request to /login using a random HTTP `Host` header value
2) Get redirected to \<random HTTP Host header value\>/login/

Image

This is a screenshot from Firefox dev tools from Firefox. I set my reverse proxy to set the `Host` header to awesome.com. Note the correct request to https://mealie.REDACTED/login which then returns a Location/redirect to https://awesome.com/login/

### Mealie Version

Docker Image - ghcr.io/mealie-recipes/mealie:latest   d163d4dbe45d
Version - v3.12.0
Build - 5aafebb7a6e893da95cb3239e806240c3b5347fe

### Deployment

Docker (Linux)

I have two domains for Mealie:

- mealie.external.com
- mealie.internal.lan

Caddy proxies for mealie.external.com and sends it to mealie.internal.lan with a `Host` header of mealie.internal.lan. This redirect problem causes a request to https://mealie.external.com/login to be redirected to https://mealie.internal.lan/login/.

lotusk:

In the production container, the frontend is served by `SPAStaticFiles`, a subclass of Starlette's `StaticFiles(html=True)`. Prerendered routes such as `/login` exist as directories (`login/index.html`). When a request omits the trailing slash, `StaticFiles` answers with a 307 whose `Location` is an **absolute** URL built from the request's `Host` header (`URL(scope=scope)`). Two problems follow:

- Behind a reverse proxy that rewrites `Host`, the browser is sent to the internal host name. This breaks the OIDC login flow, whose documented redirect URI is `/login`.
- The client controls where it is redirected: a request to `/login?direct=1` with `Host: evil.example` returns `Location: https://evil.example/login/?direct=1`.

image

This PR rewrites these redirects to a relative `Location` that keeps only the path and query. The browser resolves it against the URL it actually requested, so it works regardless of `Host`, proxies or `BASE_URL`. Leading slashes are collapsed to one, so a request like `//login` can't turn into a protocol-relative redirect (`//login/` would send the browser to a host named `login`).

- `mealie/routes/spa/__init__.py`: in `SPAStaticFiles.get_response`, rewrite `RedirectResponse` locations to a relative path. `StaticFiles` only returns a redirect for this directory case, so other responses are unaffected.
- `tests/unit_tests/test_spa_static_files.py`: add regression tests for the spoofed-Host case (query string preserved) and the leading-double-slash case.

Fixes #7251

- I chose a relative redirect over building the URL from `BASE_URL`: it needs no configuration, can't be wrong when `BASE_URL` is unset or differs from the address the user actually uses, and relative `Location` values are valid per RFC 9110.
- This only affects production (single-container) deployments. In development, `/login` is handled by the Nuxt dev server.

## Testing

- New unit tests fail on `mealie-next` (`Location` is `https://evil.example/login/?direct=1` and `http://mealie.example//login/` respectively) and pass with this change.
- End-to-end through a FastAPI app with `SPAStaticFiles` mounted, using `TestClient`:
  - `/login?direct=1` with a spoofed Host returns 307 to `/login/?direct=1`, and following it serves the login page.
  - Percent-encoded query strings are preserved.
  - `/_nuxt/*` assets and client-side routes such as `/g/home/r/<slug>` still return 200.
- Full `tests/unit_tests` suite passes; `ruff check`, `ruff format --check` and `mypy` pass on the touched files.
- Not tested against a production frontend build behind a real reverse proxy.

## AI / LLM Assistance

Used Claude Code to trace the redirect to Starlette's `StaticFiles`, reproduce it with a spoofed Host header, write the fix and regression tests, and verify the tests fail without the fix.


michael-genson (approved):
Thanks!

PR #8414, opened 2026-09-15
fix: fallback to json_object for providers without structured outputs

rtalmaci:

When using OpenAI-compatible AI providers that only support standard JSON mode (e.g. DeepSeek via `https://api.deepseek.com/v1`) and do not support Structured Outputs (`response_format={"type": "json_schema", ...}`), requests fail with a `400 BadRequestError` indicating unsupported `response_format`.

This PR introduces an automatic one-shot downgrade mechanism in `OpenAIService._get_raw_response`:
- When a `BadRequestError` specifically reporting unsupported `response_format` / `json_schema` / structured outputs is encountered, it logs the downgrade and retries the request using `response_format={"type": "json_object"}`.
- Unrelated 400 errors (e.g., parameter errors) are never retried.
- Default system prompts have been updated to explicitly instruct models to return JSON, satisfying requirements of JSON-mode-only providers.

**File Changes:**
- `mealie/services/openai/openai.py`: Added `_is_unsupported_response_format_error` helper and fallback logic in `_get_raw_response` to retry with `{"type": "json_object"}` upon `BadRequestError`.
- `mealie/services/openai/prompts/*`: Updated system prompt templates (`build-recipe.txt`, `parse-recipe-ingredients.txt`, `parse-recipe-video.txt`, `resolve-organizers.txt`, `debug.txt`, `transcribe-audio.txt`) to explicitly mention JSON output.
- `tests/unit_tests/services_tests/test_openai_service.py`: Added comprehensive unit tests covering:
  - Successful structured outputs path remains unchanged.
  - DeepSeek-style 400 error retrying with `json_object` and parsing properly.
  - Handling and stripping of markdown JSON fences on downgraded responses.
  - Non-retry on unrelated 400 bad request errors.

_None / fixes compatibility with DeepSeek and JSON-mode-only OpenAI-compatible providers_

- Transcription / audio paths remain untouched as they do not use `response_format`.
- Image provider paths go through `_get_raw_response` and benefit from the same fallback.

## Testing

- Added unit tests in `tests/unit_tests/services_tests/test_openai_service.py` testing the structured output path, downgrade fallback path, markdown fence stripping, and non-retry of unrelated 400s.
- Validated with:
  - `uv run ruff check mealie/services/openai tests/unit_tests/services_tests/test_openai_service.py` (passed)
  - `uv run ruff format --check mealie/services/openai tests/unit_tests/services_tests/test_openai_service.py` (passed)
  - `uv run mypy mealie/services/openai` (passed)
  - `uv run pytest tests/unit_tests/services_tests/test_openai_service.py` (13 passed)

## AI / LLM Assistance

Assisted by AI coding assistant (Junie / JetBrains) in analyzing the error logs, implementing the fallback logic and prompt adjustments, writing unit tests, and verifying code quality.


michael-genson:
Hey there, thanks for the PR! I'm going to close this one, for the same reason as #8358: on the fallback path the response shape comes from prompting rather than from the spec, which isn't something we can rely on. Structured outputs exist so the provider guarantees the schema, and that's what we build against.

To be clear about the underlying limitation, since this keeps coming up: DeepSeek's API only supports `response_format: {"type": "json_object"}` and rejects `json_schema` outright ([their docs](https://api-docs.deepseek.com/guides/json_mode/)). That's a gap on their side, so Mealie's AI features need a provider that implements structured outputs. Ollama, OpenAI, and most OpenAI-compatible servers do.

PR #8437, opened 2026-09-17
fix: report image upload failures instead of failing silently

### LINKED ISSUE #4494: [BUG] - Missing error message on image upload
### First Check

I tried to upload photo of my meal, but photo was to large. At first I wasn't aware what was the issue. Ideally Mealie would resize image to some maximum size, or at least show error so user is aware what is the problem. 

### Steps to Reproduce

1. Create new recepie 
2. Try to add large photo eg. https://unsplash.com/photos/cooked-food-and-gold-colored-saucer-Q5mPPGld5Zk (Download original size)
3. Nothing will happen. 

Not needed. 

### Mealie Version

v2.1.0

### Deployment

Docker (Linux)

_No response_

henry1113nz:

Uploading an image that fails leaves you with no idea anything went wrong. The picture just
doesn't change.

The axios interceptor does raise a toast, but only when the error carries a Mealie error body:

    if (error?.response?.data?.detail?.message) {
      alert.error(error.response.data.detail.message as string);
    }

Image size limits aren't Mealie's, they're enforced by whatever web server sits in front of it,
so a rejected upload comes back as a plain 413 with no Mealie body and the interceptor stays
quiet. Dropped connections and timeouts land in the same place.

The handlers then discard what's left. `uploadImage` in `RecipePageEditorToolbar.vue` only ever
looked at `data`:

    const newVersion = await api.recipes.updateImage(recipe.value.slug, fileObject);
    if (newVersion?.data?.image) {
      recipe.value.image = newVersion.data.image;
    }
    imageKey.value++;

On a failure `data` is null, the `if` doesn't fire, and the function carries on as though the
upload had worked. `getImageFromURL` in `RecipeImageUploadBtn.vue` is the same shape.

Changes:

- `composables/use-toast.ts` — adds `alertUnreportedError`, which shows a generic message only
  when the interceptor hasn't already reported the error. Without that check a backend error
  would raise two toasts instead of one.
- `RecipePageEditorToolbar.vue` and `RecipeImageUploadBtn.vue` — the image calls now check the
  returned `error` and report it.

Fixes #4494

**One change here is slightly outside the issue and I'd rather point it out than have you find
it.** `deleteImage`, in the same component, already had a `try`/`catch` with an `alert.error`
for exactly this. It can never run: `request.safe` in `api-client.ts` catches everything and
resolves, so nothing is ever thrown and the only signal is the returned `error` field. I've
changed it to check `error` like the other two. It felt wrong to leave a `catch` that can't fire
sitting next to two paths I'd just fixed, but I'm happy to pull it out into its own PR if you'd
prefer this one stay strictly to the upload.

On the wording — I've gone with the existing generic `events.something-went-wrong` rather than
branching on the status code. A 413-specific message would be more useful, but you mentioned on
the issue that the limit isn't Mealie's to know, and guessing at "it may be too large" felt like
your call rather than mine. Easy to change if you'd rather have it.

## Testing

Three unit tests on `alertUnreportedError` covering the three cases it has to separate: an error
the interceptor already reported (stays quiet), an error with no Mealie body such as a web server
413 (reports), and no response at all such as a dropped connection (reports).

The call sites themselves aren't covered - mounting `RecipePageEditorToolbar` needs the page
state, user store and household store mocked, which seemed like a lot of scaffolding for a
three-line change. Happy to add it if you want it.

Full frontend suite passes (394 tests), as does eslint on the changed files.

## AI / LLM Assistance

I found and traced this one myself, including reading through the interceptor and the request
wrapper to work out why nothing was surfacing. An LLM helped draft the change and the tests, and
I reviewed and ran them.


michael-genson (approved):
Looks good thanks!

PR #7912, opened 2026-07-20
fix: Account for on hand items when adding a recipe to a shopping list

miah120:
## What this PR does / why we need it:
Previously, if 
* a recipe was added to a shopping list, 
* that recipe had on-hand items, and
* the user increased the count of that recipe to the shopping list, 
 the on hand items would be added to the shopping list as well, even though they shouldn't be. 

Now, on hand items will be ignored. 

## Which issue(s) this PR fixes:
fixes: #4547

## Testing
Tested in docker for both incrementing and decrementing the reference recipe count. 

## AI / LLM Assistance
No AI used.

PR #8441, opened 2026-09-17
fix: add explicitly selected on-hand ingredients to the shopping list

### LINKED ISSUE #8439: [BUG] - "On Hand" Items are not added to Shopping List
### First Check

When adding a recipe to a shopping list foods that are "on hand" are unchecked by default.
The expected behavior is that if I check "on hand foods", they will be added to the shopping list.

Currently, this does not work. I have to add the foods manually later. It used to work before. Unfortunately, I cannot tell which version broke this.

I could replicate this behavior on the demo site.

### Steps to Reproduce

1. Create a food item in the food database with the "on hand" flag set
2. Create a recipe using this item (and maybe other items without the "on hand" flag)
3. Add the recipe to a shopping list
4. Check the "on hand" food
5. Click "Add to list"
6. Food does not appear in the shopping list

I do not have any logs

### Mealie Version

3.26.0

### Deployment

Docker (Linux)

_No response_

chiliec:

Since #7912, `get_shopping_list_items_from_recipe` drops every ingredient whose food is "on hand" for the list's household. That is correct for the default path (adding a recipe by id, incrementing a recipe already on the list), but it also runs when the client sends an explicit `recipe_ingredients` selection. The "Add to shopping list" dialog unchecks on-hand foods by default and lets the user re-check them, and that selection is sent as `recipe_ingredients` — the backend then silently discards those items again, so checking an on-hand food does nothing.

- `mealie/services/household_services/shopping_lists.py`: the on-hand skip now only applies when no explicit ingredient list was provided (default `skip_on_hand = recipe_ingredients is None`). The flag is passed down through nested sub-recipes so their behavior matches the top-level call. `remove_recipe_ingredients_from_list` is unchanged.
- `tests/integration_tests/user_household_tests/test_group_shopping_lists.py`: new `test_shopping_lists_add_recipe_on_hand_food` — adds a recipe with one on-hand and one regular food by id (expects only the regular food) and then with an explicit `recipe_ingredients` selection (expects both).

Fixes #8439

## Testing

- `pytest tests/integration_tests/user_household_tests/test_group_shopping_lists.py -k on_hand_food` — fails on `mealie-next` (only the non-on-hand food ends up on the list), passes with this change.
- `pytest tests/integration_tests/user_household_tests/test_group_shopping_lists.py` — 27 passed.
- `ruff check` / `ruff format --check` / `mypy` on the touched files — clean.

## AI / LLM Assistance

An LLM assistant was used to trace the regression and draft the change and tests; everything was run and reviewed locally before opening this PR.


michael-genson (approved):
Thanks!

PR #8468, opened 2026-09-20
fix: allow re-importing a recipe zip into the same instance

Rouzax:

Exporting a recipe as a zip and importing it back into the same Mealie instance fails with HTTP 500, `UNIQUE constraint failed: recipes.id`. The export carries the id of the recipe it came from, and the importer passes that id straight through, so it collides with the recipe you exported. Duplicate slugs are already handled this way, ids were not.

While fixing that I found a second problem on the same round trip: importing writes `group_id` and `household_id` into the recipe's `extras`. Export `{"my_key": "my_value"}` and it comes back as `{"my_key": "my_value", "group_id": "...", "household_id": "..."}`.

- `mealie/services/recipe/recipe_service.py`
  - `_replace_taken_id` gives the recipe a new id when the one in the file is already in use. When the id is free it is kept exactly as before, so importing a file from another instance is unaffected. The lookup deliberately runs without the group and household filters, because recipe ids are unique across every group, and passes `key="id"` because `RepositoryRecipes` is keyed by slug.
  - `_process_recipe_data` no longer walks into `extras`. It stamps the current group and household onto every nested object it visits, which is right for foods and units but wrong for a free-form field. This also covers the nested `extras` on foods and units.
  - Added the missing return annotation on `_process_recipe_data`.

Relates to #6887

The narrowness of the id change is the part worth checking: `test_create_recipe_from_zip_keeps_unused_id` passes both with and without the fix, so a file whose id is not already present still imports under its original id. Only the case that previously returned a 500 behaves differently.

Two pre-existing problems I left alone, happy to do either as a follow-up:

- `create_from_zip` binds `data_service` inside `if recipe and recipe.id:` but uses it in a separate `if recipe_image:` block, so a recipe with no id and an image would raise `NameError`.
- A malformed id in the file returns HTTP 500. I confirmed this on an unmodified checkout, so it is not introduced here.

## Testing

Six tests added to `tests/integration_tests/user_recipe_tests/test_recipe_crud.py`. With the fix stashed, four of them fail and `keeps_unused_id` passes, which is the intended split.

- `reimport_into_same_instance` exports through the real endpoint, imports it back, and checks the new recipe has a different id while the original is untouched
- `reimport_preserves_contents` does the same with a fully populated recipe and asserts every part survives: description, yield, servings, prep/cook/perform/total times, rating, original URL, categories, tags, tools, nutrition, notes, settings, extras, ingredient section heading, quantity, unit, food, note, ingredient substitutions (both a substitute food and a note-only one), step heading, summary and text, step to ingredient links, and step to note links
- `keeps_unused_id` guards that a free id is still preserved
- `without_id` covers a file with no id
- `id_taken_in_another_group` covers a collision outside the importing user's group, and checks the other group's recipe is untouched
- `preserves_extras` covers the extras fix

Run: `test_recipe_crud.py` 67 passed, `ruff check` clean, `ruff format` clean, `mypy` clean on the changed module. Not run: the full backend suite and any Postgres run, since the change is confined to the zip import path.

I also checked by hand that the image in the archive survives, that importing the same zip twice yields two distinct recipes, and that re-importing after deleting the original keeps the original id.

## AI / LLM Assistance

This PR was written with Claude Code. I directed the work and reviewed every change. The two bugs were found by testing the export and import round trip end to end, and the behaviour claims in this description are backed by test runs rather than reading the code.


Rouzax:
@michael-genson, are you happy with this shape or do you want me to look at it again?

michael-genson (dismissed):
Thanks!

michael-genson:
Looks like some test failures, this is probably related to the 3.14 upgrade. I had to address similar issues here:
https://github.com/mealie-recipes/mealie/pull/8521/changes/ed978deba90ab59ad5ef97461433051cfaa5526c
https://github.com/mealie-recipes/mealie/pull/8521/changes/97fcbd1c4e09e434d6aaf6acd6bc36c96565943e

Rouzax:
Thanks, that was the 3.14 temp file leak in one of my new tests (plus the dropped `shutil` import). Rebased on mealie-next and fixed, ready for the queue again.

michael-genson (approved):

PR #8473, opened 2026-09-20
chore: run tests in parallel with pytest-xdist

bestxmg:

The Python test suite takes about 15 minutes to run one test at a time. This PR runs it in parallel with pytest-xdist. On a 16-core machine the full suite (2982 passed, 18 skipped) goes from about 15 min to about 7 min.

- `Taskfile.yml`: `task py:test` now runs `pytest -n auto --dist loadfile`. `loadfile` keeps each test file on one worker, in order, because tests in a file share module-level state and fixtures.
- `pyproject.toml`, `uv.lock`: add `pytest-xdist` to the dev group.
- `tests/conftest.py`: each worker uses its own data dir (`tests/.temp/gwN`), so workers don't share one SQLite file. A plain `pytest` run still uses `tests/.temp`.

Two tests must be fixed for parallel runs:

- `test_recipe_crud.py`: `zip_recipe` called `shutil.make_archive` without `root_dir`, which zipped the whole repo (about 100s per zip test) and then overwrote the result. It now only uses `ZipFile`.
- `test_user_api_token.py`: `test_delete_token` deleted hardcoded token ids 1 and 2. If another file that creates tokens (e.g. `test_user_login.py`) runs first on the same worker, those ids belong to other tokens and the test fails with a 404. xdist runs bigger files first, so this order can now happen. The test now deletes the tokens it creates.

No issue. Related discussion: https://github.com/mealie-recipes/mealie/discussions/8198

- Zipped file and all the content inside before and after my commit.
image

- If test_user_login.py runs prior to test_user_api_token.py in the same worker. The conserquence be like this:
image

## Testing

Ran the full suite with `task py:test`: 2982 passed, 18 skipped, 0 failed.

## AI / LLM Assistance

Claude Code


bestxmg:
Hi mates, Hi @michael-genson
May I get some feedback in terms of this PR? Because I wanna ensure that I am working on the right track if possible. Thanks alot. 😋

By the way, this screenshot is the comparison of execution time between my PR and a recently merged PR with regard to "[PR CI / Backend Server Tests / tests (sqlite) "

It shows my PR gained expected performance improvement. However, the backend test CI for postgreSQL failed. I am expected to fix the problem this week.
image

michael-genson (approved):
Nice work, and thanks for the thorough benchmarking write-up.

Checked the isolation story specifically, since that's where parallelizing a suite usually goes wrong:

- `DATA_DIR=tests/.temp/gwN` resolves correctly — `determine_data_dir()` joins a relative `DATA_DIR` onto `BASE_DIR` under `TESTING`, so it lands in the same place `_TEMP_DIR` cleans.
- The env vars are set before `mealie.app` is imported, so `config.DATA_DIR` (read at module import) picks them up.
- Per-worker Postgres databases are created from the default DB connection before `main()` runs, and each worker cleans up its own temp dir via `global_cleanup`.
- The `test_delete_token` rewrite is a strict improvement — it no longer depends on hardcoded ids that were only ever correct by accident of ordering.
- `--dist loadfile` is the right choice here given how much module-level state the fixtures carry.

Approving.

PR #8496, opened 2026-09-21
fix: show error feedback when a recipe asset upload fails

### LINKED ISSUE #8494: [BUG] - Recipe asset upload fails silently (no error shown)
### First Check

When the server rejects a recipe asset upload, the UI shows no feedback, so nothing tells the user the upload failed or why.

Cause, from reading the code (mealie-next at ec9d05abd):
- frontend/app/plugins/axios.ts only toasts when error.response.data.detail.message exists.
- POST /api/recipes/{slug}/assets raises HTTPException(400, detail="Unsupported file extension"). The detail is a plain string, so no toast is raised.
- RecipeAssets.vue addAsset(): `if (data) { ... }` has no else branch, and the form state is reset whether or not the upload succeeded.
- RecipePageInstructions.vue handleImageDrop(): `if (!data) { return; // TODO: Handle error }`.

#8437 added alertUnreportedError for the hero image upload but did not cover the asset paths. Related: #5580 (similar symptom, different trigger).

Expected: an error toast, so the user knows the upload did not succeed.

I'm happy to submit a PR using alertUnreportedError.

### Steps to Reproduce

1) Open a recipe > Edit.
2) Assets > + > choose a file with an unsupported extension (e.g. rename a file to test.exe) > Create.
3) Nothing is shown. The Network tab shows POST /api/recipes/<slug>/assets returning 400 {"detail":"Unsupported file extension"}.

Also reproduced on demo.mealie.io.

Backend log (mealie-next @ ec9d05abd, running from source):

INFO     2026-09-21T11:50:28 - [<client>] 400 Bad Request "POST /api/recipes/classic-apple-tart/assets HTTP/1.1"

Browser Network tab (same request):
Status: 400 Bad Request
Response: {"detail":"Unsupported file extension"}

No toast or other message is shown in the UI in either case.

### Mealie Version

demo.mealie.io (nightly), reproduced Sep 21, 2026.
Also mealie-next @ ec9d05abd, running from source

### Deployment

Docker (Linux)

Debian 13 LXC, running from source (uvicorn + nuxt dev).

bferd:

When a recipe asset upload fails, the user gets no feedback at all. This makes the failure visible.

**Repro (before this PR)**

1. Recipe > Edit > Assets > new asset dialog > upload a disallowed file type (e.g. a renamed `test.exe`). The server returns `400 "Unsupported file extension"`. Nothing visible happens.
2. In the step editor, drop an image with a disallowed extension (e.g. `.svg`, which passes the `image/` MIME check but isn't in `ASSET_ALLOWED_EXTENSIONS`). Nothing visible happens, and `handleImageDrop()` had a `// TODO: Handle error`.

**Cause**

`createAsset()` returns the request wrapper's `{ data, error }`, and both call sites ignored `error`. The axios interceptor only toasts when `detail.message` exists, and the asset route raises a plain-string `detail`, so nothing was ever shown.

**Changes**

- `RecipeAssets.vue` and `RecipePageInstructions.vue` now check `error` and call the existing `alertUnreportedError()` helper, following `RecipePageEditorToolbar.vue` and `RecipeImageUploadBtn.vue`. The TODO is removed.
- `alertUnreportedError()` in `use-toast.ts` now also handles a plain-string `detail`: the string is shown as the toast message, with the existing "Something Went Wrong!" as the title. Object `detail.message` (already toasted by the interceptor) is still skipped, so there is no double toast. Anything else (validation arrays, missing detail, no response) still shows the generic fallback.

**Release notes**

- Uploading a recipe asset that the server rejects (for example, an unsupported file type) now shows an error message instead of failing silently.

Fixes #8494

- No new translation strings. `events.something-went-wrong` already exists, and no locale files are touched.
- The string-detail change to `alertUnreportedError()` also applies to its existing callers (hero image upload, delete image, image from URL). None of those routes currently returns a string `detail`, so their behavior is unchanged today.
- The asset dialog still closes on submit even when the upload fails (`BaseDialog` closes on `submit` unless `loading`/`keep-open` is set), so the user's input is cleared. I left that alone to keep this PR to error feedback.
- Found while testing, and deliberately not in this PR:
  - `PUT /api/recipes/{slug}/image` returns a bare `500` for non-image files (Pillow raises `UnidentifiedImageError` during minification), so the hero image upload can only show the generic message. This needs backend validation and belongs in its own PR.
  - The asset dialog doesn't say which file types are accepted, and the icon selector doesn't affect upload behavior.
  - The server's `detail` string is not localized.

## Testing

- `pnpm exec eslint` on the four changed files: clean.
- `pnpm test:ci`: 41 files, 420 tests pass. Three new tests cover string `detail`, whitespace-only string `detail`, and array `detail` in `use-toast.test.ts`.
- Manual UI checks on a dev instance:
  - Renamed `.exe` in the asset dialog shows "Something Went Wrong! Unsupported file extension", and no asset is added.
  - `.svg` dropped on a step shows the same toast, and no `` is inserted.
  - Valid `.png` upload works in both places with no toast.
  - Hero image "from URL" with a non-image URL shows a single toast (no double toast).
  
## Screenshot

2026-09-21 12_52_42-Greenshot

## AI / LLM Assistance

Claude Code was used to implement this, following the existing `alertUnreportedError` pattern and its current call sites. All generated code was reviewed, tested manually in the UI, and verified against the codebase conventions before submission.


bferd:
While testing I found a related backend gap that I've kept out of that PR.

`PUT /api/recipes/{slug}/image` (hero image upload) has no file-type validation. Uploading a non-image (I tried `.svg`, `.pdf`, `.mp4`) writes the file, then image minification fails with `PIL.UnidentifiedImageError`, and the route returns a bare `500` with no `detail`. The frontend can only show the generic "Something Went Wrong!" toast for that case.

The asset route already returns `400 "Unsupported file extension"` for disallowed types, and the URL-based image route returns 400s with `ErrorResponse.respond(...)`. Validating the hero image upload the same way would make the reason reach the toast with no further frontend changes.

Will open a separate PR for that.

